PR #1, opened 2020-03-24
Add a basic Dockerfile

andyhd:
For a reproducible environment when testing and presumably for
deployment, we will run the app in a Docker container. Using Gunicorn
for a WSGI server, but only as a default.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+*.sw?
+.git
+.github
+venv
+.projectrc
+.python-version
+requirements-dev.txt
+
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,6 @@
-*.sw?
 .git
 .github
 venv
 .projectrc
 .python-version
-requirements-dev.txt
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.8-slim-buster
 
-LABEL maintainer="andy.driver@digital.trade.gov.uk"
+LABEL maintainer="webops@digital.trade.gov.uk"
 
 ENV DJANGO_SETTINGS_MODULE "tamato.settings"
 
@@ -12,20 +12,17 @@ USER tamato
 WORKDIR /home/tamato
 
 # install python dependencies
-COPY requirements.txt manage.py ./
-RUN python -mvenv venv && \
-    venv/bin/pip install -U pip && \
-    venv/bin/pip install -r requirements.txt
+COPY requirements.txt ./
+RUN pip install -U pip && \
+    pip install -r requirements.txt
 
-COPY tamato tamato
-COPY docker docker
-COPY static static
+COPY . .
 
 # empty .env file to prevent warning messages
 RUN touch .env
 
 # collect static files for deployment
-RUN venv/bin/python manage.py collectstatic --noinput
+RUN python manage.py collectstatic --noinput
 
 EXPOSE 8000
-CMD ["venv/bin/gunicorn", "-b", "0.0.0.0:8000", "-w", "4", "tamato.wsgi:application"]
+CMD ["/home/tamato/.local/bin/gunicorn", "-b", "0.0.0.0:8000", "-w", "1", "tamato.wsgi:application"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM python:3.8-slim-buster
+
+LABEL maintainer="andy.driver@digital.trade.gov.uk"
+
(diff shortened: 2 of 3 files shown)


.dockerignore
```diff
@@ -0,0 +1,8 @@
+*.sw?
```
NDevox: What are these files? 🤔

.dockerignore
```diff
@@ -0,0 +1,8 @@
...
+.git
+.github
+venv
+.projectrc
+.python-version
+requirements-dev.txt
```
NDevox: debateable - I like to run my tests within a docker container ideally so they're close to the prod env, so might need this for that. But up for discussion.

Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM python:3.8-slim-buster
+
+LABEL maintainer="andy.driver@digital.trade.gov.uk"
```
NDevox: Should use a group email so it's not tied to a single person.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    useradd -u 1000 -g tamato -m tamato
+USER tamato
+
+WORKDIR /home/tamato
+
+# install python dependencies
+COPY requirements.txt manage.py ./
+RUN python -mvenv venv && \
+    venv/bin/pip install -U pip && \
+    venv/bin/pip install -r requirements.txt
```
NDevox: Better to switch this to:


```docker
COPY requirements.txt
RUN pip install -r requirements.txt
```

We don't need to include anything but requirements here - it will give us the simplest hash for this layer of the dockerfile so we don't rebuild the layer needlessly.

We also don't need to run in a virtualenv. They're useful for local development so project requirements are isolated. However a container should only hold a single project within it and so this isolation already exists, introducing a virtualenv just adds extra overhead and complexity.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+COPY tamato tamato
+COPY docker docker
+COPY static static
```
NDevox: Might as well switch this one to

```docker
COPY .
```

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# empty .env file to prevent warning messages
+RUN touch .env
+
+# collect static files for deployment
+RUN venv/bin/python manage.py collectstatic --noinput
+
+EXPOSE 8000
+CMD ["venv/bin/gunicorn", "-b", "0.0.0.0:8000", "-w", "4", "tamato.wsgi:application"]
```
NDevox: with the above removal of the `venv` we can just run `gunicorn`.

Also should specify workers as 1 in a container - containers are the scaling unit here, extra workers in a container just confuse the CPU usage:

```docker
CMD ["gunicorn", "-b", "0.0.0.0:8000", "-w", "1", "tamato.wsgi:application"]
```

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+LABEL maintainer="andy.driver@digital.trade.gov.uk"
+
+ENV DJANGO_SETTINGS_MODULE "tamato.settings"
+
+# don't run as root
+RUN groupadd -g 1000 tamato && \
+    useradd -u 1000 -g tamato -m tamato
+USER tamato
+
+WORKDIR /home/tamato
```
NDevox: Advantages of setting up a non root user are limited in an isolated env such as this container.

I'm neither here nor there on the subject.

.dockerignore
andyhd: Ah, Vim swap files - there isn't a global `.dockerignore` file like there is for `.gitignore`, otherwise I'd put this in that

.dockerignore
andyhd: Yep, I'm not wedded to having this in here at all

Dockerfile
andyhd: Agreed 👍

Dockerfile
andyhd: I've read arguments both for and against virtualenvs inside docker containers. I normally don't use virtualenv inside docker, so happy to remove it. I must have read something recently that made me assume it was the preferred method these days.

Dockerfile
andyhd: This is just good security practice in Docker containers. If an application manages to break out of the container, it can have root privileges on the host. This is a belt-and-braces approach (similar to the virtualenv-inside-docker thing), but I'd usually err on the side of caution.

Dockerfile
NDevox: 🤔 I'm all good with everything else. We should just get an email for this and it's good to go.

Maybe squashing commits as well.

PR #3, opened 2020-03-25
refactor(structure): Updated directory structure for TaMaTo.

NDevox:
This creates a structure which splits at the top level making things immediately
easy to navigate as opposed to having to dig through further folder levels.

A setup.cfg is added to simplify coverage. Two empty apps are added as a base.
commodities to cover commodity codes and measures.


NDevox:
I think that was the biggest issue for me with this one as well 😄. Not entirely sure where else to put it, but since it essentially is a setting... Cheers. I'll merge yours into mine and then merge it all.

PR #4, opened 2020-03-25
Fix settings and Dockerfile to work with changed directory structure

andyhd:
Refactor of the project layout left a few module paths and directories
broken in settings and the Dockerfile.

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -2,10 +2,10 @@
 Django settings for tamato project.
 """
 import os
+import sys
 from os.path import abspath
 from os.path import dirname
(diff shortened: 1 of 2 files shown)


settings/common.py
```diff
@@ -2,10 +2,10 @@
 Django settings for tamato project.
 """
 import os
+import sys
```
andyhd: ugh, i have a plugin which auto-formats imports, but it does some odd things. i'll fix this.

NDevox:
my bad, lgtm

PR #5, opened 2020-03-30
feature(models): Add Commodity and Measure models to the Tariff tool.

NDevox:
Commodities and Measures are some of the core components for the Tariff tool. They
also form some of the most significant complexities. Commodities are designed in
life to follow a hierarchical system. Measures on the other hand have incredibly
complex relations pertaining to their conditions, duties and geographical areas.
This is combined with the need for a "time machine" allowing a user to see the state
of the database as it was at any point of time.

This commit creates the Commodity model as a hierarchical model using
Django-Treebeard. The commodities follow the Materialised Path design as it is
efficient for read operations (this dataset will be edited infrequently) and
allows multiple roots (each "section" for commodity codes would form a root). The
code also uses the default treebeard setup of a 4-character step length with a 36
character alphabet - giving each node the possibility of having roughly 1.6 million
children.

To facilitate the time machine system the design approach that has been decided on
is an append-only log style system - in that no row should be deleted, any
object that needs to be replaced should just be superseded with a new row. To
facilitate this a "valid_between" DateTimeRangeField has been added. This enables
the code to use database queries that find data as it was at specific dates with
ease. Some convenience methods implementing this have been added. An Exclusion
Constraint has also been added to enforce commodities with the same code cannot have
overlapping dates. This as a result requires the BTreeGIST extension installed on
postgresql - which is included in the migration. The system for enforcing
append-only updates has not yet been designed, however a concept of "successor"
and "predecessor" has been introduced, producing a link to the last used and next
used row for a commodity.

A simple Measure model has been added as well to showcase how this would connect
to a commodity as well and how it can be used in queries looking for measure in a
commodity tree.


commodities/models.py
```diff
@@ -1,3 +1,46 @@
...
+
+    description = models.TextField()
+    predecessor = models.OneToOneField(
+        "self",
+        on_delete=models.PROTECT,
+        null=True,
+        blank=True,
+        related_name='successor'
+    )
+    version = models.PositiveIntegerField()
```
andyhd: I assume this is just a stub - what is the logic supposed to be around `version`? Is it an auto-increment?

NDevox: Kind of but not really. It should increment within a commodity but not across the whole table

commodities/models.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def get_live_children(self, **kwargs):
+        return self.get_children().filter(live=True, **kwargs)
```
andyhd: To implement the time machine functionality, "liveness" would need to take the requested timestamp into account - so is the `live` field useful?

commodities/models.py
andyhd: Ah, yes, that's what I meant

commodities/models.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def get_measure(self):
+        if hasattr(self, 'measure'):
+            return self.measure
+        query = self.objects.get_ancestors().filter(measure__isnull=False)
+        if query:
+            return query.first().measure
```
andyhd: Is it true that a commodity has only (zero or) one measure? The existing data model seems to be a one to many relation.

matthewford: A GN can have many measures.

NDevox: 👍 my bad, I negledcted to check my notes on this one. Updated.

commodities/models.py
NDevox: Live is useful for drafting. If we want to go that way.

I.e. we might be able to allow overlapping commodity codes if one is still a draft, i.e. not live. Up for debate on how this is done though.

commodities/models.py
NDevox: It could be seen as slightly more read friendly than `successor`/`predecessor`. The table wide "version" would just be the PK in this case (which defaults to an autoincrement in django). Drafts would have a version IMO - but obviously that's not been designed at all yet so is open for discussion.

simonwo:
Also, another general question: in all the places where we're making changes, are we also aiming to keep compability with the EU schema so that we can pull in existing new EU data for doing comparisons? Or are we intending to have a different schema entirely for that?

commodities/models.py
NDevox: I have a few thoughts on this one.

I essentially see 3 options:

1) Have "draft" rows that are stored within the tables. This is what the current design is leaning towards but I'm not married to it, nor am I confident it's a good approach. This would essentially mean we have to filter for live (and current) vs draft rows depending on what we are doing - but that's not a huge ask and very indexable. Might be a bit confusing when looking at multiple drafts at once - will probably need to include a concept of "ownership" for whoever is responsible for that row change.

2) Have a separate drafting table for each model - this might be a bit tidier in terms of separating draft from live data - easier to query as well. But a lot more code and doesn't feel clean or reusable.

3) Have a global drafting table - I'd imagine this essentially being the workbasket table. I'd probably propose linking the workbasket to a user so we know who owns it, and then store a JSON (or similar) blob which holds the actions proposed (Model types, IDs of the object and fields to change). We could then link rows for models (e.g. comm codes) to workbaskets and not users. Would also help with storing a users changes before publishing a workbasket. I quite like this approach as it completely segregates the drafting and workbasket process from the rest of the system keeping things less tightly coupled.

In theory 4 could probably be replaced with a cache or similar, but I'd rather store in a DB for persistence and auditing.

commodities/models.py
```diff
@@ -1,3 +1,46 @@
...
+from django.contrib.postgres.fields import RangeOperators
 from django.db import models
+from treebeard.mp_tree import MP_Node
 
-# Create your models here.
+from common.models import TimestampedMixin, ValidityMixin
+
+
+class Commodity(MP_Node, TimestampedMixin, ValidityMixin):
+    code = models.CharField(max_length=12)
```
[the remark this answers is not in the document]
NDevox: Lack of faith is the reasoning here 😄. We can be fairly certain they'll be 10 digits, but I wouldn't be overly surprised if this changes or is extended - we lose nothing by keeping it slightly longer.

I would keep the suffix separate. What I'm less convinced about is what we do with additional codes.

measures/models.py
```diff
@@ -1,3 +1,8 @@
 from django.db import models
 
-# Create your models here.
+from commodities.models import Commodity
+
+
+class Measure(models.Model):
+    commodity = models.ForeignKey(Commodity, on_delete=models.PROTECT, related_name='measures')
```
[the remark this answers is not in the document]
NDevox: Yeah this is a question I have momentarily discussed with @andyhd but without any real conclusion being brought from that.

When a new commodity is created we need to create a new tree... Does that mean we should also recreate all the dependencies? i.e. recreate all measures as well? Or do we just shift that measure to point at the new commodity instead of recreating it?

The former is better for the time machine approach. Not sure what the argument for the latter really is.

TBD. I'm leaving measures intentionally vague right now.

commodities/migrations/0001_initial.py
```diff
@@ -0,0 +1,40 @@
...
+        BtreeGistExtension(),
+        migrations.CreateModel(
+            name='Commodity',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('path', models.CharField(max_length=255, unique=True)),
+                ('depth', models.PositiveIntegerField()),
+                ('numchild', models.PositiveIntegerField(default=0)),
+                ('created_at', models.DateTimeField(auto_now_add=True)),
+                ('updated_at', models.DateTimeField(auto_now=True)),
```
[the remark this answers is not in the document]
NDevox: Depends on how we carry this forward - I wouldn't immediately expect it to change but there might be some niche cases where it does. I've generally found it better to have both of these fields than not, just in case, and it costs us very little to do so. It won't matter if we never use it. But if we want to use it and don't have it it'll be painful.

matthewford:
> Also, another general question: in all the places where we're making changes, are we also aiming to keep compatibility with the EU schema so that we can pull in existing new EU data for doing comparisons? Or are we intending to have a different schema entirely for that?

If you do not keep compatibility with the EU schema, you will be giving yourselves a lot more work generating the TARIC3 XML's for ECDP.

NDevox:
The plan as it stands I believe is to not copy the EU schema perfectly as the design is... not ideal. We'd want to store the data in a way that keeps all the information that the EU schema does. The difference is that it will be more robust and simpler to manage (from a dev perspective). We should be able to conform to the TARIC3 XML given that we have all the data required. One significant body of work down the road will be transforming the EU data to the model we are building. (another will be producing the TARIC3 XML...)

With this in mind, depending on how we want to proceed, we should be able to take any EU export at any time and run it through our transform script to suit our model.

PR #8, opened 2020-04-02
Add GOV.UK PaaS deployment manifest

andyhd:
To deploy the app to GOV.UK PaaS, we need to specify the application name and various
metadata in a manifest file. The manifest specifies to use the Python buildpack, which
gets the required version of Python from the runtime.txt file.


urls.py
```diff
@@ -13,9 +13,21 @@
...
+    path("measures/", include("measures.urls")),
     path("admin/", admin.site.urls),
 ]
+
+if settings.DEBUG:
+    import debug_toolbar
+
+    urlpatterns = [path("__debug__/", include(debug_toolbar.urls)), *urlpatterns]
+
+urlpatterns += staticfiles_urlpatterns()
```
NDevox, line 33: I wouldn't let django serve static files - it's not very good at doing it. If we want to be lazy around this I'd put whitenoise in and serve them that way: http://whitenoise.evans.io/en/stable/django.html

andyhd: Yeah, I would serve them separately rather than use Whitenoise. But this is just temporary.

andyhd: Maybe I can put this behind `if settings.DEBUG`?

NDevox: 🤔

This is because we are serving with gunicorn in dev... Is there a good reason for recommending this over runserver in dev? If so then yes lets put it behind `if debug`, we should also add the `--reload` flag to the recommended dev setup IMO.

andyhd: Good point - there's no reason to use gunicorn in dev. I'll remove that.

PR #9, opened 2020-04-07
Add GOV.UK Frontend integration

andyhd:
To make the webapp look and feel like GOV.UK, and to benefit from their accessibility work, we want to integrate with the GOV.UK Frontend framework.

This change adds GOV.UK Frontend styles, scripts, fonts and images via npm, webpack and django-webpack-loader - see https://readme.trade.gov.uk/docs/howtos/govuk-frontend-webpack-integration.html

This also adds GOV.UK Design System component templates using https://github.com/alphagov/govuk-frontend-jinja


common/jinja2/layouts/_generic.jinja
```diff
@@ -0,0 +1,19 @@
+{% extends "template.njk" %}
```
NDevox, line 1: I'm guessing this comes from git+https://github.com/alphagov/govuk-frontend-jinja.git

common/jinja2/common/index.jinja
```diff
@@ -0,0 +1,7 @@
+{% extends "layouts/layout.jinja" %}
+
+{% set page_title = "Hello World" %}
```
NDevox, line 3: Probably want to change this

common/jinja2/common/index.jinja
```diff
@@ -0,0 +1,7 @@
...
+
+{% block content %}
+  <h1 class="govuk-heading-xl">Hello World</h1>
```
NDevox, line 6: and this

common/jinja2/layouts/_generic.jinja
```diff
@@ -0,0 +1,19 @@
...
+  <!--[if !IE 8]><!-->
+    {{ render_bundle("main", "css") }}
+  <!--<![endif]-->
+  <!--[if IE 8]>
+    {# render_bundle("ie8", "css") #}
+  <![endif]-->
+  {% block styles %}{% endblock %}
+{% endblock %}
+
+{% set htmlClasses = "" %}
```
NDevox, line 13: Do we want to be picky over camelCase vs snake_case?

andyhd: This template variable is used in the GOV.UK template, so we have to keep it camelCase, I'm afraid. Otherwise I would always use snake_case.

common/jinja2/layouts/_generic.jinja
andyhd: Yes, this template is in `node_modules/govuk-frontend/govuk/template.njk` and govuk-frontend-jinja allows us to extend it like this.

PR #10, opened 2020-04-08
Compile SCSS and Javascript assets in Dockerfile

andyhd:
The webapp now depends on NodeJS to fetch and compile GOV.UK Frontend and its dependencies.

This change introduces a multi-stage build, which first uses a NodeJS image to install GOV.UK Frontend, then passes the compiled assets to the Python image.

The container is now run with docker-compose, as it depends on a Postgres database.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,21 @@
+FROM node:13.12-buster-slim AS jsdeps
+
+COPY . .
+
+RUN npm install && npm run build
+
+
 FROM python:3.8-slim-buster
 
 LABEL maintainer="webops@digital.trade.gov.uk"
 
 ENV DJANGO_SETTINGS_MODULE "settings"
+
+# add git client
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -16,7 +29,10 @@ COPY requirements.txt ./
 RUN pip install -U pip && \
     pip install -r requirements.txt --no-warn-script-location
 
-COPY . .
+COPY --chown=tamato:tamato . .
+COPY --chown=tamato:tamato --from=jsdeps node_modules node_modules
```
NDevox, line 33: Wonder if we can bundle the assets we need from here as well...

If not maybe we should just copy the specific path we need?

andyhd: Ah yeah, good point. We don't want to bundle them, because the intent here is to copy the GOV.UK Nunjucks templates for Python to use. I'll copy just the specific path.

PR #11, opened 2020-04-14
Enable deployment onto GOV.UK PaaS

andyhd:
* Adds [`dj-database-url`](https://github.com/jacobian/dj-database-url) to get the connected database service credentials from the `DATABASE_URL` environment variable.
* Adds [`whitenoise`](http://whitenoise.evans.io/en/stable/) to serve static assets
* Adds a [`Procfile`](https://docs.cloudfoundry.org/buildpacks/prod-server.html) to run migrations and start the app on deployment
* Updates the Github Actions workflow to deploy to GOV.UK PaaS on passing tests

### DIFF
--- a/.cfignore
+++ b/.cfignore
@@ -1,0 +1,131 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+pip-wheel-metadata/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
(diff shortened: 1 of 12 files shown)


NDevox:
Are we not using the node stuff anymore then?

andyhd:
I thought we might run `npm install` before `cf push` and vendor the node modules, but now I'm having second thoughts.

simonwo:
Is there a way to get GOV.UK styles stuff without any Node stuff?

andyhd:
Yes, we can fetch the compiled CSS and JS files directly from Github. But we're also using their templates, which we could also get directly from Github in a different way. I had written some code to do this, but it seemed more sensible to use npm and fetch everything at once - plus we'll likely write our own CSS using SCSS and our Javascript with Babel, so we'll end up needing npm anyway.

simonwo:
For me it's just about trying to keep things as simple as possible, both for us now and devs in the future. I'd be happy to have some static stuff committed into the repo if it means we don't have to pull in Node and solve these problems around getting stuff in the right place for the deployment.

Although as you say we might be pulling in some sort of transpile at some point. I know there is a SCSS package for Python, but less support for anything JS related – certianly might need that when we come to use stuff from [the other repo](https://github.com/uktrade/trade-tariff-management) which uses node already.

So it's your call on how best to keep it simple – happy to just commit some static stuff now and park using Node for another day if you think your template copying isn't equally complex, and we can replace it with the frontend package when we come to actually use a Node for something else. Or go ahead and plumb Node in if we're sure we'll go beyond vanilla CSS and JS!

NDevox:
https://uktrade.atlassian.net/browse/TP-100

PR #15, opened 2020-04-21
Add BDD test framework

andyhd:
### Requirement
We want to write tests to prove that the TAMATO app functions equivalently to the old
TAP project that it replaces.

### Context
* We have a large amount of pre-existing documentation of functional requirements in the
form of test scenarios and data.
* We have already chosen pytest as our testing framework, due to the development team's familarity with it.

### Decision
* Acceptance criteria for user stories will be written in [Gherkin format](https://en.wikipedia.org/wiki/Cucumber_(software)#Gherkin_language). We will write tests that interpret these criteria using [pytest-bdd](https://github.com/pytest-dev/pytest-bdd#readme).
* Test result reports will be automatically generated using [Allure Framework](https://github.com/allure-framework/allure2) and published somewhere stakeholders can view them (currently attached to Github Actions artifacts).

### Consequences
* Collaboration between business analysts and the development team will be easier due to a shared use of Gherkin formatted feature specifications.
* The development team can continue to use pytest and not have to learn a new test
framework or tool.
* Stakeholders can see easy to understand reports on test results and requirements
coverage.

A simple example is included, testing the stub footnotes app.


footnotes/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
NDevox, line 3: We should delete this file

footnotes/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
```
NDevox, line 3: Would argue for deleting this as well

tests/bdd/test_browse_footnotes.py
```diff
@@ -0,0 +1,42 @@
...
+import pytest
+from django.urls import reverse
+from pytest_bdd import given
+from pytest_bdd import scenarios
+from pytest_bdd import then
+from pytest_bdd import when
+
+
+@pytest.fixture(autouse=True)
+def enable_db(db):
```
NDevox, line 11: Can't we just use `pytestmark = pytest.mark.django_db`?

tests/bdd/test_browse_footnotes.py
```diff
@@ -0,0 +1,42 @@
...
+    pass
+
+
+class FootnoteFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = "footnotes.Footnote"
+
+    description = factory.Faker("paragraph")
```
NDevox, line 19: I'd argue to put factories in their own factory file - these are likely to be used all over the place in tests.

tests/bdd/test_browse_footnotes.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@when("I go to the footnotes page")
+def footnotes_page(footnotes, client, response):
+    response["content"] = client.get(reverse("footnotes-list")).content.decode("utf-8")
+
+
+@then("I should see a list of footnotes")
+def footnotes_list(footnotes, response):
+    assert all(footnote.description in response["content"] for footnote in footnotes)
```
NDevox, line 42: Ok having not done BDD before this looks really weird for me. But I kind of love that Arrange, Act and Assert are split like this.

tests/bdd/test_browse_footnotes.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@given("some footnotes exist")
+def footnotes():
+    return FootnoteFactory.create_batch(10)
+
+
+@pytest.fixture
+def response(client):
+    return {}
```
NDevox, line 32: We should group fixtures at the top IMO - makes it more readable.

Also this is a weird usage of mutability - is this the accepted way of doing things?

requirements.txt
```diff
@@ -13,3 +13,5 @@ psycopg2-binary
 sentry-sdk
 werkzeug
 whitenoise
+pytest-bdd
+allure-pytest-bdd
```
NDevox, line 17: Can we order alphabetically. Also shouldn't these be in requirements-dev.txt?

footnotes/views.py
```diff
@@ -0,0 +1,10 @@
+from django.shortcuts import render
+
+from .models import Footnote
+
+
+# Create your views here.
```
NDevox, line 6: Remove comment

requirements.txt
andyhd: Yep, will do. Actually, I think we need to move all the pytest dependencies in here, or at least into a requirements-test.txt

footnotes/views.py
andyhd: Yep

tests/bdd/test_browse_footnotes.py
andyhd: Sure, I'll try to come up with something less surprising

andyhd:
My thinking around having the BDD tests at project level was that they are more like integration/functional tests, testing across apps. But features map quite cleanly to the apps, so either way would probably be fine. I think keeping unit tests within the apps is logical, so I guess putting BDD tests in with them makes sense.

requirements.txt
NDevox: That works for me - although I'm happy with using requirements-dev as well given we'll almost always have to install both as devs.

NDevox:
Yeah actually you're argument for having bdd tests top level makes some sense... Especially as many will cover multiple aspects of the system... I think maybe as the tests are currently obviously related to footnotes we can put them in that folder. We can probably punt redesigning test layout to later, when the issue of where to place tests becomes murkier.

requirements.txt
andyhd: It's just that django-debug-toolbar isn't needed in testing environments - but maybe we should remove that and just put it in our own (gitignored) local settings overrides.

PR #16, opened 2020-04-24
Add browsing and finding footnotes [MTP-2]

andyhd:
Tariff Managers need to be able to read and edit footnotes.

This PR adds some API endpoints using [Django REST Framework](https://www.django-rest-framework.org/), to enable listing footnotes
and footnote types in the database (currently only in JSON format) and a web page which
displays a list of footnotes and allows filtering by footnote ID.

See [MTP-2 Browse and find footnotes - Search only](https://uktrade.atlassian.net/browse/TP-91?atlOrigin=eyJpIjoiMTRiZDJhN2MzNTc0NDFhZTg1MmFkODY3YzU1NDE3ODkiLCJwIjoiaiJ9)


common/tests/factories.py
```diff
@@ -1,11 +1,41 @@
...
+
 import factory
+from psycopg2.extras import DateTimeTZRange
+
+
+BREXIT_DATE = datetime(2021, 1, 1).replace(tzinfo=timezone.utc)
+
+
+class ValidityFactoryMixin(factory.django.DjangoModelFactory):
+    valid_between = factory.LazyFunction(lambda: DateTimeTZRange(BREXIT_DATE, None))
```
NDevox, line 13: If the date never changes is it worth putting it in a lazy function?

common/serializers.py
```diff
@@ -0,0 +1,20 @@
...
+from django.contrib.auth.models import User
+from drf_extra_fields.fields import DateTimeRangeField
+from rest_framework import serializers
+
+
+class ValiditySerializerMixin(serializers.HyperlinkedModelSerializer):
+    valid_between = DateTimeRangeField()
+
+
+class UserSerializer(serializers.HyperlinkedModelSerializer):
```
NDevox, line 11: Do we need users and groups in the API?

footnotes/models.py
```diff
@@ -1,5 +1,35 @@
 from django.db import models
 
+from common.models import ValidityMixin
 
-class Footnote(models.Model):
+
+class NationalMixin(models.Model):
+    national = models.BooleanField(default=False)
```
NDevox, line 7: national?

footnotes/models.py
```diff
@@ -1,5 +1,35 @@
...
+
+    class Meta:
+        abstract = True
+
+
+class FootnoteType(ValidityMixin, NationalMixin):
+    id = models.CharField(primary_key=True, max_length=2, null=False, blank=True)
```
NDevox, line 14: I'd argue against overriding the generated PK and instead just having a unique field for the str (could probably be seen as a slug).

footnotes/models.py
```diff
@@ -1,5 +1,35 @@
...
+    footnote_type = models.ForeignKey(FootnoteType, on_delete=models.PROTECT)
+    description = models.TextField()
+
+
+class FootnoteDescriptionPeriod(ValidityMixin, NationalMixin):
+    footnote = models.ForeignKey(Footnote, on_delete=models.PROTECT)
+    footnote_type = models.ForeignKey(FootnoteType, on_delete=models.PROTECT)
+    footnote_description = models.ForeignKey(
+        FootnoteDescription, on_delete=models.PROTECT
+    )
```
NDevox, line 35: Don't think it serves any purpose to separate these does it?

footnotes/models.py
```diff
@@ -1,5 +1,35 @@
...
+
+
+class Footnote(ValidityMixin, NationalMixin):
+    id = models.CharField(primary_key=True, max_length=5, null=False, blank=True)
+    footnote_type = models.ForeignKey(FootnoteType, on_delete=models.PROTECT)
+
+
+class FootnoteDescription(NationalMixin):
+    footnote = models.ForeignKey(Footnote, on_delete=models.PROTECT)
+    footnote_type = models.ForeignKey(FootnoteType, on_delete=models.PROTECT)
```
NDevox, line 26: Do we need to put footnote_type in every one of these? They should match surely, so we end up with a ton of repetition.

footnotes/models.py
```diff
@@ -1,5 +1,35 @@
...
+
+
+class FootnoteType(ValidityMixin, NationalMixin):
+    id = models.CharField(primary_key=True, max_length=2, null=False, blank=True)
+    application_code = models.IntegerField()
     description = models.TextField()
+
+
+class Footnote(ValidityMixin, NationalMixin):
+    id = models.CharField(primary_key=True, max_length=5, null=False, blank=True)
```
NDevox, line 20: Same as above, I would avoid overriding the generated PK. Natural PKs can be a pain to work with and are unreliable. It costs nearly nothing to use the autogenerated PK.

footnotes/models.py
```diff
@@ -1,5 +1,35 @@
...
+    footnote_type = models.ForeignKey(FootnoteType, on_delete=models.PROTECT)
+
+
+class FootnoteDescription(NationalMixin):
```
NDevox, line 24: Given the way these models are laid out does it make sense to keep them separate?

With this way we can have multiple descriptions to a single footnote - but I thought we might also want to share a description between footnotes. But as we aren't storing anything language specific this feels a bit obtuse right now.

Even with language being stored as a separator I'd argue for a single model at the moment. To me the APIs essentially end up as the following:

```python
# With separate models.

lang_code = get_language()

description = foonote.foonote_description_set.filter(language=lang_code).first().description

# Using a single model with a JSONField

description = footnote.description[lang_code]
```

This would also be more performant from a DB perspective (JSONFields are still indexed in psql if we want to search them) and simpler to handle on the frontend.

Are we sure we want to go with the current setup?

footnotes/views.py
```diff
@@ -1,9 +1,45 @@
...
-def list_footnotes(request):
-    return render(
-        request, "footnotes/list.jinja", context={"footnotes": Footnote.objects.all(),}
-    )
+class FootnoteViewSet(viewsets.ReadOnlyModelViewSet):
+    """
+    API endpoint that allows footnotes to be viewed or edited.
+    """
+
+    queryset = Footnote.objects.all().order_by("id")
```
NDevox, line 19: I think this is the default order isn't it?

footnotes/views.py
```diff
@@ -1,9 +1,45 @@
...
+    renderer_classes = [renderers.TemplateHTMLRenderer]
+
+    def list(self, request):
+        queryset = self.filter_queryset(self.get_queryset())
+        return render(request, "footnotes/list.jinja", context={"footnotes": queryset})
+
+
+class FootnoteTypeViewSet(viewsets.ReadOnlyModelViewSet):
+    """
+    API endpoint that allows footnote types to be viewed or edited.
```
NDevox, line 40: ReadOnly can't be edited

footnotes/views.py
```diff
@@ -1,9 +1,45 @@
...
+    """
+
+    queryset = FootnoteType.objects.all().order_by("id")
```
NDevox, line 43: same as above, not sure this order_by is useful.

footnotes/views.py
```diff
@@ -1,9 +1,45 @@
...
+class FootnoteUIViewSet(FootnoteViewSet):
+    """
+    UI endpoint that allows footnotes to be viewed or edited.
+    """
+
+    renderer_classes = [renderers.TemplateHTMLRenderer]
+
+    def list(self, request):
+        queryset = self.filter_queryset(self.get_queryset())
+        return render(request, "footnotes/list.jinja", context={"footnotes": queryset})
```
NDevox, line 35: This is awesome and I've never seen a DRF view be turned back into a Django CBV like this.

common/tests/factories.py
andyhd: good point

common/serializers.py
andyhd: Yes, temporarily, because one of the acceptance criteria is that the request is made by a valid user. I'm assuming we will replace the user authentication with API tokens later.

footnotes/models.py
andyhd: I just copied the existing data model, but I realise that is unnecessary, so I'll remove that 👍

footnotes/models.py
andyhd: Again, just copying the existing data model, but no reason to do that 👍

common/serializers.py
NDevox: Good point.. we will need a whole auth discussion at some point (although I assume DIT has a standard way of doing this).

footnotes/models.py
andyhd: Same as above - I'll remove it

footnotes/models.py
NDevox: That is fair, nice opportunity to come up with a new schema though.

footnotes/models.py
andyhd: As we're still working on the data model, I just temporarily went with the existing one for the sake of having some models to use. I should have made the simplest models that work, especially as I was assuming we would be changing them soon anyway - my mistake!

footnotes/views.py
andyhd: Oops, not sure why I put that there

footnotes/views.py
andyhd: Ha, good point!

footnotes/views.py
NDevox: 😄

footnotes/models.py
NDevox: This is true

footnotes/views.py
andyhd: I wanted to reuse the filtering, but I think this might end up causing problems - I already ran into some issues with view name clashes.

footnotes/models.py
simonwo, line 20: Agreed with @NDevox that most likely we'll need to change this for a different primary key, for append-only log purposes, but maybe we revisit that when we've got a good view on it.

common/serializers.py
```diff
@@ -0,0 +1,20 @@
+from django.contrib.auth.models import Group
+from django.contrib.auth.models import User
+from drf_extra_fields.fields import DateTimeRangeField
+from rest_framework import serializers
+
+
+class ValiditySerializerMixin(serializers.HyperlinkedModelSerializer):
+    valid_between = DateTimeRangeField()
```
simonwo, line 8: So I assume this validates that start dates < end dates? What about `<=`?

footnotes/views.py
NDevox: Ah interesting. It was a nice idea.

footnotes/models.py
```diff
@@ -1,5 +1,35 @@
...
+    footnote = models.ForeignKey(Footnote, on_delete=models.PROTECT)
+    footnote_type = models.ForeignKey(FootnoteType, on_delete=models.PROTECT)
+    description = models.TextField()
```
simonwo, line 27: Presumably this should not be `NULL`?

footnotes/views.py
andyhd: I resolved those issues, of course, and I like this approach, but I am wary...

common/serializers.py
simonwo: End dates can be unspecified… does this handle that? (is that just `MAX_DATE`?)

footnotes/models.py
simonwo: Footnote ID length must be specifically 5 _or_ 3 digits – can we validate that?

footnotes/models.py
simonwo, line 24: So we've not included languages as a concept, which I know we've talked about. Do we have a plan for how we will hold EU data accurately without them?

footnotes/models.py
simonwo: It also needs to be limited to 500 characters as per the TARIC spec.

footnotes/models.py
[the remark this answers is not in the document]
simonwo: Footnote type ID can be 2 or 3 characters from A-Z – can we validate that?

common/urls.py
```diff
@@ -1,8 +1,10 @@
+from django.urls import include
 from django.urls import path
 
 from . import views
 
 
 urlpatterns = [
     path("", views.index, name="index"),
+    path("api-auth/", include("rest_framework.urls")),
```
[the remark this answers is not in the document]
andyhd: The acceptance criteria say "Given I am logged in with a valid user". This is a temporary measure to meet the criteria, with the knowledge that we will implement API tokens (or similar) and integrate with DIT SSO at a later date.

footnotes/migrations/0002_footnotes_and_footnote_types.py
```diff
@@ -0,0 +1,151 @@
...
+        migrations.AddField(
+            model_name="footnote",
+            name="national",
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AddField(
+            model_name="footnote",
+            name="valid_between",
+            field=django.contrib.postgres.fields.ranges.DateTimeRangeField(
+                default=DateTimeTZRange(BREXIT_DATE, MAX_DATE)
```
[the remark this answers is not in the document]
andyhd: This is a default value, only used if a footnote is entered with no start date. EU data imports would have a start date already set.

footnotes/models.py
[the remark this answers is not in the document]
andyhd: I will add that validation in. It should probably be in the acceptance criteria.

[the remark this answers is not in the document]
andyhd: That blank should not be there, I'll remove it.

footnotes/views.py
NDevox: Yeah me too. Lets keep it for now - no reason to do more work. But with this in mind.

footnotes/models.py
```diff
@@ -1,5 +1,35 @@
...
+class NationalMixin(models.Model):
+    national = models.BooleanField(default=False)
+
+    class Meta:
+        abstract = True
+
+
+class FootnoteType(ValidityMixin, NationalMixin):
+    id = models.CharField(primary_key=True, max_length=2, null=False, blank=True)
+    application_code = models.IntegerField()
```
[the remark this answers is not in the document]
andyhd: This is just a stub, copied from the existing data model. As I said in response to @NDevox's comments, copying the existing data model was the wrong approach - I'll remove this.

footnotes/models.py
andyhd: Can you give me a link to the validation details? There's nothing in the ticket

common/serializers.py
andyhd: Yes, the `DateTimeRangeField` validates that the start <= end. End dates can be `None`.

footnotes/models.py
andyhd: They are not in scope for this ticket

footnotes/models.py
andyhd: The default is that fields cannot be `NULL`

footnotes/models.py
andyhd: As I understand it, the footnote ID is only 3 alphanumeric characters and is usually displayed with the footnote type ID prefixed, to make 5 characters.

common/urls.py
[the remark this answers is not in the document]
simonwo: Ah yes, understood.

footnotes/migrations/0002_footnotes_and_footnote_types.py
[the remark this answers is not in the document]
simonwo: Ah, apologies, I assumed it was a validation range.

footnotes/models.py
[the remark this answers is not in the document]
simonwo: Yeah agreed, we haven't got all the detail into the acceptance criteria yet. We'll keep improving this.

common/serializers.py
simonwo: Ace. Some other points to note, start dates cannot be unspecified AFAICS, and also the XML only specifies a `YYYY-MM-DD` – might a `DateTimeRange` be a choice for that?

andyhd: Yeah, maybe a `DateRange` is all that is required here. I am reusing @NDevox's `ValidityMixin`, which uses a `DateTimeRange`, and I didn't change it in case it affected his stuff. It provides a superset of the required functionality, so I don't think it matters that much. The `DateTimeRange` does also require a start date, so that's already covered.

NDevox: TBF I don't think the extra granularity is an issue.

footnotes/filters.py
```diff
@@ -0,0 +1,38 @@
...
+
+from django_filters import rest_framework as filters
+
+from footnotes.models import Footnote
+
+
+log = logging.getLogger(__name__)
+
+COMBINED_FOOTNOTE_AND_TYPE_ID = re.compile(
+    r"(?P<footnote_type_id>[A-Z0-9]{2})(?P<footnote_id>[A-Z0-9]{3})"
```
NDevox, line 12: I forgot the footnote ID was a concatenation like this.

ugh OK.

footnotes/models.py
```diff
@@ -1,5 +1,17 @@
...
+
+class FootnoteType(ValidityMixin):
+    footnote_type_id = models.CharField(unique=True, max_length=2)
+    description = models.CharField(max_length=500)
+
+
+class Footnote(ValidityMixin):
+    footnote_id = models.CharField(unique=True, max_length=5)
+    footnote_type = models.ForeignKey(FootnoteType, on_delete=models.PROTECT)
+    description = models.CharField(max_length=500)
```
NDevox, line 14: Some debate on this being a textfield, not important for the current PR though.

requirements.txt
```diff
@@ -3,14 +3,18 @@ black
...
 django-health-check
 django-treebeard
 django-webpack-loader
 django_extensions
+djangorestframework
 factory-boy
 git+https://github.com/alphagov/govuk-frontend-jinja.git
 gunicorn
 jinja2
+markdown
```
NDevox, line 17: Might be worth deleting this?

footnotes/filters.py
simonwo, line 12: So the XSD pattern for footnote types is: `[A-Z]{2}|[A-Z]{3}|[0-9]{2}` so oddly, only two digits are allowed, but two or three letters.
And for footnote IDs is: `[0-9]{5}|[0-9]{3}` so 3 or 5 digits only.
Could we reflect that here?

footnotes/models.py
```diff
@@ -1,5 +1,17 @@
 from django.db import models
 
+from common.models import ValidityMixin
 
-class Footnote(models.Model):
-    description = models.TextField()
+
+class FootnoteType(ValidityMixin):
+    footnote_type_id = models.CharField(unique=True, max_length=2)
```
simonwo, line 7: Length can go from 2 to 3 chars.

footnotes/validators.py
```diff
@@ -0,0 +1,24 @@
+"""
+Validators for footnotes
+"""
+import re
+
+from rest_framework import serializers
+
+
+FOOTNOTE_ID = re.compile(r"^[A-Z0-9]{3}$")
+FOOTNOTE_TYPE_ID = re.compile(r"^[A-Z]{2,3}$")
```
simonwo, line 10: See other comment for amended regexes.

PR #19, opened 2020-05-18
TP-118 Add WorkBaskets for grouping tariff edits

andyhd:
## Why
A change to the tariff may often involve creating or modifying several different records
(measures, quotes, regulations, etc).

Tariff Managers may collaborate and split the work among themselves, so that a single
conceptual change to the tariff may be composed of several smaller edits by multiple
users.

Changes must be approved before they are allowed to be published, and it is helpful for
reviewers to understand the reason for a change.

## What
Screenshot 2020-06-05 at 11 26 18

This PR adds WorkBaskets to group related tariff edits and provide metadata for
explaining the reason for a change and for recording the approval and publishing
workflow.

The approval/publish workflow is modeled as a Finite State Machine, using [django-fsm](https://github.com/viewflow/django-fsm).

WorkBaskets are assigned a Transaction ID when they are submitted for approval.

## See
https://uktrade.atlassian.net/browse/TP-118


workbaskets/models.py
```diff
@@ -0,0 +1,167 @@
...
+    author = models.ForeignKey("auth.User", on_delete=models.PROTECT)
+    approval = models.ForeignKey(ApprovalDecision, on_delete=models.PROTECT, null=True)
+
+
+class WorkBasketItem(TimestampedMixin):
+    "A WorkBasketItem represents draft changes to records (measures, quotas, etc)"
+    workbasket = models.ForeignKey(
+        WorkBasket, on_delete=models.CASCADE, related_name="items"
+    )
+    author = models.ForeignKey("auth.User", on_delete=models.PROTECT)
```
NDevox, line 150: As above

workbaskets/models.py
```diff
@@ -0,0 +1,167 @@
...
+    approval = models.ForeignKey(ApprovalDecision, on_delete=models.PROTECT, null=True)
```
NDevox, line 151: if this is connected to a workbasket, this should be the same as `workbasket.approval`. Do we need further approvals for each individual part of the workbasket?

workbaskets/fixtures/test_workbaskets.json
```diff
@@ -0,0 +1,66 @@
+
```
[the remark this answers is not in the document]
NDevox: Actually do we even use these? Tests below use the factories.

workbaskets/models.py
```diff
@@ -0,0 +1,167 @@
...
+    object_id = models.PositiveIntegerField(null=True)
+
+    "The diff field encodes the changes to be applied to the record, for audit logging"
+    diff = JSONField(default=dict)
+
+    """
+    The draft field encodes the record data with the changes applied, for efficiency
+    when previewing the changes
+    """
+    draft = JSONField(default=dict)
```
NDevox, line 164: Query on how much a purpose it serves to have both of these.

workbaskets/fixtures/test_workbaskets.json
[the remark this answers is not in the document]
andyhd: These fixtures are only for convenience when running locally

footnotes/jinja2/includes/footnotes_list.jinja
```diff
@@ -0,0 +1,28 @@
+
```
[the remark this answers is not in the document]
andyhd: Yes, the workbasket detail page uses this template to list footnotes being edited

footnotes/jinja2/footnotes/list.jinja
```diff
@@ -39,33 +39,7 @@
     </form>
```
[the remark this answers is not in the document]
andyhd: This has only changed because the footnote list code has been factored out into a partial template

workbaskets/models.py
```diff
@@ -0,0 +1,167 @@
...
+    @transition(
+        field=status,
+        source=WorkflowStatus.SENT_TO_CDS,
+        target=WorkflowStatus.CDS_ERROR,
+    )
+    def cds_error(self):
+        pass
+
+
+class ApprovalDecision(TimestampedMixin):
```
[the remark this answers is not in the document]
andyhd: I'm all for getting rid of this class and replacing with a foreign key to User (which is what I had on the first pass at this), but I think the explicit rejection indication is necessary to notify the Tariff Manager that their submitted workbasket needs attention. In fact, I think this ApprovalDecision model should also have a reason given, otherwise most of the benefit (of not having to follow-up with the reviewer out-of-band) is lost.

workbaskets/models.py
andyhd: There is a feature for partial approval/rejection of a workbasket in the prototype. I think we could probably drop it, at least for launch.

workbaskets/models.py
andyhd: I agree - I thought it might be an optimisation. The TAP model has `original_data`, `new_data` and `changed_values` JSON fields.

workbaskets/models.py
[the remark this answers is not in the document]
NDevox: So I have two thoughts on this... Which maybe warrant wider discussion.

My first is whether this is necessary. We've already heard the original system was slow and over complicated, so starting with a more streamlined system may be ideal. Building review systems before I've erred on the "humans first, machines second" side of things - in the sense that people can talk to each other and that conversation doesn't have to be recorded. E.g. a tariff manager can email the workbasket owner with changes, or go and speak to them. This makes the system simpler from our standpoint.

The second thought is maybe this is too simple for what we're doing and we really do want to track these conversations. But I'm not convinced that the comments that went into a workbasket need to be auditable, we just need to know who made it and who approved it.

workbaskets/models.py
NDevox: I would be in favour of dropping this for now.

Honestly if we have time I'm pro building an interface like this - but if we can keep things streamlined for now it helps.

workbaskets/models.py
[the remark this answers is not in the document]
andyhd: That's a good point about simplifying the process. If we don't need to track any metadata about approval and assume the submitter and reviewer can resolve things offline, then you're right and we can lose this model. I assumed this is going to be required since explicit rejection is in TAP and the prototype even after they simplified the workflow. But perhaps this is a question for @simonwo?

workbaskets/fixtures/test_workbaskets.json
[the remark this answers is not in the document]
andyhd: Am I abusing fixtures? Perhaps I should remove them.

workbaskets/models.py
andyhd: Yeah, I'll drop this for now.

workbaskets/fixtures/test_workbaskets.json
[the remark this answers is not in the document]
NDevox: 😆

I'm unsure how I feel about it so was deferring comment.

We can keep them in, although maybe it's worth .gitignoring them or something?

I'm usually pro letting devs have their own setups so we can create data how we want etc.

But tbf this doesn't mean I won't ask to borrow your fixtures down the line. It's just non-critical to the app.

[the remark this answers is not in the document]
andyhd: In fact, I already had a test fail because they were applied to the test database. (Although the test assumed an empty db, which was wrong too).

workbaskets/models.py
simonwo: Agreed, that partial accept feature raises a lot of questions, and it's not clear that the complexity it brings in is justified by a real need. Let's keep it simple until later in the year.

workbaskets/models.py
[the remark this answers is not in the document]
simonwo: The use case for advanced conversations is basically replacing _all_ out of band discussion – if people are using the feature and then have to resort back to e-mail anyway, we lose any benefit from having written the feature. In short, can we do better than an e-mail thread? 

So I think I'm agreeing that the two options are either:

1. No conversation support at all – there is just approval and rejection, we must log who approves the workbasket, but that's it. No rejection reason. 
2. Full conversation support Github style – there is a full conversation, including rejections, seeing intermediate data changes, notifications etc.

I think 2 is much too much to worry about for now and needs to be guided really tightly by user research, so let's just do 1. 

There are some bonus points we can pick up with some really simple stuff based on UR too e.g. something like a button that the rejector can press after rejection saying "e-mail the workbasket author" and it's just a `mialto:` link with a prefilled subject. 

A final question is – once approved, is it possible to go backwards? Can a single workbasket be approved twice under any circumstances?

simonwo:
> @NDevox:
> I do have an internal query around using a generic FK system to link workbasketitems vs using a custom JSON structure that could be held within the workbasket object itself.
> 
> This way, with the workbasket items is more secure, more auditable and likely less work for us tbh. The custom JSON however I reckon would be more performant.

There's loads of trade-offs to think carefully about here:
* Would doing things in JSON give us all the validation constraints that we currently implement in the DB e.g. field types, lengths, uniqueness? The FK approach would obviously implement all of these by virtue of them being in the same schema.
* How important do we think it will be to (at a later date) have "partially valid" entities? E.g. it may be that someone who is working on a draft workbasket will want to leave data in an invalid state because it's a work in progress. With the JSON model, we do have the flexibility to do that and implementing it in the FK model later would be more challenging. This is a general feature of data editors but not seen any clamour for it yet, as by the time things come into TAP they've generally already been decided.
* How hard will it be to mix workbasket items and live data? One requirement is that should be no(t much) extra dev overhead to include data in workbaskets alongside live data in every UI view. I can be persuaded that there is some model magic that could make that work with a mixin or something, such that when we query a model with an active workbasket it also includes items from that workbasket automatically, but it would be good to understand it before we commit to it. 
* I'm concerned that with JSON, performance would be compromised on database queries that will need to query the data to do ordering or grouping in the db, although I'll openly admit I've got no idea how performant the Postgres support is for native JSON fields for this stuff and could believe they do something behind the scenes to make these sorts of queries performant. Can we find any info on this?

> e.g. if we query for changes since x we only need to get a list of workbaskets approved since then with the JSON approach. With this approach we need to query for those workbaskets, and then all of their workbasket items as well.

We should not forget about the append-only log model here – I know we're still in progress on that. But we would get a lot of features for free if we separate out the workbasket ID from the "transaction ID" of live data – the two are directly correlatble, but not the same. Workbasket IDs are arbitrary, whereas transaction IDs are sequential. That makes it easy to say, "get me all of the changes since x" if x is a transaction ID, because then it's just `WHERE transaction_id > x`. Workbasket IDs don't have this property though, because they can't be guaranteed to be approved in order…


Great progress on thinking about both of these. Let's talk about them at 4pm?

workbaskets/models.py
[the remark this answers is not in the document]
NDevox: As this is an FK it can't. Would need to be an M2M if we wanted multiple approvals.

The only way this would work is if we wiped the approval before reapproving. So still effectively a single approval.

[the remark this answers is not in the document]
andyhd: @simonwo I think option 1 was actually approved or not approved (yet). Once we have an explicit "rejected" state, we need to a) store that, b) wipe the decision before reapproving, and c) maybe provide a reason for rejection.

andyhd:
I forgot to add the transaction ID field. This would need to be populated just before exporting to CDS.

This model already allows partially valid entities, as the edits are in JSON and do not have to be subject to the same constraints as live data.

Mixing draft entities as JSON fields with live data is admittedly not trivial (eg `Footnote.objects.all()` can't include a new footnote you are working on). The alternative I considered is to have an `is_draft` field (or rather, an optional `workbasket_id` field) on every record and store draft entities mixed in with live data - but then you have to relax the constraints on the whole table to permit partially valid entities. Another alternative is to have a draft table for each entity, with relaxed constraints on the draft table. Then you need to JOIN the tables to query all data including drafts. 

In any case, we can trivially apply the same validation to draft entities (by applying the diffs to a new entity not saved in the db) as to existing entities.

NDevox:
My thinking on mixing the data is essentially to make it so that the system which will run the workbasket changes can run in a transaction that gets rolled back.

That way we can run the changes, serve the data in this state, and then rollback the changes without applying it to the DB. Depending on how we do this it may not be too complex.

My concern on the performance with workbasket items is that a transaction ID will pertain to a workbasket, to get all the workbasket items we'll have to do something like

```sql
SELECT workbasket_items.* 
  FROM workbasket_items
  JOIN workbaskets
    ON workbaskets.id = workbasket_items.id
 WHERE workbaskets.transaction > {some ID}
```

Whereas with JSON this would be

```sql
SELECT * 
  FROM workbaskets
 WHERE transaction_id > {some ID}
```

Honestly before this comment I was about 65-70% leaning towards the current schema. Now I'm closer to about 85%. Performance difference should be minimal whereas the validation gains we get from having a workbasket_items table is huge.

workbaskets/models.py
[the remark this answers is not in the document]
andyhd: I think this is resolved then? Option 2 sounds like a "Would like to have" for later (if at all). I'll remove the `ApprovalDecision` class and just have a foreign key to the approving user.

workbaskets/models.py
```diff
@@ -0,0 +1,147 @@
...
+
+class WorkflowStatus(models.TextChoices):
+    # Newly started, but not yet submitted into workflow
+    NEW_IN_PROGRESS = "NEW_IN_PROGRESS", "New - in progress"
+    # Existing item, already on CDS, being edited but not yet submitted into workflow
+    EDITING = "EDITING", "Editing"
+    # Submitted for approval, pending response from Approver
+    AWAITING_APPROVAL = "AWAITING_APPROVAL", "Awaiting approval"
+    # Was not approved, returned to submitter
+    APPROVAL_REJECTED = "APPROVAL_REJECTED", "Failed approval"
```
NDevox, line 17: Do we need this?

workbaskets/models.py
```diff
@@ -0,0 +1,147 @@
...
+from common.models import TimestampedMixin
+
+
+class WorkflowStatus(models.TextChoices):
+    # Newly started, but not yet submitted into workflow
+    NEW_IN_PROGRESS = "NEW_IN_PROGRESS", "New - in progress"
+    # Existing item, already on CDS, being edited but not yet submitted into workflow
+    EDITING = "EDITING", "Editing"
+    # Submitted for approval, pending response from Approver
+    AWAITING_APPROVAL = "AWAITING_APPROVAL", "Awaiting approval"
```
NDevox, line 15: Do we need this?

workbaskets/models.py
```diff
@@ -0,0 +1,147 @@
...
+        related_name="approved_workbaskets",
+    )
+
+
+class Transaction(TimestampedMixin):
+    """A Transaction is created once the WorkBasket has been sent for approval"""
+
+    workbasket = models.OneToOneField(
+        WorkBasket, on_delete=models.PROTECT, editable=False,
+    )
```
NDevox, line 147: Was thinking this would be a nullable positive integer on the workbasket. But actually this could give us a nicer API - especially with the dates. I like it.

NDevox: An approved flag, I suppose `READY_FOR_EXPORT`, serves the same purpose the positive integer would have as well if we want to avoid joins.

workbaskets/models.py
andyhd: I think yes - we don't want the approvers to have to poll tariff editors to find out when they need to approve workbaskets, or to mistakenly approve one before it is ready. This provides a simple way to list workbaskets-to-approve for approvers.

workbaskets/models.py
```diff
@@ -0,0 +1,147 @@
...
+from django_fsm import transition
+
+from common.models import TimestampedMixin
+
+
+class WorkflowStatus(models.TextChoices):
+    # Newly started, but not yet submitted into workflow
+    NEW_IN_PROGRESS = "NEW_IN_PROGRESS", "New - in progress"
+    # Existing item, already on CDS, being edited but not yet submitted into workflow
+    EDITING = "EDITING", "Editing"
```
andyhd, line 13: I'm not sure if we need to differentiate between `NEW` and `EDITING`

workbaskets/models.py
andyhd: I'm not sure. Potentially not, as we've discussed, because a lack of approval is the same thing as a rejection. Except if we want to make it obvious in the UI that the workbasket needs attention.

workbaskets/models.py
```diff
@@ -0,0 +1,147 @@
...
+    # Edited item approved and scheduled for sending to CDS, existing version will be updated
+    AWAITING_CDS_UPLOAD_OVERWRITE = (
+        "AWAITING_CDS_UPLOAD_OVERWRITE",
+        "Awaiting CDS upload - overwrite",
+    )
+    # Delete instruction approved and scheduled for sending to CDS
+    AWAITING_CDS_UPLOAD_DELETE = (
+        "AWAITING_CDS_UPLOAD_DELETE",
+        "Awaiting CDS upload - delete",
+    )
```
andyhd, line 36: These states are used internally in the old TAP - not sure if we should keep them.

andyhd, line 36: These states are used internally in the old TAP - not sure if we should keep them.

workbaskets/models.py
```diff
@@ -0,0 +1,147 @@
...
+    # Sent to CDS, waiting for response
+    SENT_TO_CDS = "SENT_TO_CDS", "Sent to CDS"
+    # Delete instruction sent to CDS, waiting for response
+    SENT_TO_CDS_DELETE = "SENT_TO_CDS_DELETE", "Sent to CDS - delete"
```
andyhd, line 40: Not sure if we need to differentiate - or if we'll be sending deletions

workbaskets/models.py
simonwo: Let's leave it in just for that use case, and make sure we test that showing "approcal rejected" will be better UX than just mysteriously "awaiting approval".

workbaskets/models.py
simonwo: Yeah I wasn't clear on why these are different. Well, the comment sort of makes it clear, `EDITING` is about something "already on CDS". 

So I've just checked out the original docs and `EDITING` is very much a "post-CDS" step. So the intention is that `NEW_IN_PROGRESS` covers everything before approval, and `EDITING` is meant to cover things being edited after rejection. We've decided not to go that route – but we still need to firm up our understanding of what CDS expects here.

workbaskets/models.py
simonwo: Yeah, we should check out when these get used.

andyhd:
Merging for now. I expect we'll come back and tweak the states later.

workbaskets/models.py
```diff
@@ -0,0 +1,147 @@
...
+    approver = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.PROTECT,
+        editable=False,
+        null=True,
+        related_name="approved_workbaskets",
+    )
+
+
+class Transaction(TimestampedMixin):
```
simonwo, line 142: Cool! Do we need to add some sort of step to `approve` which actually makes this?

workbaskets/serializers.py
```diff
@@ -0,0 +1,14 @@
+from django.contrib import auth
+from rest_framework import serializers
+
+from common.serializers import UserSerializer
+from workbaskets import models
+
+
+class WorkBasketSerializer(serializers.ModelSerializer):
```
simonwo, line 8: Do we need a `TranasctionSerializer`?

workbaskets/models.py
NDevox: Definitely. But maybe think about this implementation in a coming sprint.

workbaskets/serializers.py
NDevox: As above with the approve mechanic.

andyhd: Oops - I've added the serialization for the transaction ID here https://github.com/uktrade/tamato/pull/23

PR #20, opened 2020-05-18
TP-130 Ensure Footnote and Footnote Type business rules

andyhd:
## Why
There are a number of TARIC processing rules pertaining to Footnotes (FO1 - FO16) and
Footnote Types (FOT1 - FOT3).

## What
This change adds database constraints and model validation to ensure the relevant
processing rules are applied.

## See
* [TARIC3 IDS Processing Rules](https://uktrade.atlassian.net/wiki/download/attachments/521273352/TARIC3_IDS_Processing%20Rules_v0.49%20reduced.docx?version=2&modificationDate=1533044571627&cacheVersion=1&api=v2)
* https://uktrade.atlassian.net/browse/TP-130


footnotes/models.py
```diff
@@ -13,10 +14,25 @@ class FootnoteType(ValidityMixin):
...
 
     class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["footnote_id", "footnote_type"], name="FO2",
+            ),
+        ]
         ordering = ["footnote_type__footnote_type_id", "footnote_id"]
+
+    def clean(self):
```
NDevox, line 30: `clean` isn't automatically invoked on `save`. We will need to override the `save` method to run `full_clean` for this to be run as well.

footnotes/tests/test_business_rules.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def test_fo3(validity_range):
+    "The start date must be less than or equal to the end date"
+    range, expected = validity_range
+    with raises_if(django.db.utils.DataError, expected):
+        factories.FootnoteFactory.create(valid_between=range)
+
+
+@pytest.mark.skip()
```
NDevox, line 97: Need a reason for this one

footnotes/models.py
andyhd: Yes, I know. I think explicitly calling `full_clean` is probably better than overriding `save`, though.

footnotes/tests/test_business_rules.py
andyhd: I think this shouldn't be skipped actually - I need to implement multiple descriptions.

footnotes/models.py
NDevox: We only do that in a test though don't we? or does it not matter for the time being and we'll call it in a view when is appropriate.

footnotes/tests/test_business_rules.py
NDevox: 👍

footnotes/models.py
andyhd: The UI will likely use a ModelForm which calls `full_clean` in the validation of the form.
Perhaps this validation code should be factored out into a validator function/class that can be used both here and in the API Serializer.

NDevox: Ok this makes sense to me.

simonwo: Key question from me is, is it possible to forget to call this and accidentally skip the validation?

andyhd: It is always possible to forget to do things.

andyhd:
In what context will footnote type ids have trailing spaces? Wouldn't the XML parser strip those?

andyhd:
Oops, forgot to add the application code field 👍

simonwo:
> In what context will footnote type ids have trailing spaces? Wouldn't the XML parser strip those?

It's more in the context of creating new footnote types ourselves. If a tariff manager or a migration script specifies a type ID as the same as another one but with a trailing space, when that goes into CDS they'll be treated the same which is wrong. Not likely, but we should have some validation around that to make sure.

andyhd:
The UI form code should strip whitespace and so should the API (Postel's law), but you're right, the footnote type id should have a `RegexValidator` too. 👍

footnotes/models.py
simonwo: Ha, well, yes. What I'm really asking is, can someone write to the db and skip this validation, and if so can we avoid that?

andyhd: The users will only be able to write to the database via the UI or the API, so those routes are covered. If someone is accessing the database programmatically, there isn't much we can do beyond overriding the `save` method to enforce the application-level validation, but they shouldn't be doing that. When we write the data migration scripts, we should just remember to explicitly call `full_clean`. The `save` method is not called by Django in some circumstances, so explicit is better than implicit.

footnotes/models.py
```diff
@@ -1,22 +1,77 @@
...
+        "Meursing Heading",
+        "Dynamic footnote",
+    ],
+)
+
+
 class FootnoteType(ValidityMixin):
     footnote_type_id = models.CharField(
-        unique=True, max_length=3, validators=[validators.valid_footnote_type_id]
+        unique=True, max_length=3, validators=[validators.FootnoteTypeIDValidator]
```
NDevox, line 32: This can't be unique if we're using the append only implementation.

andyhd: OK, but they do have to be temporally unique. I think we need to implement a temporal database (maybe with something like https://github.com/arkhipov/temporal_tables or https://github.com/nearform/temporal_tables), adding an additional `DateTimeRangeField` for when a FootnoteType (or any model) is valid. Then we can have a uniqueness constraint across `footnote_type_id` and this new validity period.

andyhd: No need to add the complexity of temporal tables here - we can just move the business rule from a unique constraint to a validation method which needs to be called before saving the model.

PR #21, opened 2020-05-26
Rough idea of how the tracking will work.

NDevox:
This is a rough concept of how introducing polymorphic models for tracking will work.

It's easiest to redo all the migrations at this point as we're yet to commit to a data model and changing the base model, especially in this kind of way is a bit gnarly with pre-existing migrations.

Most of the complexity will arise from using an MPTT system combined with the polymorphic models. However, despite django-polymorphic leaning heavily into metaclasses, django-treebeard does not use metaclasses at all. Therefore combining them seems simpler than expected. Most of the conflicts arise in the querysets and managers which is resolved in this example.

The other main point of contention I have with this is the data stored on the `TrackedModel`, or the data not stored on it. Personally I would prefer to take all fields from the `ValidityMixin` into the `TrackedModel` - however this disables the ability to properly use exclusion constraints with the validity date range. Currently a compromise is used to store the live and draft flags on the `TrackedModel`, which makes it easier to query things from the `TrackedModel`. But it might be preferable to store all the fields in `ValidityMixin` instead and simply use `TrackedModel` as an intermediate M2M table for `WorkBaskets`.

This branch is not useful until `WorkBaskets` are merged in as well - as `TrackedModel` will need to link to them.


simonwo:
Some thoughts:
1. Does this include the concept of the transaction ID? Or is that to go somewhere else?
2. Does the `TrackedModel` create a new database table, and link off to other tables? What's the performance of this like – e,g, if we want to get all of the _N_ objects changed in a transaction, does this require _N+1_ queries? Does it make certian things slow or fast? How does it compare with having a `TrackedMixin` and storing columns on each model table?
3. I think we're overloading the `ValidityMixin` to do two things that need to be separate. The `validity_start_date` and `validity_end_date` found on most models represents the date that entities are live from an operational border perspective. There are a separate set of dates that represent when an entity is live from a TAP database perspective. The difference is that if we create a new version of e.g. a footnote, and only modify its description, it's `validity_*_date` columns should remain unchanged but it's TAP tracking dates should be updated (e.g. the old version now has an end date and the new version has a start which is `now()`). Note that I think the tracking dates can be conceptual and don't necessarily need to be represented as columns on every entity, but there might be a date column on the `TrackedModel` for this?

NDevox:
@simonwo 

> Does this include the concept of the transaction ID? Or is that to go somewhere else?

Not right now, I'm of the opinion that this should go on the workbasket model, so am not touching until we come to an equilibrium with Andys work.

> Does the TrackedModel create a new database table, and link off to other tables? What's the performance of this like – e,g, if we want to get all of the N objects changed in a transaction, does this require N+1 queries? Does it make certian things slow or fast? How does it compare with having a TrackedMixin and storing columns on each model table?

This is the key complexity of the workbasket system - it needs to be able to link out to a multitude of objects of varying types which is not something easily handled in an RDBMS (a generic foreignkey).

The short answers are:
- Yes it does create a new database table which links to the other tables.
- It does require N+1 queries, *however N does not represent the number of objects in this workbasket*, it represent the _number of types_ in a workbasket - i.e. if we have a workbasket changing 20 commodity codes and 5 regulations, the number of queries will be 3. If we have a workbasket changing 1 commodity code, 1 regulation, 1 footnote and 1 footnote type the number of queries will be 5. There isn't really a simpler and more optimised way of doing this that I can think of..
- Compared to a TrackedMixin there is significant code complexity introduced as there is no simple way of telling the WorkBasket what other models it is connected to - we'd have to have the WorkBasket inspect every possible option which will require a query for _every_ single type of object which can feasibly be edited (this also increases the possibilities of bugs being introduced). This solution allows the WorkBasket to only care about a single model which automatically knows all the models it is connected to.

> I think we're overloading the ValidityMixin to do two things that need to be separate. The validity_start_date and validity_end_date found on most models represents the date that entities are live from an operational border perspective. There are a separate set of dates that represent when an entity is live from a TAP database perspective. The difference is that if we create a new version of e.g. a footnote, and only modify its description, it's validity_*_date columns should remain unchanged but it's TAP tracking dates should be updated (e.g. the old version now has an end date and the new version has a start which is now()). Note that I think the tracking dates can be conceptual and don't necessarily need to be represented as columns on every entity, but there might be a date column on the TrackedModel for this?

I did put a validity range on the TrackedModel - this created some complexity around selecting for objects within a range (as the date range is now stored on the TrackedModel table and not the model we are interested ins table). So there are some unresolved issues here about where we put dates and how to handle them, but I think it warrants further discussion about how we want to handle the various date concepts in the system (and if we need to in some cases). I don't think it needs to block this current work as it can be easily added in down the line.

common/models.py
```diff
@@ -12,7 +41,44 @@ class Meta:
...
+    live = models.BooleanField(default=False)
+    predecessor = models.OneToOneField(
+        "self",
+        on_delete=models.PROTECT,
+        null=True,
+        blank=True,
+        related_name="successor",
+    )
+
+    def new_draft(self, workbasket, save=True, **kwargs):
```
NDevox, line 61: FYI @simonwo @andyhd this is a primitive concept on making new drafts of a specific objects. Works fine for anything without much complexity. Commodity codes will require some extra work around this.

Will also likely want to override the `save` method so it either forces the creation of a new draft on save, or throws and error if the object isn't in draft state.

And also overriding the `update` method on the queryset to also make new drafts instead of updating current rows... This is for the Append Only implementation though.

common/models.py
```diff
@@ -1,5 +1,117 @@
...
+        Latest is defined as the last row relating to the history of an object.
+        this may include current operationally live rows, it will often also mean
+        rows that are not operationally live yet but may be soon. It could also include
+        rows which were operationally live, are no longer operationally live, but have
+        no new row to supercede them.
+
+        If done from the TrackedModel this will return the latest objects for all tracked
+        models.
+        """
+        return self.filter(predecessor__isnull=True)
```
simonwo, line 28: To be fixed as discussed?

common/models.py
simonwo: Is part of the answer for each model to have it's own `clone` method which can handle the model-specific work?

andyhd: Should this be an override of `TrackedModel.objects.create()`?

NDevox: Maybe `.save`? We don't want it to interfere with brand new models however.

Could probably override `.save` to check for a PK and then run this method instead if a PK exists?

NDevox:
This pull request has been superceded by #25

PR #22, opened 2020-06-04
TP-123 Data model for regulations

andyhd:
Elements of the tariff are added, modified and removed by legal acts known as regulations. Regulations are published (by the EU in the Official Journal) and have a number and role type, which may be one of several, including Base regulation, Modification regulation and Abrogation regulation. TAMATO will only create and modify Base regulations, however we need to store historical data.

Screenshot 2020-06-02 at 16 30 24

Regulations are stored in a single table (in contrast to the TARIC data model), with the relationships of the various types of regulation to each other represented as a number of reflexive many-to-many database relations.

A Modification regulation amends a single Base regulation.

A Prorogation regulation extends one or more Base or Modification regulations. Each regulation may be extended to different dates.

A Full Temporary Stop (FTS) regulation suspends one or more Base or Modification regulations. Each regulation is suspended to a single effective end date. A Partial Temporary Stop regulation is not recorded directly in the regulations table, but instead is referenced by the measures that are suspended.

An Abrogation regulation terminates a (Base/Prorogation/Modification/FTS) regulation. An Explicit Abrogation regulation specifies an effective date that the affected regulation is terminated. A Complete Abrogation regulation does not specify a date.

See
* [TP-123 Data model for regulations](https://uktrade.atlassian.net/browse/TP-123?atlOrigin=eyJpIjoiNGJhZWYzMGFmOWI5NGEzYmJlMzgzOWMyMGE5MTM1ZWYiLCJwIjoiaiJ9)
* [TP-133 Browse and find regulations](https://uktrade.atlassian.net/browse/TP-133?atlOrigin=eyJpIjoiZjY0NTg2MDEwNjA0NDRkNmFiMjRjZWY0MWU5OGFlMjgiLCJwIjoiaiJ9)
* [TP-138 Core regulation data - Tab 1](https://uktrade.atlassian.net/browse/TP-138?atlOrigin=eyJpIjoiZTUyZjNjNTVjMDQ1NDA5YzkxYzI0YmNhZjA0NzkzODMiLCJwIjoiaiJ9)


regulations/models.py
```diff
@@ -0,0 +1,264 @@
...
+
+    """The regulation number"""
+    regulation_id = models.CharField(
+        max_length=8, editable=False, validators=[RegulationIDValidator()],
+    )
+
+    official_journal_number = models.CharField(max_length=5, null=True)
+    official_journal_page = models.CharField(max_length=4, null=True)
+    published_at = models.DateField(null=True)
+    information_text = models.CharField(max_length=500, blank=True, null=True)
```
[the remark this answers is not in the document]
simonwo: As a consequence, we'll need to validate that 

```python
len(url) + len(public_identifier) + len(information_text) <= 500
```

so that it will fit into the XML field.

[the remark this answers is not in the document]
andyhd: ah, that's where the public identifier and url come from

regulations/models.py
```diff
@@ -0,0 +1,264 @@
...
+        choices=RoleType.choices, default=RoleType["Base"],
+    )
+
+    """The regulation number"""
+    regulation_id = models.CharField(
+        max_length=8, editable=False, validators=[RegulationIDValidator()],
+    )
+
+    official_journal_number = models.CharField(max_length=5, null=True)
+    official_journal_page = models.CharField(max_length=4, null=True)
```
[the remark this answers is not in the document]
andyhd: Is it a number or a digit string? I assume since it's a page number then it is a number. I can either add a RegexValidator to check for 4 digits, or change to a IntegerField with a MaxValueValidator of 9999. Not sure which is better - IntegerField might save a few bytes per record, I suppose.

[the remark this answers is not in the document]
simonwo: it is specified as an `int` (whereas the `official_journal_number` is a `string`) so we might as well use an `IntegerField` I guess.

regulations/models.py
```diff
@@ -0,0 +1,267 @@
...
+    If approved, all measures apply.
+    Once the actual regulation is known, it replaces the draft regulation and all
+    measures are moved from the draft to the actual regulation.
+    It is possible for a draft regulation to be replaced by multiple actual regulations,
+    each one partially replacing the draft.
+    """
+    approved = models.BooleanField(default=False)
+
+    """The code which indicates whether or not a regulation has been replaced."""
+    replacement_indicator = models.PositiveIntegerField(
```
NDevox, line 106: This is an interesting sub system on top of the tracking - where it seems like we're editing regulations by adding relations - but it's more like adding regulations on top of regulations.

When is this replacement indicator changed? Does it make sense in the context of the tracking system or do we need to view them entirely separately?

regulations/models.py
```diff
@@ -0,0 +1,267 @@
...
+        choices=ReplacementIndicator.choices,
+        default=ReplacementIndicator["Not replaced"],
+    )
+
+    # Base, Modification and FTS regulations have validity periods and effective end
+    # dates
+    valid_between = DateTimeRangeField(null=True)
```
NDevox, line 113: validity mixin?

regulations/serializers.py
```diff
@@ -0,0 +1,144 @@
...
+    suspends = NestedRegulationSerializer(many=True)
+    suspensions = NestedSuspensionSerializer(many=True)
+    terminates = NestedRegulationSerializer(many=True)
+    terminations = NestedTerminationSerializer(many=True)
+    replaces = NestedRegulationSerializer(many=True)
+    replacements = NestedReplacementSerializer(many=True)
+
+    class Meta:
+        model = models.Regulation
+        fields = [
```
NDevox, line 120: `__all__`?

regulations/tests/bdd/test_regulations.py
```diff
@@ -0,0 +1,62 @@
...
+# def footnote_details(client, footnote_NC000):
+#     return client.get(reverse("footnote-detail", kwargs={"pk": footnote_NC000.pk}))
+
+
+# @then("a summary of the core information should be presented")
+# def footnote_core_data(footnote_details):
+#     result = footnote_details.json()
+#     assert {"description", "id", "valid_between"} <= set(result.keys())
+#     assert "footnote_type" in result
+#     assert {"description", "id", "valid_between"} <= set(result["footnote_type"].keys())
```
NDevox, line 62: Remove dead code?

regulations/tests/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+from django.core.management import call_command
+
+
+@pytest.fixture(scope="session")
+def django_db_setup(django_db_setup, django_db_blocker):
+    with django_db_blocker.unblock():
+        call_command("loaddata", "regulation_groups.json")
```
NDevox, line 8: Are these used? Looks like we're using factories instead.

regulations/tests/test_regulations_business_rules.py
```diff
@@ -0,0 +1,159 @@
...
+
+from common.tests import factories
+from regulations import models
+
+
+pytestmark = pytest.mark.django_db
+
+
+@contextlib.contextmanager
+def raises_if(exception, expected):
```
NDevox, line 16: We already have this in the footnotes app.

Should move it somewhere common.

regulations/tests/test_regulations_business_rules.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@pytest.fixture(
+    params=[
+        ("2020-05-18", "2020-05-17", True),
+        ("2020-05-18", "2020-05-18", False),
+        ("2020-05-18", "2020-05-19", False),
+    ]
+)
+def validity_range(request):
```
NDevox, line 34: We already have this in the footnotes app.

Should move it somewhere common.

regulations/validators.py
```diff
@@ -0,0 +1,84 @@
+"""
+Validators for regulations
+"""
+import re
+
+from django.core.exceptions import ValidationError
+from django.core.validators import RegexValidator
+
+
+class RegulationIDValidator(RegexValidator):
```
NDevox, line 10: Why not `regulation_id_validator = RegexValidator(r"(?x)(?P<prefix>C|R|D|A|I|J)(?P<year>\d{2})(?P<number>\d{4})(?P<suffix>[0-9A-Z])")`

regulations/validators.py
```diff
@@ -0,0 +1,84 @@
...
+    (A draft regulation's) flag can only change from 0='Not Approved' to 1='Approved'.
+    """
+
+    return regulation.is_draft_regulation and not regulation.approved
+
+
+def validate_official_journal(regulation):
+    """Official Journal number and page must both be set, or must both be NULL"""
+
+    return bool(regulation.official_journal_number) == bool(official_journal_page)
```
NDevox, line 73: missing `regulation.` from `official_journal_page`

`official_journal_number` can also be 0, not sure if that is an edge case we should consider.

regulations/validators.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def unique_regulation_id_for_role_type(regulation):
+    """ROIMB1
+
+    The (regulation id + role id) must be unique.
+    """
+
+    Regulation = regulation.__class__
```
NDevox, line 82: nice avoidance of circular imports

regulations/views.py
```diff
@@ -0,0 +1,42 @@
...
+
+class RegulationUIViewSet(RegulationViewSet):
+    """
+    UI endpoint that allows regulations to be viewed.
+    """
+
+    renderer_classes = [renderers.TemplateHTMLRenderer]
+
+    def list(self, request):
+        queryset = self.filter_queryset(self.get_queryset())
```
NDevox, line 32: I think the first arg needs to be `request`

common/serializers.py
```diff
@@ -4,7 +4,7 @@
 from rest_framework import serializers
 
 
-class ValiditySerializerMixin(serializers.HyperlinkedModelSerializer):
+class ValiditySerializerMixin(serializers.Serializer):
```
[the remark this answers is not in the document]
andyhd: No, I think that was a typo

common/jinja2/common/index.jinja
```diff
@@ -6,6 +6,9 @@
   <h1 class="govuk-heading-xl">{{ page_title }}</h1>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-one-third">
+      <h2 class="govuk-heading-m">Manage regulations</h2>
+      <a class="govuk-link" href="{{ url('regulation-ui-list') }}">Find and edit
+        regulations</a>
```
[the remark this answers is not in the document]
andyhd: Where do you think it should be?

regulations/models.py
andyhd: This field is from the TARIC data model. I believe it is caching information that is already available by other means. You can determine that a regulation is replaced by the existence of a replacment relation. You could determine that a regulation was partially replaced by querying the measures. The same is true for `stopped` and `effective_end_date` (and maybe even `approved`). But it is probably easier to keep these fields for storing EU data.

regulations/models.py
andyhd: ValidityMixin is not nullable.

regulations/tests/bdd/test_regulations.py
andyhd: oops, missed a test

common/jinja2/common/index.jinja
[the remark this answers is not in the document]
NDevox: I'm an idiot, never mind.

regulations/models.py
NDevox: Yeah this makes sense. And the complexity of the M2Ms does mean that caching them this way simpler is probably simpler.

regulations/models.py
NDevox: Are there regulations without start dates?

regulations/tests/conftest.py
andyhd: These are being used. Regulation groups "must not be modified", so it makes sense to prepopulate the database with the existing ones.

regulations/tests/test_regulations_business_rules.py
andyhd: Yep, I wrote this before the footnotes PR was merged.

regulations/validators.py
andyhd: IMO it reads better to have `regulation_id = models.CharField(validators=[RegulationIDValidator])` - and all this documentation would really fill up the models file.

regulations/validators.py
andyhd: Ah, I didn't consider that when changing `official_journal_page` to an `IntegerField`. It works with `"0"`, of course.

regulations/views.py
andyhd: I'm surprised this works 😖

regulations/models.py
andyhd: Yes, Complete Abrogation, Explicit Abrogation and Prorogation regulations have no validity periods.

andyhd: I suppose alternatively we could use ValidityMixin and set the validity start date to `published_at` and end date to `NULL`.

NDevox: I think what you currently have is better. But maybe flip the comment to explain what doesn't have validity periods etc.

regulations/tests/conftest.py
NDevox: OK fair enough. Should we add them to the migration as well so they get auto inputted to the DB as well? Or do we want to leave them until we migrate the TARIC db across... Might get messy.

Also, if these groups never change, do we want them to be tracked? If not that is fine, but if we do we will need to rethink these fixtures to include the tracking fields.

regulations/validators.py
NDevox: Yeah I was wondering about having that line done in this file, then can just import the var into the models file. It's not a big issue, just gave me a quick stop to think why we're overriding `__init__`.

regulations/views.py
NDevox: 😆

regulations/tests/conftest.py
andyhd: I think we will want them to be tracked - although the existing ones(?) can't be modified, I haven't seen anything that says we can't add more. I think you're right about leaving them to the migration - perhaps I'll remove these fixtures.

NDevox: Unless we start testing without migrations :-D.

TBF with the comm code situation I'm not sure we can test without migrations.

regulations/validators.py
andyhd: Ah, I can just set `regex` instead of calling `super().__init__()`

andyhd: But yes, I could also import the regex string as `REGULATION_ID_PATTERN` or similar.

regulations/views.py
andyhd: Nope, this line is correct 🤷

regulations/validators.py
NDevox: `REGULATION_ID_PATTERN` constant would be pretty acceptable yeah.

Alternative I was thinking would have been something like:

```python
# validators.py
REGULATION_ID_PATTEN = r"(?x)(?P<prefix>C|R|D|A|I|J)(?P<year>\d{2})(?P<number>\d{4})(?P<suffix>[0-9A-Z])"

regulation_id_validator = RegexValidator(REGULATION_ID_PATTERN)

# models.py
from validators import regulation_id_validator


class Regulation(models.model):
    regulation_id = models.CharField(
        max_length=8, editable=False, validators=[regulation_id_validator],
    )

File: regulations/validators.py [hunk 16]
```diff
@@ -88,4 +96,26 @@ def unique_regulation_id_for_role_type(regulation):
...
+    """Information text has a max length of 500 chars, but public_identifier and URL are
+    passed in the same field in the XML, so the total combined length must be less than
+    or equal to 500 chars.
+    """
+
+    if (
+        len(regulation.public_identifier or "")
+        + len(regulation.url or "")
+        + len(regulation.information_text or "")
+        + 2  # XXX assumes a single character delimiter - needs clarification
```
simonwo, line 114: Good catch – the delimiter is a single bar `|`. I guess we should also check for this delimiter in any of these fields… _sigh_.

PR #24, opened 2020-06-08
TP-143: Create the data model and validations for geographical areas

NDevox:
### Why
All trade is carried out between geographical areas, however the
definition of a geographical area spans 3 concepts:

1) A Country
2) A Region (a trading area which is not recognised as a country)
3) A grouping of the above two concepts

Countries and Regions can be members of multiple groups whereas groups
follow somewhat of a hierarchical structure (there is a parent/child
system although this is rarely used).

### What
This commit introduces the data model to cover geographical areas.
As all models linking to a geographical area are naive to what type
of area the area is (country, region or group) all these concepts are
stored in a single table. Meanwhile the group memberships are stored in
a through table so that memberships can be tracked separately of areas
with their own validity periods.

Schema:
Screenshot 2020-06-08 at 17 25 08

### See 
https://uktrade.atlassian.net/browse/TP-143


geo_areas/models.py
```diff
@@ -0,0 +1,114 @@
...
+                expressions=[
+                    ("valid_between", RangeOperators.OVERLAPS),
+                    ("area_id", RangeOperators.EQUAL),
+                ],
+            ),
+            CheckConstraint(
+                name="only_groups_have_parents",
+                check=Q(area_code=1, parent__isnull=False)
+                | (~Q(area_code=1) & Q(parent__isnull=True))
+                | Q(area_code=1, parent__isnull=True),
```
andyhd, line 67: I'm not certain I understand how multiple arguments to a `Q` object are combined - I'm assuming they are ANDed. It might be clearer to separate these out and use `&`.
Also, does this check reduce to
```python
check=Q(area_code=1) | Q(parent__isnull=True)
```
?

geo_areas/validators.py
```diff
@@ -0,0 +1,89 @@
...
+        overall_range[0] <= contained_range[0] or overall_range[1] >= contained_range[1]
+    )
+
+
+def validate_area_description_is_not_null(area):
+    """
+    GA3
+    """
+    if not area.area_description:
+        raise ValidationError("An area must have a description")
```
andyhd, line 47: You can associate the error message with the field by passing a dict to `ValidationError`, eg:
```python
raise ValidationError({"area_description": "An area must have a description"})
```

geo_areas/validators.py
```diff
@@ -0,0 +1,89 @@
...
+    GA19
+    """
+    if not hasattr(area_membership.group, "parent"):
+        return
+    parent = area_membership.group.parent
+
+    if not area_membership.member.groups.filter(group=parent).exists():
+        raise ValidationError(
+            "Members of a Group, where the Group has a parent Group, must also be members of the parent Group"
+        )
```
andyhd, line 89: Do we need to handle groups nested to a depth greater than one?

geo_areas/models.py
NDevox: Yep it works as an `AND` when in the same Q, like a filter statement.

And looking at it... yes I think it does boil down to that check... That is far simpler, thanks!

geo_areas/validators.py
NDevox: Good tip, thanks!

geo_areas/validators.py
NDevox: Yes, but given this check will occur at every membership level there is no need to recurse.

i.e. say we have 3 groups linked as granparent-parent-child. If we try to create a membership on the child it will raise an error for the parent. If we then try to create a membership on the parent it will raise an error for the grandparent.

It's a lazy validation - you may need to go all the way up the tree step by step to get all the errors - but in reality we don't actually have any multi level parent-child relations so it's a rare scenario.

geo_areas/models.py
```diff
@@ -0,0 +1,114 @@
...
+    As a country or region can belong to a group there is a self-referential many-to-many
+    field which is restricted. Yet groups can also have parent groups - in which case
+    measures must of the parent must also apply to a child. To accomodate this there is a
+    separate foreign key for group to group relations.
+    """
+
+    sid = models.PositiveIntegerField()
+    area_id = models.CharField(max_length=6)
+    area_code = models.PositiveSmallIntegerField(
+        choices=validators.AreaCode.choices, validators=[MaxValueValidator(2)]
```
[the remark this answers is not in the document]
NDevox: I am being a bit over secure on this one. Can remove.

geo_areas/models.py
```diff
@@ -0,0 +1,114 @@
...
+    area_description = models.TextField(blank=False, null=False)
+
+    # This deals with countries and regions belonging to area groups
+    memberships = models.ManyToManyField("self", through="GeographicalMembership")
+
+    # This deals with subgroups of other groups
+    parent = models.ForeignKey("self", on_delete=models.PROTECT, null=True, blank=True)
+
+    def save(self, *args, **kwargs):
+        self.clean()
```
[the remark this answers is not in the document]
NDevox: Yep changed

geo_areas/validators.py
andyhd: I guessed it would be unnecessary 👍

geo_areas/models.py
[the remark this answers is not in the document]
andyhd: I was thinking you might end up with multiple error messages or a confusing error message in the API. It doesn't matter in the UI, because you physically can't enter an incorrect choice.

geo_areas/models.py
```diff
@@ -0,0 +1,114 @@
...
+    the 3 types are stored as one for relational purposes.
+
+    As a country or region can belong to a group there is a self-referential many-to-many
+    field which is restricted. Yet groups can also have parent groups - in which case
+    measures must of the parent must also apply to a child. To accomodate this there is a
+    separate foreign key for group to group relations.
+    """
+
+    sid = models.PositiveIntegerField()
+    area_id = models.CharField(max_length=6)
```
simonwo, line 31: The XSD specifies this can only be 2 or 4 digits or capitals:
> `[A-Z0-9]{2}|[A-Z0-9]{4}`

geo_areas/models.py
```diff
@@ -0,0 +1,114 @@
...
+    As a measure does not care to distinguish between a country, region or group,
+    the 3 types are stored as one for relational purposes.
+
+    As a country or region can belong to a group there is a self-referential many-to-many
+    field which is restricted. Yet groups can also have parent groups - in which case
+    measures must of the parent must also apply to a child. To accomodate this there is a
+    separate foreign key for group to group relations.
+    """
+
+    sid = models.PositiveIntegerField()
```
simonwo, line 30: Worth noting that SIDs have a max value of `99999999`.

geo_areas/models.py
```diff
@@ -0,0 +1,114 @@
...
+    area_id = models.CharField(max_length=6)
+    area_code = models.PositiveSmallIntegerField(
+        choices=validators.AreaCode.choices, validators=[MaxValueValidator(2)]
+    )
+    area_description = models.TextField(blank=False, null=False)
```
simonwo, line 33: This has a max length of 500 chars.

geo_areas/models.py
```diff
@@ -0,0 +1,110 @@
...
+
+    group = models.ForeignKey(
+        GeographicalArea, related_name="members", on_delete=models.PROTECT
+    )
+    member = models.ForeignKey(
+        GeographicalArea, related_name="groups", on_delete=models.PROTECT
+    )
+
+    def save(self, *args, **kwargs):
+        self.clean()
```
andyhd, line 87: Should this one be `full_clean()` too?

geo_areas/validators.py
```diff
@@ -0,0 +1,98 @@
...
+
+class AreaCode(models.IntegerChoices):
+    COUNTRY = 0, "Country"
+    GROUP = 1, "Geographical Area Group"
+    REGION = 2, "Region"
+
+
+def validity_range_contains_range(overall_range, contained_range):
+    return (
+        overall_range[0] <= contained_range[0] or overall_range[1] >= contained_range[1]
```
simonwo, line 38: Shouldn't this be an `and`? Both overall bounds need to be beyond or the same as the contained bounds?

geo_areas/models.py
NDevox: oops

geo_areas/models.py
```diff
@@ -0,0 +1,110 @@
...
+    As a country or region can belong to a group there is a self-referential many-to-many
+    field which is restricted. Yet groups can also have parent groups - in which case
+    measures must of the parent must also apply to a child. To accomodate this there is a
+    separate foreign key for group to group relations.
+    """
+
+    sid = models.PositiveIntegerField()
+    area_id = models.CharField(max_length=6)
+    area_code = models.PositiveSmallIntegerField(choices=validators.AreaCode.choices)
+    area_description = models.TextField(blank=False, null=False)
```
simonwo, line 33: I don't think this captures the full capability? The description needs to have a start date of it's own, so that we can start a description at some point in the future. Hence we need to support >= 1 "database live" descriptions – one that is current, and then N more with future start dates.

geo_areas/validators.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+
+def validate_area_description_is_not_null(area):
+    """
+    GA3
+    """
+    if not area.area_description:
+        raise ValidationError({"area_description": "An area must have a description"})
```
simonwo, line 47: How come this isn't handled by the `null=False` condition?

NDevox: A blank string `""` technically is not null.

geo_areas/models.py
NDevox: I was wondering this, and do note we do this in other Footnotes.

Happy to implement that way, although I do wonder how different it is just to create a new instance of the geographical area with the updated description starting in the future? Less things to manage.

geo_areas/validators.py
NDevox: Yes that is true, was initially doing the inverse and forgot to switch the operator.

geo_areas/models.py
NDevox: Perfect, will update

geo_areas/models.py
NDevox: Updated

geo_areas/models.py
NDevox: Done

NDevox:
Tests have been added @andyhd if you are interested.

geo_areas/models.py
simonwo: If you mean creating a new instance with a new SID, then that's a "new object" in the business domain, disconnected from the existing one, and we'd have to migrate all the measures associated with it, by cloning measures and playing with their start dates so that they start when the geo-area description changes. Possible, but sounds like a lot of work.

If you mean creating a new version with the same SID, that will override the previous version and you'd end up taking an area that is currently live and making it start in the future, nullifying it's measures.

Hm, I don't think either of these are that palatable? Unless there's another option I've missed!

NDevox: Second was what I was thinking.

This is a similar problem we're going to have when updating comm codes as well...

But I can see how splitting out the description stops us from having this headache later on.

simonwo: Potentially less of an issue with comm codes if we limit ourselves to just splitting a code into two or more children, as any measures on a parent code will inherit downwards and we'd only need to create any exceptions on the new code. But yeah, in general there's some upheaval involved with that.

NDevox: I've added this now.

One issue however is checking for a description existing. One of the requirements is an area _must_ have a description. However the area has to exist before a description can be created - so in terms of db integrity this demand is impossible at first. I've left it out for now, we can circle back and implement it at a later date IMO.

PR #25, opened 2020-06-12
TP-117: Add system for tracking models and changes.

NDevox:
This Pull Request supercedes #21 

# Why
The project requires all changes to be tracked throughout the lifetime
of the system. It must be easy to query the latest versions of an object,
the currently active (based on a date) version of an object, as well as
to traverse it's entire history.

# What
This commit introduces a TrackedModel base model. This is not an
abstract base model but instead is meant to be inherited concretely
making use of Django's Multi-Table Inheritance mechanism. Combined with
django-polymorphic this creates a polymorphic parent TrackedModel table
which efficiently fetches all the relevant children in any query
performed by the user.

Schema:

Tracked Model Schema

## Note on testing

Turns out testing historical models isn't overly pretty. I'm eager for suggestions on how to clean this up.

I have also foregone testing the tracked system in combination with the hierarchical trees. There is only a couple lines of code which merge these two systems but the amount of testing would likely double the testing code below... I'm open to doing it, just not late on a Friday. May pick up again on Monday.

## ADR

And ADR has been added which is a mildly modified version of the doc in drive. Probably worth re-reviewing it to make sure it matches our thinking.


commodities/models.py
```diff
@@ -3,22 +3,23 @@
...
 
     description = models.TextField()
-    predecessor = models.OneToOneField(
-        "self",
-        on_delete=models.PROTECT,
-        null=True,
-        blank=True,
-        related_name="successor",
-    )
     version = models.PositiveIntegerField()
```
simonwo, line 19: Is this still a thing we need?

common/exceptions.py
```diff
@@ -0,0 +1,6 @@
+class NoIdentifyingValuesGivenError(Exception):
```
simonwo, line 1: Do we need any docs about what these are and how to fix them?

common/tests/test_models.py
```diff
@@ -0,0 +1,193 @@
...
+pytestmark = pytest.mark.django_db
+
+
+def generate_model_history(factory, number=5, **kwargs) -> List:
+    objects = []
+    current = factory(**kwargs)
+    objects.append(current)
+    for _ in range(number - 1):
+        kwargs["predecessor"] = current
+        current = factory(**kwargs)
```
simonwo, line 19: Do we need some convenience methods for this sort of thing? E.g. `current.create_new_version(**kwargs)` or something

regulations/models.py
```diff
@@ -158,35 +157,35 @@ class Regulation(models.Model):
         "Regulation",
         through="Amendment",
         related_name="+",
-        through_fields=("regulation", "amended"),
+        through_fields=("enacting_regulation", "amended"),
```
simonwo, line 160: What's the purpose of this rename? Potentially `amended_regulation` or something might be more appropriate? @andyhd

docs/adr/0004-use-django-polymorphic.md
```diff
@@ -0,0 +1,311 @@
+# 3. Use Django REST Framework
```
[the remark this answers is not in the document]
NDevox: 😆 I'll remove that one...

common/exceptions.py
NDevox: Not convinced we need it here. Normally you'd want everything to be explained in the error message itself (i.e. when this exception is actually thrown).

regulations/models.py
NDevox: Yeah I'm really open to suggestions here, I'm very much not convinced these are the right names.

There's a conflict in the relation names as `TrackedModel` takes on a relation name for any model inheriting it - therefore `regulation` is now taken up by the `TrackedModel`.

Might be an alternative option but I don't know what it is.

andyhd: Enacting regulation is correct, if a little verbose. The second field here would be `amended_regulation`, but I have a change I'm going to make to standardize them all (`amended`, `extended`, etc) to `target_regulation`.

common/tests/test_models.py
NDevox: I suppose this is kind of what `new_draft` is doing, although the logic here is a bit more concise (because we don't care about almost all the fields).

I'm not sure I want to introduce a convenience method just for tests. But it's a good point, as we develop the base model further we'll probably introduce methods which ease this a lot more.

regulations/models.py
NDevox: `target_regulation` sounds a lot better to me.

Shall I wait for that change?

andyhd: no

commodities/models.py
NDevox: good catch. Removing it.

regulations/models.py
NDevox: This change has been implemented.

andyhd: thanks! now i just need to remember why i was going to do it...

PR #26, opened 2020-06-16
TP-148: Add frontend view to browse geographical areas.

NDevox:
# What

It is required for tariff managers to be able to view, browse and search
for geographical areas in order to edit and update them.

# Why

This commit introduces a frontend page for viewing and searching lists
of geographical areas. The search input utilises postgreSQLs full-text
search features and so can run a search over both the geographical areas
ID and its description set.

### DIFF
--- a/geo_areas/models.py
+++ b/geo_areas/models.py
@@ -37,6 +37,9 @@ class GeographicalArea(TrackedModel, ValidityMixin):
     # This deals with subgroups of other groups
     parent = models.ForeignKey("self", on_delete=models.PROTECT, null=True, blank=True)
 
+    def get_description(self):
+        return self.geographicalareadescription_set.last()
+
     def save(self, *args, **kwargs):
         self.full_clean()
         return super().save(*args, **kwargs)
@@ -129,4 +132,4 @@ class GeographicalAreaDescription(TrackedModel, ValidityMixin):
         validators.validate_description_is_not_null(self)
 
     def __str__(self):
-        return f'description - "{self.description}" for {self.area}>'
+        return f'description - "{self.description}" for {self.area}'
--- a/geo_areas/tests/bdd/features/geo_areas.feature
+++ b/geo_areas/tests/bdd/features/geo_areas.feature
@@ -1,0 +1,16 @@
+Feature: Geographical areas
+
+
+Background:
+    Given a valid user named "Alice"
+    And geographical_area 1001 with a lorem ipsum description
+
+Scenario Outline: Searching for a geographical_area
+    Given I am logged in as Alice
+    When I search for a geographical_area using a <search_term>
+    Then the search result should contain the geographical_area searched for
+
+    Examples:
+    | search_term |
+    | 1001        |
+    | ipsum       |
--- a/geo_areas/views.py
+++ b/geo_areas/views.py
@@ -1,0 +1,34 @@
+from django.shortcuts import render
+from rest_framework import permissions
+from rest_framework import renderers
+from rest_framework import viewsets
+
+from geo_areas.filters import GeographicalAreaFilter
+from geo_areas.models import GeographicalArea
+from geo_areas.serializers import GeographicalAreaSerializer
+
+
+class GeoAreaViewSet(viewsets.ReadOnlyModelViewSet):
+    """
+    API endpoint that allows geographical areas to be viewed.
+    """
+
+    queryset = GeographicalArea.objects.all().prefetch_related(
+        "geographicalareadescription_set"
+    )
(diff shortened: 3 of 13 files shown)


geo_areas/views.py
```diff
@@ -0,0 +1,35 @@
...
+    filterset_class = GeographicalAreaFilter
+    search_fields = ["sid", "area_code"]
+
+
+class GeoAreaUIViewSet(GeoAreaViewSet):
+    """
+    UI endpoint that allows geographical areas to be viewed.
+    """
+
+    renderer_classes = [renderers.TemplateHTMLRenderer]
```
andyhd, line 28: I think this might be redundant as we don't return a `Response` object from the view method - my mistake for leaving it in footnotes.

NDevox: Good point, I've removed it - works fine.

geo_areas/models.py
```diff
@@ -37,6 +37,10 @@ class GeographicalArea(ValidityMixin):
     # This deals with subgroups of other groups
     parent = models.ForeignKey("self", on_delete=models.PROTECT, null=True, blank=True)
 
+    @property
+    def description(self):
+        return self.geographicalareadescription_set.last()
```
simonwo, line 42: Hm. We might need to be a bit smarter here… is this:
- the most recent "database live" description?
- the most recent "database live" description that is `PUBLISHED` (so ignoring errors)?
- the "operationally live" description?

Do we know what the prototypes do for this?

geo_areas/tests/bdd/features/geo_areas.feature
```diff
@@ -0,0 +1,12 @@
...
+
+Background:
+    Given a valid user named "Alice"
+    And geographical_area 1001
+
+Scenario: Searching by geographical_area ID
+    Given I am logged in as Alice
+    When I search for a geographical_area using a geographical_area ID
+    Then the search result should contain the geographical_area searched for
+
```
simonwo, line 12: Can we test the description search and all the other search features too?

urls.py
```diff
@@ -22,6 +22,7 @@
     path("", include("common.urls")),
     path("commodities/", include("commodities.urls")),
     path("", include("footnotes.urls")),
+    path("", include("geo_areas.urls")),
```
simonwo, line 25: How come this isn't `/geoareas` or something? &lt;/ignorance&gt;

NDevox: The magic of django rest framework: https://www.django-rest-framework.org/api-guide/routers/

It generates all the URLs for you in a standardised way.

geo_areas/tests/bdd/features/geo_areas.feature
NDevox: Sure! I missed it out as it wasn't in the spec.

geo_areas/models.py
NDevox: Agreed, we're going to have to deal with this at some point. But I don't think we can until #25 is merged in - until then there is no concept of live, published etc. Only an ordered set, admittedly with a date attached.

I'd argue database live is what we want - considering that the users will be concerned with updating the latest data, not with what is operational.

Again can't do this till #25 is merged, so bit of a chicken and egg scenario. If #25 is merged I'll make the change. If this one goes first we can put it in a ticket.

NDevox: And as I say this we merge in #25, I'll move it to database live - assuming that works for you.

simonwo: I think we are going to have some more analysis to do here regardless. I think all of the above suggestions have problems! With "database live" there's a subtle line to walk about what states are included in that (maybe we could do with firming up that definition more generally). E.g. is `CDS_ERROR` included? If so, does that mean things that have errored (and under our model this is a permanent state) will for ever be hanging around in search results? 

Also, are we including stuff from the "active workbasket" here or is that to come later?

I think this is something in the medium term we need to revisit with user research, and ask what users actually expect to see. In the short term, we could decide to do what the current prototype does, but might find that makes no sense given that we've shuffled the way the states interact. So it sounds like we need to do some analysis work on this to work out if that still makes sense.

Let's not block this PR or more data model work on that though. Happy to go with something that seems sensible for now and I'll put a ticket in to revisit this.

geo_areas/tests/bdd/features/geo_areas.feature
simonwo: Yeah let's get them in – if we've written the feature we should be testing it somehow, at least for whatever the intention was when it was written! And these BDD style tests will be useful later to compare.

NDevox: Added it in.

PR #32, opened 2020-06-26
Make compatible with tamato-taric-import

andyhd:
The TAMATO TARIC importer script has a dependency on TAMATO's serializers and models to
validate and modify the TAMATO DB.

This change adds a `setup.py` to allow TAMATO to be added as a pip dependency

It also adds preliminary serializers for Commodity and Measure models, to allow the
importer to deserialize these models.

Finally, the ValiditySerializerMixin is modified to handle TARIC date format strings
(YYYY-MM-DD) as well as ISO 8601 format (the default), for convenience.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python
+from distutils.core import setup
+
+import setuptools
+
+setup(
+    name="tamato",
+    version="0.0.1",
+    description="UK Tariff Management Tool",
+    maintainer="Department for International Trade",
+    maintainer_email="webops@digital.trade.gov.uk",
+    url="https://github.com/uktrade/tamato",
+    packages=setuptools.find_packages(),
+    install_requires=[
+        "dj-database-url",
+        "django",
+        "django-dotenv",
+        "django-extra-fields",
+        "django-filter",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+from distutils.core import setup
+
```
NDevox, line 3: Should include the requirements

andyhd: Oops!

setup.py
```diff
@@ -0,0 +1,36 @@
...
+setup(
+    name="TAMATO",
+    version="0.0.1",
+    description="UK Tariff Management Tool",
+    maintainer="Department for International Trade",
+    maintainer_email="webops@digital.trade.gov.uk",
+    url="https://github.com/uktrade/tamato",
+    packages=["tamato"],
+    package_dir={"tamato": ""},
```
andyhd, line 12: I was trying to get it so that you can do `from tamato.workbaskets import models`, but this doesn't seem to do it - any ideas?

NDevox:
```
    packages=setuptools.find_packages(),
```

I think should work.

PR #33, opened 2020-06-26
TP-65: Create the data model and validations for certificates

NDevox:
Many measures applied against goods require a certificate of some
type to allow the goods import into the country.

This commit introduces a data model which covers all certificates,
their various types, and their descriptions.


certificates/models.py
```diff
@@ -0,0 +1,92 @@
...
+                name="exclude_overlapping_certificates",
+                expressions=[
+                    ("valid_between", RangeOperators.OVERLAPS),
+                    ("sid", RangeOperators.EQUAL),
+                ],
+            ),
+        )
+
+
+class CertificateDescription(TrackedModel, ValidityMixin):
```
simonwo, line 74: If we're collapsing the Description Period model, then this will need to have a `sid` to populate the `certificate.description.period.sid` XML field.

certificates/models.py
```diff
@@ -0,0 +1,92 @@
+from django.contrib.postgres.constraints import ExclusionConstraint
+from django.contrib.postgres.fields import RangeOperators
+from django.db import models
+
+from certificates import validators
+from common.models import TrackedModel, ValidityMixin
+
+
+class CertificateType(TrackedModel, ValidityMixin):
+    sid = models.CharField(
```
simonwo, line 10: I see what you're doing here… but is it worth calling this a `code`? As SIDs everywhere else are of the integer to 8 or 9 digits variety.

certificates/models.py
simonwo, line 74: Do we deal with the extra end date on this `ValidityMixin` anywhere? My understanding was that we'd need to check/make sure that the end date of description N-1 is equal to certificate N's start date, and the start date of certificate N+1 is equal to certificate N's end date.

certificates/tests/test_business_rules.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+@pytest.mark.skip(
+    reason="description existence depends on certificate existence - this is a chicken and egg scenario"
+)
+def test_ce6_one_description_mandatory():
+    """
+    At least one description record is mandatory
+    """
+    assert False
```
simonwo, line 83: Couldn't we also trigger this by creating a certificate and a description, and then deleting the description?

certificates/models.py
simonwo: We also need to check the the latest most description has "no" end date (or DATE_MAX, or however we implement that).

certificates/tests/test_business_rules.py
NDevox: We need to make the certificate before we can make the description. As the description has to link to the certificate.

So if we have a validation invalidating any certificates without descriptions - we end up in a paradox.

certificates/models.py
NDevox: Can add in an exclusion constraint so no descriptions with the same sid can overlap.

And another validation so the last description cannot have an end date

NDevox: 👌

certificates/models.py
NDevox: I did have a small internal debate on this one.

I'm not sure if there's much value in constraining SIDs to only be 8-9 digit integers as opposed to them just being "the identifying value".

But I already put a flexible system into the TrackedModel system for defining what the identifying values are. I just like the consistency of terminology.

Can change if we think it is preferable.

certificates/tests/test_business_rules.py
simonwo: Oh, I see. We have to be able to create a valid certificate first anyway?

So this is one of the validations that needs to happen at the end of a transaction.  There are a few other business rules that work like that… like the infamous ME32. Have we done any of these yet and can do something similar here?

certificates/models.py
simonwo: I guess the only issue for me is that SID certainly has a well-constrained meaning to the XML, so if you're au fait with that it might be suprising. But yeah, I'm sure this isn't as a major issue.

andyhd: I have the same issue in Additional Code Types - they have no SID in the TARIC spec, but the ID is the SID. TARIC defines a SID as
> A unique number generated by the system as an internal access key. This will be used in areas where there is a need to change access key data dynamically, or where the logical key is too long.

which means that a SID here would be redundant, as the key is just the ID.

TL;DR - I agree with @NDevox

certificates/models.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class Certificate(TrackedModel, ValidityMixin):
+    sid = models.CharField(
+        max_length=3, validators=[validators.certificate_sid_validator]
+    )
+
+    certificate_type = models.ForeignKey(
+        CertificateType, related_name="certificates", on_delete=models.PROTECT
+    )
```
andyhd, line 44: I'm probably confused, but does this foreign key need to have `to_field="sid"`? (which isn't possible, because the `to_field` must be `unique`)

common/tests/factories.py
```diff
@@ -11,6 +12,8 @@
 
 BREXIT_DATE = datetime(2021, 1, 1).replace(tzinfo=timezone.utc)
 
+alphanumeric_id_generator = partial(product, "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789")
```
andyhd, line 15: Small tip - you could `import string` and do `list(string.ascii_uppercase + string.digits)` instead of `"ABCD...789"` here.

common/tests/test_util.py
```diff
@@ -27,3 +30,155 @@
...
+    [
+        (
+            DateTimeTZRange(
+                datetime(2016, 1, 1, tzinfo=timezone.utc),
+                datetime(2016, 2, 1, tzinfo=timezone.utc),
+            ),
+            DateTimeTZRange(
+                datetime(2016, 1, 10, tzinfo=timezone.utc),
+                datetime(2016, 1, 20, tzinfo=timezone.utc),
+            ),
```
andyhd, line 46: Why aren't these the named ranges from the `date_ranges` fixture?

certificates/models.py
andyhd: I have the same issue in Additional Code Types - they have no SID in the TARIC spec, but the ID is the SID. TARIC defines a SID as
> A unique number generated by the system as an internal access key. This will be used in areas where there is a need to change access key data dynamically, or where the logical key is too long.

which means that a SID here would be redundant, as the key is just the ID.

TL;DR - I agree with @NDevox

certificates/models.py
NDevox: Yeah this is where there's the confusion between the SID and the ID autofield.

I think we're OK to just point to the ID field. 99% of the time you will likely care only about the version of the object it was pointing to at that time anyway. The only case I can think of for that is when creating a new version - then you likely want the latest one. Which should be as simple as:

`self.certificate_type.objects.get_latest_version(sid=self.certificate_type.sid)`

certificates/tests/test_business_rules.py
NDevox: Yeah I don't think it's reasonable to do the validation at the data model level. Transaction/Workbasket level makes sense.

And that gives me an idea.

BTW RE ME32. With the commodity code MPTT system it should be something like this:

```python
commodity_tree = measure.commodity.get_ancestors() | measure.commodity.get_descendants()

if commodity_tree.filter(measure__valid_between__overlaps=measure.valid_between):
    raise ValidationError("Cannot have multiple measure in a single commodity code hierarchy")
```

certificates/models.py
NDevox: Implemented

NDevox: Adjacent validity dates has been implemented, as well as no overlapping.

The no end date hasn't been implemented though. There's still some working to do around how we handle updates to rows with the workbaskets and append only logs (i.e. we will have multiple DB records with no end date but this is only because they will be "updated" by rows replacing them).

This has some complexity to it so would rather defer until a bit later.

common/tests/factories.py
NDevox: I was being a bit lazy here... Done

common/tests/test_util.py
NDevox: I was thinking about doing that, but not sure how I could manage it via parametrize.

Alternative in my mind is to not use parametrize and outline the scenarios in the test - but then it's all seen as a single test and the logs are less clear exactly what we've failed on. Is there a way to switch this out?

I could feasibly put the DateRanges outside the fixture so it can be imported. Feels a bit dirty importing from a conftest though

andyhd: Ah yeah, OK. You could move the `Dates` class into `common/tests/util.py` and import it into `conftest.py`, but maybe it's not worth it.

certificates/tests/test_business_rules.py
simonwo: Cool, so is that part of this PR? I guess I'd rather get it done whilst it's fresh before we forget about it?

NDevox: I'm unconvinced if it should be done here. There are multiple models affected by this so we need a generic system for validating at the transaction level.

I don't think building that would properly fit the scope of this PR, as we'd need to update all those other models as well. But we should definitely do it sooner than later. I'd propose it be put into a ticket for the next sprint. Would be a bit cleaner in terms of commit history and there would be less context switching for myself 😉

certificates/migrations/0001_initial.py
```diff
@@ -0,0 +1,149 @@
...
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        ("common", "0001_initial"),
+    ]
+
+    operations = [
+        BtreeGistExtension(),
```
andyhd, line 20: I found that this needs to be applied before any other trackedmodels migrations - I moved it to `common/migrations/0001_initial.py`.

PR #34, opened 2020-07-02
TP-63: Data model for additional codes

andyhd:
Measures sometimes are applied to sub-types of commodities, and these are identified by
additional codes.

This commit introduces a new data model which covers additional codes, their various
types, and their descriptions.

⚠️ **WARNING**: This commit modifies a migration that most likely has already been applied. You may need to drop and recreate your database.


setup.py
```diff
@@ -13,7 +13,7 @@
     packages=setuptools.find_packages(),
     install_requires=[
         "dj-database-url",
-        "django",
+        "Django>=3.0,<3.1",
```
NDevox, line 16: I've been leaving this off for a while, but was going to suggest we pin everything when we are closing in on being ready.

Is there an issue with 3.1?

andyhd: Ah, I made this change while working on the import script - distutils failed to install the Django dependency when it was just "django", so I made it more specific and it worked. But I put that in the setup.py for the import script, so it doesn't need to be here.

additional_codes/models.py
```diff
@@ -0,0 +1,158 @@
...
+
+# class FootnoteAssociationAdditionalCode(TrackedModel, ValidityMixin):
+#     """A footnote may be associated with an additional code for part of the validity
+#     period of the footnote and part of the validity period of the additional code.
+#     """
+
+#     # This is not used and here only for historical data
+
+#     additional_code = models.ForeignKey(AdditionalCode, on_delete=models.CASCADE)
+#     footnote = models.ForeignKey("footnotes.Footnote", on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
andyhd: Oops, I need to uncomment this - we don't use footnote association apparently, but probably still need this model for storing historical data.

NDevox:
Can we add record code and sub record code to the model, and then put them in as fields for the serializers

PR #35, opened 2020-07-03
TP-189: adds the XML exporter API

NDevox:
The main consumers of this service require a TARIC3 compliant XML document
to be produced for any changes generated. These changes are generally grouped
at several levels. The top level is an `envelope`, which is the wrapper for the
document being sent. The next level is the `transaction`, which groups a block
of changes together - this is represented by the `WorkBasket` in this project.
The next group is the message which is just a wrapper around a record. The record
representing the individual object being changed.

This commit introduces an implementation to produce these compliant XML
documents. The idea behind this implementation is that, as all models
are children of `TrackedModel`, all models can follow the template method
design pattern. This enables us to have a standardised method for finding
and rendering templates relevant to each record. These record specific templates
can then in turn be included into a transaction template, building up the
TARIC document. In doing this the process offloads the rendering of each
individual model to the model itself. Giving a more granular and simplified way
to create and update how these records are presented. Thus in this case every
model is paired with a template based off of the model name (this is overrideable
with the use of either a `taric_template` attribute, or overriding the
`get_taric_template` method on the model. The template for the transaction
then adds the common transaction data, as defined in the workbasket and the
transaction model. Each record in the workbasket then has it's template included
with the data from the record inputted into it.

To facilitate this a PolymorphicModelSerializer is introduced. This requires
a mapping of models against their serializers. The mapping is managed by
using a decorator attached to the PolymorphicModelSerializer. When defining a
serializer for a model, decorating this serializer will automatically add the
serializer to the mapping, with it's relevant model.

Alongside this a new testing utility has been introduced, a decorator, which
will automatically generate an instance of the model required (via a factory)
and check if the TARIC XML generated from the model is valid against the
taric.xsd. This then returns the given XML for further testing if required.

Testing the generated data this way has shown that a couple tweaks were required
for several serializers and models, most notably the Regulation model,
these have been included in this commit.


common/renderer.py
```diff
@@ -0,0 +1,13 @@
+from rest_framework.renderers import TemplateHTMLRenderer
+
+
+class TaricXMLRenderer(TemplateHTMLRenderer):
+    media_type = "application/xml"
+    format = "xml"
```
andyhd, line 6: So this lets us go to `/api/workbaskets/<id>?format=xml` to get the taric update?

geo_areas/models.py
```diff
@@ -27,6 +27,10 @@ class GeographicalArea(TrackedModel, ValidityMixin):
     separate foreign key for group to group relations.
     """
 
+    record_code = "250"
+    subrecord_code = "00"
+    taric_template = "geo_areas/taric/geographical_area.xml"
```
andyhd, line 32: Would it be worth doing a convention-over-configuration thing here and have the templates named `taric/{model.__class__.__name__.lower()}.xml` by default? You could put them in `<app>/jinja2/taric/` to avoid having to specify the template directory name.

common/renderer.py
NDevox: `/api/workbaskets/<id>format=xml`, `/api/worbaskets/<id>.xml` and `/api/workbaskets/<id>` with an Accept header of `application/xml` will all return taric style xml

I think I made it a default so it would be available on all models, but will probably break on most of them, so should really limit it to just workbaskets...

geo_areas/models.py
NDevox: An emphatic yes from me. I was split between the two, but the more I had to write `taric_template = ` etc. the less I liked going with the config.

Just mildly worried about newcomers figuring out how to add new templates. But I'll document it and should be clear enough.

NDevox:
FYI to fix the message ID issue and multiple records needed for the validity periods with descriptions (and other merged data types) I'm probably going to make the record easily overrideable so that still a bit of fun to work around the message ID (as currently the nested records have no concept of message IDs, this will have to change somewhat.

geo_areas/models.py
andyhd: As long as you can override it by setting `taric_template` or implementing `get_taric_template()` - it's similar to how DRF does viewset template names

common/jinja2/common/taric/base.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<env:envelope xmlns="urn:publicid:-:DGTAXUD:TARIC:MESSAGE:1.0" xmlns:env="urn:publicid:-:DGTAXUD:GENERAL:ENVELOPE:1.0" id="1"  xmlns:oub="urn:publicid:-:DGTAXUD:TARIC:MESSAGE:1.0">
```
andyhd, line 2: This feels a bit of a halfway house. I get wanting to deduplicate the namespace declarations in the `transmission` element. I think we should stick to [the robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) here and have our output strictly follow either:
 - The TARIC XML produced by the EU (with namespaces declared as they were before this change) - in case HMRC or others rely (incorrectly) on this, or

 - The TARIC and ENVELOPE schemas - we either don't need to declare the MESSAGE namespace as the default namespace (because we prefix all elements), or we don't need the `oub` namespace prefix. We don't really need any namespace prefixes if we set the default namespace to ENVELOPE in the `envelope` element and to MESSAGE in the `transmission` element.

If we're going to deviate from what the EU produce, we may as well go all the way.

common/jinja2/includes/record.xml
```diff
@@ -0,0 +1,12 @@
...
+          <oub:record>
+          <oub:record.code>{{ record_code or record.record_code }}</oub:record.code>
+          <oub:subrecord.code>{{ subrecord_code or record.subrecord_code }}</oub:subrecord.code>
+          <oub:record.sequence.number>{{ index }}</oub:record.sequence.number>
+          <oub:transaction.id>{{ transaction_id }}</oub:transaction.id>
+          <oub:update.type>{{ record.update_type }}</oub:update.type>
+              {{ record_data }}
+        </oub:record>
+      </oub:transmission>
+</env:app.message>
```
andyhd, line 12: I'm not quite following how this works, but I think you can instead define a Jinja macro and use it like:
```jinja
{% macro record(some, parameters) %}
<env:app.message id="{{ message_counter.message_id }}">
  <oub:transmission>
    ...
    <oub:update.type>...</oub:update.type>
    {{ caller() }}
  </oub:transmission>
</env:app.message>
{% endmacro %}
```
and call it like
```jinja
{% from "common/includes/record_macro.jinja" import record with context %}{# probably need "with context" #}
{% call record(some, params) %}
<oub:geographical.area>
    <oub:geographical.area.sid>...
    ...
</oub:geographical.area>
{% endcall %}

File: common/models.py [hunk 5]
```diff
@@ -119,12 +121,29 @@ class Meta:
...
 
+    @property
+    def start_date(self):
+        if self.valid_between.lower:
+            return "{:%d-%b-%Y}".format(self.valid_between.lower)
+
+    @property
+    def end_date(self):
+        if self.valid_between.upper:
+            return "{:%d-%b-%Y}".format(self.valid_between.upper)
```
andyhd, line 132: Do we use these anywhere else? They might be better as a custom serializer field(s)

common/models.py
```diff
@@ -134,10 +153,46 @@ class TrackedModel(PolymorphicModel):
...
+        if self.taric_template:
+            return self.taric_template
+        class_name = self.__class__.__name__
+
+        # replace namesLikeThis to names_Like_This
+        name = re.sub(r"([a-z0-9])([A-Z])", r"\1_\2", class_name)
+        # replace names_LIKEthis to names_like_this
+        name = re.sub(r"([A-Z]{2,})([a-z0-9_])", r"\1_\2", name).lower()
+
+        template_name = f"taric/{name.lower()}.xml"
```
andyhd, line 181: Really really `lower` 😄

common/renderers.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class MessageCounter:
+    def __init__(self):
+        self.message = 0
+
+    @property
+    def message_id(self):
+        self.message += 1
+        return self.message
```
andyhd, line 11: I would have done this with a generator

common/models.py
NDevox: Yep makes sense.

common/models.py
NDevox: 🤣

common/renderers.py
NDevox: Yep was my first thought too. Not entirely sure what led me down this route...

I think I liked not have parentheses or not having to use `next`. But actually it would look a lot clearer.

common/jinja2/includes/record.xml
NDevox: Yeah macros make more sense to me on this one, I'll shift them over.

This bit got a bit messy.

I was trying to do all the blocks in the `include`d record, and then override the blocks as needed in the inheriting template. But you can't override blocks in inherited templates when the blocks are from an `include`. So ended up with this solution where you have to set the context and redo the includes which... is messy.

common/jinja2/common/taric/base.xml
NDevox: 🤔 

On reflection I'm not overly comfortable with messing with the XSD right now (which would be necessary for option 2) without confirmation that we can do it. So will revert to the first option.

Happy to discuss and change later.

common/renderers.py
NDevox: `next` isn't allowed in template, settling with a `lambda`

common/renderers.py
```diff
@@ -1,14 +1,11 @@
...
-
-    @property
-    def message_id(self):
-        self.message += 1
-        return self.message
+def message_counter():
+    x = 1
+    while True:
+        yield x
+        x += 1
```
andyhd, line 8: You can use `itertools.count(start=1)` for this

common/jinja2/common/taric/base.xml
andyhd: I tested option 2 against the XSD and it validates. The question is whether HMRC et al wpould accept it.

common/tests/factories.py
```diff
@@ -53,6 +53,7 @@ class Meta:
 
 class TrackedModelMixin(factory.django.DjangoModelFactory):
     workbasket = factory.SubFactory(WorkBasketFactory)
+    update_type = 3
```
andyhd, line 56: Would this be clearer as `UpdateType.Insert.value`?

common/renderers.py
NDevox: One day I will make a point to memorise all the tools in `itertools`

common/tests/factories.py
NDevox: Yes

common/assets/envelope.xsd
```diff
@@ -0,0 +1 @@
...

      </xs:attribute>

      <xs:attribute name="location" type="xs:string" use="required"/>

    </xs:complexType>

  </xs:element>

</xs:schema>
```
andyhd, line 1: I know this is an imported file, but maybe change the line endings - this hurts my eyes 😭

NDevox: 😆

NDevox: Done

PR #36, opened 2020-07-06
Add a hook for validating models in a workbasket

andyhd:
Some business rules require validating across models in a workbasket, for example,
Certificates must have at least one associated CertificateDescription. This cannot be
achieved with standard model validation.

This change adds a method to all TrackedModel objects, `validate_workbasket`, which may
be overridden to perform validation across all models in the same workbasket. When the
WorkBasket is validated, it calls this method on all its TrackedModels. Any
ValidationErrors raised are stored in the WorkBasket instance to enable error messaging
in the UI.

### DIFF
--- a/additional_codes/validators.py
+++ b/additional_codes/validators.py
@@ -123,3 +123,8 @@ def validate_additional_code_description_start_date_before_additional_code_end_d
                 "date of the additional code."
             }
         )
+
+
+def validate_at_least_one_description(additional_code):
+    if additional_code.descriptions.count() < 1:
+        raise ValidationError("At least one description record is mandatory.")
--- a/workbaskets/models.py
+++ b/workbaskets/models.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.db import models
 from django_fsm import FSMField
 from django_fsm import transition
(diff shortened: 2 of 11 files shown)


workbaskets/models.py
```diff
@@ -119,24 +135,15 @@ def cds_confirmed(self):
...
-        on_delete=models.PROTECT,
-        editable=False,
-        null=True,
-        related_name="approved_workbaskets",
-    )
+    def clean(self):
+        self.errors = []
+        for model in self.trackedmodel_set.all():
+            try:
+                model.validate_workbasket()
```
NDevox, line 142: Do we think there may be any cases where the tracked model needs access to the workbasket to validate itself?

Just wondering if it is worth passing in `self` to the `validate_workbasket` method.

It's not necessary now though, just a thought and easy fix.

andyhd: I was originally passing the workbasket to `validate_workbasket`, but a trackedmodel already has a reference to its workbasket, so I thought it was redundant. On retrospect, perhaps passing the workbasket will avoid a database query? I'm not certain if the workbasket will be retrieved from cache when accessing `model.workbasket`.

NDevox: huh... good point.

Maybe leave it for now unless performance becomes an issue.

additional_codes/validators.py
```diff
@@ -123,3 +123,13 @@ def validate_additional_code_description_start_date_before_additional_code_end_d
                 "date of the additional code."
             }
         )
+
+
+def validate_at_least_one_description(additional_code):
+    num_descriptions = additional_code.workbasket.trackedmodel_set.filter(
+        polymorphic_ctype__model="additionalcodedescription",
```
NDevox, line 130: Tempted to add a `of_type` or similar method to the queryset, which would allow you to simply pass in the class and let it figure out the ctype.

But again don't think this is worth doing now.

andyhd: Yeah, this does break encapsulation.

andyhd: Actually, I should be able to do `workbasket.trackedmodel_set.instance_of(AdditionalCodeDescription)`, shouldn't I?

andyhd: Or even just `.filter(described_additional_code=additional_code)` should work, since no other trackedmodels have that field.

NDevox: Mmmm, wondering what the last query would be.

You are right though, `instance_of` currently exists already. Trying to remember if that made extra queries... I think it may have done.

andyhd: Hm, this validation doesn't work for updates or deletes 🤔

andyhd:
Footnotes validation needs refactoring - I'll do it in a separate PR

PR #37, opened 2020-07-08
Fix additional code search

andyhd:
All additional codes were returned regardless of search terms.

This change adds a custom DRF filter backend, which searches for exact Additional Code
Type SID + Additional Code matches (eg: X001), and falls back to full-text search on
sid, code, type, description and type description.

### DIFF
--- a/additional_codes/filters.py
+++ b/additional_codes/filters.py
@@ -1,0 +1,23 @@
+import re
+
+from common.filters import TamatoFilterBackend
+
+
+COMBINED_ADDITIONAL_CODE_AND_TYPE_ID = re.compile(
+    r"^(?P<additional_code_type_id>[A-Z0-9])(?P<additional_code_id>[A-Z0-9]{3})$"
+)
+
+
+class AdditionalCodeFilterBackend(TamatoFilterBackend):
+    """
+    Filter that combines additional code type ID and additional code ID
+    """
+
+    search_fields = "type__sid", "code"  # XXX order is significant
+
+    def get_search_term(self, request):
+        search_term = super().get_search_term(request)
+        match = COMBINED_ADDITIONAL_CODE_AND_TYPE_ID.match(search_term.strip())
+        if match:
+            return " ".join(match.groups())
+        return search_term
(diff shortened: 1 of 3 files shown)


additional_codes/filters.py
```diff
@@ -0,0 +1,26 @@
...
+
+from additional_codes import models
+
+
+COMBINED_ADDITIONAL_CODE_AND_TYPE_ID = re.compile(
+    r"(?P<additional_code_type_id>[A-Z0-9])(?P<additional_code_id>[A-Z0-9]{3})"
+)
+
+
+class AdditionalCodeFilterBackend(filters.BaseFilterBackend):
```
NDevox, line 14: There's a lot of commonalities between all the view code we're writing at the moment, I want to see if we can start making it more and more generic so we have less to maintain.

Maybe switch to something like this?

```python3
# common/filters.py

class TamatoFilterBackend(filters.BaseFilterBackend):
    search_fields = ("sid",)

    def get_search_term(self, request):
        return request.query_params.get("search", "")

    def filter_queryset(self, request, queryset, view):
        search_term = self.get_search_term(request)
        if search_term:
            return queryset.annotate(search=SearchVector(*self.search_fields)).filter(
                search=search_term
            )
        return filters.SearchFilter().filter_queryset(request, queryset, view)


# additional_codes/filters.py

COMBINED_ADDITIONAL_CODE_AND_TYPE_ID = re.compile(
    r"(?P<additional_code_type_id>[A-Z0-9])(?P<additional_code_id>[A-Z0-9]{3})"
)

class AdditionalCodeFilterBackend(TamatoFilterBackend):
    """
    Filter that combines additional code type ID and additional code ID
    """
    search_fields = "type__sid", "code"
    def get_search_term(self, request, queryset, view):
        search_term = request.query_params.get("search", "")
        match = COMBINED_ADDITIONAL_CODE_AND_TYPE_ID.match(search_term)
        if match:
            return " ".join(match.groups())
```

andyhd: The `AdditionalCodeFilterBackend` class encapsulates the fact that the `type__sid` and `code` fields are concatenated in that order with a space in the search vector, allowing the search term `"X 000"` (for example) to match exactly.

I will implement the common filter backend and use it for the `AdditionalCodeFilterBackend` with a note that the order of `search_fields` is significant.

PR #40, opened 2020-07-10
TP-199: Fix technical debt in footnotes model

andyhd:
The footnotes data model was implemented before we developed the TrackedModel class and
implements validation and constraints incorrectly. The data model is also inconsistent
with later models with respect to descriptions handling.

This commit brings the footnotes data model into line with other components, like
certificates and additional codes.


additional_codes/tests/bdd/features/detail_additional_codes.feature
```diff
@@ -8,3 +8,8 @@ Scenario: View additional code core data
     Given I am logged in as Alice
     When I select the additional code X000
     Then the core data against the additional code should be presented
+
+Scenario: View additional code description data
+    Given I am logged in as Alice
+    When I select the additional code X000
+    Then the descriptions against the additional_code should be presented
```
simonwo, line 15: Did you mean to have this additional code stuff part of the commit?

footnotes/models.py
```diff
@@ -28,68 +26,131 @@
...
         constraints = [
-            models.UniqueConstraint(
-                fields=["footnote_id", "footnote_type"], name="FO2",
+            ExclusionConstraint(
+                name="exclude_overlapping_footnotes_FO2",
+                expressions=[
+                    ("valid_between", RangeOperators.OVERLAPS),
+                    ("footnote_id", RangeOperators.EQUAL),
+                    ("footnote_type", RangeOperators.EQUAL),
+                ],
```
simonwo, line 106: For my understanding, have we worked out the way to make these work for `TrackedModels` now? So that I can emit an update with the same data and this won't collide?

additional_codes/tests/bdd/features/detail_additional_codes.feature
andyhd: Yes, this is just a small fix I did while I noticed it

footnotes/models.py
andyhd: I think we agreed that we would set the end date on superceded TrackedModels - correct me if I'm wrong @NDevox

NDevox: Yep we discussed that as a likelihood. I kind of see this part of the design as tech debt right now.

i.e. right now it works for us and fills in a large number of the requirements we have, but I'm fairly sure that once we start seriously implementing the AO log that it'll start falling apart a bit and will require some changes.

I think it may be best to wait till we start implementing the AO log before making changes for it. It will be this sprint or next IMO, as it'll be once we start the backend for edits. I wouldn't do it in this PR as there are a lot more models that will be affected.

additional_codes/tests/bdd/features/detail_additional_codes.feature
NDevox: If it fixes my ******* tests I'm all for it.

PR #41, opened 2020-07-16
TP-204: Add Commodity Code (Goods Nomenclature) data models

NDevox:
Commodities are the fundamental objects being imported into the country
that the tariff applies to. In the taric system these are generally
represented by a "Goods Nomenclature" within a hierarchical structure.
In this way commodities belong to categories and sub-categories. Where
various other constructs, such as measures, may apply to an individual
commodity, or to an entire category of commodities. So an easily
traversible tree is important.

This commit introduces a more robust data model to accurately represent
Goods Nomenclatures, including a description model and an association
to footnotes.

In this process a Materialized Path hierarchical data structure is
used to represent the hierarchical tree due to it's fast read times.

Several legacy data structures have been wrapped up in the process.
Indent is one, due to the way the tariff has historically been stored
it has been necessary to provide an indent table to describe where
the commodity sits in the tree. With the current implementation this
table is redundant. Likewise historically there has been both an
"origin" and "successor" tables to represent an auditable history
of a commodity code, e.g. for when it was split, or a new one added.
As there are no dates related to this it is a redundant table and can
be simply represented by a single self-referential "origin"
foreign key. This way origins can be found by following the origin
foreign key, and successors can be found by querying where an item is
the origin.


commodities/models.py
```diff
@@ -1,31 +1,99 @@
...
+
+    # These are character fields as they often has leading 0s
+    item_id = models.CharField(max_length=10, validators=[RegexValidator(r"\d{10}")])
+    suffix = models.CharField(max_length=2, validators=[RegexValidator(r"\d{2}")])
+
+    statistical = models.BooleanField()
+
+    origin = models.ForeignKey(
+        "self",
+        null=True,
```
NDevox, line 79: Problem here.

Business rules requires we have an origin.

But what about the original top level codes? they can't really have an origin.

This is another chicken and egg scenario. Not sure of a solution yet so ignored for now.

commodities/models.py
```diff
@@ -1,31 +1,99 @@
...
+        blank=True,
+        on_delete=models.PROTECT,
+        related_name="succeeding_goods",
+        related_query_name="succeeding_goods",
+    )
+
+    indent_only_change = models.BooleanField(default=False)
```
NDevox, line 86: Query if this should be `indent_only_change` - only the indent has changed.

Or `indent_change` - the indent has changed, others may have as well.

commodities/migrations/0003_auto_20200716_0840.py
```diff
@@ -0,0 +1,69 @@
...
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("commodities", "0002_auto_20200716_0840"),
+        ("common", "0003_trackedmodel_update_type"),
+        ("measures", "0002_auto_20200716_0840"),
+        ("footnotes", "0002_refactor_footnotes"),
```
andyhd, line 14: Can these two commodities migrations be merged into one? Also, it would be nice to give proper names to the auto migrations.

commodities/tests/test_business_rules.py
```diff
@@ -0,0 +1,207 @@
...
+    """
+
+    workbasket = factories.WorkBasketFactory()
+    good = factories.GoodsNomenclatureFactory(workbasket=workbasket)
+
+    with pytest.raises(ValidationError):
+        workbasket.submit_for_approval()
+
+    factories.GoodsNomenclatureDescriptionFactory(described_goods_nomenclature=good)
+    workbasket.submit_for_approval()
```
andyhd, line 103: The description is instantiated without a workbasket, so it will generate a different one.
Is this passing? If so, there must be a bug in workbasket validation 😞

commodities/models.py
```diff
@@ -1,31 +1,99 @@
...
+    successor_record_code = "400"
+    successor_subrecord_code = "40"
+
+    sid = models.PositiveIntegerField(
+        validators=[MinValueValidator(1), MaxValueValidator(99999999)]
+    )
+
+    # These are character fields as they often has leading 0s
+    item_id = models.CharField(max_length=10, validators=[RegexValidator(r"\d{10}")])
+    suffix = models.CharField(max_length=2, validators=[RegexValidator(r"\d{2}")])
```
andyhd, line 73: I think it is better to have these regex validators in `validators.py` for easier sharing with the serializers

commodities/migrations/0003_auto_20200716_0840.py
NDevox: Can do the proper names.

The squashing thing is more of an issue though as currently we have a measures model reliant on the old commodity system.

TBF the measures model serves no purpose so we can delete it. if we're happy to rebuild databases (again)

commodities/models.py
NDevox: Fine by me

commodities/tests/test_business_rules.py
NDevox: Ah... good point.

I don't think the validation checks in the workbasket, just sees if there is any description in the db linked to it.

So when we make the description here in a new workbasket it's still OK.

This isn't OK for the actual business needs though. so the validation should be updated to include a description for the GN that has a workbasket older or equal to the current one.

commodities/migrations/0003_auto_20200716_0840.py
NDevox: Done as mentioned, but not removed the measures system. Reckon it may be more pain than it is worth right now.

andyhd: Cool, I don't think squashing the migrations is necessary.

commodities/tests/test_business_rules.py
```diff
@@ -0,0 +1,205 @@
...
+    workbasket = factories.WorkBasketFactory()
+    good = factories.GoodsNomenclatureFactory(workbasket=workbasket)
+
+    with pytest.raises(ValidationError):
+        workbasket.submit_for_approval()
+
+    factories.GoodsNomenclatureDescriptionFactory(
+        described_goods_nomenclature=good, workbasket=workbasket
+    )
+    workbasket.submit_for_approval()
```
simonwo, line 101: Do we need to test the start date condition?

commodities/models.py
NDevox: I would think so - that the SID wouldn't change

But I do wonder if this is too trusting or even desirable. The only real reason for this field is to avoid the extra query back to the origin to see what has changed - in case it was indent only.

Otherwise everytime we build the XML we have to make an extra query.

But it may just be more useful to know if the indent has changed, as opposed to _only_ the indent changing.

commodities/models.py
NDevox: Can a top level code be replaced? in which case it might have an origin.

Can definitely write the validation for that, just wondering how forgiving we should be.

commodities/tests/test_business_rules.py
NDevox: Yeah we should.

NDevox: Done

commodities/models.py
NDevox: Done

PR #42, opened 2020-07-17
TP-205: Add XML export for Goods Nomenclature.

NDevox:
A JSON API for the Goods Nomenclatures has also
been added for convenience.

### DIFF
--- a/commodities/jinja2/taric/footnote_association_goods_nomenclature.xml
+++ b/commodities/jinja2/taric/footnote_association_goods_nomenclature.xml
@@ -1,0 +1,16 @@
+{% from "macros/record.xml" import standard_record %}
+{% from "macros/record.xml" import base_record %}
+
+{% call standard_record(message_counter(), sequence(), transaction_id, record) %}
+    <oub:footnote.association.goods.nomenclature>
+        <oub:goods.nomenclature.sid>{{ record.goods_nomenclature.sid }}</oub:goods.nomenclature.sid>
+        <oub:footnote.type>{{ record.associated_footnote.footnote_type.footnote_type_id }}</oub:footnote.type>
+        <oub:footnote.id>{{ record.associated_footnote.footnote_id }}</oub:footnote.id>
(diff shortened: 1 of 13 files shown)


commodities/jinja2/taric/footnote_association_goods_nomenclature.xml
```diff
@@ -0,0 +1,22 @@
+{% from "macros/record.xml" import standard_record %}
+{% from "macros/record.xml" import base_record %}
+
+{% call standard_record(message_counter(), sequence(), transaction_id, record) %}
+    <oub:additional.code.type>
```
simonwo, line 5: This looks copy-pasted? 😜 
Shouldn't this be `footnote.association.goods.nomenclature`? And the fields are different.

commodities/tests/test_xml.py
```diff
@@ -0,0 +1,82 @@
+import pytest
+
+from commodities.tests.conftest import generate_good
+from common.tests import factories
+from common.tests.util import validate_taric_xml
+
+pytestmark = pytest.mark.django_db
+
```
simonwo, line 8: Probably need one for the footnote associate as well then?

commodities/jinja2/taric/footnote_association_goods_nomenclature.xml
NDevox: I wonder how that happened 😏 

I may have forgotten about this one after some "minor" distractions around indents...

commodities/tests/test_xml.py
NDevox: Yah that makes sense

commodities/jinja2/taric/footnote_association_goods_nomenclature.xml
NDevox: Done

commodities/tests/test_xml.py
NDevox: Done

PR #44, opened 2020-07-24
TP-203: Edit footnotes and footnote descriptions

andyhd:
The main purpose of the TAMATO app is to provide a user interface for tariff managers to
edit the UK global tariff. So far, the user interface has been read-only.

This change adds the ability to edit the details of a Footnote, and the associated
description, and to save those changes in a current WorkBasket.

WorkBasket creation and submission is not in scope of this change, but the user is
prompted to choose a current workbasket if none is already open.

This change has highlighted the requirement to remove in-database exclusion constraints
on the models, due to draft changes necessarily conflicting with them. The business
rules that these constraints enforced are now implemented in validation code instead.


andyhd:
I'm still working on the validation to replace the exclusion constraints.

common/forms.py
```diff
@@ -0,0 +1,106 @@
...
+        widgets = (
+            forms.NumberInput(attrs=attrs),
+            forms.NumberInput(attrs=attrs),
+            forms.NumberInput(attrs=attrs),
+        )
+        super().__init__(widgets=widgets, attrs=attrs)
+
+    def decompress(self, value):
+        if value:
+            return "{0.day} {0.month} {0.year}".format(value).split()
```
NDevox, line 26: `return [value.day, value.month, value.year]` is less characters?

common/forms.py
```diff
@@ -0,0 +1,106 @@
...
+            try:
+                day, month, year = data_list
+                result = from_current_timezone(
+                    datetime(int(year), int(month), int(day))
+                )
+            except ValueError:
+                valid = False
+
+            if not valid:
+                raise ValidationError("Enter a valid date.", code="invalid_date")
```
NDevox, line 53: Might be worth splitting the raise here to:

```python3
            valid = all(data_list)
            if not valid:
                raise ValidationError("Enter a valid date.", code="invalid_date")
            try:
                day, month, year = data_list
                result = from_current_timezone(
                    datetime(int(year), int(month), int(day))
                )
            except ValueError as e:
                raise ValidationError("Enter a valid date.", code="invalid_date") from e
```

On the chance that we get an error here we weren't expecting (e.g. in testing) it will be slightly clearer which condition actually failed.

NDevox: I am being a bit OTT I'll admit.

common/models.py
```diff
@@ -213,5 +215,11 @@ def new_draft(self, workbasket, save=True, **kwargs):
 
         return new_object
 
+    def get_versions(self):
+        query = Q()
+        for field in self.identifying_fields:
+            query &= Q(**{field: getattr(self, field)})
+        return self.__class__.objects.filter(query).order_by("-created_at")
+
```
NDevox, line 223: Shouldn't this be on the QuerySet object?

common/validators.py
```diff
@@ -27,3 +28,46 @@ def __eq__(self, other):
...
+            valid = self.min_value <= value <= self.max_value
+
+        if not valid:
+            raise ValidationError(self.message, code=self.code)
+
+
+class DayValidator(DatePartValidator):
+    message = "Enter a valid day of month."
+    min_value = 1
+    max_value = 31
```
NDevox, line 55: Wonder if this needs to be more complex to consider calendar month limits. e.g. Feb can only have 28 days.

footnotes/tests/bdd/test_edit_footnote.py
```diff
@@ -0,0 +1,71 @@
...
+    workbasket = factories.WorkBasketFactory()
+    session = client.session
+    session["workbasket"] = workbasket.to_json()
+    session.save()
+    return workbasket
+
+
+@given("Alice has permission to update a footnote")
+def alice_has_permission_to_update_a_footnote(valid_user, footnote_NC000):
+    assert True  # TODO implement permissions
```
NDevox, line 32: I'd make this an `assert False` and mark as an XFAIL so we know it needs changing.

footnotes/views.py
```diff
@@ -27,31 +41,149 @@ class FootnoteViewSet(viewsets.ReadOnlyModelViewSet):
...
-    renderer_classes = [renderers.TemplateHTMLRenderer]
+    def get_success_url(self):
+        return reverse("footnote-ui-confirm-update", kwargs=self.kwargs)
 
-    def list(self, request):
-        queryset = self.filter_queryset(self.get_queryset())
-        return render(
-            request, "footnotes/list.jinja", context={"object_list": queryset}
+    def form_valid(self, form):
+        self.object = self.object.new_draft(
```
NDevox, line 113: Nice to see `new_draft` being used 😄

footnotes/views.py
```diff
@@ -27,31 +41,149 @@ class FootnoteViewSet(viewsets.ReadOnlyModelViewSet):
...
+            workbasket=get_current_workbasket(self.request),
+            update_type=UpdateType.UPDATE,
```
NDevox, line 115: Wonder whether all new drafts should be an update.

Create should be a new object, not a new draft... delete is a bit of a question though.

common/forms.py
andyhd: Doh

common/forms.py
andyhd: That makes sense

common/models.py
andyhd: You mean filter the queryset by `query`, or the `order_by` (which we may want to override in the UI at some point anyway)?

common/validators.py
andyhd: The date as a whole is validated by the parent MultiValueField, which would catch Feb 30. The day subfield isn't aware of the month value. But actually, I'm not even sure these validators ever get to surface an error.

footnotes/views.py
andyhd: Yep, that makes sense. I think a delete is essentially a special case of update - you have a draft delete in the workbasket. We could maybe add a similar method to `TrackedModel` for deletions, but probably don't want to override `delete()`. Do we need anything more than `update_type=DELETE` to know a record is deleted?

NDevox: We need an end date as well.

Would be tempted to override the `delete` method purely to raise an error to use our specific delete update instead - so we don't accidentally delete things.

footnotes/tests/bdd/test_edit_footnote.py
andyhd: Except that this is a `@given` clause in all the edit BDD tests, rather than a test in its own right.

common/validators.py
NDevox: 🤔  that makes sense.

footnotes/views.py
andyhd: That's a good idea

common/models.py
NDevox: Well we already have `get_version` for the queryset already with this: https://github.com/uktrade/tamato/pull/44/files/164f49c5dd4cd332f3027241984a1cd67619a5eb#diff-e6ef395a6a8073fbfdf98ad843cebe14L64

This new method kind of copies that, but with an order_by and as a filter, not a get. And some convenience around getting the identifying fields.

I'd probably prefer to adapt the current method on the queryset to do a filter (or a get if `many=False` or similar).

Although being specific to the object instance maybe this makes more sense... But then the queryset method doesn't serve as much of a purpose.

i.e. is it worth having both?

footnotes/tests/bdd/test_edit_footnote.py
NDevox: Yeah this is true, it should return `valid_user` then surely?

andyhd: I don't think so. I think the options are to either 1. remove this clause for now, or 2. change the feature specification to something like "Given a valid user named Bob with permission to update footnotes, And I am logged in as Bob"

footnotes/validators.py
```diff
@@ -84,6 +88,26 @@ def validate_footnote_description_start_date_before_footnote_end_date(
...
+        type(footnote)
+        .objects.active()
+        .filter(
+            workbasket__status=WorkflowStatus.PUBLISHED,
+            footnote_id=footnote.footnote_id,
+            footnote_type__footnote_type_id=footnote.footnote_type.footnote_type_id,
+        )
+    )
+    if (
+        footnote.update_type == UpdateType.CREATE
```
simonwo, line 105: Makes sense, is it definitely possible to never `UPDATE` the type and ID? I.e. do we have actual code that prevents that?

footnotes/views.py
simonwo: There's a difference between "terminate" and "delete" – the former is end-dating something, which is an `UPDATE` in `TrackedModel` land and the latter is just a straight up "this object no longer exists" which is a `DELETE` in `TrackedModel` land and doesn't require things to be end-dated.

A good question is why we'd use the latter, maybe just to correct or "annul" errors? I'm still not very clear on why the EU uses them.

workbaskets/models.py
```diff
@@ -145,10 +149,46 @@ def clean(self):
...
+            status=data["status"],
+            created_at=datetime.fromisoformat(
+                data["created_at"].replace("Z", "+00:00")
+            ),
+            updated_at=datetime.fromisoformat(
+                data["updated_at"].replace("Z", "+00:00")
+            ),
+            author_id=int(data["author_id"]),
+            approver_id=int(data["approver_id"]) if data["approver_id"] else None,
+        )
```
simonwo, line 180: Purely out of interest, how come we don't store just the ID and then recover the workbasket from the db?

andyhd: This is probably premature optimization, but populating a non-database-bound object with the unserialized session JSON saves a database query on what is likely to be every request.

footnotes/tests/bdd/test_edit_footnote.py
NDevox: 2 is better imo

workbaskets/models.py
NDevox: Ultimately I think we're going to need a backend cache pretty soon for a lot of models. Things like the measurement units and geo areas - which will practically never change - just don't make sense to be constantly queried - especially when they're used _everywhere_.

Might make sense to put recent workbaskets in there as well.

footnotes/validators.py
```diff
@@ -84,6 +88,26 @@ def validate_footnote_description_start_date_before_footnote_end_date(
...
 
+def validate_unique_type_and_id(footnote):
+    """FO2"""
+
+    # check against active footnotes which are published (or sent to cds?)
+    footnotes_with_type_and_id = (
+        type(footnote)
+        .objects.active()
+        .filter(
+            workbasket__status=WorkflowStatus.PUBLISHED,
```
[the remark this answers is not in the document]
andyhd: Do we want to error if there's an `APPROVED` but not `PUBLISHED` workbasket containing a new footnote with the same type and ID? I suppose otherwise we'll get a sort of race condition when submitting workbaskets. We'll have to make sure the error message is clear. Should users browsing the tariff see `APPROVED` but not `PUBLISHED` records?

footnotes/models.py
```diff
@@ -79,33 +85,45 @@ class Footnote(TrackedModel, ValidityMixin):
...
+        )
+
+    def get_descriptions(self, workbasket=None):
+        descriptions = FootnoteDescription.objects.filter(
+            described_footnote__footnote_id=self.footnote_id,
+            described_footnote__footnote_type=self.footnote_type,
+        )
+        query = Q(workbasket__status=WorkflowStatus.PUBLISHED,)
+        if workbasket is not None:
+            query |= Q(workbasket=workbasket)
```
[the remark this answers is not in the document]
andyhd: Yes, I was thinking that the `TrackedModelQuerySet` needs a `database_live_or_in_approved_workbasket` method, but I didn't implement it yet, as I wanted to see what you and @NDevox thought about this.

footnotes/validators.py
andyhd: The form fields are disabled, and the UI POST request handler ignores anything but the start and end dates. The REST API currently allows updating everything, so we'll need to tighten that up.

PR #45, opened 2020-07-31
TP-257: Merge taric-importer into TaMaTo.

NDevox:
TP-257: Merge taric-importer into TaMaTo.

Currently the taric importer is separated from the main TaMaTo application but heavily relies on
the TaMaTo code. This commit combines the two projects so that the importing of specific
models can be handled close to where the models are defined, much in the same manner as the
exporter is handled.

The original repo was held at https://github.com/uktrade/tamato-taric-import/. The
majority of the code has been unchanged. However one important change has been made.
The individual importing requirements of each component is now decoupled from the rest of
the importer code. This is done by creating a registration mechanism on the ElementHandler.
This allows developers to simply decorate the component handler with a `register_component`
decorator from any parent component class, in doing so the current element will be included
as a potential child of the parent component.

E.g:

```python3
    class NormalChildComponent(ElementHandler):
        tag = Tag("normal-child")
    
    class ParentComponent(ElementHandler):
        tag = Tag("parent")
        normal_child = NormalChildComponent()
    
    
    @ParentComponent.register_component("registered_child")
    class RegisteredChildComponent(ElementHandler):
        tag = Tag("registered-child")
```
The above snipper will create a ParentComponent with the tag "parent", within this parent
will be both the "normal_child" component, but by using the `register_component` decorator
it will also include the "registered_child" component. Thus the code for the individual
 components is decoupled from the main importer code.

To further facilitate this the code now makes use of the `ready` hook provided on django
application configs. A common one has been provided which automatically tries to import any
import handlers in a package. The settings have been updated to ensure these configs are now
used. Without this the import_handlers files are never imported and the components therefore
don't register themselves.

The main command for importing has also been wrapped into a django management command. This
can be run with:

    python manage.py import_taric filename --status some_status --username some_username


importer/management/commands/import_taric.py
```diff
@@ -0,0 +1,70 @@
...
+
+            if event == "start_ns":
+                nsmap.update([elem])
+
+            if event == "end":
+                handler.end(elem)
+
+
+class Command(BaseCommand):
+    help = "Import a given wordpress CSV file into the DB."
```
andyhd, line 33: Not very helpful 😄

importer/taric.py
```diff
@@ -0,0 +1,101 @@
...
+    transaction_id = TextElement(Tag("transaction.id"))
+    record_code = TextElement(Tag("record.code"))
+    subrecord_code = TextElement(Tag("subrecord.code"))
+    sequence_number = TextElement(Tag("record.sequence.number"))
+    update_type = TextElement(Tag("update.type"))
+
+    def save(self, data, workbasket_id):
+        method_name = {"1": "update", "2": "delete", "3": "create",}[
+            data["update_type"]
+        ]
```
andyhd, line 26: While you are touching this, it would be good to use `common.models.UpdateType` here

measures/import_handlers.py
```diff
@@ -0,0 +1,76 @@
...
+    def create(self, data, workbasket_id):
+        serializer = MeasureSerializer(data=data)
+        serializer.is_valid(raise_exception=True)
+        data = serializer.validated_data
+        data.update(workbasket_id=workbasket_id)
+        logging.debug(f"Creating Measure: {data}")
+        serializer.create(data)
+
+
+class MeasureComponent(Writable, ElementHandler):
```
andyhd, line 42: Do the rest of these need a `@Record.register_child("")` decorator?

importer/management/commands/import_taric.py
NDevox: 🤦

importer/taric.py
NDevox: Good idea

measures/import_handlers.py
NDevox: I've intentionally not done this as it was not done in the importer repo (as in they weren't included in the Record class).

I think we can start including the rest as we write importing code for them.

PR #46, opened 2020-08-05
Fix querysets with multiple versions of models

andyhd:
This change refactors the `CurrentWorkBasketMixin`, moving the queryset filtering code to
a new method `with_workbasket()` on the `TrackedModelQuerySet` class, and ensuring that only the latest version of models in
the workbasket are added to the queryset.

Additionally, if there is already a draft of an object in the current workbasket, this is now reused, rather
than creating a new draft.

This fixes the bug seen when editing the same footnote more than once. Multiple drafts
in the workbasket were added to the queryset, which caused an exception in the
`DetailView` which expects a single object.

### DIFF
--- a/common/models.py
+++ b/common/models.py
@@ -101,7 +101,7 @@ class TrackedModelQuerySet(PolymorphicQuerySet):
         query = Q()
 
         # get models in the workbasket
-        in_workbasket = self.model.objects.filter(workbasket=workbasket)
+        in_workbasket = self.filter(workbasket=workbasket)
 
         # remove matching models from the queryset
         for instance in in_workbasket:
@@ -249,3 +249,10 @@ inherit TrackedModel must either:
 
     def validate_workbasket(self):
         pass
+
+    def add_to_workbasket(self, workbasket):
+        if workbasket == self.workbasket:
+            self.save()
+            return self
+
+        return self.new_draft(workbasket=workbasket)
--- a/common/models.py
+++ b/common/models.py
@@ -94,6 +94,10 @@ class TrackedModelQuerySet(PolymorphicQuerySet):
         """
         Add the latest versions of objects from the specified workbasket.
         """
+
+        if workbasket is None:
+            return self
+
         query = Q()
 
         # get models in the workbasket
--- a/common/models.py
+++ b/common/models.py
@@ -90,6 +90,27 @@ class TrackedModelQuerySet(PolymorphicQuerySet):
         """
         return self.get_version(predecessor__isnull=True, **kwargs)
 
+    def with_workbasket(self, workbasket):
+        """
+        Add the latest versions of objects from the specified workbasket.
+        """
+        query = Q()
+
+        # get models in the workbasket
+        in_workbasket = self.model.objects.filter(workbasket=workbasket)
+
+        # remove matching models from the queryset
+        for instance in in_workbasket:
+            query &= ~Q(
+                **{
+                    field: getattr(instance, field)
+                    for field in self.model.identifying_fields
+                }
+            )
+
+        # add latest version of models from the current workbasket
+        return self.filter(query) | in_workbasket.filter(successor__isnull=True)
+
 
 class PolymorphicMPTreeQuerySet(TrackedModelQuerySet, MP_NodeQuerySet):
     """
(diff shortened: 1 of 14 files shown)


workbaskets/views/generic.py
```diff
@@ -0,0 +1,40 @@
...
+            self.object = self.object.new_draft(
+                workbasket=workbasket,
+                update_type=UpdateType.UPDATE,
+                **{
+                    field_name: form.cleaned_data.get(field_name)
+                    for field_name in self.update_fields
+                },
+            )
+
+        return HttpResponseRedirect(self.get_success_url())
```
simonwo, line 40: Won't this also apply if there is an `UpdateType.CREATE`? We just want to update the created model?

common/models.py
```diff
@@ -90,6 +90,31 @@ def get_first_version(self, **kwargs):
...
+        Add the latest versions of objects from the specified workbasket.
+        """
+
+        if workbasket is None:
+            return self
+
+        query = Q()
+
+        # get models in the workbasket
+        in_workbasket = self.model.objects.filter(workbasket=workbasket)
```
NDevox, line 104: Is there a reason we can't just do `self.filter`? Wondering if there is a reason to start with a brand new queryset.

footnotes/views.py
```diff
@@ -89,16 +70,20 @@ def get_object(self, queryset=None):
...
 
 
-class FootnoteDetail(CurrentWorkBasketMixin, FootnoteMixin, generic.DetailView):
-    model = models.Footnote
+class FootnoteDetail(WithCurrentWorkBasket, FootnoteMixin, generic.DetailView):
     template_name = "footnotes/detail.jinja"
 
+    @property
+    def queryset(self):
+        # laxy evaluation to allow freezegun to work
```
NDevox, line 78: typo

workbaskets/models.py
```diff
@@ -179,6 +179,13 @@ def from_json(cls, data):
             approver_id=int(data["approver_id"]) if data["approver_id"] else None,
         )
 
+    @classmethod
+    def current(cls, request):
+        """Get the current workbasket in the session"""
+
+        if "workbasket" in request.session:
+            return cls.from_json(request.session["workbasket"])
```
NDevox, line 187: Wonder if this can be made smarter to try and get a workbasket from the DB if a session one isn't available.

Probably not worth bothering now until we have a better idea.

workbaskets/views/generic.py
NDevox: If it's a `CREATE` we won't be making a new draft. So the `else` block won't work. The other will though...

Interesting idea, may be able to break this down into a `handle_preexisting_draft` and `handle_new_draft` or similar methods.

footnotes/views.py
andyhd: how do you know i didn't mean to type laxy?

NDevox: 🤣

workbaskets/models.py
andyhd: What would the current workbasket from the database be?

common/models.py
andyhd: It seems to work with `self.filter`, so I'll change it to that.

workbaskets/models.py
NDevox: Might be able to do something based on user and workbaskets which are not published or errors.

I don't have any good suggestions, so happy to ignore.

workbaskets/views/generic.py
andyhd: This view is only for updates, so perhaps when we come to implement the "create footnote" view, we can refactor this. Actually, this code would not work if there was a `CREATE` already in the workbasket, because it changes the `update_type` to `UPDATE`, so I'll fix that now.

@simonwo moving the logic to `TrackedModel.save()` might be a good idea, I'll try it out. I would hope that the migrations wouldn't have multiple edits/creates/deletes of the same object in a single workbasket/transaction, though. Still, it's good to be defensive!

common/models.py
NDevox: I think if we're doing `.active().in_workbasket(workbasket)` we're probably only wanting active instances in the given workbasket anyway, so should still work.

workbaskets/models.py
andyhd: Interesting suggestion - so if a user has only 1 unsubmitted workbasket, default to that one?

NDevox: Something like that.

I think this falls under user needs. Maybe a good UX question for Hannah

PR #47, opened 2020-08-05
TP-206: Add importer for Goods Nomenclature and related models.

NDevox:
It is imperative that legacy data from the TARIC can be imported at will into the
TaMaTo database. Given the Goods Nomenclature models are now ready it is necessary
that the importer is setup for these models.

As this is the first set of models using the integrated importer there are a number
of changes made to the importing system to remove a few bugs and make it easier to
use the system for more models in the future.

A few discoveries of varying complexity have also been found. Firstly merged
models do not fit well with the current importer architecture. These instances have
been passed over to be done later as a way of adapting the importer to all models
affected by this (the majority of models in the system). The second is that, as
the hierarchical tree is largely unrepresented by the legacy TARIC system, this
will possibly require an entire implementation unto itself, again to be implemented
at a later date.

Tests have been added for all GN importer handlers - those which are currently not
working have been skipped.

### DIFF
--- a/commodities/import_handlers.py
+++ b/commodities/import_handlers.py
@@ -1,20 +1,47 @@
+import logging
+
+from commodities import serializers, models
+from common.validators import UpdateType
+from footnotes.models import Footnote, FootnoteType
 from importer.handlers import ElementHandler
 from importer.handlers import TextElement
 from importer.handlers import ValidityMixin
 from importer.handlers import Writable
 from importer.namespaces import Tag
+from importer.taric import Record
 
 
+logger = logging.getLogger(__name__)
+
+
+@Record.register_child("goods_nomenclature")
 class GoodsNomenclature(ValidityMixin, Writable, ElementHandler):
+    serializer_class = serializers.GoodsNomenclatureSerializer
+
     tag = Tag("goods.nomenclature")
     sid = TextElement(Tag("goods.nomenclature.sid"))
     item_id = TextElement(Tag("goods.nomenclature.item.id"))
-    productline_suffix = TextElement(Tag("producline.suffix"))  # XXX not a typo
+    suffix = TextElement(Tag("producline.suffix"))  # XXX not a typo
     valid_between_lower = TextElement(Tag("validity.start.date"))
-    statistical_indicator = TextElement(Tag("statistical.indicator"))
+    valid_between_upper = TextElement(Tag("validity.end.date"))
+    statistical = TextElement(Tag("statistical.indicator"))
 
 
+@Record.register_child("goods_nomenclature_origin")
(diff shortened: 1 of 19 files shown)


commodities/tests/test_importer.py
```diff
@@ -0,0 +1,96 @@
...
+from common.tests.factories import FootnoteAssociationGoodsNomenclatureFactory
+from common.tests.util import generate_test_import_xml, requires_interdependent_export
+from importer.management.commands.import_taric import import_taric
+from workbaskets.models import WorkflowStatus
+
+pytestmark = pytest.mark.django_db
+
+
+def test_goods_nomenclature_importer_create(valid_user):
+    good = factories.GoodsNomenclatureFactory.build(update_type=3)
```
andyhd, line 14: Should use `UpdateType.CREATE` here

measures/import_handlers.py
```diff
@@ -1,6 +1,7 @@
 import logging
 
 from importer.taric import Record
+from measures import serializers
```
andyhd, line 4: This makes the following import redundant, doesn't it? And is the `Measure.create` method also redundant?

commodities/import_handlers.py
```diff
@@ -46,13 +77,48 @@ class GoodsNomenclatureIndents(ValidityMixin, Writable, ElementHandler):
     goods_nomenclature_item_id = TextElement(Tag("goods.nomenclature.item.id"))
     productline_suffix = TextElement(Tag("productline.suffix"))
 
+    def clean(self):
+        super().clean()
+        good = models.GoodsNomenclature.objects.get_latest_version(
+            sid=self.data.pop("goods_nomenclature_sid")
+        )
+        self.data["indented_goods_nomenclature"] = good
```
andyhd, line 85: Not a problem with this code specifically, but I've been second-guessing the way we use foreign keys on our models. Do we need to update foreign keys when we create a new version of a model?

NDevox: I've been having the same thought. There aren't any other unique fields to attach an FK to so don't think there's a good alternative.

Currently thinking we should keep the FK, but maybe change how the relation manager works - make it so instead of just getting the FK object via the PK, it should instead get the directly related one, fetch the identifying fields and then make another query to fetch a range of the objects based on the validity periods. Same would have to be done for the reverse manager.

NDevox: Doesn't feel particularly efficient to me though

commodities/tests/test_importer.py
NDevox: Yeah, it should.

measures/import_handlers.py
NDevox: Yep, good point.

PR #49, opened 2020-08-06
New ADRs for discussion

simonwo:
Here are some new ADRs that are the result of TP-156 exploration. They are "discussion-ready" but not necessarily completely thought through yet.

0008 is probably not hugely controversial.
0009 is a bit moreso, and likely less well motivated too.

### DIFF
--- a/docs/adr/0009-record-generated-envelopes.rst
+++ b/docs/adr/0009-record-generated-envelopes.rst
@@ -1,0 +1,173 @@
+.. _9-generate_envelope_files:
+
+9. Generate envelope files
+==========================
+
+Date: 2020-10-14
+
+Status
+------
+
+Proposed
+
+Context
+-------
+We must geneate XML envelope files as a way of communicating tariff data to HMRC
+and our other consumers. The spec for envelope files is defined as an XSD
+schema. An envelope file contains multiple transactions.
+
+The way in which database models are packed into transactions is partly defined
+by the TARIC spec. HMRC systems place additional requirements on the
+transactions that are still partially undefined.
+
+We have known since the beginning that CDS would accept or reject changes sent
+from TaMaTo. We have also had some notion that this happens in two stages with
+different validations applied at each stage. The first stage will reject an
+entire envelope if it contains erroneous data. The second stage will only reject
+specific transactions. It was, until recently, unknown what CDS expects as
+replacement data in either of these failure modes.
+
+We have since learnt more about interacting with the CDS system and the use of
+TARIC3 XML. Particularly around the metadata associated with the interface in
+the form of a gateway API call, envelopes, transactions and messages. This has
+come from a better understanding of the new HMRC design and through further
+review of the pre-existing Trade Tariff Management (TTM) codebase.
+
+The envelope, transaction or message ID are always set automatically in TTM and
+can't be manually controlled. The envelope ID is set based on the current date
+and a sequence number. The transaction and message IDs are both set sequentially
+on a per-envelope basis. When a validation error occurs at CDS a new envelope ID
+is used.
+
+HTTPS interface
+~~~~~~~~~~~~~~~
+
+HTTPS clients already expect TARIC3 XML over an HTTPS API. The existing HTTPS
+API in TTM provides TARIC3 XML envelopes and metadata about them including the
+envelope ID. There may be little to substantially change about this API – it
+returns a list of envelopes generated per day.
+
+There is no limit on the size of returned envelopes and there is no limit on the
+number of envelopes generated in one day.
+
+HMRC interface
+~~~~~~~~~~~~~~
+
+According to HMRC envelope and transaction IDs affect the sequence numbers used
+by CDS for processing. TaMaTo should use the same envelope ID when an envelope
+is rejected. Transaction IDs should also be reused when individual transactions
+are errored. It is possible in the past that data that errored in CDS was
+corrected manually. As we are using a new API pattern from HMRC it may be more
+difficult to correct things manually.
+
+Metadata submitted to HMRC must include a checksum and file size. The filename
+should match the envelope ID.
+
+The files transferred have a maximum file size. Rather than trying to ensure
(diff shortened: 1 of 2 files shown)


docs/adr/0009-record-generated-envelopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+Date: 2020-08-05
+
+## Status
+
+Proposed
+
+## Context
+
+We have known since the beginning that CDS would accept or reject changes sent from TaMaTo, and have had some notion that this happens in two stages with different validations applied at each stage – the first stage will reject an entire envelope if it contains erroneous data, whilst the second stage will only reject specific transactions. It has been unknown what CDS expects as replacement data in either of these failure modes.
```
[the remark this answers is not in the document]
matthewford: I no longer have access to the DIT drive files, but this was discussed within DIT and with HMRC extensively a few years back.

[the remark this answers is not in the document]
simonwo: Thanks @matthewford – we’ve talked with them again about it and got some additional info, and actually I’m just working on updating this. Spurred on by knowing that this was discussed I’ll also go and search again and see what I can find with that extra context. Cheers!

docs/adr/0009-record-generated-envelopes.rst
```diff
@@ -0,0 +1,172 @@
...
+can't be manually controlled. The envelope ID is set based on the current date
+and a sequence number. The transaction and message IDs are both set sequentially
+on a per-envelope basis. When a validation error occurs at CDS a new envelope ID
+is used.
+
+HTTPS interface
+~~~~~~~~~~~~~~~
+
+HTTPS clients already expect TARIC3 XML over an HTTPS API. The existing HTTPS
+API in TTM provides TARIC3 XML envelopes and metadata about them including the
```
stuaxo, line 46: Is there somewhere to learn about the existing API in Ruby Tap, could be worth linking to a footnote here?

Either link to docs on confluence, or it would be nice to know about concrete examples.

docs/adr/0009-record-generated-envelopes.rst
```diff
@@ -0,0 +1,172 @@
...
+~~~~~~~~~~~~~~
+
+According to HMRC envelope and transaction IDs affect the sequence numbers used
+by CDS for processing. TaMaTo should use the same envelope ID when an envelope
+is rejected. Transaction IDs should also be reused when individual transactions
+are errored. It is possible in the past that data that errored in CDS was
+corrected manually. As we are using a new API pattern from HMRC it may be more
+difficult to correct things manually.
+
+Metadata submitted to HMRC must include a checksum and file size. The filename
```
stuaxo, line 63: Do they specify any kind of checksum or do we have a choice of implementation ?

stuaxo: Edit:  I noticed they have an implementation field, but not sure what they accept, but maybe that's outside the scope of the ADR.

simonwo: Yeah we have some separate documentation about that, will send you a link.

docs/adr/0009-record-generated-envelopes.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Consequences
+------------
+
+As the rules for forming transactions are complex, and partially dependent on
+what has already been added to an envelope, it does not necessarily make sense
+to keep track internally of what models have ended up in what transactions.
+However, we still expect both transactions and envelopes to come back as
+"ERRORED" from HMRC. It might make sense to retire our current table of
+transactions and only use workbaskets for internal grouping purposes.
```
NDevox, line 157: I think the discussion on this was not entirely conclusive, but (I) was leaning towards shifting the relationships around so we had 1-many relationships for Transaction-TrackedModels, and then 1-many relationships for Workbasket-Transactions. With the models being re-sorted among the transactions as required before being submitted to HMRC (which can be done as often as possible as this is outside the AO system).

Given this statement in the ADR is also inconclusive, I don't think anything needs to be added. Just thought this would be worth stating in case we need to clarify our understandings here..

NDevox (approved):
All good. One point of clarification which I don't think needs to be added, just wondering if it merits further discussion outside the ADR.

docs/adr/0009-record-generated-envelopes.rst
```diff
@@ -0,0 +1,172 @@
...
+generated to wholly contain the passed workbaskets and this might require
+multiple files. E.g:
+
+.. code:: python
+
+    def generate_envelopes(
+        envelope_id: int,
+        transaction_counter: Callable[[], int],
+        *workbaskets: WorkBasket
+    ):
```
stuaxo, line 116: envelope_id is the first id here ?

Since envelopes may end up being split up I guess this would be actually be a counter ?

I'm a little bit confused how this transaction counter thing is supposed to work.... you get a sequence of a list of callables that you call one per transaction ?


It would probably make more sense to describe the expected output here, e.g.  for a set of transactions 1, 2, 3 - they would have ids X, Y, Z.

docs/adr/0009-record-generated-envelopes.rst
simonwo: Yeah I didn't really want to add any more detail until we'd done some more work in this area, which is the stuff we're doing now. So I'm happy to merge it and remember this as something to revisit.

docs/adr/0009-record-generated-envelopes.rst
simonwo: You are quite right, really envelope ID should be a counter.

The coutner interface is basically anything that works like [this function](https://github.com/uktrade/tamato/blob/master/common/renderers.py#L6) – i.e. it's a function you can call which will return the next `int` in the sequence, starting at some value.

But this function signature is just meant to be illustrative – it might make sense for the function itself to call out to the database where we have stored these things, etc…

PR #50, opened 2020-08-12
Add quotas data model

andyhd:
A major component of the UK tariff is quotas, or quantitative restrictions, which allow
goods to be imported at a modified tariff (usually lower) up to a certain volume or
value.  Quotas are used to balance protecting domestic producers of goods from
competition from imports with allowing exporters access to the market and allowing
domestic industries to import manufacturing resources.

An importer can specify which quota they wish to make use of by providing the Quota
Order Number. The order number is associated with one or more Quota Definitions, which
detail the volumes and periods for which the quota rate is available. The order number
is also associated with an origin (usually a country, or group of countries) to enable
setting different quota volumes and tariffs for different origins.

A quota can also be subdivided into sub-quotas, which may have varying validity periods,
origins or tariffs.

A quota can also be suspended, or blocked.

This PR introduces the data model for quotas and the business rules to support data
integrity.

In addition, a couple of common data model field types are introduced - NumericSID and
ShortDescription - which represent frequently used data types in the TARIC data model.

Also, the WorkflowStatus enum class is moved to prevent a circular dependency.
Unfortunately, this changed touched a lot of files.


common/models.py
```diff
@@ -115,6 +116,17 @@ def with_workbasket(self, workbasket):
...
+    def approved(self):
+        """
+        Get objects which have been approved/sent-to-cds/published
+        """
+        return self.exclude(
+            workbasket__status__in=(
+                WorkflowStatus.NEW_IN_PROGRESS,
+                WorkflowStatus.EDITING,
+            )
+        )
```
NDevox, line 128: 🤔  this feels backwards, should we not filter for status in the approved states? As opposed to not in unapproved states

common/models.py
```diff
@@ -256,3 +268,24 @@ def add_to_workbasket(self, workbasket):
...
+        del kwargs["validators"]
+        return name, path, args, kwargs
+
+
+class ShortDescription(models.CharField):
+    def __init__(self, *args, **kwargs):
+        kwargs["max_length"] = 500
+        kwargs["blank"] = True
+        kwargs["null"] = True
+        super().__init__(*args, **kwargs)
```
NDevox, line 291: This is pretty sensible, wish we'd done it earlier!

quotas/models.py
```diff
@@ -0,0 +1,242 @@
...
+    qualifier code) are mutually exclusive.
+    """
+
+    record_code = "370"
+    subrecord_code = "00"
+
+    sid = NumericSID()
+    order_number = models.ForeignKey(QuotaOrderNumber, on_delete=models.PROTECT)
+    volume = models.DecimalField(max_digits=14, decimal_places=3)
+    initial_volume = models.DecimalField(max_digits=14, decimal_places=3)
```
NDevox, line 104: Always a bit hesitant to use decimals over plain old integer fields.

I suppose it makes sense with the measurement units. Just a thought.

common/models.py
andyhd: It is backwards, but only because there are more approved states than unapproved. I'll be more explicit and less lazy.

quotas/models.py
```diff
@@ -0,0 +1,242 @@
...
+    quota_critical_threshold = models.PositiveSmallIntegerField(
+        validators=[validators.validate_percentage]
+    )
+    description = ShortDescription()
+    mechanism = models.PositiveSmallIntegerField(
+        choices=validators.AdministrationMechanism.choices
+    )
+    category = models.PositiveSmallIntegerField(
+        choices=validators.QuotaCategory.choices
+    )
```
simonwo, line 132: So when we discussed we decided to put them on the order number and not the definition – can't forsee any issue with them being here really, but any reason why we've gone with this model instead?

quotas/models.py
andyhd: I would normally stick to integers, but decimals are the best fit to the TARIC data type. Happy to change if we have a good reason.

quotas/models.py
```diff
@@ -0,0 +1,242 @@
...
+        return super().save(*args, **kwargs)
+
+    def validate_workbasket(self):
+        validators.validate_exclusion_only_from_group_origin(self)
+        validators.validate_excluded_geo_area_must_be_member_of_origin_group(self)
+
+
+class QuotaDefinition(TrackedModel, ValidityMixin):
+    """Defines the validity period and quantity for which a quota is applicable. This
+    section defines as well quota as sub-quota as such.
```
simonwo, line 92: My English parser cannot grok this

quotas/models.py
simonwo: Does max digits include the decimal digits? XML specifies 11 + 3.

Based on the current data we don't get more than 10 integer digits and 2 decimal digits though. There are only 255 instances where the decimal digits are not `00`.

quotas/models.py
```diff
@@ -0,0 +1,242 @@
...
+    # )
+    # measurement_unit = models.ForeignKey(
+    #     "measures.MeasurementUnit", on_delete=models.PROTECT, null=True, blank=True,
+    # )
+    # measurement_unit_qualifier = models.ForeignKey(
+    #     "measures.MeasurementUnitQualifier",
+    #     on_delete=models.PROTECT,
+    #     null=True,
+    #     blank=True,
+    # )
```
simonwo, line 117: We will need to validate that the combination of unit and qualifier appears in the `measurements` table.

quotas/validators.py
```diff
@@ -0,0 +1,315 @@
...
+            raise ValidationError(
+                "Whenever a sub-quota is defined with the 'equivalent' type, it must have "
+                "the same volume as the ones associated with the parent quota."
+            )
+
+        if association.coefficient == Decimal("1.00000"):
+            raise ValidationError(
+                "A sub-quota defined with the 'equivalent' type must have a "
+                "coefficient not equal to 1"
+            )
```
simonwo, line 279: Do we need the opposite check for Normal quotas?

quotas/models.py
```diff
@@ -0,0 +1,242 @@
...
+    description = ShortDescription()
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    def __str__(self):
+        return self.description
+
+    def validate_workbasket(self):
```
simonwo, line 202: Do we also need to validate that it has only been applied to a FCFS quota?

quotas/models.py
```diff
@@ -0,0 +1,242 @@
...
+    blocking_period_type = models.PositiveSmallIntegerField(
+        choices=validators.BlockingPeriodType.choices
+    )
+    description = ShortDescription()
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
+    def validate_workbasket(self):
```
simonwo, line 223: Do we also need to validate that it has only been applied to a FCFS quota?

quotas/models.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class QuotaEvent(TrackedModel):
+    """We do not care about quota events, except to store historical data. So this model
+    stores all events in a single table."""
+
+    record_code = "375"
+    subrecord_code = models.CharField(
+        max_length=2, choices=validators.QuotaEventType.choices
+    )
```
simonwo, line 234: Spicy! 👍🏻

quotas/validators.py
```diff
@@ -0,0 +1,315 @@
...
+    order_numbers_with_overlapping_validity = (
+        type(order_number)
+        .objects.approved()
+        .filter(
+            order_number=order_number.order_number,
+            valid_between__overlap=order_number.valid_between,
+        )
+    )
+    if (
+        order_number.update_type == UpdateType.CREATE
```
simonwo, line 118: Can we not `UPDATE` the dates? What about correcting a typo?

importer/management/commands/import_taric.py
```diff
@@ -1,9 +1,10 @@
-from django.core.management import BaseCommand
 import xml.etree.ElementTree as etree
 
-from workbaskets.models import WorkflowStatus
-from importer.taric import Envelope
+from django.core.management import BaseCommand
+
 from importer.namespaces import nsmap
+from importer.taric import Envelope
+from workbaskets.validators import WorkflowStatus
```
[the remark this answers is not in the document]
andyhd: It is, and I'm sure I can add it to the git hooks - we already run black.

quotas/models.py
```diff
@@ -0,0 +1,242 @@
...
+
+from common.models import NumericSID
+from common.models import ShortDescription
+from common.models import TrackedModel
+from common.models import ValidityMixin
+from quotas import validators
+
+
+class QuotaOrderNumber(TrackedModel, ValidityMixin):
+    """The order number is an alternative identification of a measure. It is defined for
```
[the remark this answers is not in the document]
andyhd: This is copy-pasta from the TARIC documentation.

quotas/models.py
andyhd: Again copy-pasted from the TARIC docs. I'll see if I can reword it.

quotas/models.py
andyhd: Yes, `max_digits` includes the decimal digits.

quotas/models.py
andyhd: Yep, but the measurements table doesn't exist yet

quotas/models.py
andyhd: Oops, my mistake!

quotas/validators.py
andyhd: If we are doing an `UPDATE` then, unless we are changing the validity period, we will overlap with the `CREATE` version.

But now that I look at this again, the correct thing to do i filter out the previous versions of this order_number when checking for overlaps.

quotas/validators.py
andyhd: You would think so...

I'll add it

quotas/models.py
andyhd: Is it worth reducing this to 10 + 2?

importer/management/commands/import_taric.py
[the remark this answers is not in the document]
andyhd: Done in https://github.com/uktrade/tamato/pull/53

quotas/models.py
simonwo: I think leave it as it is, given that the schema is specified like that.

PR #56, opened 2020-08-19
TP-271: Implement interdependent and linked models importer

NDevox:
The current importer, whilst very effective at extracting data from XML
in a simple to use manner, is naive to a some of the complexities
around the TARIC dataset and the changes that have been made to the
data models. Two prime examples of this are models which have been merged
and how foreign keys are transmitted.

Merged models has the consequence of the required data for each models being
split across several records within the incoming data. Whereas for Foreign Keys
have no guarantee of being processed in the correct order.

As a result there is no guarantee that at any one time the data given from a
TARIC record will be complete or ready for processing.

To account for this a pipeline has been constructed which consists of three new
concepts introduced and the parser concept partially redefined:

1. Parsers

Parsers have now been relegated to only handle extracting data from the XML
and packaging it for processing elsewhere. This keeps the parser classes
lightweight and simple to design.

2. Handlers

Handlers are at the other end of the pipeline from parsers. They take the parsed
data and match it with other data until a complete dataset for an object has been
obtained. They also handle fetching foreign keys. The end result of a handler should
be a database object. This is the most complex new concept introduced and comes with
a lot of customisability for specific models, however at it's simplest it should be
trivial to define.

3. The Nursery

The nursery is a caretaker process which effectively orchestrates the pipelines.
It receives data from parsers and matches them to the handlers. If the data turns
out to be incomplete (determined by how successful a handler is) it stores the data
to be matched against later inputs.

4. The Cache

The nursery needs to store the incomplete data for later processing, it does this in
a "cache". Currently this is implemented as a process level dictionary, but can easily
be adapted to some more persistent or distributed storage solutions down the road.

All the above works together to make an asynchronous data pipeline where partial data
for an object can be sent at any time and can be patched together at a later point.

Some example handlers and parsers have been defined to showcase this, specifically
the `FootnoteAssociationGoodsNomenclatureHandler` in the commodities handlers
showcasing how links work. And the `FootnoteTypeHandler` together with the
`FootnoteTypeDescriptionHandler` showcasing how interdependent models work.


commodities/import_handlers.py
```diff
@@ -1,124 +1,64 @@
...
-    goods_nomenclature_item_id = TextElement(Tag("goods.nomenclature.item.id"))
-    productline_suffix = TextElement(Tag("productline.suffix"))
+    tag = parsers.GoodsNomenclatureDescriptionPeriodParser.tag.name
 
 
-@Record.register_child("goods_nomenclature_indent")
-class GoodsNomenclatureIndents(ValidityMixin, Writable, ElementHandler):
+class GoodsNomenclatureIndentsHandler(BaseHandler):
     serializer_class = serializers.GoodsNomenclatureIndentSerializer
+    tag = parsers.GoodsNomenclatureIndentsParser.tag.name
```
NDevox, line 34: These aren't yet handled

certificates/import_handlers.py
```diff
@@ -0,0 +1,28 @@
+from certificates import import_parsers as parsers, serializers
```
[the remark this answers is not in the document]
andyhd: how did this import get past the formatter? should be
```python
from certificates import import_parsers as parsers
from certificates import serializers
```

NDevox:
# Continuing problems:

Ignoring the "special case" that is the commodity code tree, there is also still an issue with optional but existing foreign keys.

It's kind of hard to decide whether a key is optional or not in some cases. For example `GoodsNomenclatureOrigin`. There are definitely cases there where the key should be null. But the majority of cases it is required. But the null requirement means we don't check for it. More importantly the key comes in an "interdependent" record - this is a table that's been squashed into a single field. We don't really have a mechanism for waiting for interdependent records which are actually foreign keys. It's kind of a confusing state to be in.

I reckon there will be a relatively obvious path around more complex checks for whether a link is optional or not and how to wait (e.g. making the `optional` value either a bool or a callable which returns a bool). I don't really want to confront this issue here because it's a bit more niche and I want to keep things moving (ironic considering how long this branch took to get up).

# Possible changes

There is some meta programming in this implementation... specifically in handlers. Effectively all handlers need to be registered with the nursery, so to cater for this we do it on class definition via the BaseHandlers metaclass. This is actually also a reasonable route for handling the forward-referenced dependency issue. If a Handler defines another Handler as a dependency, then the metaclass can inspect these dependencies and auto-register the reverse dependency relationship. It would reduce boilerplater, but I'm already well aware the behaviour appears magical to anyone unaware of the metaclass. So I hesitate to introduce more magic, but honestly I think I'd prefer it.

certificates/import_handlers.py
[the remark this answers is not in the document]
NDevox: Ack.

Fixed

docs/adr/0010-convert-the-importer-to-a-pipeline.md
```diff
@@ -0,0 +1,264 @@
...
+
+The flow between these three components looks roughly like the following diagram: 
+
+![Importer Pipeline Flow Diagram](./images/importer-pipeline-flow.png)
+
+#### The XML Parser
+
+Initially an XML file is passed to the XML parser.
+
+The parser reflects the XML file structure with each elemnt being a `Parser` class. Each `Parser` class contains other
```
stuaxo, line 67: "with each element"

docs/adr/0010-convert-the-importer-to-a-pipeline.md
```diff
@@ -0,0 +1,264 @@
...
+`Parser` classes as attributes which reflect the elements expected to be  contained within the "parent" `Parser`s tag.
```
stuaxo, line 68: "xpected to be  contained"

docs/adr/0010-convert-the-importer-to-a-pipeline.md
```diff
@@ -0,0 +1,264 @@
...
+At the `Record` level of the TARIC3 XML the data is extracted, converted into a python dictionary, and emitted to the
+Nursery with the Workbasket ID and XML tag.
+
+The important factor here is how the data is extracted and put into the dictionary. The naming of each piece of data
+is integral to how the handler uses that data later on.
+
+##### Code example
```
stuaxo, line 75: I'd have the data that is going to be parsed before the code, so the reader doesn't have to scroll down and then back up again.

[not going to block on things staying as they are here]

importer/handlers.py
```diff
@@ -1,210 +1,411 @@
...
+
+    def __init__(
+        self, dispatched_object: DispatchedObjectType, nursery: TariffObjectNursery
+    ):
+        self.nursery = nursery
+
+        self.data = dispatched_object["data"]
+        if not self.identifying_fields:
+            self.identifying_fields = (
+                self.serializer_class.Meta.model.identifying_fields
```
andyhd, line 220: I've not tried, but will this be a type error if `self.identifying_fields` is `List` and `...model.identifying_fields` is a tuple? Or maybe cast to `list` here, in case `.model.identifying_fields` is some other iterable?

importer/utils.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    hash_input = tag
+    for key in identifying_fields:
+        hash_input += str(data[key])
+
+    return sha256(hash_input.encode()).digest()
+
+
+"""
+The current importer, whilst very effective at extracting data from XML
```
andyhd, line 63: Why is this doc comment here?

docs/adr/0010-convert-the-importer-to-a-pipeline.md
NDevox: Good idea

importer/handlers.py
NDevox: Not unless we run a static type checker - MyPy might throw an issue.

I've updated the type to `Iterable`

importer/utils.py
NDevox: 🤣  my bad, was copying my git commit message for some editing

PR #61, opened 2020-08-28
TP-271: Apply importer to additional codes models

NDevox:


additional_codes/import_handlers.py
```diff
@@ -0,0 +1,58 @@
...
+class AdditionalCodeTypeHandler(BaseHandler):
+    serializer_class = serializers.AdditionalCodeTypeSerializer
+    tag = parsers.AdditionalCodeTypeParser.tag.name
+
+
+@AdditionalCodeTypeHandler.register_dependant
+class AdditionalCodeTypeDescriptionHandler(BaseHandler):
+    dependencies = [AdditionalCodeTypeHandler]
+    serializer_class = serializers.AdditionalCodeTypeSerializer
+    tag = parsers.AdditionalCodeTypeDescriptionParser.tag.name
```
andyhd, line 16: I wonder if changing the abstraction from "dependant" to "aggregate" or "composite" might clarify this and allow the metaprogramming solution to be more self-explanatory? Eg:
```python
class AdditionalCodeTypeHandler(BaseHandler):
    serializer_class = ...
    tag = ...

class AdditionalCodeTypeDescriptionHandler(AggregateHandler):
    aggregate = AdditionalCodeTypeHandler  # get the serializer class from this class
    tag = ...
```
or perhaps something more explicit:
```python
class AdditionalCodeTypeHandler(ComponentHandler):
    tag = ...

class AdditionalCodeTypeDescriptionHandler(ComponentHandler):
    tag = ...

class CompositeAdditionalTypeHandler(CompositeHandler):
    components = AdditionalCodeTypeHandler, AdditionalCodeTypeDescriptionHandler
    serializer_class = ...
```

andyhd (approved):
This looks like a good solution - no TARIC tags are triply nested, right?

additional_codes/import_handlers.py
andyhd: Just thinking, not asking you to change it, ofc

NDevox: Yeah I agree with this. Needing something more explicit to handle the aggregate nature of things as opposed to this weird dependency hackery going on. It would make things much clearer and I think easier to work with in the background.

I'm happy to do it tbh. But my brain has been slowly melting over the last month of looking into this so not before the bank holiday weekend. If you do have a clear idea of how to implement it I would implore you for a PR!

NDevox:
Not seen any triple nested tags but I'm not going to say with 100% certainty that that excludes them.

~Blissful~ Naive ignorance until that day arrives.

additional_codes/import_handlers.py
stuaxo: Here, an aggregate is a record made up of several different records in TARIC, that in the old system would have been individual models in the database ?

andyhd: Yes. We deviate from the TARIC data model (rightly, or wrongly) in a few ways, the most common is that we merge the separate `Description` models into the models they describe, because the only reason they exist is to provide multiple language versions of the description, and we only need English.

stuaxo: In the first proposed syntax, it's not immediately obvious which class is the group, and which are the items, the second version is clearer in this respect.

NDevox: I agree with this, the second syntax is clearer to me too.

stuaxo: [2] has the nice propery of internal consistency (`components`  /  `ComponentHandler`), here is a tweak to see what it looks like if the "`handler`ness" is emphasized instead.

I wasn't sure if ComponentHandler was needed as it's own thing, so this has `BaseHandler` instead.

```python
class AdditionalCodeTypeHandler(BaseHandler):
    tag = ...

class AdditionalCodeTypeDescriptionHandler(BaseHandler):
    tag = ...

class CompositeAdditionalTypeHandler(CompositeHandler):
    handlers = AdditionalCodeTypeHandler, AdditionalCodeTypeDescriptionHandler
    serializer_class = ...
```

PR #62, opened 2020-09-03
TP-300: Apply importer to certificates models.py

NDevox:
All certificates models now work with the importer.

Tests have been simplified as the majority of importer tests would include
significant amounts of boilerplate. To handle this a new test utility has
been added and applied to older importer tests.

Also replaced tabs in the taric3.xsd with spaces. Because copying the XML
definitions for the parser docstrings was upsetting the PyCharm linter.


stuaxo (approved):
LGTM, though obviously I'm a bit new to this codebase :)

PR #63, opened 2020-09-04
TP-297 add footnotes importer

NDevox:
Does what it says on the tin

### DIFF
--- a/common/tests/factories.py
+++ b/common/tests/factories.py
@@ -102,7 +102,7 @@ class FootnoteDescriptionFactory(TrackedModelMixin, ValidityFactoryMixin):
 
     description = short_description()
     described_footnote = factory.SubFactory(FootnoteFactory)
-    description_period_sid = numeric_sid()
+    description_period_sid = factory.Sequence(lambda x: x + 1)
 
 
 class RegulationGroupFactory(TrackedModelMixin):
--- a/footnotes/import_handlers.py
+++ b/footnotes/import_handlers.py
@@ -1,4 +1,5 @@
 from footnotes import import_parsers as parsers
+from footnotes import models
 from footnotes import serializers
 from importer.handlers import BaseHandler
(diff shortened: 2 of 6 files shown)


stuaxo (approved):

common/tests/factories.py
```diff
@@ -102,7 +102,7 @@ class Meta:
 
     description = short_description()
     described_footnote = factory.SubFactory(FootnoteFactory)
-    description_period_sid = numeric_sid()
+    description_period_sid = factory.Sequence(lambda x: x + 1)
```
andyhd, line 105: What's the reason for this change? The description SID is a standard numeric SID (1 - 99999999)

andyhd (approved):
Looks good 👍

common/tests/factories.py
NDevox: The numeric SID returns a 0-padded string. It matches the XML spec and how we receive the data, but not what's in the DB (which is just a plain int). So the difference messes up the test comparisons. I could convert to int or 0-pad the db value instead. However I don't think having the 0-padded string in the factory makes much sense as it's immediately converted to an int in the db.

TBF what I should do is check the other models that use this and if it's the same all around just convert the `numeric_sid` function to return the sequence I'm putting in..

andyhd: Ah, I see. The XML schema wants an integer, and the `NumericSID` common model is a `PositiveSmallInteger`, so that means the `numeric_sid` function is wrong to generate a string - my bad. Do we receive zero-padded number strings in the TARIC XML? That doesn't match the schema.

NDevox: I might be making an assumption there that I shouldn't be...

Looking through the code the only place we zfill is for the GN indent, so you are probably right that it should just always return an integer.

I'll put it into the next PR.

PR #64, opened 2020-09-08
TP-240: Add Measures data models

andyhd:
Measures are the detail of the tariff. They specify how much must be paid to import a
specified good and how to calculate the amount, be it by percentage of value, by weight,
a fixed amount or varying by the origin location, date, or whether a quota applies.

This commit introduces the Django model classes and validators representing measures and
their associated business rules.

Many of the business rules are similar, and this commonality is factored out into
fixture functions in the conftest.py file.

Also in this commit are some refactoring of other business rules to make them more
consistent, and in the case of commodities, to make more effective use of Factory Boy
features.


andyhd:
I think this is ready for review now. Let's call the implementation of ME32 a first pass. 😬

measures/tests/conftest.py
```diff
@@ -0,0 +1,123 @@
...
+
+from common.tests import factories
+from common.validators import ApplicabilityCode
+from common.validators import UpdateType
+
+
+# register factories and models as fixtures
+# register(CamelCaseFactory) creates fixtures:
+# - camel_case_factory (factory class)
+# - camel_case (model instance)
```
simonwo, line 14: Do we want to keep this?

measures/tests/test_business_rules.py
```diff
@@ -0,0 +1,1244 @@
...
+
+
+def test_MT7(date_ranges):
+    """A measure type can not be deleted if it is used in a measure."""
+
+    with pytest.raises(IntegrityError):
+        factories.MeasureFactory().measure_type.delete()
+
+
+@pytest.mark.skip(reason="All measure type series are open ended")
```
simonwo, line 78: Hm, does it not make sense to run this anyway given it's implemented?

measures/tests/test_business_rules.py
```diff
@@ -0,0 +1,1244 @@
...
+
+
+def test_ME7():
+    """The goods nomenclature code must be a product code; that is, it may not be an
+    intermediate line.
+    """
+
+    with pytest.raises(ValidationError):
+        factories.MeasureFactory(goods_nomenclature__suffix="00")
+
```
simonwo, line 192: Should we be testing the positive case, e.g. that there is no failure when the suffix is 80?

measures/tests/test_business_rules.py
```diff
@@ -0,0 +1,1244 @@
...
+    A measure can be assigned to:
+    - a commodity code only (most measures)
+    - a commodity code plus an additional code (e.g. trade remedies, pharma duties,
+      routes of ingress)
+    - an additional code only (only for Meursing codes, which will be removed in the UK
+      tariff).
+
+    This means that a goods code is always mandatory in the UK tariff, however this
+    business rule is still needed for historical EU measures.
```
simonwo, line 217: Do we test anywhere that it's possible to use just an additional code?

measures/tests/test_business_rules.py
```diff
@@ -0,0 +1,1244 @@
...
+        factories.MeasureExcludedGeographicalAreaFactory(
+            excluded_geographical_area=membership.member,
+            modified_measure__geographical_area=membership.geo_group,
+            modified_measure__valid_between=date_ranges.normal,
+        )
+
+
+def test_ME68():
+    """The same geographical area can only be excluded once by the same measure."""
+
```
simonwo, line 853: No test?

measures/tests/test_business_rules.py
```diff
@@ -0,0 +1,1244 @@
...
+                suffix="80",
+            ),
+        )
+
+
+@pytest.mark.skip(reason="Nothing to check")
+def test_ME72():
+    """Footnotes with a footnote type for which the application type = "measure
+    footnotes" can be associated at any level."""
+
```
simonwo, line 897: Hm should we be doing a positive test here?

measures/tests/test_business_rules.py
```diff
@@ -0,0 +1,1244 @@
...
+
+def test_ME117(approved_workbasket):
+    """When a measure has a quota measure type then the origin must exist as a quota
+    order number origin. This rule is only applicable for measures with start date after
+    31/12/2007.
+
+    Only origins for quota order numbers managed by the first come first served
+    principle are in scope; these order number are starting with '09'; except order
+    numbers starting with '094'
```
simonwo, line 1182: Does this validation still work if we've changed 09 to 05 for the UK tariff?

measures/validators.py
```diff
@@ -0,0 +1,964 @@
...
+    """
+
+    if measure.valid_between.lower < datetime(2008, 1, 1, tzinfo=timezone.utc):
+        return
+
+    if measure.order_number is None:
+        return
+
+    order_number = measure.order_number.order_number
+    if order_number.startswith("094") or not order_number.startswith("09"):
```
simonwo, line 923: We might be better off using the `QuotaOrderNumber.mechanism` field here? And delegating the validation of 09/094 to that field. In the UK tariff the 09s will be 05s.

measures/validators.py
```diff
@@ -0,0 +1,964 @@
...
+    if measure.order_number is not None:
+        matching_identity["order_number__sid"] = measure.order_number.sid
+    if measure.additional_code is not None:
+        matching_identity["additional_code__sid"] = measure.additional_code.sid
+    if measure.reduction is not None:
+        matching_identity["reduction"] = measure.reduction
+    matching_measures = type(measure).objects.filter(**matching_identity)
+
+    # all commodities in tree
+    indent = measure.goods_nomenclature.indents.active().get()
```
simonwo, line 381: Hm, I'm not sure `active` is what we want as this needs to be true for at least as long as the measure is active. After all, we could be creating a measure in the future on a commodity code that starts in the future, in which case indent wouldn't even return the commodity code this measure is on.

Should we also be selecting comm codes that have start dates in the future but that overlap with this measure? Do we in fact need `current`?

measures/validators.py
```diff
@@ -0,0 +1,964 @@
...
+
+def validate_goods_code_level_within_measure_type_explosion_level(measure):
+    """ME88"""
+
+    if not measure.goods_nomenclature:
+        return
+
+    # XXX is .active() correct here?
+    # one level of tree depth corresponds to an increment of 2 in explosion level
+    goods_code_level = measure.goods_nomenclature.indents.active().get().depth * 2
```
simonwo, line 797: Again, do we need more than `active` and instead all the codes that overlap with this measure?

simonwo (commented):
Stellar! ✨

common/tests/factories.py
```diff
@@ -240,6 +243,20 @@ class Meta:
     suffix = "80"
     statistical = False
 
+    indent = factory.RelatedFactory(
+        "common.tests.factories.GoodsNomenclatureIndentFactory",
+        factory_related_name="indented_goods_nomenclature",
+        workbasket=factory.SelfAttribute("..workbasket"),
```
NDevox, line 249: Nice, I was unaware of this feature.

measures/tests/conftest.py
andyhd: No, it didn't turn out to be useful. I'll delete this.

measures/tests/test_business_rules.py
andyhd: No, I will add it.

measures/tests/test_business_rules.py
andyhd: Huh, missed one

measures/tests/test_business_rules.py
andyhd: No. I didn't see that on the measures spreadsheet, or in the Ruby code. Where is it documented?

measures/validators.py
andyhd: Ah, I forgot about mechanism. I'll rewrite to check that.

measures/validators.py
andyhd: Yeah, I wasn't sure about `active`. I think what we actually need is "latest approved or in current workbasket".

common/tests/factories.py
andyhd: It's a nice one

measures/tests/test_business_rules.py
andyhd: Measures can only be associated with Goods with suffix "80" (ME7), so what does _level_ mean here?

measures/tests/test_business_rules.py
```diff
@@ -0,0 +1,1244 @@
...
+    specified but the order number is blocked for input.
+    """
+
+
+@requires_export_refund_nomenclature
+def test_ME21():
+    """If the additional code type has as application "ERN" then the combination of
+    goods code + additional code must exist as an ERN product code and its validity
+    period must span the validity period of the measure.
+    """
```
NDevox, line 378: Tests like this I would probably make to `assert False`. As if we flip the boolean on this they would pass and anyone who was unaware of these tests would likely be non the wiser.

measures/tests/test_business_rules.py
```diff
@@ -0,0 +1,1244 @@
...
+    order number, additional code and reduction indicator. This rule is not applicable
+    for Meursing additional codes.
+
+    This is the mother of all business rules and the one that is likely to trip DDaT and
+    Tariff Managers up more than any other. This is an extension of the previously
+    described ME1 to all commodity codes in the upward hierarchy and all commodity codes
+    in the downward hierarchy. Therefore it is critical that the development team works
+    out a way of identifying a commodity’s hierarchy from the available data. There is
+    no concept of parent / child relationship in the database, rather a list of
+    sequential numbers and indents which infer the hierarchy.
```
NDevox, line 446: This is untrue now. The database has a concept of hierarchy for comm codes.

measures/tests/test_business_rules.py
```diff
@@ -0,0 +1,1244 @@
...
+@pytest.mark.skip(
+    reason="Abrogation, modification and replacement regulations are not used"
+)
+def test_ME29():
+    """If the entered regulation is a modification regulation then its base regulation
+    may not be completely abrogated.
+    """
+
+
+def test_ME32(date_ranges):
```
NDevox, line 434: There's a lot of corner cases to add here I think. Would probably need to sit down and think about them and then at least add stubs. But could be another PR as far as I'm concerned.

measures/validators.py
NDevox: maybe:

```python3
for indent in measure.goods_nomenclature.indents.filter(workbasket__approved=True, valid_between__overlaps=measure.valid_between):
    # run below logic on all the results
```

NDevox: And this doesn't take into account future and past versions of that goods_nomenclature...

So you probably want:

```python3
for gn in GoodsNomenclature.objeects.filter(workbasket__approved=True, valid_between__overlaps=measure.valid_between, sid=measure.goods_nomenclature.sid):
    for indent in measure.goods_nomenclature.indents.filter(workbasket__approved=True, valid_between__overlaps=measure.valid_between):
        # Run below logic on all the results.

File: measures/tests/test_business_rules.py (hunk 6 above)
Comment (simonwo) [reply measures/tests/test_business_rules.py]: Ah, good question. I assumed it refered to "level of the goods nomenclature hierarchy" but with a re-read I'm now not convinced.

File: measures/validators.py (hunk 9 above)
Comment (NDevox) [reply measures/validators.py:385]: And this query probably needs to be recursively run on all the nodes you find...

Soooooo maybe:

```python3

def find_nodes(qs, measure, exclude: set = None):
    exclude = exclude or set()
    new = TrackedModelQuerySet()

    for node in qs:
        node_versions = ... # get all versions of this node within the measures validity range - excluding the excluded ids
        tree = ... # get the tree for all versions of the node - add all new nodes to the new qs where their ids aren't excluded.

        excluded.update(**tree.values_list("pk", flat=True)

        for item in tree:
            new = new | find_nodes(tree, measure, excluded)

        new = new | tree
        
    return qs | new

nodes = find_nodes(...)
if any(node.indented_goods_nomenclature.matching_measures for node in nodes):
    #...
```

measures/tests/test_business_rules.py
simonwo: OK so given that the various application types are:
```
        "CN nomenclature",
        "TARIC nomenclature",
        "Export refund nomenclature",
        "Wine reference nomenclature",
        "Additional codes",
        "CN measures",
        "Other measures",
        "Meursing Heading",
        "Dynamic footnote",
```
if "measure footnotes" relates to the `"Other measures"` application code then it'd be checking they can also apply to measures on a comm code at 10 digits or above? Whereas `"CN measures"` can only be applied to measures with a comm code at 8 digits or above?

NDevox (commented):
Codes generally pretty good IMO. There's a lot of it! but I think that's a common theme among us at the moment.

I think Simons got a keener eye for the business rules and models than me but I left a couple thoughts around Feel free to ignore.

measures/validators.py
simonwo: So I thought further about this. Basically I think that:
- the goods nomenclature is king – we can't change the nomenclature to fix an ME32 problem
- so any measure that gets added needs to work with the present and future nomenclature

So we probably need to check using the nomenclature for the entire validity period of the measure.

If the nomenclature changes 4 times over the course of the measure (e.g. because the measure is on a parent code which spans the validity but the children are scheduled to change) then we need to do the ME32 check for each of those 4 nomenclature versions. I don't know if that actually requires 4 checks or if there's an algorithm to simplify this. But TARIC is bad at anything other than point-in-time queries so it might be hard.

measures/tests/test_business_rules.py
andyhd: Ah, that makes sense 👍

stuaxo (approved):
LGTM

NDevox (approved):
There's a couple really minor things I saw. But nothing blocking IMO.

workbaskets/validators.py
```diff
@@ -37,3 +37,17 @@ class WorkflowStatus(models.TextChoices):
     PUBLISHED = "PUBLISHED", "Published"
     # Sent to CDS, but CDS returned an error
     CDS_ERROR = "CDS_ERROR", "CDS error"
+
+    @classmethod
+    def approved_statuses(cls):
```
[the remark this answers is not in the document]
andyhd: Yeah, not sure why I didn't do that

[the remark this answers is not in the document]
andyhd: Ah yes, because it's an `Enum` subclass, so the metaclass tries to interpret `approved_statuses` as another member of the enum. I was tempted to make a "classproperty", but that seemed unnecessary since `@classmethod` already exists and works.

measures/validators.py
```diff
@@ -0,0 +1,1012 @@
...
+    if measure.order_number is not None:
+        query &= Q(order_number__sid=measure.order_number.sid)
+    if measure.additional_code is not None:
+        query &= Q(additional_code__sid=measure.additional_code.sid)
+    if measure.reduction is not None:
+        query &= Q(reduction=measure.reduction)
+    matching_measures = type(measure).objects.filter(query)
+
+    # get all goods nomenclature versions associated with this measure
+    goods = type(measure.goods_nomenclature).objects.filter(
```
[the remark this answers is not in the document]
andyhd: But may as well leave it for consistency with all the other `type(model_instance).objects`

PR #65, opened 2020-09-08
TP-299 add quotas importer

NDevox:


stuaxo (approved):

andyhd (approved):
Looks good to me, except for tests failing due to breaking change in pytest-bdd 4.0.0 - you could just set the dependency to pytest-bdd<4?

NDevox:
@andyhd in retrospect I probably should have done that.

Was a bit of a wild goose chase but did manage to fix it in the end.

PR #66, opened 2020-09-15
TP-299: Apply importer to Regulations models

NDevox:
This introduces a further extendable method in the handler for the
`save` step when actually saving to the database. The default
is to save through the serializer. This allows some finer changes to
be made when saving. This is useful in the instance of regulations
so that multiple models can be saved from a single handler.


andyhd (approved):
Sorry for making your life difficult by refactoring the data model 🤦

NDevox:
🤣  not your fault, I agree with the refactoring. We were either going to handle the complexity on import (so ideally once or a very low number of times). Or deal with it permanently in how we do queries.

PR #67, opened 2020-09-16
Sphinx docs

andyhd:
Add a Github Action and some stub files to automatically generate Sphinx documentation on PRs and publish it to Github Pages on merge to master


NDevox (approved):
👌

simonwo:
Cool – does this override the stuff on the `sphinx-test` branch??

.github/workflows/django.yml
```diff
@@ -103,3 +103,25 @@ jobs:
         deployment_id: ${{steps.deployment.outputs.deployment_id}}
         env: ${{steps.deployment.outputs.env}}
         env_url: https://tamato-dev.london.cloudapps.digital
+    
+    - name: Sphinx Build
+      uses: ammaraskar/sphinx-action@0.4
+    
+    - name: Commit documentation changes
+      run: |
+        git clone https://github.com/ammaraskar/sphinx-action-test.git --branch gh-pages --single-branch gh-pages
```
simonwo, line 112: Is this the right repo??

simonwo: On the tariff data manual we also use a commit action which is quite good.
https://github.com/uktrade/tariff-data-manual/blob/master/.github/workflows/update-from-wiki.yml

andyhd:
Ah, I forgot `sphinx-test` existed

.github/workflows/django.yml
simonwo: Or indeed an action which does it in one: https://github.com/uktrade/tariff-data-manual/blob/master/.github/workflows/update-from-wiki.yml#L65

docs/Makefile
```diff
@@ -0,0 +1,20 @@
...
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     = .
+BUILDDIR      = _build
+
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
NDevox, line 13: This doesn't look right

docs/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+.PHONY: help Makefile
+
+# Catch-all target: route all unknown targets to Sphinx using the new
+# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
+%: Makefile
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
NDevox, line 20: Probably don't want to default to building docs.

Might want to copy what was done in the sphinx-build branch.

Essentially added:

```
SPHINXOPTS    ?=

.PHONY: help clean clean-bytecode clean-static collectstatic compilescss dependencies docker-image docker-test node_modules run test build-docs



help: Makefile
	@echo
	@echo " Commands in "$(PROJECT)":"
	@echo
	@sed -n 's/^##//p' $< | column -t -s ':' | sed -e 's/^/ /'
	@echo

## build-docs: Build the project documentation
build-docs:
	@sphinx-build . docs/_build -c docs/ -b html
```

.github/workflows/django.yml
andyhd: This step loads a skeleton gh-pages project, not sure if we need it

docs/conf.py
```diff
@@ -0,0 +1,54 @@
...
+copyright = '2020, UK Department for International Trade'
+author = 'DIT'
+
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
```
NDevox, line 30: Would just ask to add `"sphinx.ext.autodoc"` in here.

NDevox (changes requested):
Actually a couple changes to make!

NDevox:
This doesn't really override the `sphinx-test` branch. Added some comments to make sure the `Makefile` and `conf.py` are more aligned. But otherwise it follows the same ideas.

Would be a pretty quick PR to add in the rest of the `sphinx-test` branch, which was effectively just converting the ADRs to `rst`

docs/Makefile
andyhd: It defaults to the help message. The command to build docs is `make html`.

docs/Makefile
andyhd: I'm not familiar with `sphinx-build`, but from the documentation it doesn't look like `-M help` is a possibility, so maybe `-h`?

andyhd:
I think the easiest thing to do would be to merge my github actions config into `sphinx-test` instead.

andyhd:
Merged into `sphinx-test` instead

PR #68, opened 2020-09-16
Auto-build docs with Sphinx

andyhd:
I'm not very familiar with Sphinx, so I've not got the index file or links working right, but I think the build works now


stuaxo:
Not sure why `import dj_database_url` is failing in the build docs stage, did it not install correctly or something ?

andyhd:
Ah, it's because the `requirements.txt` in `docs/` is incomplete - probably need to use the same `requirements.txt` for the sphinx github action container

stuaxo (commented):
Wow, this works these days ?   I was on a pip bug for the longest time where you couldn't do this.  Fantastic :)

andyhd:
It doesn't appear to work here :(

stuaxo:
> It doesn't appear to work here :(

Oh no, that would probably have worked if I hadn't mentioned it.

This could either stay where it is and -
have `tamato` as a dependency, or maybe even something really ugly like `-e ..` `-e ../` may work.

Or moving it to `../requirements_doc.txt` might be practical.

simonwo:
Looks like this works now! What shall we do with the generated docs? Push them to Github Pages?

andyhd:
Yes, the docs will be automatically pushed to Github Pages on merge to master

simonwo (approved):
LGTM assuming we can sort the build issue.

PR #76, opened 2020-09-29
TP-262: Add remaining ADR docs to Sphinx build and fix minor formatting issues

NDevox:


andyhd (approved):

docs/conf.py
```diff
@@ -48,7 +48,7 @@
 # source_encoding = 'utf-8-sig'
 
 # The master toctree document.
-master_doc = "index"
+master_doc = "docs/index"
```
andyhd, line 51: When the Github action image runs, it uses `docs/` as its working directory. So this will cause an error unless there is a  `docs/docs/index.rst` file.

.github/workflows/check-pr.yml
```diff
@@ -79,4 +79,4 @@ jobs:
       with:
         pre-build-command: "apt-get update -y && apt-get install -y git-core"
         docs-folder: "docs/"
-        build-command: "sphinx-build -b html . _build"
+        build-command: "sphinx-build -b html -Q . _build"
```
andyhd, line 82: You could change `docs-folder` to `""` and either set `build-command` to point to `docs/`, or remove `build-command` and make sure the `Makefile` has an `html` target which runs the correct sphinx-build command.

PR #77, opened 2020-09-29
TP-262: Fix prod build for docs

NDevox:


.github/workflows/django.yml
```diff
@@ -108,8 +108,8 @@ jobs:
       uses: ammaraskar/sphinx-action@0.4
       with:
         pre-build-command: "apt-get update -y && apt-get install -y git-core"
-        docs-folder: "docs/"
-        build-command: "sphinx-build -b html . _build"
+          docs-folder: "."
+          build-command: "sphinx-build -b html -c docs/ . _build"
```
andyhd, line 112: Ah, same thing tripped me up. The default `build-command` is `make html`, so maybe we should remove this from the workflow yaml and make sure the `html` target is correct in `Makefile`? Then there's only one place to keep up-to-date (not that we're likely to change this again)

andyhd (approved):

.github/workflows/django.yml
NDevox: That sounds pretty reasonable, giving it a go.

NDevox: It did not work

PR #78, opened 2020-09-30
Remove dependency on freezegun

andyhd:
We are currently using a set of fixed dates for testing purposes.  Eventually the
current date will no longer fall within those date ranges.  This commit replaces the
fixed dates with relative dates, based on the current date at runtime.  This also
removes our dependency on freezegun for the few tests that needed it.

This change has been separated out from TP-240 (measures data model), because I need it for another PR.

### DIFF
--- a/common/tests/util.py
+++ b/common/tests/util.py
@@ -8,6 +8,7 @@ from typing import Dict
 from typing import Type
 
 import pytest
+from dateutil.relativedelta import relativedelta
 from django.core.exceptions import ValidationError
 from django.template.loader import render_to_string
 from django.urls import reverse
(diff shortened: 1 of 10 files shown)


common/tests/util.py
```diff
@@ -186,73 +187,101 @@ def wraps(valid_user, *args, **kwargs):
...
-        datetime(2021, 2, 1, tzinfo=UTC),
-        datetime(2023, 3, 1, tzinfo=UTC),
+        NOW + relativedelta(months=+1),
+        NOW + relativedelta(years=+2, months=+2),
     )
     overlap_normal = DateTimeTZRange(
-        datetime(2021, 1, 15, tzinfo=UTC),
-        datetime(2022, 2, 15, tzinfo=UTC),
+        NOW + relativedelta(days=+14),
+        NOW + relativedelta(days=+14, months=+1),
```
simonwo, line 228: This also had a `year=+1` in the original.

common/tests/util.py
```diff
@@ -186,73 +187,101 @@ def wraps(valid_user, *args, **kwargs):
...
-        datetime(2021, 1, 15, tzinfo=UTC),
-        datetime(2021, 2, 1, tzinfo=UTC),
+        NOW + relativedelta(days=+14),
+        NOW + relativedelta(months=+1),
     )
     current = DateTimeTZRange(
-        datetime(2020, 8, 1, tzinfo=UTC) - timedelta(weeks=4),
-        datetime(2020, 8, 1, tzinfo=UTC) + timedelta(weeks=4),
+        NOW + relativedelta(weeks=-4),
+        NOW + relativedelta(weeks=+4),
```
simonwo, line 256: Was it significant that these were `2020`?

simonwo (approved):
LGTM assuming the very minor changes made don't really matter.

common/tests/util.py
andyhd: Hm, that's odd - good spot, I will check again.

common/tests/util.py
andyhd: No, this was what prompted me to make this change. I picked an arbitrary date for `current` and used freezegun to fix the current datetime for tests.

common/tests/util.py
andyhd: Fixed

NDevox (approved):

PR #81, opened 2020-10-08
TP-309: Ensure business rules with start dates are correctly documented

andyhd:
We have discovered that some of the business rules in TARIC have been
added to the system after data that breaks the rule has already been
present, and these have not been retroactively applied. This means that
in our implementation we need to give a free pass to data that was
inserted before the rule was formed.

The rules are:

- ME116: The quota order number on a measure must exist.
    Only applies to measures with a start date after 31/12/2007.

- ON6: The validity period of the geographical area must span the
  validity period of the quota order number origin.
    Only applies to quota order numbers with a start date after
    2006-12-31.

- ON9: When a quota order number is used in a measure then the validity
  period of the quota order number must span the validity period of the
  measure.
    Only applicable for measures with start date after 31/12/2007.

- ROIMB8: Explicit dates of related measures must be within the validity
  period of the base regulation.
    Only applicable to measure with start date after 31/12/2003.

This commit is to make sure these dates are reflected in code and
documentation.

This change also includes a decorator function to help testing business
rules which only apply after a specified date. Implementing this decorator 
required refactoring the `Dates` helper class and code which touched it.


regulations/tests/test_business_rules.py
```diff
@@ -76,6 +67,35 @@ def test_ROIMB7():
...
+    Only applicable for measures with start date after 31/12/2003."""
+
+    with pytest.raises(ValidationError):
+        factories.MeasureFactory(
+            generating_regulation__valid_between=date_ranges.starts_with_normal,
+            valid_between=date_ranges.normal,
+        )
+
+
+@pytest.mark.xfail(reason="ME87 catches this currently")
```
simonwo, line 83: Is this going to cause issues for seed file loading?

regulations/validators.py
```diff
@@ -165,3 +172,53 @@ def validate_base_regulations_have_group(regulation):
...
+def validate_regulation_validity_spans_measures_validity(regulation):
+    """ROIMB8
+
+    Explicit dates of related measures must be within the validity period of the base
+    regulation.
+
+    Only applicable for measures with start date after 31/12/2003.
+    """
+    # XXX this is covered by ME87, but ME87 does not have the cutoff date
+
```
simonwo, line 186: I think the difference is that this is talking about the `validity_start_date` and `validity_end_date` of both measures and regulations, but ME87 talks about the "effective end date" which is the earliest of `validity_end_date`, or the start date of an aborgating or prorogating regulations attached to the base regulation? Does our ME87 implementation have that logic?

quotas/validators.py
```diff
@@ -165,7 +172,7 @@ def validate_order_number_validity_spans_origin_validity(origin):
 
 
 def validate_order_number_validity_spans_quota_definition_validity(definition):
-    """ON7"""
+    """ON8"""
```
simonwo, line 175: Are we missing an implementation for ON9, (even one marked `@requires_measures`?)

simonwo (approved):

stuaxo (approved):

NDevox (approved):

PR #82, opened 2020-10-08
TP-333: Allow fetching of "latest" versions of a relation.

NDevox:

this being the case it is possible to create a row on one table and link it
to a row in another table. This row in another table can then be "updated" -
adding a new row to the table and a new version of the object. In this case
the foreign key may be considered stale, but only within the date range of
the new row that has been added.

```python3
    # must be a TrackedModel
```

And then access the latest version of the relation by simply adding `_latest`

```python3
example_model.other_model_latest  # Gets the latest version
```

This method finds all the relations of a model and for each one adds a column
populated by a SQL subquery that looks roughly like so:

```sql
SELECT Row_to_json({relation_table}) AS "json_data"
  FROM  {relation_table}
 INNER JOIN "common_trackedmodel" U1
    ON {relation_table}."trackedmodel_ptr_id" = U1."id"
 INNER JOIN "workbaskets_workbasket" U2
    ON U1."workbasket_id" = U2."id"
 WHERE {relation_table}.{identifying_field} = {outer_relation_join}.{identifying_field}
   AND U2."status" IN ({approved_statuses})
 ORDER BY {relation_table}."trackedmodel_ptr_id" DESC
 LIMIT 1
```

This caches the data for each row on the object instances in a dictionary
which is then used to create the related object in python when called upon.

To facilitate this a new `RowToJson` Func has been added, adding the ability
to call the postgres specific `row_to_json` function. This must be used in
conjunction with the `RowToJsonSubquery` which adds the proper arguments to
the function.

### DIFF
--- a/additional_codes/tests/test_business_rules.py
+++ b/additional_codes/tests/test_business_rules.py
@@ -59,20 +59,22 @@ def test_ACN1():
         )
 
 
-def test_ACN2():
+def test_ACN2(must_exist):
     """
     The referenced additional code type must exist and have as application code "non-
     Meursing" or "Export Refund for Processed Agricultural Goods”.
     """
 
-    wb = factories.WorkBasketFactory.create()
-    ac = factories.AdditionalCodeFactory.build(type=None, workbasket=wb)
-    with pytest.raises(ObjectDoesNotExist):
-        ac.save()
+    assert must_exist(
+        "type",
+        factories.AdditionalCodeTypeFactory,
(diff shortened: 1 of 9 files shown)


simonwo (approved):
Looks good (although I can't claim to understand enough about Django internals!). A shame that Django forces our hand a bit here when maybe what we really needed was more flexible foreign keys?

We've used `approved()` in a few places and it might be that we need to make that customisable for different UI views… but I'm sure there's an easy way of doing that.

PR #83, opened 2020-10-09
Measure exporter and updates

simonwo:
This pull request is an amalgam of all the various model, validator and serialiser fixes I have made as a result of building importers. We can split this down if we want, but the commits are faily logical so it might best to review on a commit-by-commit basis.

- The main event is all of the serialisers necessary to output measures. 
- This also includes tweaks to the measure models to handle prefixes and abbreviations used in duty sentences, which power the duty sentence parser coming in a later PR.
- There are also some fixes for other XML output.
- And there are some fixes to handle National Schema models and legacy data where rules are violated.

There are not enough tests, but I thought I would get some input on where coverage needs to be improved by creating this PR.


NDevox (changes requested):
All looks OK to me, as mentioned, only thing that seems to be missing are tests

measures/migrations/0008_measuretype_origin_destination_code.py
```diff
@@ -0,0 +1,53 @@
...
+        migrations.AddField(
+            model_name="measuretype",
+            name="origin_destination_code",
+            field=models.PositiveSmallIntegerField(
+                choices=[(0, "Import"), (1, "Export"), (2, "Import/Export")], null=True
+            ),
+        ),
+        migrations.RunPython(
+            code=copy_trade_movement_code,
+            reverse_code=do_nothing,
```
NDevox, line 44: You can just omit this argument entirely - otherwise it's just noise.

measures/migrations/0007_measurements_have_optional_measurement_unit.py
```diff
@@ -0,0 +1,24 @@
...
+
+    operations = [
+        migrations.AlterField(
+            model_name="measurement",
+            name="measurement_unit_qualifier",
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=django.db.models.deletion.PROTECT,
+                to="measures.MeasurementUnitQualifier",
```
NDevox, line 21: Could we merge these migrations? would be nice to reduce the amount we have early on.

Can just copy the code from the operations and shift it into the migration before (and the ones after the same).

While no one else has run them this should work without issue.

regulations/validators.py
```diff
@@ -72,10 +71,15 @@ class CommunityCode(models.IntegerChoices):
 #     one regulation)
 regulation_id_validator = RegexValidator(
```
NDevox, line 72: One of the nicer looking regexes I've seen 👍

NDevox (commented):
Nothing major, just a couple thoughts to keep things a _tad_ tidier.

NDevox (approved):

stuaxo:
:+1:  :+1: :+1:  :+1: :+1:  :+1:

PR #87, opened 2020-10-16
TP-242: Importer for measures

andyhd:
It is a requirement for the project to be able to import Measures,
Measure Types, Components, Conditions, Duty Expressions, Measurements, Monetary Units,
etc from legacy TARIC3 XML data.

This commit introduces the import parsers and handlers for these models.


stuaxo (approved):
Looks good to me.

NDevox (changes requested):
All looks good so far.

Only thing I'd ask for is tests!

NDevox (approved):

PR #91, opened 2020-10-21
Tp 342 refactor indents cleanup

NDevox:
This PR cleans up the data models from #90 and fixes the tests.


stuaxo (approved):

simonwo:
Is `depth` definitely correct? Shouldn't it still be indent + 2?

NDevox:
Good spot, and a weird one too. On depth that is.

Given we are using `add_child` from treebeard, treebeard actually imposes it's own depth on the child and ignores what we tell it (given it has to be at a certain depth to be the child of this node).

So it just needs to go altogether.

simonwo (approved):

stuaxo (approved):

PR #93, opened 2020-10-22
TP-342: Refactor GoodsNomenclatureIndents for phantom headers.

NDevox:
In cases where there are phantom headers right below a chapter
(regardless of whether the good is directly within its lineage)
some goods are still given a "00" indent, despite being 3
levels into the hierarchy.

This commit checks for phantom headers and handles this edge
case in the importer.

With this indent is more and more abstracted away from
the depth attribute of a materialized path node. Therefore
it is less and less reasonable to infer the indent from
the depth without incurring large numbers of queries.

To handle this this commit also adds the indent column to
the GoodsNomenclatureIndent table so that the TARIC defined
indent (not the tree depth) can be stored here for easier
access.


commodities/import_handlers.py
```diff
@@ -102,6 +102,27 @@ def save(self, data: dict):
 
         parent_depth = depth + 1
 
+        # In some cases, where there are phantom headers at the 4 digit level
+        # in a chapter, the depth is shifted by one more (even if the good is
+        # not directly descended from this good).
```
simonwo, line 107: So I almost agree with this – except that the goods are all always directly descended. E.g.

| Commodity code | Suffix | Indent | Real Depth |
|----------------|--------|--------|------------|
| 2900000000     | 80     | 0      | 0          |
| 2901000000     | 10     | 0      | 1          |
| 2901000000     | 80     | 0      | 2          |
| 2902000000     | 80     | 0      | 2          |
| 2903000000     | 80     | 0      | 2          |

So all codes in chapter 29 are descended from one of these headers. Nothing wrong with the code, but could we update the comment?

commodities/import_handlers.py
```diff
@@ -102,6 +102,27 @@ def save(self, data: dict):
...
+                item_id__startswith=item_id[0:2],
+                item_id__endswith="000000",
+            )
+            .exclude(suffix="80")
+            .exists()
+        )
+        if extra_headings and (
+            item_id[-6:] != "000000"
+            or data["indented_goods_nomenclature"].suffix == "80"
+        ):
```
simonwo, line 123: Would any of the other comments from `import_indents.py` be helpful?

simonwo (changes requested):
Just a comment change!

commodities/import_handlers.py
NDevox: 👍

commodities/import_handlers.py
NDevox: I was tempted by the comment left immediately above the `if`, but it's more or less explained by the comment above. All that is left to do with the rest of the comments is to say what the code is doing, which is described by the code itself - best to let the code do the talking, comments can be more misleading sometimes (as just pointed out in your comment above about the phantoms always being an actual ancestor).

simonwo: Yeah I did also think that when I read them properly so fair enough.

simonwo (approved):

PR #94, opened 2020-10-23
Handle errors in legacy data in indent importer

simonwo:
We discovered that a number of commodity codes
don't obey the rule "my indent is +1 of my parent
indent". We think these are errors as there is a
small nubmer of them and no discernible pattern to
them, so we have udpated the indent importer to
detect their specific SIDs and correct the data.

I have also converted `indent` to a numeric field,
– it should always contain numeric data, even if
we have to serialize it specially in the XML. We do
the same thing for `DutyExpression`.


stuaxo (approved):

commodities/models.py
```diff
@@ -80,7 +80,7 @@ class GoodsNomenclatureIndent(TrackedModel, ValidityMixin):
         validators=[MinValueValidator(1), MaxValueValidator(99999999)], db_index=True
     )
 
-    indent = models.CharField(max_length=2, validators=[MinLengthValidator(2)])
+    indent = models.PositiveIntegerField(validators=[MinValueValidator(0)])
```
NDevox, line 83: Now I know what I missed in the above comment!

Why do we want to do this? If indent doesn't really represent the tree, just what we got on import then wouldn't it be simpler to store it as the 2 char sequence?

NDevox (approved):

simonwo:
I think there are a couple of good reasons why we should store this is as the `int` that it is:

1. One is just to have confidence that data is correct. If we start importing indents from an EU taric file that is not a positive integer, we do need to know about it.
2. The other reason was that when we start making our own commodity codes we do want to be making sure these are positive ints.
3. If we move to a JSON API, we might choose to serialise the indent as a number rather than a string, and we should be doing that verification as we store the data rather than as we serialise it.

Again, not impossible with a char field at all. But just making the system work for us in certain ways and putting the logic for how it's displayed near the place where that matters (i.e. in the importer and serializers).

PR #95, opened 2020-10-26
TP-357: trade disputes importer

kristof-bb:
The importer takes as inputs the existing and new trade dispute measures in Excel format and generates the corresponding Taric3 XML transactions. The resulting transactions are added to a workbasket and the underlying models are updated.


importer/management/commands/tests/test_importers.py
```diff
@@ -0,0 +1,189 @@
...
+            regulation_group=regulation_group,
+            regulation_id='R0000002'
+        )
+
+        # Setup certificates
+        certificate_type = factories.CertificateTypeFactory.create(sid="N")
+        factories.CertificateFactory.create(
+            sid="990",
+            certificate_type=certificate_type
+        )
```
andyhd, line 70: As you never use the `certificate_type` variable after this point, and `CertificateFactory` already uses `CertificateTypeFactory` as a `SubFactory` for the `certificate_type` field, you might like to know that you can reduce this to:
```python
factories.CertificateFactory(
    sid="990",
    certificate_type__sid="N",
)
```

simonwo (approved):

PR #97, opened 2020-10-26
Handle partially-loaded elements in XML parsing

simonwo:
The `etree` streaming API will sometimes emit start
events for elements without their children if
passed a IO object or the Pull mecahnism is used.
This is because the API will consume input in
blocks and return as many events as it can, rather
than waiting for elements to be complete.

So we are not able to check what elements an
element has when handling a start event. To get
around this, this commit removes the need for
looking at if an element is a leaf node and
instead introduces extra state fields to keep
track of if the parser is currently parsing an
element or not, which is sufficient for selecting
the right parser. (This should also remove the
limitation on only having a single level of
nesting.)

### DIFF
--- a/importer/parsers.py
+++ b/importer/parsers.py
@@ -83,6 +83,7 @@ class ElementParser:
         self.many = many
         self.parent = None
         self.text = None
+        self.started = False
 
         if tag:
             self.tag = tag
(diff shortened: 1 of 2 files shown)


NDevox (approved):
This diff is awfully confusing, git apparently doesn't handle docstrings well.

stuaxo (approved):

importer/parsers.py
```diff
@@ -101,13 +102,20 @@ def _field_lookup(self) -> Dict[ElementParser, str]:
     def is_parser_for_element(
         self, parser: ElementParser, element: etree.Element
     ) -> bool:
-        """
-        Check if the parser matches the element.
+        """Check if the parser matches the element."""
+        return parser.tag == element.tag
```
andyhd, line 106: This method is redundant now, I think

importer/parsers.py
```diff
@@ -164,12 +165,14 @@ def end(self, element: etree.Element):
...
 
         # leaving this element, so marshal the data
-        if self.is_parser_for_element(self, element):
-            if element.text:
-                self.text = element.text.strip()
-            self.data.update(element.attrib.items())
-            self.clean()
-            self.validate()
+        else:
+            if self.is_parser_for_element(self, element):
```
andyhd, line 169: These should be combined into `elif`

andyhd (approved):

PR #98, opened 2020-10-27
TP-328: Support negative SIDs

NDevox:
HMRC has proposed allowing SIDs for the data models to vary from
9,999,999,999,999,999,999 to -9,999,999,999,999,999,999. A key
aspect of this is to allow negative SIDs which could be used to
distinguish UK data from historically EU data. As it is
unlikely that so many measures will be introduced at once it has
been proposed that instead a signed 32-bit integer is used,
allowing a range from -2,147,483,648 to +2,147,483,647. This still
allows orders of magnitude more of each data type than currently
allowed by the EU and can be easily updated to a 64-bit integer in
future if required.

This commit implements this by adding a new field called `SignedIntSID`
which replaces the `NumericSID` field most models used. This is similar
to the `NumericSID` field, but it has no validators and inherits from
`IntegerField` instead of `PositiveIntegerField`. `NumericSID` is kept
so that old migrations still work. The XSD has been updated to match
these new limits.


simonwo (approved):

andyhd:
Just too late! I was going to ask about naming migrations. Perhaps it's not important.

NDevox:
Sorry!

PR #99, opened 2020-11-03
Update and Delete Importer system

NDevox:
It is required that the importer not only allows for creating items, but updating and deleting them as well.

Given that this system is intended to be an Append-Only-Log it should not be possible for Update and Delete events to actually update or delete a row in the database. Therefore it is necessary to provide a method of linking rows across the database so that they can be identified as versions of the same item. This was initially intended to be done with the `predecessor` One-to-One link on `TrackedModel`. However the One-to-One nature is not conducive to various scenarios (errors and mistakes, the possibility of multiple versions stemming from one previous version).

This commit introduces a more robust Append-Only-Log by overriding the `save` method of `TrackedModel`. This disables saving on a single row once that row has been created and approved. A new model has also been introduced, termed `VersionGroup`. `TrackedModel` holds a Foreign Key to `VersionGroup` giving a One-to-Many relationship between the two models. On saving a new row, if it is a creation event, a new `VersionGroup` is created and attached to the `TrackedModel`. On update and deletion events a new `VersionGroup` is not made, instead an existing one is found using the identifying fields of the `TrackedModel`, this found `VersionGroup` is considered the grouping for that `TrackedModel` and is attached to the new row. In turn the `VersionGroup` is updated outside of the Append-Only-Log to reflect the most recent `TrackedModel` row it has been attached to.

In tandem with this some refactoring has taken place:

- With the removal of `predecessor` and `successor` on `TrackedModel` most of the querying around versions had to be updated to reflect the use of `VersionGroup` instead.
- The querysets have been moved to their own folder as well as the custom fields.
- Factories have been updated to default to approved workbaskets.
- Some common testing utilities were simplified.
- Importer testing was unified across the models.


common/models.py
```diff
@@ -176,25 +34,32 @@ class Meta:
...
+    current_version = models.ForeignKey(
+        "common.TrackedModel",
+        on_delete=models.SET_NULL,
+        unique=True,
+        null=True,
+        related_name="+",
+        related_query_name="is_current",
+    )
+
+
```
simonwo, line 47: So is the idea that the `VersionGroup` is always updated to point at the correct version? That's going to lead to problems because (as ever) the "current" version depends on context. E.g. when we are editing a workbasket the current version might be the version just added to the workbasket. Can we instead just use the `MAX(tracked_model.id)` or alternatively add a sequence number to the `TrackedModel`? That way we can select the "latest" version depending on context, right?

common/models.py
```diff
@@ -264,15 +121,17 @@ def new_draft(self, workbasket, save=True, **kwargs):
...
     def get_versions(self):
-        query = Q()
-        for field in self.identifying_fields:
-            query &= Q(**{field: getattr(self, field)})
-        return self.__class__.objects.filter(query).order_by("-created_at")
+        if hasattr(self, "version_group"):
+            return self.version_group.versions.all()
+
+        query = Q(**self.get_identifying_fields())
+        return self.__class__.objects.filter(query).order_by("-pk")
```
simonwo, line 128: …That's basically what we're doing here. I think what I'm proposing above is to use this same code for `get_latest_version` but just select the last item in the list.

common/querysets.py
```diff
@@ -0,0 +1,133 @@
...
+        return self.filter(workbasket__transaction__id__gt=transaction_id)
+
+    def as_at(self, date: datetime) -> QuerySet:
+        """
+        Return the instances of the model that were represented at a particular date.
+
+        If done from the TrackedModel this will return all instances of all tracked models
+        as represented at a particular date.
+        """
+        return self.filter(valid_between__contains=date)
```
simonwo, line 48: Does this return all the old versions too? Or does it only consider latest versions?

common/querysets.py
```diff
@@ -0,0 +1,133 @@
...
+        Get objects which have been approved or are in the specified workbasket.
+        """
+
+        return self.filter(
+            Q(workbasket=workbasket)
+            | Q(
+                workbasket__status__in=WorkflowStatus.approved_statuses(),
+                workbasket__approver__isnull=False,
+            )
+        )
```
simonwo, line 133: …Unless I'm wrong, and this will return latest versions?

common/querysets.py
```diff
@@ -0,0 +1,133 @@
...
+                raise exceptions.NoIdentifyingValuesGivenError(
+                    f"Field {field} expected but not found."
+                )
+        return self.filter(**kwargs)
+
+    def get_latest_version(self, **kwargs):
+        """
+        Gets the latest version of a specific object.
+        """
+        return self.get_versions(**kwargs).current().get()
```
simonwo, line 71: …I think my question boils down to: if I have footnote `FO123` and I add a new version of it into my workbasket, when I call:

    Footnote.objects.filter(footnote_type__id="FO", footnote_id="123").get_latest_version()

Does this return the version in the workbasket or not? In some cases that's what I want (such as when looking at a workbasket) but in other cases it's not (such as when I'm writing XML for public consumption).

common/querysets.py
NDevox: this will only consider objects that are relevant at this time, without extra filters it will provide all versions, approved and unapproved. (but you can limit this to specific versions, approved only, etc whatever you want)

common/querysets.py
NDevox: This returns any and all objects which are approved and/or are in a given workbasket. Again can be combined with any other filter as is required.

common/models.py
NDevox: This here is purely about "sensible" defaults and giving us an easy way to keep queries simple for what could be a large number of cases (pure speculation on my part).

It is worth baring in mind that just because we may provide what we feel is a sensible default does not mean we have to use it for everything. Any and all queries are still available to us and we can achieve more or less anything we want to without the use of this FK (as you point out, we can use groupings quite simply in a variety of ways).

My thought here is that it probably makes sense to log a "current" version just to provide the simplest of queries (direct joins, no need to analyse/sort/scan the tables). I'd assume this should be _the latest row for an object which has been approved_ but there's no reason that definition can be changed (it also doesn't actually fit the current code, which is just "the latest row for an object" essentially).

There is also no reason that this would limit the queries we need to do in the future.

common/querysets.py
NDevox: This one specifically will return a single version of an object, which is the most recently entered row in the database for that object. And yes it uses the convenience FK in the new model for that.

If the most recent row for the object is that in the workbasket, it will return it. However as mentioned above I think this definition should be changed a bit so the FK is only really used for approved rows (highly debateable).

Might be worth adding another convenience query (something like `.get_latest_version_with_workbasket`)

common/models.py
stuaxo: This is good, as long as there is a docstring by the `current` version explaining that some contexts idea of current may vary (so that future developers stop and think before using this).

common/models.py
stuaxo: Why does this use self.__class__.objects, not self.objects?

common/querysets.py
stuaxo: Does `workbasket=workbasket` work here ?  ISTR having to do this by the primary key in the past.

NDevox: I believe it works fine. This isn't actually new and is just copied over from the `models.py` file (separating concerns and all)

NDevox: ISTR similar though, I think it may be a slightly different usecase that requires this.

common/models.py
simonwo: Agreed with @stuaxo – if we want to keep this then it would make sense to be very clear what sort of current we are talking about. I'm mainly just worried that future devs will use this without thinking.

common/models.py
```diff
@@ -176,25 +34,32 @@ class Meta:
...
 
+class VersionGroup(models.Model):
+    current_version = models.ForeignKey(
+        "common.TrackedModel",
+        on_delete=models.SET_NULL,
+        unique=True,
+        null=True,
+        related_name="+",
+        related_query_name="is_current",
+    )
```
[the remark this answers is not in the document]
simonwo: I think I agree with this one?

common/querysets.py
stuaxo: Possibly reverse fk or m2m sets.

common/models.py
NDevox: Yep. I do realise I haven't actually put any time into documenting this yet. I should do that.

It probably needs an ADR as well. Although I'm tempted to do this after the fact.

common/models.py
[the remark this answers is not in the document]
NDevox: Blank dictates generally what is going to go into an HTML form and whether it gets the `required` attribute.

For this one I'd argue:

1) It's never going into a form nor should it ever be seen by a human as something they can directly edit.
2) It should realistically be considered pseudo NOT NULL. It's just a chicken-egg scenario. The `TrackedModel` needs a `VersionGroup` to reference, so we make a `VersionGroup` first. But the `VersionGroup` wants a `TrackedModel` to reference so we would want to make the `TrackedModel` first. In the end the `current_version` field is useful but not strictly necessary. So we make the `VersionGroup` first with a NULL `current_version`. Then make the `TrackedModel`, which as soon as it is saved gets put into the `current_version` column. At which point it should never be NULL again.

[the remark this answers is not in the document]
andyhd: Could make it `editable=False` so it will never be automatically included in any forms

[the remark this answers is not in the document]
NDevox: I think the point is that VersionGroup should never be seen in its entirety

PR #100, opened 2020-11-06
Fixes from the field

simonwo:
Another mixed bag of fixes for core:

- XSD change to ensure envelopes are produced according to our ID standard.
- Fix quota definition XML to include units.
- Don't use the quota description as a string representation as it may be `None`.
- Allow Python from a envrionment variable in the Makefile.

### DIFF
--- a/measures/serializers.py
+++ b/measures/serializers.py
@@ -11,6 +11,9 @@ from footnotes.serializers import FootnoteSerializer
 from geo_areas.serializers import GeographicalAreaSerializer
 from measures import models
 from measures import validators
+from measures.unit_serializers import MeasurementSerializer
+from measures.unit_serializers import MonetaryUnitSerializer
+from quotas.serializers import QuotaOrderNumberSerializer
 from regulations.serializers import RegulationSerializer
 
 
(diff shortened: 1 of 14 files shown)


measures/serializers.py
```diff
@@ -279,6 +278,8 @@ class Meta:
 
 @TrackedModelSerializer.register_polymorphic_model
 class MeasureSerializer(TrackedModelSerializerMixin, ValiditySerializerMixin):
+    from quotas.serializers import QuotaOrderNumberSerializer
+
```
NDevox, line 282: Circular imports are generally code smell.

I won't deny that this is because the way the XML is made to be transmitted has created the circular dependency...

But instead of reproducing it in the code we can do something like:

```python


class MeasureSerializer(TrackedModelSerializerMixin, ValiditySerializerMixin):
    # ...
    order_number = QuotaOrderNumberSerializer(read_only=True)
    # ...

    def get_order_number(self, obj):
        if obj.order_number:
            return obj.order_number.order_number
```

That way the serializer gives direct access to the order_number value that is actually needed and there is no need to import serializer in this module like so (and can therefore remove the circular nature of the imports in the Quotas module. (could arguably do the same for that serializer too).

The XML would need to be updated to:

```xml
      {%- if record.order_number %}
      <oub:ordernumber>{{ record.order_number }}</oub:ordernumber>
      {%- endif -%}
```

NDevox (changes requested):
Mostly good, just want to avoid circular imports.

measures/serializers.py
simonwo: I implemented this and then thought about it again, and realised that there is only a circular import issue because we grouped units together with measures. There's no real need to group them together like this – we could even go as far as splitting them out into a separate app. But for now I've split them into a separate serializer file and this avoids any import tomfoolery and also lets us retain the proper serializer for the quota order number. What do you think?

NDevox (approved):

measures/serializers.py
NDevox: 👍  this works for me

NDevox: I agree it may make sense to move these to a separate file. Looks like we put them together because of the commonality shared with the word "measure" but that's not really a reason to keep them in the same app.

PR #101, opened 2020-11-08
TP-340: Seperate out goods origins and successors

simonwo:
We has assumed that these two models were symmetrical,
i.e. that when one code is an origin for another
then the reverse relationship is a successor. This
is not the case (by oberservation of real data).

Instead:
- a good can have multiple origins
- a good can have multiple successors
- only codes with an end date can have successors


stuaxo:
Bikeshed:
Naming things is hard, I guess `absorbed_into` is a concept it should be possible to find by searching the code for the comment that explains it, `old` is pretty generic, is a more descriptive name we can do for this ?

simonwo:
> Bikeshed:
> Naming things is hard, I guess `absorbed_into` is a concept it should be possible to find by searching the code for the comment that explains it, `old` is pretty generic, is a more descriptive name we can do for this ?

In the instance we could probably go with `replaced`? The `new` is a bit harder though.

stuaxo:
> > Bikeshed:
> > Naming things is hard, I guess `absorbed_into` is a concept it should be possible to find by searching the code for the comment that explains it, `old` is pretty generic, is a more descriptive name we can do for this ?
> 
> In the instance we could probably go with `replaced`? The `new` is a bit harder though.

replaced definitely sounds better.

commodities/import_handlers.py
```diff
@@ -32,9 +33,90 @@ class GoodsNomenclatureHandler(BaseHandler):
...
+    )
+
+    def get_derived_from_goods_nomenclature_link(self, model, kwargs):
+        return kwargs
+
+    def pre_save(self, data: dict, links: dict) -> dict:
+        # To work out the correct goods nomenclature to link to,
+        # we need to know the start date of the other goods nomenclature
+        # which we don't have when we look up the link. So instead we
+        # assume the link is present and look it up in the pre-save step
```
NDevox, line 66: Nice, this is... an interesting workaround. But it would work 😅 

Thinking on it though are we sure `derived_from_goods_nomenclature` doesn't exist yet? The importer should go through the links in the order defined and will store them in `self.resolved_links` until `pre_save`.

You can see how `pre_save` is normally called which makes use of `self.resolved_links` here: https://github.com/uktrade/tamato/blob/464bda8b8969e2e95ebf0b7a4240de35e6b13821/importer/handlers.py#L392

So in theory you should be able to do `self.resolved_links["new_goods_nomenclature"].valid_between.lower`

commodities/import_handlers.py
```diff
@@ -32,9 +33,90 @@ class GoodsNomenclatureHandler(BaseHandler):
...
+    def pre_save(self, data: dict, links: dict) -> dict:
+        # To work out the correct goods nomenclature to link to,
+        # we need to know the end date of the other goods nomenclature
+        # which we don't have when we look up the link. So instead we
+        # assume the link is present and look it up in the pre-save step
+        data = super().pre_save(data, links)
+        must_be_active_on_date = data[
+            "old_goods_nomenclature"
+        ].valid_between.upper + timedelta(days=1)
+        data["absorbed_into_goods_nomenclature"] = models.GoodsNomenclature.objects.get(
```
NDevox, line 114: As above, are we sure this doesn't work in the get link system using `self.resolved_links`?

NDevox (commented):
if `self.resolved_links` doesn't work for this then it's 👍  for me. But just want to verify that doesn't work first!

Also, I take this to mean we don't need a separate origin/successor table? (if so 🎉 )

NDevox:
Ok somehow the diff didn't load fully for me 😬 . Now checking the rest...

commodities/models.py
```diff
@@ -231,6 +217,53 @@ class Meta:
...
+    )
+    absorbed_into_goods_nomenclature = models.ForeignKey(
+        GoodsNomenclature,
+        on_delete=models.PROTECT,
+    )
+
+    def clean(self):
+        validators.validate_absorbed_by_code_applicable_after_closing_date(self)
+
+
```
NDevox, line 266: The weird thing for me here is that this gives rise to the following possibility a good having an origin link to another good, but that other good not reciprocating with a successor link. I.e. Good B has an origin link to Good A, But Good A does not have a successor link to Good B.

Is this a reasonable scenario? Because they sound like they should very much be two sides of the one coin.

Having multiple origins _definitely_ forces the need for an M2M relationship so that's 100% clear to me now. My question is if we need both tables, or if they're both just one side of the same coin. If every successor record must be reciprocated in an origin record as well, then having both tables just creates more work.

From your comment it sounds like this is not the case (in which case what on earth is the intent of this within the TARIC system!). But I'd just like to double check 😇

commodities/import_handlers.py
NDevox: I just tried the code as:

```python3



class GoodsNomenclatureOriginHandler(BaseHandler):
    serializer_class = serializers.GoodsNomenclatureOriginSerializer
    tag = parsers.GoodsNomenclatureOriginParser.tag.name

    identifying_fields = (
        "new_goods_nomenclature__sid",
        "derived_from_goods_nomenclature__item_id",
        "derived_from_goods_nomenclature__suffix",
    )

    links = (
        {
            "model": models.GoodsNomenclature,
            "name": "new_goods_nomenclature",
            "optional": False,
        },
        {
            "model": models.GoodsNomenclature,
            "name": "derived_from_goods_nomenclature",
            "optional": False,
            "identifying_fields": ("item_id", "suffix"),
        },
    )

    def get_derived_from_goods_nomenclature_link(self, model, kwargs):
        must_be_active_on_date = self.resolved_links[
            "new_goods_nomenclature"
        ].valid_between.lower - timedelta(days=1)
        return model.objects.get(
            valid_between__contains=must_be_active_on_date,
            **kwargs,
        )


class GoodsNomenclatureSuccessorHandler(BaseHandler):
    serializer_class = serializers.GoodsNomenclatureSuccessorSerializer
    tag = parsers.GoodsNomenclatureSuccessorParser.tag.name

    identifying_fields = (
        "old_goods_nomenclature__sid",
        "absorbed_into_goods_nomenclature__item_id",
        "absorbed_into_goods_nomenclature__suffix",
    )

    links = (
        {
            "model": models.GoodsNomenclature,
            "name": "old_goods_nomenclature",
            "optional": False,
        },
        {
            "model": models.GoodsNomenclature,
            "name": "absorbed_into_goods_nomenclature",
            "optional": False,
            "identifying_fields": ("item_id", "suffix"),
        },
    )

    def get_absorbed_into_goods_nomenclature_link(self, model, kwargs):
        must_be_active_on_date = self.resolved_links[
            "old_goods_nomenclature"
        ].valid_between.upper + timedelta(days=1)
        return model.objects.get(
            valid_between__contains=must_be_active_on_date,
            **kwargs,
        )
```

and it works fine

simonwo: Ah great, I didn’t spot that. I had assumed that it wouldn’t store any state for some reason. I’ll switch it over to this way!

commodities/models.py
simonwo: Yes, that right – they are not two sides of the same coin. What it boils down to is:

- An “origin” represents where a code “came from” but not that a code is being replaced. A high level heading (like 2900000000) will have many codes mark is an origin over the course of its life but it won’t have any successors, because…
- A “successor” represents a new code that is *replacing* an old code. So they only get added to codes that have an end date. In the example above 2900000000 will have many codes mark it as origin but will have no successors because those codes are not replacing it… they are just formed out of it.

So I guess the key difference is that one is explicitly about replacing whereas the other is not, hence they’re not symmetrical.

NDevox: That actually makes sense... Should we use origin to aid in the algorithm to find a codes parent?

simonwo: Hm, I think it'll be hard to use the data in a meaningful way… I think being created from a parent is one case but there are potentially others. I don't think I still understand it well enough yet to know what they all are, but I'm sure there will be cases where codes are "derived from" siblings or grandparents. We'll also still have to handle the case where parents end and the code gets a new parent without any help. So I think what we've got for parent detection is probably still the best without some very radically different approach.

NDevox (approved):

andyhd:
I haven't had a chance to review this properly, but I would point out that the naming should be self-consistent at least: you have origins and successors, so there should be "originating_good" and "succeeding_good" and not "derived from" or "absorbed by".

andyhd (approved):
Looks good to me - the TARIC data model is still surprising us :/

PR #102, opened 2020-11-10
TP-295: Notify HMRC of data uploads

andyhd:
The second stage of our data journey to HMRC is notifying their HTTPS API that files are
available in our bucket.

This commit implements an API client class, which sends the notification to the HMRC
Secure Data Exchange Notification API. It also records the time of notification in the
database.

See https://uktrade.atlassian.net/browse/TP-295

See HMRC API documentation at https://developer.service.hmrc.gov.uk/api-documentation/docs/api/service/secure-data-exchange-notifications/1.0


stuaxo (approved):
LGTM (review was quick, but I already read the changes before it was officially requested :) )

common/tests/factories.py
```diff
@@ -78,7 +79,7 @@ class Meta:
 
 class TrackedModelMixin(factory.django.DjangoModelFactory):
     workbasket = factory.SubFactory(WorkBasketFactory)
-    update_type = UpdateType.UPDATE.value
+    update_type = UpdateType.UPDATE
```
NDevox, line 82: I may have updated this to `.CREATE` in my latest PR, as it doesn't make sense to have UPDATE as the default when making an initial object. Hopefully this doesn't impact you.

hmrc_sdes/api_client.py
```diff
@@ -0,0 +1,68 @@
+from apiclient import APIClient
```
NDevox, line 1: This looks like quite a dead project, which also warns off of using it as a dep. Do we definitely want to use it?

Especially considering the code is just:

```python3
class APIClient(object):
    BASE_URL = 'http://localhost:5000/'

    def __init__(self, rate_limit_lock=None, encoding='utf8'):
        self.rate_limit_lock = rate_limit_lock
        self.encoding = encoding
        self.connection_pool = self._make_connection_pool()

    def _make_connection_pool(self):
        return urllib3.PoolManager( cert_reqs='CERT_REQUIRED',
                                    ca_certs=certifi.where())

    def _compose_url(self, path, params=None):
        return self.BASE_URL + path + '?' + urlencode(params)

    def _handle_response(self, response):
        return json.loads(response.data.decode(self.encoding))

    def _request(self, method, path, params=None):
        url = self._compose_url(path, params)

        self.rate_limit_lock and self.rate_limit_lock.acquire()
        r = self.connection_pool.urlopen(method.upper(), url)

        return self._handle_response(r)

    def call(self, path, **params):
        return self._request('GET', path, params=params)
```

hmrc_sdes/api_client.py
```diff
@@ -0,0 +1,68 @@
...
+
+        return self.post(
+            f"{self.base_url}/organisations/notification/files/transfer/complete/{self.srn}",
+            data=upload.notification_payload,
+        )
+
+    def notify_transfer_ready(self, upload):
+        """Notifies that a given bulk file transfer is ready for processing"""
+
+        print(upload.notification_payload)
```
NDevox, line 63: print

hmrc_sdes/models.py
```diff
@@ -0,0 +1,54 @@
...
+        try:
+            client = HmrcSdesClient()
+            client.notify_transfer_ready(self)
+        except Exception:
+            raise
+        else:
+            if now is None:
+                now = timezone.now()
+            self.notification_sent = now
+            self.save()
```
NDevox, line 37: What are we trying to do here?

hmrc_sdes/views.py
```diff
@@ -0,0 +1,3 @@
+from django.shortcuts import render
+
+# Create your views here.
```
NDevox, line 3: DEL

NDevox (changes requested):
Small things, otherwise all ok.

How does this gel with `common.Transaction` at the moment?

hmrc_sdes/api_client.py
andyhd: I don't see any warnings? It seems to be active - last commit was 2 months ago. https://github.com/MikeWooster/api-client

I just wanted to remove the boilerplate from a HTTP JSON API client which this module seems to do - except for the OAuth2 stuff that I had to hand-roll.

common/tests/factories.py
andyhd: Shouldn't impact me - this was just to pacify mypy

hmrc_sdes/models.py
andyhd: This could probably just be a boolean. I was trying to anticipate unhappy path scenarios - when retrying failed notifications we probably want to avoid resending successful ones. I was imagining we would have some async process maybe.

hmrc_sdes/api_client.py
NDevox: 🤦  I was looking at https://github.com/shazow/apiclient

This seems fine.

andyhd:
The `taric` app models are stubs for the work on TP-371. This task was higher priority, but really depends on that and TP-112.

hmrc_sdes/models.py
NDevox: I think `else` is only called when an exception is not raised though? in this case the `try/except` doesn't actually do anything and could all just be top level

NDevox:
Yep stubs makes sense. ok

hmrc_sdes/models.py
andyhd: 🤦

andyhd:
I really only needed to include the Envelope model stub, but I accidentally started working on TP-371 before switching to this task and this was left-over code.

NDevox (approved):
Honestly having the stubs in actually works better for what I'm doing at the moment. LGTM

taric/models.py
```diff
@@ -0,0 +1,64 @@
...
+    envelope_id = EnvelopeId()
+    transactions = models.ManyToManyField(
+        Transaction, related_name="envelopes", through="EnvelopeTransaction"
+    )
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+    def __str__(self):
```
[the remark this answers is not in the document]
stuaxo: Hi @richtier it would be great if django-doctor could link to source docs for it's suggestions like this, e.g.   

https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#model-style.

I opened a ticket in Django about more granular links in the docs themselves
https://code.djangoproject.com/ticket/32185#ticket

PR #103, opened 2020-11-11
TP-333: Allow fetching of "latest" versions of a relation in a single query

NDevox:

```python3
    # must be a TrackedModel
```

```python3
```


stuaxo (approved):

simonwo (approved):
Big improvement!

PR #110, opened 2020-11-19
Updates to FTA importer

simonwo:
Updates the FTA spreadsheet importer to handle more exotic quotas and also to correctly understand dates across the Brexit boundary.

- Handle origin certificates in origin quotas.
- Actually output multiple countries into one envelope.
- Fix date handling in the FTA importer
- Handle quota suspensions in the FTA importer.
- Handle excluded quota origins in the FTA importer.
- Handle quota associations in the FTA importer.
- Import universal style quotas into the FTA importer.
- Don't output licensed quotas from the FTA importer.
- Untie date boundary handling from parsing of rates

### DIFF
--- a/common/tests/util.py
+++ b/common/tests/util.py
@@ -383,6 +383,20 @@ class Dates:
         )
 
     @classmethod
+    def short_after(cls, dt):
+        return DateTimeTZRange(
+            dt + relativedelta(days=+14),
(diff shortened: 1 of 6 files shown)


simonwo:
Closing – we need a proper decision about what to do with this stuff.

PR #111, opened 2020-11-19
Add a test hitting the HMRC sandbox API

andyhd:
Questionable things ahead:
* This commit adds a test which is skipped by default.
* This test makes real HTTP requests to https://test-api.service.hmrc.gov.uk
* This test relies on a `.env` file in the project root directory

Run the test with `pytest hmrc_sdes --hmrc-live-api`.
The test will be skipped if the command line argument is omitted.

### DIFF
--- a/hmrc_sdes/api_client.py
+++ b/hmrc_sdes/api_client.py
@@ -26,7 +26,7 @@ class HmrcSdesClient(APIClient):
             ],
         )
         params = dict(
-            token_url=settings.HMRC["token_url"],
+            token_url="{base_url}{token_url}".format(**settings.HMRC),
             client_id=settings.HMRC["client_id"],
             client_secret=settings.HMRC["client_secret"],
             include_client_id=True,
@@ -41,11 +41,22 @@ class HmrcSdesClient(APIClient):
         self.srn = settings.HMRC["service_reference_number"]
 
     def get_default_headers(self) -> dict:
+        # TODO The correct values for these headers are to be confirmed by HMRC
         return dict(
             **super().get_default_headers(),
             **{
                 "Content-Type": "application/vnd.hmrc.1.0+json; charset=UTF-8",
                 "Accept": "application/vnd.hmrc.1.0+json",
+                "Gov-Client-Connection-Method": "OTHER_DIRECT",
+                "Gov-Client-Device-ID": settings.HMRC["device_id"],
+                "Gov-Client-Local-IPs": "127.0.0.1",
+                "Gov-Client-MAC-Addresses": "00%3A00%3A00%3A00%3A00%3A00",
+                "Gov-Client-Multi-Factor": "",
+                "Gov-Client-Timezone": "UTC+00:00",
+                "Gov-Client-User-Agent": "Linux/Debian%20Buster%2010.6 (Docker/container)",
+                "Gov-Client-User-IDs": "os=test",
+                "Gov-Vendor-License-IDs": "",
+                "Gov-Vendor-Version": "tariff-management-tool=0.0.0",
             },
         )
 
--- a/hmrc_sdes/tests/test_client.py
+++ b/hmrc_sdes/tests/test_client.py
@@ -1,5 +1,7 @@
 import json
 import os
+import uuid
+from hashlib import md5
 from unittest.mock import Mock
 
 import dotenv
@@ -9,14 +11,10 @@ from common.tests import factories
 from hmrc_sdes.api_client import HmrcSdesClient
 
 
-def test_sdes_client(responses, settings):
-    settings.HMRC = {
-        "client_id": "test",
-        "client_secret": "test",
-        "service_reference_number": "test-srn",
-        "token_url": "https://test-api.service.hmrc.gov.uk/oauth/token",
-    }
+pytestmark = pytest.mark.django_db
 
+
+def test_sdes_client(responses):
     responses.add(
         responses.POST,
         url="https://test-api.service.hmrc.gov.uk/oauth/token",
@@ -61,3 +59,34 @@ def test_sdes_client(responses, settings):
             "checksumAlgorithm": "MD5",
         },
     }
+
+
+@pytest.mark.hmrc_live_api
+def test_api_call(responses, settings):
+    responses.add_passthru(settings.HMRC["base_url"])
+
+    # reload settings from env, overriding test settings
+    dotenv.read_dotenv(os.path.join(settings.BASE_DIR, ".env"))
+    settings.HMRC["client_id"] = os.environ.get("HMRC_API_CLIENT_ID")
+    settings.HMRC["client_secret"] = os.environ.get("HMRC_API_CLIENT_SECRET")
+    settings.HMRC["service_reference_number"] = os.environ.get(
+        "HMRC_API_SERVICE_REFERENCE_NUMBER"
+    )
+
+    # fetches OAuth2 access token on instantiation
+    client = HmrcSdesClient()
+    assert client.get_session().token is not None
+
+    # check fraud prevention headers
+    result = client.get(
+        f"{client.base_url}/test/fraud-prevention-headers/validate",
+    ).json()
+    assert result.get("errors") is None
+
+    # generate a dummy upload of an empty file with a valid checksum
+    upload = factories.UploadFactory()
+    upload.file = Mock(size=0)
+    upload.checksum = md5("".encode("utf-8")).hexdigest()
+
+    response = client.notify_transfer_ready(upload)
+    assert response.status_code == 204  # no data on success
--- a/settings/common.py
+++ b/settings/common.py
@@ -3,6 +3,7 @@ Django settings for tamato project.
 """
 import os
 import sys
+import uuid
 from os.path import abspath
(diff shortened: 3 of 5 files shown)


hmrc_sdes/api_client.py
```diff
@@ -46,6 +46,16 @@ def get_default_headers(self) -> dict:
...
                 "Content-Type": "application/vnd.hmrc.1.0+json; charset=UTF-8",
                 "Accept": "application/vnd.hmrc.1.0+json",
+                "Gov-Client-Connection-Method": "OTHER_DIRECT",
+                "Gov-Client-Device-ID": settings.HMRC["device_id"],
+                "Gov-Client-Local-IPs": "127.0.0.1",
+                "Gov-Client-MAC-Addresses": "00%3A00%3A00%3A00%3A00%3A00",
+                "Gov-Client-Multi-Factor": "",
+                "Gov-Client-Timezone": "UTC+00:00",
+                "Gov-Client-User-Agent": "Linux/Debian%20Buster%2010.6 (Docker/container)",
+                "Gov-Client-User-IDs": "os=test",
```
simonwo, line 56: Do what extent does this stuff need to be correct? E.g. do we need to be pulling `Debian/Buster` from some OS call? And does `os=test` need to change in a prod environment? Maybe not in scope for this PR but just asking.

hmrc_sdes/tests/test_client.py
```diff
@@ -61,3 +68,41 @@ def test_sdes_client(responses, settings):
             "checksumAlgorithm": "MD5",
         },
     }
+
+
+def test_api_call(responses, settings, pytestconfig):
+    if not pytestconfig.getoption("--test-hmrc-api"):
+        pytest.skip(msg="Not hitting HMRC API for real")
```
simonwo, line 75: How sensible would it be in the future to run this in CI? Assuming we can load test secrets in CI… or is that against the rules? Would we need to use a private CI instance for that?

simonwo (approved):
I'm a fan.

hmrc_sdes/api_client.py
andyhd: Good question, I constructed these values just to avoid errors from the Fraud Prevention Test API and I don't know what these values should be. I don't even think that these are meaningful, as we don't have an "originating device" and we're not making the request on the behalf of a user, but the test API complains if these headers are missing or blank.

hmrc_sdes/tests/test_client.py
andyhd: For infrequent builds (eg merging to master) I think it should be fine. It's only hitting the HMRC sandbox API, so no real data or effects. Just so long as we avoid spamming their API and the test doesn't run too slowly. Which reminds me - I should check that the request has a timeout.

hmrc_sdes/tests/test_client.py
```diff
@@ -61,3 +68,41 @@ def test_sdes_client(responses, settings):
             "checksumAlgorithm": "MD5",
         },
     }
+
+
+def test_api_call(responses, settings, pytestconfig):
+    if not pytestconfig.getoption("--test-hmrc-api", False):
+        pytest.skip(msg="Not hitting HMRC API for real")
```
NDevox, line 75: I did something similar for an old project where we split tests into unit and integration tests.

We marked all integration tests like so:

```python3
@pytest.mark.integration
def test_integration():
    # do something
```

Then if we wanted to run only the unit tests we could run `pytest -m "not integration"`. If we wanted to only run integration tests we could run `pytest -m "integration"` and so on.

I do like the custom flag approach similar to this however - seems a bit clearer. Pytest have an interesting example [here](https://docs.pytest.org/en/stable/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs) where you won't have to write the skip logic within the individual test. Maybe it would be better to do that?

hmrc_sdes/tests/test_client.py
```diff
@@ -61,3 +68,41 @@ def test_sdes_client(responses, settings):
...
+
+    responses.add_passthru("https://test-api.service.hmrc.gov.uk")
+
+    dotenv.read_dotenv(os.path.join(os.path.dirname(__file__), "..", "..", ".env"))
```
NDevox, line 79: `settings.BASE_DIR`?

hmrc_sdes/tests/test_client.py
```diff
@@ -61,3 +68,41 @@ def test_sdes_client(responses, settings):
...
+
+    # fetches OAuth2 access token on instantiation
+    client = HmrcSdesClient()
+    assert client.get_session().token is not None
+
+    # check fraud prevention headers
+    result = client.get(
+        f"{client.base_url}/test/fraud-prevention-headers/validate",
+    ).json()
+    pprint(result)
```
NDevox, line 98: print

hmrc_sdes/tests/test_client.py
```diff
@@ -61,3 +68,41 @@ def test_sdes_client(responses, settings):
...
+    assert result.get("errors") is None
+
+    # generate a dummy upload of an empty file with a valid checksum
+    upload = factories.UploadFactory()
+    upload.file = Mock(size=0)
+    upload.checksum = md5("".encode("utf-8")).hexdigest()
+    pprint(upload.notification_payload)
```
NDevox, line 105: print

hmrc_sdes/tests/test_client.py
```diff
@@ -61,3 +68,41 @@ def test_sdes_client(responses, settings):
...
+    dotenv.read_dotenv(os.path.join(os.path.dirname(__file__), "..", "..", ".env"))
+    settings.HMRC = {
+        "client_id": os.environ.get("HMRC_API_CLIENT_ID"),
+        "client_secret": os.environ.get("HMRC_API_CLIENT_SECRET"),
+        "token_url": os.environ.get(
+            "HMRC_API_TOKEN_URL", "https://test-api.service.hmrc.gov.uk/oauth/token"
+        ),
+        "service_reference_number": os.environ.get("HMRC_API_SERVICE_REFERENCE_NUMBER"),
+        "device_id": str(uuid.uuid4()),
+    }
```
NDevox, line 88: Aren't these already defined in the settings?

settings/common.py
```diff
@@ -289,4 +290,5 @@
         "HMRC_API_TOKEN_URL", "https://test-api.service.hmrc.gov.uk/oauth/token"
     ),
     "service_reference_number": os.environ.get("HMRC_API_SERVICE_REFERENCE_NUMBER"),
+    "device_id": str(uuid.uuid4()),
```
NDevox, line 293: This will change for every process. Is that ok?

NDevox (changes requested):
Some minor comments otherwise all OK.

I don't have an issue with skipping a test by default - I think it's good to have tests even if they're only run in specific circumstances.

For the `.env` thing, bleh. I've added a blank `.env` file to my project now as we have a few other things requiring it.

The test currently failing would ideally be fixed 😇 .

For the test making a real HTTP request against a 3rd party. Whilst I think it's fine to have as default skipped, I wouldn't run it under any CI. I've worked on a few projects with tests like these that run on every test run. Although they are useful with the initial setup when run on every test run they become a burden. It generally only acts as a signal for one of two things:

1) The 3rd party service is down.
2) The 3rd party deprecated some feature we were using and we weren't paying attention enough to realise.

Both will be signalled, at the same time if not earlier, by production failing to communicate with the API. Whereas the test hitting the API just slows things down and adds network costs (luckily for us not so bad, I've seen a client spend thousands on tests hitting 3rd parties 🤦 ).

If you want to test that you are sending the right kind of request to the API, then a regression test will do just as well. Whereas if you want to assert the communication with the API still works truly, it can be done independently by a dev running the tests with the flag provided outside of CI.

hmrc_sdes/api_client.py
stuaxo: I'd add a short TODO comment along those lines, it's possible in the future we may have a better understanding of what this API wants.

stuaxo (commented):
Overall I reckon hitting their API now makes sense (not only do we keep our understanding of their API up to date, but if they change it we will know quickly).

Once we have a good understanding of how the real API acts it would be good to replace this with something that doesn't hit an external API - along the lines of mocking the return value of request.get, using real data we have retrieved from their API).

settings/common.py
```diff
@@ -289,4 +290,5 @@
...
 }
```
stuaxo, line 294: If there are docs for these data structures from HMRC then it would be good to mention them in the comments somewhere near where they are created (further up in this file, but github makes it tricky for me to place the comment there).

settings/common.py
andyhd: Unknown. We don't really have a "device", as I believe it's intended to identify the client. I'm waiting for a response from HMRC on the correct values for these headers.

hmrc_sdes/tests/test_client.py
andyhd: That is nice 👍

settings/common.py
NDevox: 👍

NDevox (approved):

PR #112, opened 2020-11-23
Regulation information text should contain URL and public ID

simonwo:
This is the format tentatively agreed with the Trade Tariff Service for adding extra information about UK legislation. We now output the pipe format in the order:

    information_text "|" public_identifier "|" url

And we also detect this in XML so that we can load any TAP-produced regulations in this format.

CC @matthewford – does this still match the format expected by the Trade Tariff Service?

### DIFF
--- a/regulations/tests/test_importer.py
+++ b/regulations/tests/test_importer.py
@@ -36,7 +36,13 @@ def create_and_test_m2m_regulation(
             "replacement_indicator": test_regulation.replacement_indicator,
             "community_code": test_regulation.community_code,
             "stopped": test_regulation.stopped,
-            "information_text": test_regulation.information_text,
+            "information_text": "|".join(
+                [
+                    test_regulation.information_text,
+                    test_regulation.public_identifier,
+                    test_regulation.url,
+                ]
+            ),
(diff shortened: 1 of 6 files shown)


regulations/tests/test_importer.py
```diff
@@ -36,7 +36,13 @@ def create_and_test_m2m_regulation(
...
             "community_code": test_regulation.community_code,
             "stopped": test_regulation.stopped,
-            "information_text": test_regulation.information_text,
+            "information_text": "|".join(
+                [
+                    test_regulation.information_text,
+                    test_regulation.public_identifier,
+                    test_regulation.url,
+                ]
+            ),
```
simonwo, line 45: Any idea why we manually generate XML here and don't use the serializers?

NDevox: I can't remember in detail but from memory the serializers don't play nicely when dealing with non-existent relations (i.e. unsaved objects via FKs). They like to insist that the relations actually exist. So it was easier to manually do generate the data.

NDevox (approved):
LGTM

PR #113, opened 2020-11-24
Tp 409 add celery

stuaxo:
Initially add celery, properly integrating this with the previous upload work.


docker-compose.yml
```diff
@@ -41,6 +49,32 @@ services:
       - runserver
       - 0.0.0.0:8000
 
+  celery_worker:
+    image: tamato
+    restart: always
+    env_file: .env
+    entrypoint: celery
+    command: -A celery_app.celery beat --loglevel=info
```
NDevox, line 57: That's not a worker!

common/celery.py
```diff
@@ -0,0 +1,17 @@
...
+# set the default Django settings module for the 'celery' program.
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "settings")
+
+app = Celery("tamato")
+
+# Using a string here means the worker doesn't have to serialize
+# the configuration object to child processes.
+# - namespace='CELERY' means all celery-related configuration keys
+#   should have a `CELERY_` prefix.
+app.config_from_object("django.conf:settings", namespace="CELERY")
```
NDevox, line 14: I don't think there are any Celery settings included in this PR?

docker-compose.yml
```diff
@@ -9,6 +9,14 @@ services:
...
       - "5432:5432"
 
+  redis:
+    image: redis
+    restart: always
+    ports:
+      - "6378:6379"
+    volumes:
+      - .:/redis.conf
+
```
NDevox, line 19: Do we want to use redis over rabbitmq for this? Given https://docs.celeryproject.org/en/stable/getting-started/brokers/redis.html#redis-caveats.

I do want to say I would like to use redis for caching anyway 😄  although I don't know how well we can handle mixing the concerns.

NDevox (changes requested):
Small things and asking to discuss redis vs rabbit

docker-compose.yml
stuaxo: Oh no!  oops!

common/celery.py
stuaxo: There are not, but... I probably should... I'll add some logging settings to justify this..

docker-compose.yml
stuaxo: Maybe... not, or at least YAGNI for now so I'll switch to RabbitMQ.

Good point about separation of concerns, have used this before when the concern was more "we have redis already and getting anything else is difficult" - it turns out redis was the only backend that supported some of the chord/group primitives in 4.0, so worked out well.

simonwo (commented):
LGTM, but can we add some info to the README about needing to set up Redis/Rabbit? https://github.com/uktrade/tamato#running-locally

docker-compose.yml
stuaxo: After a conversation on Teams, we decided to keep Redis as it will be needed for other things fairly soon as well.

stuaxo:
> LGTM, but can we add some info to the README about needing to set up Redis/Rabbit? https://github.com/uktrade/tamato#running-locally

Will do, same for celery itself.

docker-compose.yml
stuaxo: Sorted.

common/celery.py
stuaxo: Sorted

stuaxo:
> > LGTM, but can we add some info to the README about needing to set up Redis/Rabbit? https://github.com/uktrade/tamato#running-locally
> 
> Will do, same for celery itself.

Ran this locally and updated, let me know if anything is missing or could be improved for clarity.

NDevox (approved):
LGTM

README.md
```diff
@@ -82,6 +133,30 @@ Run the script to see the command line arguments:
...
+The celery job UUID is output and the command quits.  To see output switch to the celery workers console.
+A more ergonomic way of launching the celery job is to launch the management command:
+
+    ./manage.py upload_workbaskets
+
+### Dump workbaskets
+
+Dump workbaskets to stdout or to a file.
+
+     ./manage.py dump_workbaskets [-o filename]
```
andyhd, line 155: I'm not sure workbaskets should be the term used. Workbaskets are purely a user interface concept. We upload bunches of transactions.

andyhd (commented):
Still reviewing...

conftest.py
```diff
@@ -18,6 +18,17 @@
 from common.tests.util import Dates
 from exporter.storages import HMRCStorage
 
+pytest_plugins = ("celery.contrib.pytest",)
```
andyhd, line 21: Just for consistency with other pytest plugins, maybe prefer adding `pytest-celery` to `requirements.txt` to this.

exporter/tasks.py
```diff
@@ -0,0 +1,10 @@
+from celery import shared_task
+from django.core.management import call_command
+
+
+@shared_task
+def upload_workbaskets():
+    # TODO the code in upload_workbaskets should be split out
+    #      and called from both, instead of calling the management
+    #      command from here.
```
andyhd, line 9: This could be clearer. Does it mean refactor the `upload_workbaskets` command `handle` method to call a new function and replace `call_command("upload_workbaskets")` here with a call to that new function?

andyhd (approved):
LGTM, except that naming quibble

exporter/tasks.py
stuaxo: Yes, thanks for deciphering that :)

README.md
stuaxo: Probably the best thing is if I change the name to `transactions` here and continue in the next ticket, which integrates this with the notification work you've done.

I think there will be further integration work once the ticket you're currently working on is in.

Dockerfile
```diff
@@ -25,7 +25,8 @@ RUN groupadd -g 1000 tamato && \
     useradd -u 1000 -g tamato -m tamato
 USER tamato
 
-WORKDIR /home/tamato
+RUN mkdir /home/tamato/app
```
andyhd, line 28: Not needed - `WORKDIR` creates the directory if it doesn't exist. Best to keep the number of layers in the Dockerfile to a minimum.

Dockerfile
```diff
@@ -43,5 +44,7 @@ RUN touch .env
 # collect static files for deployment
 RUN python manage.py collectstatic --noinput
 
+ADD ./ /home/tamato/app/
```
andyhd, line 47: Is this not already done in line 36?

andyhd: Also, `COPY` is preferred over `ADD` unless you need the extra functionality

stuaxo: I'll fix these in TP-464.   Need to check the difference between ADD and COPY.  Not sure if this was the thing that let me edit the files in the web app live or not, but that's what I was aiming for.

PR #115, opened 2020-12-10
Tp 464 fixup exporter tests

stuaxo:
Re-add exporter tests.
Catch up changes to docker-compose, Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:13.12-buster-slim AS jsdeps
 
-RUN apt update -y
-RUN apt install  -y g++ build-essential python3
+RUN apt-get update -y
+RUN apt-get install -y g++ build-essential python3
 
 COPY . .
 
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -38,13 +43,8 @@ COPY --chown=tamato:tamato --from=jsdeps node_modules/govuk-frontend/govuk node_
...
 COPY --chown=tamato:tamato --from=jsdeps webpack-stats.json ./
 
-# empty .env file to prevent warning messages
-RUN touch .env
-
 # collect static files for deployment
-RUN python manage.py collectstatic --noinput
-
-ADD ./ /home/tamato/app/
+#RUN python manage.py collectstatic --noinput
```
NDevox, line 47: Do we not want this now? We do have something of a frontend

NDevox (commented):
All fine, just one q

NDevox (approved):
LGTM, squash commits though please!

Dockerfile
stuaxo: We do want this, I've just pushed the fix to re-enable it :)

stuaxo: I knew something would go-awry patching this in the night.

stuaxo:
> LGTM, squash commits though please!

Will do, hope this isn't because I forgot to do this before - luckily squash and merge seems to be the default :)

PR #117, opened 2020-12-14
Fix failing tests

andyhd:
* GOV.UK Frontend change to table template breaking Jinja - fixed
* xlrd no longer supports xlsx format - skipping tests for now

### DIFF
--- a/common/jinja2.py
+++ b/common/jinja2.py
@@ -32,6 +32,7 @@ class GovukFrontendExtension(NunjucksExtension):
             # fix nested attribute access
             # NunjucksUndefined ought to catch these, but doesn't
             nested_attrs = [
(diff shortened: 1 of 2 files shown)


stuaxo:
>     * xlrd no longer supports xlsx format - skipping tests for now

I'm a bit shocked at this (I do believe it - because stranger things have happened) I wonder what they were thinking.

andyhd:
I think the maintainer is encouraging us to use a different library.
I should just pin the previous version for now.

stuaxo (approved):

stuaxo:
Approved (Would have approved earlier but my comment went in before the build had completed).

NDevox (approved):

PR #118, opened 2020-12-22
TP-396 - Trigger HMRC async upload

stuaxo:
Move upload and dump envelopes to be based around Envelopes rather than Workbaskets.

- Rename upload_workbaskets, dump_workbaskets -> upload_transactions dump_transactions
- Make upload_transactions create concrete Envelope instances in the database for upload with correct enevelope ids.
- Add new_envelope to populate Envelope.envelope_id and test.
- Add `-l` option to upload_envelope to optionally bypass celery.
- Add Envelope templates (based on Workbasket templates)
- Move id padding to templates.
- Refactor EnvelopeSerializer to separate concerns (rendering / writing) and make this easier for the coming work on splitting envelopes.
- Workbasket validation: fix the ordering check, ability to ignore XML namespace declaration.

Not in scope / Moved out of scope:
- Uploading again on failure (will need another ticket)
- Transactions ordering: moved to another ticket, due to complexity.
- Moving EnvelopeSerializer out of the importer app (this PR will follow, it would have impeded readability).
- Outputting split envelopes to multiple files, moved out of this PR, will be in a following PR


stuaxo:
Probably could use some more detailed tests around serialization, this will make sense at the point of splitting envelopes into the right size, since a little restructuring may be needed at this point.

stuaxo:
I've taken this off "draft" so this can start to be reviewed.   There's a little bit around envelope_ids that will be pushed shortly.

andyhd:
This PR needs a description of what changes are contained and why.

workbaskets/models.py
```diff
@@ -6,32 +6,46 @@
...
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
-from django.db.models import Manager
 from django.db.models import Prefetch
 from django.db.models import QuerySet
 from django_fsm import FSMField
 from django_fsm import transition
 
-from common.models import TimestampedMixin
+from common.models import TimestampedMixin, Transaction
```
andyhd, line 14: One import per line

workbaskets/tests/test_models.py
```diff
@@ -1,6 +1,8 @@
 import pytest
+from pytest_django.asserts import assertQuerysetEqual
```
andyhd, line 2: Unused import

andyhd (approved):
Minor things to tidy up, but looks good to me overall.

exporter/apps.py
```diff
@@ -0,0 +1,5 @@
+from common.app_config import CommonConfig
+
+
+class ExporterConfig(CommonConfig):
+    name = "envelopes"
```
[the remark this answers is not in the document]
stuaxo: Oh, good catch !

exporter/management/commands/util.py
```diff
@@ -1,52 +1,26 @@
...
             try:
                 schema.assertValid(xml)
             except etree.DocumentInvalid as err:
-                self.stderr.write("Envelope did not validate against XSD:")
-                self.stderr.write(str(err.error_log))
-                sys.exit(1)
+                error_log = err.error_log
+                raise CommandError(
+                    f"Envelope did not validate against XSD:\n{error_log}"
+                )
```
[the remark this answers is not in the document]
stuaxo: Ah, this is mis-remembering the django-style guide on fstrings will fix all these :)

exporter/tasks.py
```diff
@@ -1,10 +1,79 @@
...
+
+        content_file = ContentFile(envelope_data)
+        upload = Upload()
+        upload.envelope = envelope
+        upload.file = content_file
+        upload.checksum = md5("".encode("utf-8")).hexdigest()
+
+        upload.notify_hmrc()
+
+        # destination = storage.save(full_filename, content_file)
```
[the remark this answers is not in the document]
stuaxo: Doh

workbaskets/tests/test_models.py
```diff
@@ -26,3 +28,23 @@ def test_workbasket_transactions():
...
+
+
+def test_envelope_of_transactions_populates_envelope(
+    approved_workbasket, populate_workbasket
+):
+    """
+    Workbasket.envelope_of_transactions create an envelope containing
+    the same transactions as the workbasket.
+    """
+    populate_workbasket(approved_workbasket)
```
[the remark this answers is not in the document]
stuaxo: Certainly can do, I wasn't sure if we have a preference for fixtures over calling factories.

exporter/tasks.py
```diff
@@ -1,10 +1,79 @@
...
+            raise
 
 
 @shared_task
-def upload_workbaskets():
-    # TODO the code in upload_workbaskets should be split out
-    #      and called from both, instead of calling the management
-    #      command from here.
-    call_command("upload_transactions")
+def upload_workbaskets() -> (Optional[str], Optional[BaseException]):
```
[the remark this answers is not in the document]
stuaxo: Ah, good catch it doesn't any more, will fix.

exporter/tasks.py
```diff
@@ -1,10 +1,79 @@
...
+    with atomic():
+        envelope = WorkBasket.objects.filter(
+            status=WorkflowStatus.READY_FOR_EXPORT
+        ).envelope_of_transactions()
+        envelope_data = serialize_envelope_as_xml(envelope)
+        try:
+            validate_envelope(envelope_data)
+        except (TaricDataAssertionError, etree.DocumentInvalid):
+            # Nothing to log here - validate_envelope has already logged the issue.
+            raise
```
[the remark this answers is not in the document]
stuaxo: Yes - and this must be why I initially returned the Exception instead of raising it ... something to reinstate, and maybe a policy of reading my own docstrings.

exporter/views/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+from common.renderers import TaricXMLRenderer
+from exporter.models import Envelope
+from exporter.serializers import EnvelopeSerializer
+
+
+class EnvelopeViewSet(viewsets.ReadOnlyModelViewSet):
+    """
+    API endpoint that allows envelopes to be viewed.
+
+    The API is for convenience, mostly used
```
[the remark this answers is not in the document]
stuaxo: My worst vice :-/ I'm sure there is a tool that can find these..

exporter/views/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+        renderers.JSONRenderer,
+        renderers.BrowsableAPIRenderer,
+        TaricXMLRenderer,
+    ]
+    template_name = "exporter/taric/envelope_list.xml"
+
+    def get_template_names(self, *args, **kwargs):
+        if self.detail:
+            return ["exporter/taric/envelope_detail.xml"]
+        return ["exporter/taric/envelope_list.xml"]
```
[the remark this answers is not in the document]
stuaxo: At least part of it.. I've cargo culted this from workbaskets, so will apply any fix there.

exporter/serializers.py
```diff
@@ -0,0 +1,42 @@
...
+            "order",
+        ]
+
+
+class EnvelopeSerializer(serializers.ModelSerializer):
+    envelope_id = serializers.SerializerMethodField(read_only=True)
+    transactions = EnvelopeTransactionSerializer(read_only=True, many=True)
+
+    def get_envelope_id(self, object: models.Envelope):
+        return str(object.pk).zfill(6)
```
simonwo, line 34: I think you're already working on this, but obviously need a real thing here.

taric/models.py
```diff
@@ -1,10 +1,36 @@
...
+        """Return all envelopes for a year, defaulting to this year.
+
+        :param year: int year, defaults to this year.
+
+        Limitation:  This queries envelope_id which only stores two digit dates.
+        """
+        if year is None:
+            date = datetime.now(tz=timezone.utc).replace(
+                hour=0, minute=0, second=0, microsecond=0
+            )
```
simonwo, line 24: Not just a `date` class? Then you wouldn't have to faff with zeroing or worry about timezones.

taric/models.py
```diff
@@ -30,11 +56,48 @@ class Envelope(models.Model):
...
+            counter = 1
+        else:
+            year = int(envelope.envelope_id[:2])
+            counter = int(envelope.envelope_id[2:]) + 1
+
+            if counter > 9999:
+                raise ValueError(
+                    "Cannot create more than 9999 Envelopes on a single day."
```
simonwo, line 83: Should be `a single year`?

taric/models.py
```diff
@@ -30,11 +56,48 @@ class Envelope(models.Model):
     tariff in the sequence defined by the transaction IDs.
     """
 
+    objects = EnvelopeQuerySet.as_manager()
+
     envelope_id = EnvelopeId(unique=True)
```
simonwo, line 61: So I think unfortunately there are two unhelpul facts of life:

- Envelope IDs will not be unique – e.g. if an error is encountered the expectation is that we will submit an envelope with the same ID agian.
- Envelope IDs should represent the order in which data leaves the system for HMRC. Does a new `Envelope` only get created when exporting? If so there's no problem, but if not we may need to make sure the ID only gets set at that point.

workbaskets/models.py
```diff
@@ -1,37 +1,52 @@
...
+        return self.prefetch_related(
             Prefetch("tracked_models", queryset=q_annotate_record_code)
         )
 
+    def envelope_of_transactions(self) -> Optional[Envelope]:
+        """Return an Envelope populated with the transactions from this queryset."""
+        envelope = Envelope.new_envelope()
+
+        workbasket_pks = self.values_list("pk", flat=True)
+        transactions = Transaction.objects.filter(workbasket__pk__in=workbasket_pks)
```
simonwo, line 40: Can we make sure that the `transactions` are entered into the envelope using their own `order` field? I.e. add an `order_by("order")` to this line.

common/tests/util.py
```diff
@@ -117,7 +121,7 @@ def validate_taric_xml_record_order(xml):
         subrecord_code = record.findtext(".//subrecord.code", namespaces=xml.nsmap)
         full_code = record_code + subrecord_code
         if full_code < last_code:
-            raise AssertionError(
+            raise TaricDataAssertionError(
                 f"Elements out of order in XML: {last_code}, {full_code}"
             )
```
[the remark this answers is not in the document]
stuaxo: Thanks for tracking this down, this bug has been a pain !

exporter/serializers.py
stuaxo: Yes !

I'm going to leave your comment so I don't forget.

TBH, I think the zero padding should happen in the templates.

taric/models.py
stuaxo: It would make more sense.

taric/models.py
stuaxo: Envelopes are only created on export,  but the will need a status along the lines of workbaskets so retrying can work.

Based on our keep-everything approach, upload failure we save the old Envelope with a failed status and create a new one on the next upload.

Edge cases:
If we upload to s3, but then we don't get the right acknowledgement from their API should we:
a - delete the item.
b - leave the item, and next time just call their API.

workbaskets/models.py
stuaxo: Yes, good catch.

taric/models.py
stuaxo: Oops - spent too long away from the spec and wrote this message how I thought it should work, not how it *does* work.

taric/models.py
simonwo: Excellent thoughts, agreed.

For the edge case, I'd think option B – we should be resilient to API failures. So any state machine may need a `UPLOADED` state this is distinct from a `SENT` state…

exporter/management/commands/upload_transactions.py
```diff
@@ -1,34 +1,62 @@
...
+    user needs to run a task, but celery infrastructure is unavailable.
+    """
+    try:
+        if local:
+            return task.apply(args=args, kwargs=kwargs).get()
+        return task.apply_async(args=args, kwargs=kwargs).get()
+    except kombu.exceptions.OperationalError:
+        # OperationalError here usually indicate that it was not possible to
+        # connect to the celery backend, e.g. redis is not accessible/up.
+        sys.exit("Timeout connecting to celery.")
```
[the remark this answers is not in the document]
stuaxo: Fair enough, though I'll add a log with the hint, since almost every time it's because the user didn't run celery or redis (or on live there is a connectivity issue with same)

Almost every time it's because the user forgot to run celery or redis.   On live it's usually a connectivity issue with the same.

taric/models.py
stuaxo: I guess SENT is after UPLOADED ?

READY_FOR_UPLOAD->UPLOADED->SENT.

Once we find out more about their API SENT may change to something ACCEPTED, or REFUSED_WITH_ERRORS .

exporter/management/commands/dump_transactions.py
```diff
@@ -1,10 +1,16 @@
-from exporter.management.commands.util import get_envelope_of_active_workbaskets
-from exporter.management.commands.util import WorkBasketBaseCommand
+import sys
+
+from django.db.transaction import atomic
+
+from exporter.management.commands.util import (
+    TransactionsBaseCommand,
+)
```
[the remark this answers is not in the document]
stuaxo: It turns out nothing, but I'm going to remove this anyway.

I don't mind that form for multiple imports, but for just one it doesn't make any sense !

exporter/models.py
```diff
@@ -1,25 +1,27 @@
...
 from hmrc_sdes.api_client import HmrcSdesClient
 from taric.models import Envelope
 
 
 def to_hmrc(instance: "Upload"):
     """Generate the filepath to upload to HMRC"""
-
-    return f"tohmrc/staging/{instance.filename}"
+    full_filename = str(Path(settings.HMRC_STORAGE_DIRECTORY) / instance.filename)
+    return full_filename
```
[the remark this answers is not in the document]
stuaxo: Fixed the assignment, I need to check if this is duplication now, so won't hit *Resolve* just yet.

exporter/management/commands/upload_transactions.py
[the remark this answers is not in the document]
stuaxo: Exception raised, with log message.

simonwo:
Where are we at with this? Looks like its getting large!

stuaxo:
@simonwo yes, this is true - should have a better idea on sizing things in this project after this at least :)

We are at:
In the bits I need to push up now, it serializes everything into 40mb temp files + validates those

I am:
Finishing moving it from uploading single envelopes (these originally just a ContentFile) to multiple envelopes - the temporary files.

Will push these shortly, then there are some tests + it's ready for review.

taric/models.py
stuaxo: Error handling / retry can move to it's own ticket so this has a chance of landing :)

common/jinja2/common/taric/start_envelope.xml
```diff
@@ -1 +1 @@
-<env:envelope xmlns="urn:publicid:-:DGTAXUD:TARIC:MESSAGE:1.0" xmlns:env="urn:publicid:-:DGTAXUD:GENERAL:ENVELOPE:1.0" id="{{ envelope_id }}">
+<env:envelope xmlns="urn:publicid:-:DGTAXUD:TARIC:MESSAGE:1.0" xmlns:env="urn:publicid:-:DGTAXUD:GENERAL:ENVELOPE:1.0" id="{{ '%06d' % envelope_id | int }}">
```
andyhd, line 1: `'{:06d}'.format(envelope_id | int)` would be better here - the `%` operator is pretty much deprecated and there's no confusion over the precedence of `%` and `|`.

stuaxo: I'll double check, `int` vs `str` I seem to recall something about jinja turning them into strings (though - I should double check this, as I'm not sure I even use this in Jinja any more)

importer/management/commands/utils.py
```diff
@@ -658,75 +659,102 @@ def __init__(
...
+            },
+        )
+
+    def render_envelope_end(self) -> str:
+        return render_to_string(template_name="common/taric/end_envelope.xml")
 
-    def new_envelope(self) -> None:
-        self.end_envelope()
+    def start_next_envelope(self) -> None:
+        # Setup data for the next envelope.
```
NDevox, line 725: docstring instead of comment?

workbaskets/models.py
```diff
@@ -29,6 +32,15 @@ def prefetch_ordered_tracked_models(self) -> QuerySet:
             Prefetch("tracked_models", queryset=q_annotate_record_code),
         )
 
+    def ordered_transactions(self):
+        """This Workbaskets transactions in creation order.
+
+        Note: tracked_models are ordered by record_code, subrecord_code by TransactionManager"""
+        workbasket_pks = self.values_list("pk", flat=True)
+        return Transaction.objects.filter(workbasket__pk__in=workbasket_pks).order_by(
```
NDevox, line 40: Could do this as

```python
Transaction.objects.filter(workbasket__pk__in=Subquery(workbasket_pks))
```

Just to make it a bit neater SQL wise.

NDevox:
Squash the commits!

exporter/management/commands/dump_transactions.py
```diff
@@ -1,12 +1,18 @@
...
-from exporter.management.commands.util import get_envelope_of_active_workbaskets
+import io
+import os
+import sys
+
+from django.core.management import BaseCommand
+from django.db.transaction import atomic
+
+from exporter.tasks import validate_envelope
+from importer.management.commands.utils import EnvelopeSerializer
```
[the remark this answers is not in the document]
stuaxo: Will do this in a following PR, as it will bloat this one a lot.

common/tests/util.py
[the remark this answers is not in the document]
stuaxo: Fixed.

exporter/tasks.py
```diff
@@ -1,10 +1,122 @@
...
+def validate_envelope(envelope_file, skip_declaration=False):
+    """Validate envelope content for XML issues and data order issues.
+
+    raises DocumentInvalid | TaricDataAssertionError
+    """
+    with open(settings.TARIC_XSD) as xsd_file:
+        if skip_declaration:
+            declaration_line = envelope_file.readline()
+            if declaration_line != '<?xml version="1.0" encoding="UTF-8"?>\n':
+                logger.error(
```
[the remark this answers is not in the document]
stuaxo: Changed to warning :)

exporter/management/commands/dump_transactions.py
```diff
@@ -28,25 +37,37 @@ def add_arguments(self, parser):
...
 
-        f = self.get_output_file(options["filename"])
-        f.write(envelope.decode("utf-8"))
-        if f not in (self.stdout, self.stderr):
+        output_file = self.get_output_file(options["filename"])
+        output_file.write(buffer.getvalue())
+        if output_file not in (self.stdout, self.stderr):
             # Closing stdout or stderr can make a capturing pytest unhappy.
-            f.close()
+            output_file.close()
```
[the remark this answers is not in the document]
stuaxo: Went with making it more friendly to `with` by making a wrapper.

exporter/tasks.py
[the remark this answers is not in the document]
stuaxo: Logic is different now.

PR #120, opened 2020-12-26
TP-332: Implement Update and Deletion importer system (with TP-371)

NDevox:
It is required that the importer not only allows for creating
items, but updating and deleting them as well.

Given that this system is intended to be an Append-Only-Log it
should not be possible for Update and Delete events to actually
update or delete a row in the database. Therefore it is necessary
to provide a method of linking rows across the database so that
they can be identified as versions of the same item. This was
initially intended to be done with the predecessor One-to-One link
on TrackedModel. However the One-to-One nature is not conducive to
various scenarios (errors and mistakes, the possibility of
multiple versions stemming from one previous version).

This commit introduces a more robust Append-Only-Log by overriding
the save method of TrackedModel. This disables saving on a single
row once that row has been created and approved. A new model has
also been introduced, termed VersionGroup. TrackedModel holds a
Foreign Key to VersionGroup giving a One-to-Many relationship
between the two models. On saving a new row, if it is a creation
event, a new VersionGroup is created and attached to the
TrackedModel. On update and deletion events a new VersionGroup is
not made, instead an existing one is found using the identifying
fields of the TrackedModel, this found VersionGroup is considered
the grouping for that TrackedModel and is attached to the new row.
In turn the VersionGroup is updated outside of the Append-Only-Log
to reflect the most recent TrackedModel row it has been attached to.

In tandem with this some refactoring has taken place:

- With the removal of predecessor and successor on TrackedModel most
of the querying around versions had to be updated to reflect the
use of VersionGroup instead.
- The querysets have been moved to their own folder as well as the
custom fields.
- Factories have been updated to default to approved workbaskets.
- Some common testing utilities were simplified.
- Importer testing was unified across the models.
- GoodsNomenclatureIndent importer now handles copying children on update.


conftest.py
```diff
@@ -214,7 +236,135 @@ def check(dependency_name, dependency_factory, dependent_factory):
...
+
+    return check
+
+
+@pytest.fixture(params=[UpdateType.UPDATE, UpdateType.DELETE])
+def update_imported_fields_match(
+    imported_fields_match,
+    date_ranges,
+    request,
+):
```
stuaxo, line 304: Missing a docstring.

PR #121, opened 2020-12-26
TP-333: Allow fetching of "latest" versions of a relation in a single query (with TP-371)

NDevox:

```python3
    # must be a TrackedModel
```

```python3
```


commodities/models.py
```diff
@@ -153,6 +167,31 @@ def has_measure_in_tree(self):
...
+        ):
+            new_valid_between = DateTimeTZRange(
+                valid_between.lower, new_valid_between.upper
+            )
+        if not new_valid_between.upper or (
+            valid_between.upper and new_valid_between.upper > valid_between.upper
+        ):
+            new_valid_between = DateTimeTZRange(
+                new_valid_between.lower, valid_between.upper
+            )
```
stuaxo, line 185: ^^ hopefully we have tests for each of these `if`s .. :)   [I haven't scrolled down yet]

commodities/tests/test_importer.py
```diff
@@ -198,21 +276,42 @@ def test_goods_nomenclature_indent_importer_create_with_parent_high_indent(valid
...
-    assert (
-        db_indent.indented_goods_nomenclature.sid
-        == indent.indented_goods_nomenclature.sid
+
+@pytest.mark.parametrize("update_type", [UpdateType.UPDATE, UpdateType.DELETE])
+def test_goods_nomenclature_indent_importer_update_with_parent_high_indent(
+    valid_user, update_type, date_ranges
+):
+    parent_indent = None
+    for idx in range(1, 6):
```
stuaxo, line 287: A quick comment about what the range here is (same for other hardcoded numbers around here).

commodities/tests/test_importer.py
```diff
@@ -198,21 +276,42 @@ def test_goods_nomenclature_indent_importer_create_with_parent_high_indent(valid
...
+        valid_between=date_ranges.adjacent_no_end,
+        update_type=update_type,
     )
-    assert db_indent_node.get_parent() == parent_indent
-    assert db_indent.valid_between.lower == indent.valid_between.lower
 
+    db_indent = make_and_get_indent(updated_indent, valid_user, depth=4)
 
-def test_goods_nomenclature_indent_importer_multiple_parents(valid_user, date_ranges):
+    compare_indent(updated_indent, db_indent, 4, 1, 6)
```
stuaxo, line 309: Docstring for this test, and a comment about these numbers would be nice.

commodities/tests/test_importer.py
NDevox: Done

commodities/tests/test_importer.py
NDevox: Switched over to using keyword args. Added a docstring

PR #123, opened 2020-12-29
TP-462: Add a distributed cache for the nursery and fix handlers

NDevox:
The TaMaTo app is most likely going to run on a distributed system
once deployed. As it stands the cache for both the django system
and importer nursery rely on localised stateful systems (e.g. a file
based or in memory cache).

This commit adds redis as the default cache for django, and adds
several cache "engines" for the nursery, including a redis engine
which can be used. The default is left as the in memory engine.

On top of this the importing system has been improved to better
handle foreign key references - especially those handlers which
rely on foreign keys for identifying fields. All new models classed
as "current" are also now cached  based on their identifying fields,
storing both the primary key and the class name in the cache. This
is to reduce the number of queries required when importing models
as the cache can be used instead of the database.

Measures have been updated to better reflect their effective end date
(this can now be generated within a queryset reducing query number again).

Database indexes have been added for many fields - focusing on fields
used in validator queries - this reduces query time significantly.


common/models/records.py
```diff
@@ -498,7 +495,63 @@ class ExampleModel(TrackedModel):
                 ) from e
         return getattr(self, relations[item]).current_version
 
+    @classmethod
+    def get_handler_link_fields(cls, link_fields=None):
```
stuaxo, line 499: Docstring please :)

common/tests/factories.py
```diff
@@ -698,7 +703,7 @@ class Meta:
     measure_component_applicability_code = FuzzyChoice(ApplicabilityCode.values)
     origin_destination_code = FuzzyChoice(ImportExportCode.values)
     order_number_capture_code = OrderNumberCaptureCode.NOT_PERMITTED
-    measure_explosion_level = FuzzyChoice(range(2, 11, 2))
+    measure_explosion_level = 2
```
stuaxo, line 706: Just a comment to appreciate naming here.

common/tests/factories.py
```diff
@@ -748,15 +753,39 @@ def terminating_regulation(self):
...
+        if "measure_type" in kwargs and not leave_measure:
+            kwargs["measure_type"] = cls.measure_type_explosion(
+                kwargs["measure_type"], kwargs.get("goods_nomenclature")
+            )
+        obj = model_class(*args, **kwargs)
+        obj.save()
+        return obj
+
+    @staticmethod
+    def measure_type_explosion(measure_type, goods_nomenclature):
```
stuaxo, line 768: measure_type_explosion, or explode_measure_type ?

conftest.py
```diff
@@ -186,7 +186,7 @@ def must_exist():
     def check(dependency_name, dependent_factory):
         non_existent_id = -1
 
-        with pytest.raises(ValidationError):
+        with pytest.raises((ValidationError, ObjectDoesNotExist)):
```
stuaxo, line 189: OK, can see this working...  but this seems a bit of a test-smell that that this test can be either.

docker-compose.yml
```diff
@@ -5,11 +5,19 @@ services:
...
 
   redis:
+    image: redis
+    restart: always
+    ports:
+      - "6379:6379"
+    volumes:
+      - .:/redis.conf
+
+  redis2:
```
stuaxo, line 20: OK, redis/redis2 ... but maybe we can name them for their purposes later.

docker-compose.yml
```diff
@@ -28,60 +36,35 @@ services:
...
-    env_file: .env
-    entrypoint: /home/tamato/.local/bin/celery
-    command: -A common.celery beat -l info
-    links:
-      - celery_worker
-      - redis
-    depends_on:
-      - redis
-      - tamato
-
```
stuaxo, line 84: Poor celery ?

docker-compose.yml
```diff
@@ -28,60 +36,35 @@ services:
...
 volumes:
   tamato-data-volume:
   tamato-s3-volume:
+#  tamato-data-volume:
```
stuaxo, line 69: Commented... hopefully sorted in the washup.

common/tests/factories.py
NDevox: 😄

common/models/records.py
NDevox: Gets one! https://github.com/uktrade/tamato/pull/124/files#diff-5a0dbdf1479a8cadde19a65a656007be46fb83c95df0ef25139afc6bf5e8fe76R133-R138

docker-compose.yml
NDevox: This gets cleaned up later to just `redis` (in 124). I've not actually tested to see how well that works, but will find out soon.

docker-compose.yml
NDevox: Hmmm... Not sure why that got knocked out on this one, will have to add it back later.

docker-compose.yml
NDevox: yes quite

stuaxo (approved):

common/tests/factories.py
NDevox: I'm going to say `measure_type_explosion` 😉

importer/cache/redis.py
```diff
@@ -0,0 +1,46 @@
...
+
+        return result
+
+    @prefix_key_decorator
+    def put(self, key, obj):
+        cache.set(key, obj, timeout=None)
+
+    def keys(self):
+        prefix = getattr(settings, "IMPORTER_CACHE_PREFIX", "__IMPORTER_CACHE")
+        return cache.keys(f"{prefix}*") + cache.keys(f"'b{prefix}*")
```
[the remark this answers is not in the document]
NDevox: This is a legacy from before when I was using binary keys. Need to remove it now.

measures/querysets.py
```diff
@@ -0,0 +1,83 @@
...
+                        Func(F("valid_between"), function="LOWER"),
+                        F("generating_regulation__effective_end_date"),
+                        Value("[]"),
+                        function="TSTZRANGE",
+                    ),
+                ),
+                default=F("valid_between"),
+                output_field=TaricDateTimeRangeField(),
+            )
+        )
```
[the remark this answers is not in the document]
NDevox: 😄  I've got to say I'm getting quite well versed with the more intricate query system in Django.

I really wish I didn't need to be.

importer/taric.py
```diff
@@ -84,14 +98,55 @@ def save(
         :param workbasket_id: The primary key of the workbasket to add transactions to
         """
         logging.debug(f"Saving transaction {self.data['id']}")
-        transaction = workbasket.get_transaction(
-            import_transaction_id=int(data["id"]),
+
+        composite_key = str(envelope.id) + self.data["id"]
+        try:
+            cursor = connection.cursor()
+            while 1:
```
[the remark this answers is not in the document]
NDevox: Huh, no idea. Going to survey that one

importer/management/commands/split_taric.py
```diff
@@ -0,0 +1,199 @@
...
+        output_dir,
+        files_in_progress,
+        len(file_map[record_code]),
+    )
+
+    with open(filepath, "ab") as code_file:
+        code_file.write(
+            etree.tostring(
+                transaction
+            )  # pythons XML doesn't write namespaces back correctly.
```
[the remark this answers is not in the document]
NDevox: I haven't actually found any docs saying this, it just ended up being what happened.

Looks like there may be something to do with registering namespaces which I thought we did (not convinced we do now) which may fix it.

Will look into that in a future commit.

importer/management/commands/split_taric.py
```diff
@@ -0,0 +1,199 @@
...
+            "env:app.message/ns2:transmission/ns2:record/ns2:goods.nomenclature/ns2:producline.suffix",
+            nsmap,
+        )
+        item_id = min(item.text for item in item_ids) if item_ids else "999999999999"
+        indent = min(indent_obj.text for indent_obj in indents) if indents else "99"
+        suffix = (
+            min("00" if suffix.text != "80" else "80" for suffix in suffixes)
+            if suffixes
+            else "99"
+        )
```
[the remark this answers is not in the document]
NDevox: Done

importer/management/commands/import_taric.py
```diff
@@ -10,27 +16,113 @@
...
-            if event == "start_ns":
-                nsmap.update([elem])
+        if event == "end":
+            if handler.end(elem):
+                elem.clear()
 
-            if event == "end":
-                handler.end(elem)
+
+def build_file_map(dir_name):
```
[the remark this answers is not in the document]
NDevox: This function doesn't survive the next PR

importer/taric.py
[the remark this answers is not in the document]
NDevox: Survey says you are correct.

This doesn't survive the next PR however so no need to change.

importer/handlers.py
```diff
@@ -388,8 +402,14 @@ def dispatch(self) -> TrackedModel:
 
         logger.debug(f"Creating {self.model}: {data}")
         data = self.pre_save(data, self.resolved_links)
-        obj = self.save(data)
-        self.post_save(obj)
+        try:
+            obj = self.save(data)
+            self.post_save(obj)
+        except AttributeError as e:
```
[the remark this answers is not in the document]
NDevox: I get the feeling this is left over debug. Not saying it isn't useful (and as the error is reraised it would include the original traceback). But I've removed it anyway as I don't think it ever gets raised with the current imports.

conftest.py
NDevox: Yeah it depends on how the factory is functioning. In some cases we've overridden the save functionality of the model, or of the factory. In this case it raises `DoesNotExist`. bit of a pain but they both validate the same thing really.

PR #126, opened 2020-12-29
TP-333: Allow fetching of "latest" versions of a relation in a single…

NDevox:
… query.

```python3
    # must be a TrackedModel
```

```python3
```

### DIFF
--- a/additional_codes/validators.py
+++ b/additional_codes/validators.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.exceptions import ValidationError
 from django.core.validators import RegexValidator
 from django.db.models import IntegerChoices
@@ -37,6 +38,11 @@ def validate_additional_code_type(obj):
     The referenced additional code type must exist and have as application code
     "non-Meursing" or "Export refund for processed agricultural good".
     """
+    try:
+        if obj.type is None:
+            return
+    except ObjectDoesNotExist as e:
+        raise ValidationError("The referenced additional code type must exist.") from e
 
     permitted_codes = [
         ApplicationCode.ADDITIONAL_CODES,
--- a/common/models/records.py
+++ b/common/models/records.py
@@ -1,15 +1,20 @@
-"""TrackedModel and supporting manager."""
+from __future__ import annotations
+
 import re
 from datetime import datetime
 from typing import Any
 from typing import Dict
 from typing import Iterable
+from typing import List
 from typing import Optional
+from typing import Tuple
+from typing import Type
 
 from django.contrib.postgres.fields import DateTimeRangeField
 from django.db import models
 from django.db.models import Case
 from django.db.models import F
+from django.db.models import Field
 from django.db.models import Q
 from django.db.models import QuerySet
 from django.db.models import Value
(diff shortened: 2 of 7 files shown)


additional_codes/validators.py
```diff
@@ -37,6 +38,11 @@ def validate_additional_code_type(obj):
     The referenced additional code type must exist and have as application code
     "non-Meursing" or "Export refund for processed agricultural good".
     """
+    try:
+        if obj.type is None:
+            return
+    except ObjectDoesNotExist as e:
+        raise ValidationError("The referenced additional code type must exist.") from e
```
stuaxo, line 45: This should be BusinessViolation ?

common/models/records.py
```diff
@@ -171,6 +176,38 @@ def annotate_record_codes(self) -> QuerySet:
             ),
         )
 
+    def _get_related_lookups(
+        self, model, *lookups, prefix="", recurse_level=0
+    ) -> List[str]:
```
stuaxo, line 181: Docstring ?

common/querysets.py
```diff
@@ -110,6 +113,38 @@ def with_workbasket(self, workbasket):
...
+                    self._get_related_lookups(
+                        model,
+                        *lookups,
+                        prefix=f"{prefix}{relation.name}__version_group__current_version__",
+                        recurse_level=recurse_level - 1,
+                    )
+                )
+        return related_lookups
+
+    def with_latest_links(self, *lookups, recurse_level=0) -> QuerySet:
```
stuaxo, line 140: Docstring ?

common/tests/test_models.py
```diff
@@ -219,3 +220,37 @@ def test_trackedmodel_can_attach_record_codes(workbasket):
         ),
         ordered=False,
     )
+
+
+def test_get_latest_relation_with_latest_links(
+    model1_with_history, django_assert_num_queries
+):
```
stuaxo, line 227: Can we have a docstring with a bit of an explanation ?

common/tests/test_models.py
```diff
@@ -219,3 +220,37 @@ def test_trackedmodel_can_attach_record_codes(workbasket):
...
+
+def test_get_latest_relation_without_latest_links(
+    model1_with_history, django_assert_num_queries
+):
+    oldest_link = model1_with_history.all_models[0]
+    latest_link = model1_with_history.all_models[-1]
+
+    factories.TestModel3Factory.create(linked_model=oldest_link)
+
+    with django_assert_num_queries(4):
```
stuaxo, line 250: Definitely need a comment to explain the 4 here.

common/tests/test_models.py
```diff
@@ -219,3 +220,37 @@ def test_trackedmodel_can_attach_record_codes(workbasket):
...
+        fetched_oldest_link = instance.linked_model
+        fetched_latest_link = instance.linked_model_current
+
+    assert oldest_link.pk == fetched_oldest_link.pk
+    assert latest_link.pk == fetched_latest_link.pk
+
+
+def test_get_latest_relation_without_latest_links(
+    model1_with_history, django_assert_num_queries
+):
```
stuaxo, line 244: Can we get a bit of a docstring here ?

stuaxo (approved):
Approved, but please add any docstrings / comments that come to mind.

NDevox:
Also see https://github.com/uktrade/tamato/pull/121 (same PR, some merge issues on GH)

additional_codes/validators.py
NDevox: Looking at it this shouldn't even be in the `validators` - `business_rules` covers it

PR #129, opened 2020-12-31
TP-451: Update business rules to be inline with TP-398

NDevox:
TP-371 included an overhaul of the business rules system to make it
more generic an inheritable. Whilst a sensible move, it also rendered
much of the work in TP-398 too difficult to merge. The end result was
a refactored and improved business rules system, with outdated business
rules.

This commit updates the business rules to be inline with TP-398 whilst
also conforming to the new business rules system. It also adds the
business rules to the relevant models so that they are used when
adding transactions.


commodities/business_rules.py
```diff
@@ -19,7 +20,7 @@ class NIG1(BusinessRule):
     def validate(self, good):
         if (
             type(good)
-            .objects.filter(
+            .objects.filter.current()(
```
stuaxo, line 23: This may-be probably fine, but looks a little strange, probably because I don't know how current works.

NDevox: Bit surprised this works... doesn't in my command line

NDevox: Fixed

common/business_rules.py
```diff
@@ -96,8 +98,7 @@ def validate(self, model):
         identifying_fields = self.identifying_fields or model.identifying_fields
         query = dict(get_field_tuple(model, field) for field in identifying_fields)
 
-        # TODO need to handle DELETE update type
-        if model.__class__.objects.filter(**query).count() > 1:
+        if model.__class__.objects.filter(**query).current().count() > 1:
```
stuaxo, line 101: Maybe a silly question, but isn't this the same thing as `.exists()` ?

NDevox: Technically it is different, but it is the same in what it achieves. Worth updating to regardless.

stuaxo (approved):
(Hit approve, but pls check comment)

common/business_rules.py
NDevox: Done

measures/models.py
```diff
@@ -179,7 +192,8 @@ class MeasureType(TrackedModel, ValidityMixin):
         choices=validators.OrderNumberCaptureCode.choices
     )
     measure_explosion_level = models.PositiveSmallIntegerField(
-        validators=[validators.validate_measure_explosion_level]
+        choices=[(2, 2), (4, 4), (6, 6), (8, 8), (10, 10)],
```
andyhd, line 195: Maybe move this to validators as an `models.IntegerChoices`

andyhd (approved):
This makes me happy 👍

measures/models.py
NDevox: Done

PR #135, opened 2021-01-06
Remove BtreeGistExtension migrations

andyhd:
1. These require superuser access to the database, which the staging deployment does not have.
2. This extension is already enabled in RDS Postgres.
3. For running locally, the extension should be installed manually with
`CREATE EXTENSION IF NOT EXISTS btree_gist;`


andyhd:
Not necessary

PR #137, opened 2021-01-07
Add linting steps

andyhd:


stuaxo (approved):

andyhd:
I'm not convinced it is a good idea to add this 😰

PR #140, opened 2021-01-12
TP-542: Update frontend views to be inline with current relations.

NDevox:
Historically descriptions didn't have a named `related_name`. This has
changed recently and so the frontend queries were prefetching the
wrong related name and erroring as a result.

This commit updates the queries to use the correct related_name for
descriptions - it also improves the caching setup for fetching the
latest descriptions. This may need to be made more generic and more
flexible in the near future.

### DIFF
--- a/geo_areas/models.py
+++ b/geo_areas/models.py
@@ -56,7 +56,13 @@ class GeographicalArea(TrackedModel, ValidityMixin):
     )
 
     def get_description(self):
-        return self.geographicalareadescription_set.last()
+        if (
+            hasattr(self, "_prefetched_objects_cache")
+            and "descriptions" in self._prefetched_objects_cache
+        ):
+            descriptions = list(self.descriptions.all())
+            return descriptions[-1] if descriptions else None
+        return self.descriptions.last()
 
     def get_descriptions(self, workbasket=None):
         return (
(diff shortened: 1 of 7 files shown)


geo_areas/models.py
```diff
@@ -56,7 +56,13 @@ class GeographicalArea(TrackedModel, ValidityMixin):
...
 
     def get_description(self):
-        return self.geographicalareadescription_set.last()
+        if (
+            hasattr(self, "_prefetched_objects_cache")
+            and "descriptions" in self._prefetched_objects_cache
+        ):
+            descriptions = list(self.descriptions.all())
+            return descriptions[-1] if descriptions else None
+        return self.descriptions.last()
```
andyhd, line 65: This looks like it could be pulled out into a mixin class maybe

DIT-CJWaszczuk (approved):

andyhd (approved):
Nice 👍

geo_areas/models.py
stuaxo: Could this `return self.descriptions.last() if self.descriptions else None` ?

NDevox: I entirely agree with this. In the near future. I think realistically we need a common queryset class for things with descriptions.

NDevox: `.last` is what we were doing before, sadly prefetch doesn't work well with anything other than `.all`, and so this forces a new set of queries for every object.

stuaxo (approved):

PR #142, opened 2021-01-13
TP-534: Integrate DIT staff SSO

andyhd:
It is a requirement that changes to the tariff can be attributed to the user who made
them, for auditing purposes. It may also be required later that different groups of
users have different permissions, for example only some users may be permitted to
approve workbaskets.

This commit integrates the tariff editor with the DIT single sign-on authentication
broker service (https://sso.trade.gov.uk). All pages in the tariff editor require an
authenticated user, so if not currently authenticated, the user is redirected on any
page request to the DIT SSO login screen. After successfully logging in, they are
redirected back to the tariff editor.

Credentials for the DIT SSO broker service are secret, so not included in this commit.
The credentials are passed into the tariff editor via environment variables. If you need
the credentials for running locally, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,9 @@ When running tests the settings module defaults to settings.test
 
 | Name | Description |
 | ---- | ----------- | 
+| AUTHBROKER_URL           | Base URL of the OAuth2 authentication broker (default https://sso.trade.gov.uk) |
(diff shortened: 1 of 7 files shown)


NDevox (approved):
LGTM

PR #144, opened 2021-01-13
Restore substantive parts of the TARIC3 XSD

simonwo:
The XSD we have been using as been modified at
some point in the past for unknown reasons and no
longer matches what HMRC use. This commit updates
the XSD from an earlier version on Confluence that
applies the correct checks.


common/assets/taric3.xsd
```diff
@@ -589,8 +589,8 @@
     </xs:simpleType>
     <xs:simpleType name="SID">
         <xs:restriction base="xs:int">
-            <xs:minInclusive value="-2147483648"/>
-            <xs:maxInclusive value="2147483647"/>
```
NDevox, line 593: This one is from https://github.com/uktrade/tamato/commit/e697c36142d2d7d0095444b5c4c3f7752c01f792#diff-e44f6f44a4c4d96eec98a83eef5f7e82fa883a5243ccf5f2505d5af0866cf60bR592

As I believe we wanted to support negative SIDs (https://uktrade.atlassian.net/browse/TP-328). Do we definitely want to change it back?

NDevox (changes requested):
1 Q

common/assets/taric3.xsd
stuaxo: Side note:   

If we do want to support negative numbers, we should have a comment to say why we're different.

Or better:  keep the original TARIC3 XSD immutable, and our changes as a separate XML, then generate our own version of the XSD during collectstatic (i.e. a very simple XSLT).

simonwo: Yeah, all good points, and I did think that's where we made the change.

So we made the original change thinking that we need to support national measures in TaMaTo. Turns out we don't! As we've migrated all the national measures as part of the Jan 1st process, and any that are left are solely the domain of HMRC. Further, HMRC are clear they don't want anything conforming to the national schema to be coming across their interface, and nethier do JE/GG. So we should (unfortunately) start unwinding that support, and this is one of the first places to do that.

TL;DR: yes, I think we do want to undo our previous change.

simonwo: Maybe we need to document that reverse of intention somewhere…

NDevox: Could probably doc it in the commit which reverses the support.

We should definitely have a ticket to reverse the support!

NDevox (approved):

PR #145, opened 2021-01-14
TP-567: Fix search on current frontend views.

NDevox:
The old views have used a mix of Rest Framework UI views and
standard Django Generic views. As a result the filtering setups
have become a bit confusing - they also simply don't work as
descriptions have been updated.

This commit fixes the reverse relations to descriptions and creates
a more generic filtering setup which serves both standard Django views
and Rest Framework UI views.


andyhd (commented):
wot, no tests?

DIT-CJWaszczuk (approved):

andyhd (approved):
Looks good to me

andyhd:
Oops, too slow!

PR #148, opened 2021-01-14
Allow disabling SSO

andyhd:
When running locally, during development, it may be useful to disable
SSO and avoid requiring an internet connection.

This change adds an environment variable `SSO_ENABLED` which switches
on the SSO authentication flow if truthy (eg "on", or "true"), and
disables it if falsy - falling back to Django's auth ModelBackend, which
authenticates users in the database.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ When running tests the settings module defaults to settings.test
 
 | Name | Description |
 | ---- | ----------- | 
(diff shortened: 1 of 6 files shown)


NDevox (approved):

simonwo (commented):
Do we need anything in the README?

stuaxo (approved):
+1 nice !

PR #149, opened 2021-01-14
Fix `dump_transactions` and accept an envelope ID.

simonwo:
Looks like this got broken at some point in the past. This now successfully outputs any workbasket currently in ready to export status, and accepts an envelope ID.

We'll be using this command for as long as we have a manual interface with HMRC.


stuaxo (approved):
Being able to pass envelope id is useful :)

PR #150, opened 2021-01-18
Handle successors deleted after their replacing code is unterminated

simonwo:
The EU TARIC we have recently processed contains a
number of instances of codes being end dated and
then subsequently updated to remove the end date.
In this case, the code is first updated and then
the successor removed. This breaks the import of
the successor because the importer expects an end
date to be on the replacing code in order to look
up the correct absorbing code, but it has just
been removed.

Instead, in this specific case we now look up the
absorbed link using the previous version of the
successor. As this is a deletion, the data should
in the XML should not have changed since the last
update so this is safe to do.

### DIFF
--- a/commodities/import_handlers.py
+++ b/commodities/import_handlers.py
@@ -109,16 +109,32 @@ class GoodsNomenclatureSuccessorHandler(BaseHandler):
         else:
             good = self.resolved_links["replaced_goods_nomenclature"]
 
-        must_be_active_on_date = good.valid_between.upper + timedelta(days=1)
-
-        return (
-            model.objects.filter(
-                valid_between__contains=must_be_active_on_date,
-                **kwargs,
+        # If this successor is being deleted, the replaced goods nomenclature
+        # will no longer have an end date, so we can't use that to look up the
+        # correct absorbing goods nomenclature. Instead, we retrieve that
+        # link from the previous version of the successor.
+        if (
+            good.valid_between.upper is None
+            and self.data["update_type"] == UpdateType.DELETE
+        ):
+            previous = (
+                models.GoodsNomenclatureSuccessor.objects.filter(
+                    **{key: self.data[key] for key in self.identifying_fields}
+                )
+                .current()
+                .get()
             )
-            .current()
-            .get()
-        )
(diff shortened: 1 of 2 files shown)


commodities/import_handlers.py
```diff
@@ -109,16 +109,25 @@ def get_absorbed_into_goods_nomenclature_link(self, model, kwargs):
...
+        else:
+            must_be_active_on_date = good.valid_between.upper + timedelta(days=1)
+            return (
+                model.objects.filter(
+                    valid_between__contains=must_be_active_on_date,
+                    **kwargs,
+                )
+                .current()
+                .get()
             )
```
NDevox, line 130: Only thought here is that, as you've returned in the `if` the `else` is redundant and you can dedent this - making it slightly more readable.

NDevox (commented):
Black failed too.

commodities/import_handlers.py
simonwo: Fixed the `black` issues. Personally I prefer having the `if`/`else` set out explicitly but I don't mind changing it if we think it's better style.

stuaxo: I used to always have an `else` to my `if` - BUT I got switched over by the standard on my last project.  It's nice having one less statement, and a bit less indentation.

NDevox: Yeah I understand the otherside of the argument as well. And for bits of code like this (small) it doesn't matter all that much. I've just found for bigger blocks it helps with the mental load. Without the else it's a bit easier to follow the logic as separate paths - as you can simply forget about the if block once it ends. You can do this with the else there as well of course, it's just a visual cue in my book that the blocks ended so you can forget about it.

simonwo: Ok… if there aren't any strong opinions about this, can we leave this as is and I'll make sure to bear it in mind for any major changes in the future?

andyhd: I think [Sourcery](https://sourcery.ai/) would pick this sort of thing up

andyhd (approved):

NDevox (approved):

PR #151, opened 2021-01-18
TP-568 Search Additional Codes

DIT-CJWaszczuk:
## Summary 
Implements the search features for additional codes, alongside making various code changes to make everything a little cleaner. 
## Screenshots
image

### DIFF
--- a/additional_codes/filters.py
+++ b/additional_codes/filters.py
@@ -1,14 +1,31 @@
 import re
+from datetime import date
+from datetime import datetime
+from datetime import timezone
 
+from crispy_forms_gds.choices import Choice
+from django import forms
 from django.contrib.postgres.aggregates import StringAgg
+from django.db.models import Q, DateTimeField
+from django.db.models.functions import Extract, Lower
+from django_filters import CharFilter
+from django_filters import MultipleChoiceFilter
+from django_filters import ChoiceFilter
+
 
 from additional_codes.models import AdditionalCode
+from additional_codes.forms import SearchFilterForm
+from additional_codes.validators import TypeChoices
 from common.filters import TamatoFilter
 from common.filters import TamatoFilterBackend
 from common.filters import TamatoFilterMixin
+from common.util import TaricDateTimeRange
+
+from typing import Callable
+
 
 COMBINED_ADDITIONAL_CODE_AND_TYPE_ID = re.compile(
-    r"^(?P<additional_code_type_id>[A-Z0-9])(?P<additional_code_id>[A-Z0-9]{3})$"
+    r"^(?P<type__sid>[A-Z0-9])(?P<code>[A-Z0-9]{3})$"
 )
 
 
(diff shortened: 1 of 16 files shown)


DIT-CJWaszczuk:
Wanted to get some early looks at list, There's quite a bit going on that me and Nick have changed, would be good to get some eyes on it.

stuaxo:
On first look, LGTM - obv will get to know some of this more as time goes on.

additional_codes/filters.py
```diff
@@ -35,7 +48,87 @@ class AdditionalCodeFilterBackend(TamatoFilterBackend, AdditionalCodeFilterMixin
...
+        required=False,
+    )
+
+    active_state = ChoiceFilter(
+        choices=ACTIVE_STATE_CHOICES,
+        empty_label="Both",
+        widget=forms.RadioSelect,
+        method="filter_active_state",
+        label="Active state",
+        help_text="Select all that apply",
```
NDevox, line 90: This is single choice now so the label is misleading.

additional_codes/filters.py
```diff
@@ -1,14 +1,27 @@
...
+from django_filters import ChoiceFilter
 
 from additional_codes.models import AdditionalCode
+from additional_codes.validators import TypeChoices
 from common.filters import TamatoFilter
 from common.filters import TamatoFilterBackend
 from common.filters import TamatoFilterMixin
+from common.util import TaricDateTimeRange
+
+from .forms import SearchFilterForm
```
NDevox, line 21: Would probably be better to do an explicit import (as done above, `from additional_codes.forms import SearchFilterForm`. It will have to go between `AdditionalCode` and `TypeChoices`

additional_codes/jinja2/additional_codes/list.jinja
```diff
@@ -72,12 +64,12 @@
                 {% endif %}
               </li>
             {% endfor %}
-            {% if object_list.has_next() %}
+            {% if page_obj.has_next() %}
               <li>
                 <a 
-                  href="?page={{ object_list.next_page_number() }}" 
+                  href="?page={{ page_obj.next_page_number() }}"
```
NDevox, line 70: Loses the filters when going to next page

additional_codes/jinja2/additional_codes/list.jinja
```diff
@@ -23,47 +23,39 @@
...
+            Page {{page_obj.number}} of {{ paginator.num_pages}}
           </div>
           <ul class="govuk-list pagination__links"> 
-            {% if object_list.has_previous() %}
+            {% if page_obj.has_previous() %}
               <li>
                 <a 
                   class="govuk-link govuk-!-margin-right-1" 
-                  href="?page={{ object_list.previous_page_number() }}" 
+                  href="?page={{ page_obj.previous_page_number() }}"
```
NDevox, line 49: Loses the filters when going to previous page

NDevox (changes requested):
Couple bits and pieces

additional_codes/filters.py
```diff
@@ -35,7 +48,87 @@ class AdditionalCodeFilterBackend(TamatoFilterBackend, AdditionalCodeFilterMixin
...
+    )
+
+    def filter_start_year(self, queryset, name, value):
+        """
+        Provides a query for filtering on possible start years.
+
+        DateTimeRanges sadly don't have a direct method for checking the start year
+        (valid_between__lower__year= would have made sense but doesn't exist).
+        Instead this creates a query to make sure the range contains Dec 31st of the
+        given year, but not of the previous year.
```
andyhd, line 101: Postgres does provide the [datetime EXTRACT function](https://www.postgresql.org/docs/9.1/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT), which you could potentially use instead?

additional_codes/jinja2/additional_codes/list.jinja
NDevox: You should be able to get these from the request object (`request.GET.search` and so on.)

additional_codes/urls.py
```diff
@@ -1,20 +1,30 @@
 from django.urls import include
 from django.urls import path
+from django.urls import re_path
 from rest_framework import routers
 
 from additional_codes import views
-
+from additional_codes.filters import COMBINED_ADDITIONAL_CODE_AND_TYPE_ID
```
andyhd, line 7: Imported but not used?

common/static/common/scss/_layout.scss
```diff
@@ -4,6 +4,7 @@
 
 .filter-layout__filters {
   @extend .govuk-grid-column-one-quarter;
+  background: #f3f2f1;
```
andyhd, line 7: Should this be `govuk-colour("light-grey")`?

andyhd (approved):
Looks good to me, only very minor comments

additional_codes/filters.py
```diff
@@ -35,7 +48,89 @@ class AdditionalCodeFilterBackend(TamatoFilterBackend, AdditionalCodeFilterMixin
     pass
 
 
+START_YEAR_CHOICES = tuple(  # Filter on the last 10 years
+    Choice(str(year), str(year))
+    for year in range(date.today().year, date.today().year - 10, -1)
```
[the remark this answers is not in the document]
NDevox: Good point on the constants, I prefer this.

additional_codes/filters.py
NDevox: This is far better.

Equivalent is:

```python3
if value:
    queryset = queryset.objects.annotate(start_year=Extract(Lower("valid_between", output_field=DateTimeField()), "year")).filter(start_year__in=value)
return queryset
```

This replaces the whole function.

andyhd: Ah thanks, I was too lazy to look up how to actually use the Extract function in Django 😴

additional_codes/forms.py
```diff
@@ -0,0 +1,31 @@
...
+from django.urls import reverse
+
+from crispy_forms_gds.helper import FormHelper
+from crispy_forms_gds.layout import Field, Layout, Size, Submit, Button, HTML
+from crispy_forms_gds.templatetags.crispy_forms_gds import button_link
+
+
+class SearchFilterForm(forms.Form):
+    def __init__(self, *args, **kwargs):
+        super(SearchFilterForm, self).__init__(*args, **kwargs)
```
andyhd, line 11: You can omit the arguments to `super()` in Python 3

additional_codes/jinja2/additional_codes/list.jinja
andyhd: It may be useful to add a custom filter or global function to the Jinja environment to do allow updating the current URL with the previous or next page number. [This example is of a Django template tag](https://stackoverflow.com/a/16609498), but could be easily adapted.

additional_codes/filters.py
```diff
@@ -31,11 +46,112 @@ class AdditionalCodeFilterMixin(TamatoFilterMixin):
...
+    additional_code_type = LazyMultipleChoiceFilter(
+        choices=TypeChoices.choices,
+        widget=forms.CheckboxSelectMultiple,
+        field_name="type__sid",
+        label="Additional Code Type",
+        help_text="Select all that apply",
+        required=False,
+    )
+    start_year = LazyMultipleChoiceFilter(
+        choices=last_10_years(),
```
NDevox, line 99: You want to pass this in without the parens, i.e. `choice=last_10_years`. So we pass the function instead of the result of the function.

additional_codes/filters.py
NDevox: @CJWaszczuk shall we implement this one?

additional_codes/jinja2/additional_codes/list.jinja
```diff
@@ -5,7 +5,7 @@
 {% from "components/table/macro.njk" import govukTable %}
 
 {% set page_title = "Find and edit additional codes" %}
-
+{{update_params(request.args)}}
```
NDevox, line 8: What does this do?

NDevox (approved):
LGTM

PR #153, opened 2021-01-20
TP-526 Add validation for non-printable whitespace character in legislation text

simonwo:
We have previously extended the system to include
the public ID and URL in the `information_text`
field for the Trade Tariff Serivce in 97ea76a6.

It turns out that the HMRC system filters this out
and replaces it with some whitespace character:
`0xA0` in UTF-8.

We should continue to output pipes and add a
validation that this character isn't present
anywhere else in the final text string, so that
the Trade Tariff Service can split on that
character instead.


andyhd (approved):
Looks good to me

NDevox (approved):

stuaxo (approved):
LGTM :)

PR #154, opened 2021-01-20
TP-517 Refactor details tab

DIT-CJWaszczuk:
## Summary 
Refactor the details page and tab to fit with the current designs 
## Changes
- Move the tabs into their own files, takes some of the pollution out of the main details template. 
- Styling and content changes to the core data tab
- Added a new breadcrumb block, so the alpha banner doesn't get overwritten 
- Removed sidebar until we have actions, left the container at 3/4 width however.
## Screenshots 
image


andyhd (approved):
Nice

NDevox (approved):

PR #155, opened 2021-01-20
TP-502: And a frontend for the importer and make the templates more generic

NDevox:
It is required that a user can login to the tool and upload a TARIC XML
file to be loaded. This commit introduces a frontend at `/importer` which
allows this functionality.

To facilitate these views the templates for additional codes have been made
more generic so that they can be used across the app. Likewise the filtering
capabilities have been updated to allow handling of GDS crispy forms formatting.

## List View:

Screenshot 2021-01-20 at 13 59 00

## Create view:

Screenshot 2021-01-20 at 13 59 31


additional_codes/filters.py
```diff
@@ -115,6 +112,8 @@ class AdditionalCodeFilter(TamatoFilter, AdditionalCodeFilterMixin):
         required=False,
     )
 
+    clear_url = reverse_lazy("additional_code-ui-list")
```
andyhd, line 115: When the user clicks the "Clear" button, they are taken back to the list view? That doesn't feel right to me - maybe the button should be labelled "Cancel" or it should clear the form on click instead. (I realize this hasn't changed in this PR).

common/jinja2/layouts/list.jinja
```diff
@@ -0,0 +1,99 @@
...
+      {"text": page_title}
+    ]
+  }) }}
+{% endblock %}
+
+{% block content %}
+  <h1 class="govuk-heading-xl">{{ page_title }}</h1>
+  <p class="govuk-body">
+    Enter criteria to help find an {{ object_type }}.
+    Alternatively, <a href="{{ create_url }}">create a new {{ object_type }}</a>.
```
DIT-CJWaszczuk, line 25: Do we want to just hide this if there is no create functionality for a given type?

common/jinja2/layouts/list.jinja
```diff
@@ -0,0 +1,99 @@
...
+{% set page_title = "Find and edit " + (object_type_plural or object_type + "s") %}
+
+{% block beforeContent %}
+  {{ govukBreadcrumbs({
+    "items": [
+      {"text": "Home", "href": url("index")},
+      {"text": page_title}
+    ]
+  }) }}
+{% endblock %}
```
DIT-CJWaszczuk, line 19: This might be out of date with master, changed to a breadcrumb block

common/jinja2/layouts/list.jinja
```diff
@@ -0,0 +1,99 @@
+{% extends "layouts/layout.jinja" %}
+
+{% from "components/button/macro.njk" import govukButton %}
+{% from "components/input/macro.njk" import govukInput %}
+{% from "components/table/macro.njk" import govukTable %}
+
+{% set object_type_plural = object_type_plural or object_type + "s" %}
+{% set create_url = create_url or "create" %}
+
+{% set page_title = "Find and edit " + (object_type_plural or object_type + "s") %}
```
DIT-CJWaszczuk, line 10: Could this not just be:

`{% set page_title = "Find and edit " + object_type_plural %}`

common/jinja2/layouts/list.jinja
NDevox: Maybe? I can't think of any types we wouldn't want this for at the moment. So I'd argue YAGNI for now.

common/jinja2/layouts/list.jinja
NDevox: Can't see this in my master, have you got a link?

common/jinja2/layouts/list.jinja
NDevox: This allows us to set a reasonable default - and so means we don't _have_ to set `object_type_plural` in other templates. I've updated the additional codes list template to reflect this if you want to see.

common/jinja2/layouts/list.jinja
andyhd: Yes, in fact, we will want to wrap a lot of these with something like 
```jinja2
{% if request.user.has_perm("app_name.add_object_type") %}
Alternatively, <a href="{{ create_url }}">create a new {{ object_type }}</a>.
{% endif %}

simonwo:
Looks cool. I assume we're just targeting data/dev users with this right now, rather than tariff managers (else we might want to visit the content). 

Also, I can see the value in letting people choose a status for now… but imagine we might want to lock that down sometime so we can't just skip through approval steps.

common/jinja2/layouts/list.jinja
NDevox: Can probably do something generic on that such as `.has_perm(content_type)` or something? I can see us adding stuff in the view for context.

common/jinja2/layouts/list.jinja
DIT-CJWaszczuk: Apologies, it was the wrong file, you can see it here https://github.com/uktrade/tamato/blob/c38ac3f0ab922c094132a05c18881ff3f710e8f5/additional_codes/jinja2/additional_codes/detail.jinja

NDevox:
@simonwo yep agreed. Needing a bit more direction on that but happy to punt off to the future. Also, whilst the page is there to see, there is no link on it from any menu - you need to know in advance. Not really security in anyway but the point being we don't need to publicise it at all for now.

common/jinja2/layouts/list.jinja
DIT-CJWaszczuk: Makes sense!

common/jinja2/layouts/list.jinja
NDevox: Ah ok. Makes sense. We should probably move them all over in another commit as we standardise the templates a bit more. I can see a few things like this happening pretty soon!

simonwo:
Hm, is there any SSO control around it? If this is the first time we'll have a page that will allow people to edit the data stuff I'd prefer to put at least some kind of access control. Does the django admin method already handle that?

NDevox:
@simonwo good point. I'd assumed SSO was on for all pages but that isn't the case.

I've added a new mixin which requires users to be superusers on the import views - means not just anyone with an account can go to them.

DIT-CJWaszczuk (approved):

PR #157, opened 2021-01-21
TP-568: Small patch to fix search functionality.

NDevox:
Currently if an regexable piece of data exists within a search
query the rest of the search query will be removed in favour of it
which is not desired behaviour. It also expected anything regexable
to be at the front of the query.

This commit fixes this behaviour to prepend the query with the regexed
data and provide the rest of the search query as well. re.search has also
replaced re.match which enables the term to be found anywhere in the query.

In addition to this the Additional Codes regex expected the search term
to _only_ be the ID, which is wrong. This has been updated.

### DIFF
--- a/additional_codes/filters.py
+++ b/additional_codes/filters.py
@@ -22,7 +22,7 @@ from common.util import TaricDateTimeRange
 
 
 COMBINED_ADDITIONAL_CODE_AND_TYPE_ID = re.compile(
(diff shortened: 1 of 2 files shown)


DIT-CJWaszczuk (approved):

stuaxo (approved):
Nice :)

PR #159, opened 2021-01-22
TP-544: add more tests + minor refactor.

NDevox:
This is essentially a never ending ticket, but here are some extra tests.

Also moved the workbasket manager to be from queryset - should be a bit more robust. Removed an extra save method on trackedmodel. And removed some unused methods.


common/models/records.py
```diff
@@ -401,21 +401,15 @@ def get_versions(self):
         query = Q(**self.get_identifying_fields())
         return self.__class__.objects.filter(query)
 
-    def get_latest_version(self):
```
NDevox, line 404: This did the exact same thing as `current_version`.

common/models/records.py
```diff
@@ -457,26 +451,6 @@ def get_identifying_fields(
 
         return fields
 
-    @atomic
-    def save(self, *args, force_write=False, **kwargs):
```
NDevox, line 461: Almost the exact same as the other `save` method below.

workbaskets/tests/test_models.py
```diff
@@ -26,3 +43,207 @@ def test_workbasket_transactions():
...
+    ],
+)
+def test_workbasket_submit(new_workbasket, start_status, fails):
+    new_workbasket.status = start_status
+    try:
+        new_workbasket.submit_for_approval()
+    except TransitionNotAllowed:
+        if fails:
+            return
+        raise
```
stuaxo, line 72: The `return` looks a bit weird here ?

NDevox: In this case we were expecting the error so can exit the test here - we've verified what we wanted to.

andyhd: Would it be clearer as 
```python
except TransitionNotAllowed:
    if not fails:
        raise
else:
    assert new_workbasket.status == ...
```
maybe `fails` should be something like `should_fail` or (inverted) `transition_allowed`?

workbaskets/tests/test_models.py
```diff
@@ -26,3 +43,207 @@ def test_workbasket_transactions():
...
+    assert new_workbasket.status == models.WorkflowStatus.CDS_ERROR
+
+
+def test_get_tracked_models(new_workbasket):
+    for _ in range(10):
+        factories.FootnoteFactory.create()
+
+    assert TrackedModel.objects.count() > 10
+    assert new_workbasket.tracked_models.count() == 10
```
stuaxo, line 249: This is fine here (though obv in other places checking count can be dubious), it could be better to be pedantic enough to check the ids are the same.

I'm assuming 10 is arbitrary, it's probably more than is needed.  2 is always good when checking something that it's a collection, since you at least verify it can hold multiple things.

workbaskets/tests/test_models.py
```diff
@@ -26,3 +43,207 @@ def test_workbasket_transactions():
     assert assoc.transaction == tx2
     assert assoc.associated_footnote.transaction == tx2
     assert workbasket.transactions.count() == 2
+
+
+@pytest.mark.parametrize(
+    "start_status,fails",
```
andyhd, line 49: I think you can collapse all these tests into a single parametrized `test_workbasket_transition` test. If you have `start_status, target_status, transition_allowed` as parameters, you can do something like
```python
from contextlib import nullcontext as does_not_raise
@pytest.mark.parametrize(
    "start_status, transition, target_status, expect_error",
    [
        (NEW_IN_PROGRESS, "submit_for_approval", AWAITING_APPROVAL, does_not_raise()),
        (PUBLISHED, "submit_for_approval", AWAITING_APPROVAL, pytest.raises(TransitionNotAllowed)),
        ...
    ]
)
def test_workbasket_transition(new_workbasket, start_status, transition, target_status, expect_error):
    new_workbasket.status = start_status
    transition_method = getattr(new_workbasket, transition)
    with expect_error:
        transition_method()
        assert new_workbasket.status == target_status
```
and maybe you can stack parametrize decorators to get a test matrix, eg
```python
@pytest.mark.parametrize("start_status", WorkflowStatus.values)
@pytest.mark.parametrize("transition, target_status", [
    ("submit_for_approval", WorkflowStatus.AWAITING_APPROVAL),
    ("publish", WorkflowStatus.PUBLISHED),
    ...
])
def test_workbasket_transitions(new_workbasket, start_status, transition, target_status):
    ...
```

stuaxo: Should the assertion be outside of the `with` in this example ?    Otherwise it will never be called for the tests that cause the exception to be raised.

workbaskets/tests/test_models.py
stuaxo: Your suggestion of does_not_raise() and pytest.raises() in the parameters in the [following comment](https://github.com/uktrade/tamato/pull/159#discussion_r566303232) is probably the best solution.

workbaskets/tests/test_models.py
NDevox: In the case where we call the assertion outside of the with block - if the transition is expected to error then the assertion should fail - so I don't think this is desirable sadly.

The test matrix idea is nice as well - if we knew which statuses would cause a failure ahead of time. I've gone with a massive parametrize block instead!

workbaskets/tests/test_models.py
NDevox: I've switched these over to 2 now

workbaskets/tests/test_models.py
NDevox: Followed the below comment

workbaskets/tests/test_models.py
NDevox: Done as described

andyhd: I did realize that the test matrix would mean you can't pass in the expected results. But I think you can work it out with `django_fsm.can_proceed(workbasket_transition_method)`. But if you do that, are you just testing that `django_fsm` works?

NDevox: Yeah exactly, I don't really want to test `django_fsm` I want to check that `status_x` can('t) transition to `status_y`.

andyhd (approved):

PR #162, opened 2021-01-26
Fix regulation attributes and measure tag

kristof-bb:
- Wrong xml tag used for additional code in measure xml. Unfortunately it seems the TARIC import didn't import any additional codes on measures due to this bug.
- Fix some attributes in the regulation serialiser


simonwo (approved):
LGTM. I wonder how we ever got anything done with these issues?

NDevox:
That's painful. Yes this will require a re-import most likely. Surprised this got past the xsd.

PR #164, opened 2021-01-27
Uplift measure creation pattern for production use

simonwo:
This pattern was produced during the Transition Period data migration as a simple way to create a new measure along with all of the associated models from a single function call.

This commit uplifts this to use the proper TaMaTo database, save the models and remove some complex logic around dates that is no longer needed.

This pattern is still useful for migrations and may also be useful as a way of creating measures from the UI.

## Review questions worth answering:

- Do we think this pattern is too big? Should it be split up? Can this be done whilst still giving an easy interface for migrations?
- Do we think this sort of object is useful for the UI? The point of this is to encode the domain logic around certain tasks that transcend beyond the business rules. I.e. to give the system a simpler way to produce data that doesn't break them.


andyhd:
You need to install https://pre-commit.com/

stuaxo:
Where in the code is this ?

common/util.py
```diff
@@ -11,6 +13,36 @@ def is_truthy(value: str) -> bool:
...
+
+
+def maybe_min(*objs: Optional[TypeVar("T")]) -> Optional[TypeVar("T")]:
+    """Return the lowest out of the passed objects that are not None,
+    or return None if all of the passed objects are None."""
+    present = [d for d in objs if d is not None]
+    if any(present):
+        return min(present)
+    else:
+        return None
```
andyhd, line 33: No big deal, but you're looping through the `objs` three times (`for`, `any` and `min`), and it's considered more Pythonic to check for exceptions, eg:
```python
try:
    return min(d for d in objs if d)
except ValueError:
    return None
```
Also, since `*objs` is a argument list, it is not `Optional` and will always be a tuple (or list?) with zero or more elements.

commodities/util.py
```diff
@@ -0,0 +1,30 @@
...
+class InvalidItemId(Exception):
+    pass
+
+
+def clean_item_id(value: Union[str, int, float]) -> str:
+    """Given a value, return a string representing the 10-digit item id
+    of a goods nomenclature item taking into account that the value may
+    have a leading zero or trailing zeroes missing."""
+    item_id = strint(value)
+    if type(value) in (int, float) and len(item_id) % 2 == 1:
```
andyhd, line 15: Is the type check for `value` necessary?

common/util.py
```diff
@@ -11,6 +13,36 @@ def is_truthy(value: str) -> bool:
...
 
 
+def strint(value: Union[int, str, float]) -> str:
+    """If the passed value is a number type, return the
+    number as a string with no deciaml point or places.
+    Else just return the string."""
+    if type(value) in (int, float):
+        return str(int(value))
+    else:
+        return str(value)
```
andyhd, line 23: This could be
```python
if isinstance(value, float):
    value = int(value)
return str(value)
```

commodities/util.py
```diff
@@ -0,0 +1,30 @@
...
+        item_id = "0" + item_id
+
+    item_id = item_id.replace(" ", "").replace(".", "")
+
+    # We need a full 10 digit code so padd with trailing zeroes
+    if len(item_id) % 2 != 0:
+        raise InvalidItemId(f"Item ID {item_id} contains an odd number of characters")
+
+    item_id += "0" * (10 - len(item_id))
+    if len(item_id) != 10:
```
andyhd, line 27: This check can never fail. Also, you can zero pad with `item_id = f"{item_id:010}"`.

andyhd (commented):
Still reviewing, sorry. Some minor comments so far:

measures/util.py
```diff
@@ -0,0 +1,16 @@
...
+
+def clean_duty_sentence(value: Union[str, int, float]) -> str:
+    """Given a value, return a string representing a duty sentence
+    taking into account that the value may be storing simple percentages
+    as a number value."""
+    if type(value) in (float,):
+        # This is a percentage value that Excel has
+        # represented as a number.
+        return "{:.3%}".format(value)
+    elif type(value) in (int,):
```
andyhd, line 12: `elif` can just be `if` here, because of the `return` above

measures/util.py
```diff
@@ -0,0 +1,16 @@
...
+        return "{:d}%".format(value)
+    else:
```
andyhd, line 14: `else` is unnecessary here

andyhd (approved):
Just minor stylistic quibbles, but nothing that needs to fixing as far as I can see.

commodities/util.py
simonwo: It's trying to differentiate between when a leading zero has been lost (which _only_ happens when the value comes in as a number) and when a commodity code is just malformed. So it's just trying to be as specific as possible about when to try and automatically error-correct.

andyhd: I only ask because `strint` does the same check, so it seemed redundant. But I get it.

PR #169, opened 2021-01-28
TP-580 Refactor footnote list page

DIT-CJWaszczuk:
## Summary 
Refactored the list page to look more like the designs, also moved a few common search fields into common, rather than have repetitive code. 
## Changes
- General styling tweaks
- Moved start year and active year into the common filters
- Added JS button for showing more than 10 checkboxes, similar to the implementation at https://data.trade.gov.uk/
## Screenshots
**_(The grey bit of box at the bottom left is the screen capture extension not working properly, not an actual bug)_**

image

### DIFF
--- a/common/filters.py
+++ b/common/filters.py
@@ -1,7 +1,12 @@
 import re
 from collections import OrderedDict
+from datetime import date
+from datetime import datetime
+from functools import cached_property
+from typing import Callable
 from typing import Optional
 
+from crispy_forms_gds.choices import Choice
 from crispy_forms_gds.helper import FormHelper
 from crispy_forms_gds.layout import Button
 from crispy_forms_gds.layout import Field
@@ -10,11 +15,18 @@ from crispy_forms_gds.layout import Layout
 from crispy_forms_gds.layout import Size
 from django import forms
 from django.contrib.postgres.search import SearchVector
+from django.db.models import DateTimeField
+from django.db.models import Q
 from django.urls import reverse
 from django_filters import CharFilter
 from django_filters import FilterSet
+from django_filters import MultipleChoiceFilter
 from rest_framework import filters
 from rest_framework.settings import api_settings
+
+from common.util import TaricDateTimeRange
+
+ACTIVE_STATE_CHOICES = [Choice("active", "Active"), Choice("terminated", "Terminated")]
 
 
 def field_to_layout(field_name, field):
@@ -29,6 +41,28 @@ def field_to_layout(field_name, field):
         return Field.checkboxes(field_name, legend_size=Size.SMALL)
 
     return field_name
+
+
+def last_10_years():
+    current_year = date.today().year
+    return [
+        Choice(str(year), str(year))
+        for year in range(current_year, current_year - 10, -1)
+    ]
+
+
+class LazyMultipleChoiceFilter(MultipleChoiceFilter):
+    def get_field_choices(self):
+        choices = self.extra.get("choices", [])
+        if isinstance(choices, Callable):
+            choices = choices()
+        return choices
+
+    @cached_property
+    def field(self):
+        field_kwargs = {**self.extra.copy(), "choices": self.get_field_choices()}
+        field_kwargs.update(choices=self.get_field_choices())
+        return self.field_class(label=self.label, **field_kwargs)
 
 
 class TamatoFilterForm(forms.Form):
@@ -108,6 +142,7 @@ class TamatoFilter(FilterSet, TamatoFilterMixin):
     """
 
     search = CharFilter(method="filter_search", label="Search")
+
     clear_url = None
 
     def filter_search(self, queryset, name, value):
--- a/common/static/common/js/showHideCheckboxes.js
+++ b/common/static/common/js/showHideCheckboxes.js
@@ -1,0 +1,74 @@
+let showMoreClicked = {};
+
+const nodeListForEach = (nodes, callback) => {
+    if (window.NodeList.prototype.forEach) {
+      return nodes.forEach(callback)
+    }
+    for (let i = 0; i < nodes.length; i++) {
+      callback.call(window, nodes[i], i, nodes)
+    }
+}
+
+class FilterShowMore {
+    constructor(module) {
+        this.module = module;
+        this.button = this.module.getElementsByClassName('js-filter-show-more')[0];
+    }
+    init() {
+        if (!this.module || !this.button) {
+            return
+        }
+    
+        if (this.hasBeenRevealed()) {
+            this.revealChoices();
+        } else {
+            this.button.classList.remove('js-hidden');
+            this.button.addEventListener('click', () => this.revealChoices());
+        }
+    }
+    hasBeenRevealed() {
+        return showMoreClicked[this.module.getAttribute('data-show-more-id')] !== undefined;
+    }
+    revealChoices() {
+        this.button.classList.add('js-hidden');
+
+        let choices = Array.from(this.module.getElementsByClassName('js-hidden'));
+        choices.forEach((choice) => {
+            choice.style.display = 'block';
+        });
+        showMoreClicked[this.module.getAttribute('data-show-more-id')] = true;
+    }
+}
+
+const createShowButton = () => {
+    const button = document.createElement('button');
+    button.classList.add('govuk-button', 'govuk-button--secondary', 'js-filter-show-more');
+    button.innerText = 'Show more';
+    button.type = 'button'
+    return button;
+}
+
+const setupCheckBoxes = () => {
+    const checkboxSets = Array.from(document.getElementsByClassName('govuk-checkboxes'));
+    checkboxSets.forEach((checkboxSet) => {
+        if (checkboxSet.children.length > 10) {
+            Array.from(checkboxSet.children).forEach((checkbox, index) => {
+                if (index > 9) {
+                    checkbox.classList.add('js-hidden')
+                }
+            })
+            checkboxSet.dataset.module="filter-show-more";
+            checkboxSet.append(createShowButton());
+        }
+    })
+}
+
+const installFilterShowMore = () => {
+    setupCheckBoxes();
+    let modules = document.querySelectorAll('[data-module="filter-show-more"]');
+    nodeListForEach(modules, (module) => {
+        new FilterShowMore(module).init()
+    });
+}
+
+export default installFilterShowMore;
--- a/footnotes/filters.py
+++ b/footnotes/filters.py
@@ -1,10 +1,23 @@
 import re
+from datetime import datetime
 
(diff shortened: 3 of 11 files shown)


andyhd:
If you install https://pre-commit.com/, it will run black on your files and prevent you committing files that need reformatting.

DIT-CJWaszczuk:
Will sort it now, keep meaning to and then forgetting

common/static/common/js/showHideCheckboxes.js
```diff
@@ -0,0 +1,79 @@
...
+
+    if (this.hasBeenRevealed()) {
+        this.revealChoices();
+    } else {
+        let that = this;
+
+        this.button.classList.remove('hidden');
+        this.button.addEventListener('click', function () {
+        that.revealChoices();
+        });
```
andyhd, line 30: I feel like this would be better as an arrow function
```javascript
this.button.addEventListener('click', () => { this.revealChoices(); });
```
although could this be even more concise as
```javascript
this.button.addEventListener('click', this.revealChoices);
```

common/static/common/js/showHideCheckboxes.js
```diff
@@ -0,0 +1,79 @@
...
+      callback.call(window, nodes[i], i, nodes)
+    }
+  }
+  
+let showMoreClicked = {};
+
+function FilterShowMore(module) {
+    this.module = module;
+    this.button = this.module.getElementsByClassName('js-filter-show-more')[0];
+}
```
andyhd, line 15: Is there a reason to not to use `class` style here?
```javascript
class FilterShowMore {
    constructor(module) {
        this.module = module;
        ...
    }
    init() {
        if (!this.module || !this.button) {
    ...
}
```
Even if we need to support older browsers or Internet Explorer, we can transpile.

common/static/common/js/showHideCheckboxes.js
```diff
@@ -0,0 +1,79 @@
...
+    return showMoreClicked[this.module.getAttribute('data-show-more-id')] !== undefined;
+}
+
+FilterShowMore.prototype.revealChoices = function (e) {
+    this.button.classList.add('hidden');
+
+    let choices = this.module.getElementsByClassName('app-js-hidden');
+    for (let i = 0; i < choices.length; i++) {
+        choices[i].style.display = 'block';
+    }
```
andyhd, line 44: This could be
```javascript
let choices = Array.from(this.module.getElementsByClassName('app-js-hidden'));
choices.forEach((choice) => {
    choice.style.display = 'block';
});
```

footnotes/validators.py
```diff
@@ -19,6 +20,33 @@ class ApplicationCode(models.IntegerChoices):
...
+    COMBINED_NOMENCLATURE = "NC", "NC - Combined Nomenclature"
+    CN_MEASURE = "NM", "NM - CN measure"
+    EXPORT_REFUND_NOMENCLATURE = "NX", "NX - Export Refund Nomenclature"
+    OZONE_DEPLETING_SUBSTANCES = "OZ", "OZ - Ozone-depleting substances"
+    PUBLICATION = "PB", "PB - Publication"
+    SEE_ANNEX = "PN", "PN - See annex"
+    TARIC_MEASURE = "TM", "TM - Taric Measure"
+    TARIC_NOMENCLATURE = "TN", "TN - Taric Nomenclature"
+    DYNAMIC_FOOTNOTE = "TP", "TP - Dynamic footnote"
+    TORTURE_REPRESSION = "TR", "TR - Torture and repression"
```
andyhd, line 47: These don't seem correct. For a start, these values don't all pass the regex validator for footnote_type_id (on line 50 below). Also if/when they are correct, it would be good to change the model field to a ChoiceField using these values.

andyhd (commented):
Odd variety of Javascript styles in one module, but no real problems. I would approve, but I'm not sure the list of footnote type IDs is correct.

common/filters.py
```diff
@@ -31,6 +41,32 @@ def field_to_layout(field_name, field):
...
+
+class LazyMultipleChoiceFilter(MultipleChoiceFilter):
+    def get_field_choices(self):
+        choices = self.extra.get("choices", [])
+        if isinstance(choices, Callable):
+            choices = choices()
+        return choices
+
+    @property
+    def field(self):
```
stuaxo, line 60: It looks like you are reimplementing @cached_property.

```python
    @cached_property
    def field(self):
        field_kwargs = {**self.extra.copy(), choices=self.get_field_choices()}
        field_kwargs.update(choices=self.get_field_choices())
        self.field_class(label=self.label, **field_kwargs)
```

I haven't looked at too long, but you may not even need field_kwargs as it's own variable.

common/static/common/js/showHideCheckboxes.js
```diff
@@ -0,0 +1,76 @@
...
+            return
+        }
+    
+        if (this.hasBeenRevealed()) {
+            this.revealChoices();
+        } else {
+            let that = this;
+    
+            this.button.classList.remove('hidden');
+            this.button.addEventListener('click', () => that.revealChoices());
```
andyhd, line 28: Since this is now an arrow function, it isn't bound to its own `this`, so you don't need to `let that = this;` and you can do `() => this.revealChoices()` (and can that be shortened to `this.revealChoices`?)

footnotes/validators.py
```diff
@@ -4,7 +4,7 @@
 from django.core.exceptions import ValidationError
 from django.core.validators import RegexValidator
 from django.db import models
-
+from django.db.models import TextChoices
```
andyhd, line 7: This isn't used any more - were the choices incorrect?

common/static/common/js/showHideCheckboxes.js
```diff
@@ -0,0 +1,76 @@
...
+    return button;
+}
+
+const setupCheckBoxes = () => {
+    const checkboxSets = Array.from(document.getElementsByClassName('govuk-checkboxes'));
+    checkboxSets.forEach((checkboxSet) => {
+        if (checkboxSet.children.length > 10) {
+            Array.from(checkboxSet.children).forEach((checkbox, index) => {
+                if (index > 9) {
+                    checkbox.classList.add('app-js-hidden')
```
andyhd, line 59: We have both `app-js-hidden` and `hidden` - can we do with just one?

andyhd (approved):
Looks good to me.

footnotes/validators.py
DIT-CJWaszczuk: We chose to get them straight from the db to avoid choices with no results.

andyhd: Much better!

DIT-CJWaszczuk: Also, does black not catch unused imports?

common/static/common/js/showHideCheckboxes.js
DIT-CJWaszczuk: Switching all to just 'js-hidden', feel it's the most appropriate name.

DIT-CJWaszczuk:
Made all the  required changes, also tampered with  the table columns widths in the PR to avoid future asks. Mind taking a final check? @andyhd @stuaxo

footnotes/filters.py
```diff
@@ -38,7 +52,72 @@ class FootnoteFilterBackend(TamatoFilterBackend, FootnoteFilterMixin):
...
+            ),
+        )
+        for footnote_type in footnote_types
+    ]
+
+
 class FootnoteFilter(TamatoFilter, FootnoteFilterMixin):
+
+    footnote_type = LazyMultipleChoiceFilter(
+        choices=footnote_type_choices(),
```
NDevox, line 71: We shouldn't call the function here (it make the "lazy" part of the filter redundant).

Instead remove the parens like so:

```python3

footnote_type = LazyMultipleChoiceFilter(
    choices=footnote_type_choices,
    ...
)
```

footnotes/filters.py
```diff
@@ -38,7 +52,72 @@ class FootnoteFilterBackend(TamatoFilterBackend, FootnoteFilterMixin):
...
+        widget=forms.CheckboxSelectMultiple,
+        field_name="type__sid",
```
NDevox, line 73: I think it should be `field_name="footnote_type__footnote_type_id"`

NDevox (changes requested):
Couple minor things

NDevox (approved):
LGTM

common/static/common/js/showHideCheckboxes.js
```diff
@@ -0,0 +1,74 @@
+let showMoreClicked = {};
+
+const nodeListForEach = (nodes, callback) => {
+    if (window.NodeList.prototype.forEach) {
+      return nodes.forEach(callback)
+    }
+    for (let i = 0; i < nodes.length; i++) {
+      callback.call(window, nodes[i], i, nodes)
+    }
+}
```
andyhd, line 10: You could have added `forEach` to the prototype. There's a documented [polyfill](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#polyfill).

footnotes/filters.py
```diff
@@ -38,7 +51,72 @@ class FootnoteFilterBackend(TamatoFilterBackend, FootnoteFilterMixin):
...
 
 
+def footnote_type_choices():
+    footnote_types = models.FootnoteType.objects.current()
+    return [
+        Choice(
+            footnote_type.footnote_type_id,
+            "{0} - {1}".format(
+                footnote_type.footnote_type_id, footnote_type.description
+            ),
```
andyhd, line 61:
```python
f"{footnote_type.footnote_type_id} - {footnote_type.description}"
```

footnotes/filters.py
```diff
@@ -38,7 +51,72 @@ class FootnoteFilterBackend(TamatoFilterBackend, FootnoteFilterMixin):
...
 class FootnoteFilter(TamatoFilter, FootnoteFilterMixin):
+
+    footnote_type = LazyMultipleChoiceFilter(
+        choices=footnote_type_choices,
+        widget=forms.CheckboxSelectMultiple,
+        field_name="footnote_type__footnote_type_id",
+        label="Footnote Type",
+        help_text="Select all that apply",
+        required=False,
+    )
```
andyhd, line 76: Since the choices now come from the database, you should use a [ModelMultipleChoiceFilter](https://django-filter.readthedocs.io/en/stable/ref/filters.html#modelchoicefilter-and-modelmultiplechoicefilter-arguments), eg:
```python
from django_filters import ModelMultipleChoiceFilter
...
class FootnoteFilter(...):
    footnote_type = ModelMultipleChoiceFilter(
        field_name="footnote_type__footnote_type_id",
        queryset=models.FootnoteType.objects.current(),
        label="Footnote Type",
        help_text="Select all that apply",
        required=False,
    )

PR #171, opened 2021-02-01
Documentation upgrades

simonwo:
- Add business rules to the documentation.
- Fix up documentation to not emit so many warnings
    - Updates to Pygments that are [about to land](https://github.com/pygments/pygments/pull/1657) will
      take care of the remaining issue with `dot`.

### DIFF
--- a/importer/handlers.py
+++ b/importer/handlers.py
@@ -34,8 +34,8 @@ class BaseHandlerMeta(type):
 
     Firstly there are two required attributes:
 
-    1) "tag" - a string which is what matches the handler against the incoming data.
-    2) "serializer_class" - a ModelSerializer which is used to validate and create the database object.
+    1. "tag" - a string which is what matches the handler against the incoming data.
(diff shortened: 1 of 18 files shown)


importer/handlers.py
```diff
@@ -79,16 +79,19 @@ class BaseHandler(metaclass=BaseHandlerMeta):
     This effectively takes place in 8 stages:
 
     Init:
+
     1) The handler is initialised with the initial data.
```
NDevox, line 83: If we `)` -> `.` the above, should we not do that for these?

NDevox (approved):
Yep looks good to me. One minor comment about consistency, but not hugely important.

importer/handlers.py
simonwo: I think it turns out that the `)` -> `.` is not actually meaningful – initially tried this as a fix when I hadn't read the docs fully.

simonwo: So perhaps should change it back, but I'll leave that to be done when we update Pygments.

PR #173, opened 2021-02-02
Tiny PR to fix utils and add tests

simonwo:
The tests picked up on a few edge cases that should be taken care of.


stuaxo (approved):
LGTM

PR #175, opened 2021-02-02
Make workbaskets.xml great again.

stuaxo:
This is a small PR to make /api/workbaskets.xml output data again.

It changes the key for the list of results to be `results` - which is where Pagination outputs them.

This changes EnvelopeSerializer to use `TrackedModelSerializer(models, many=True, read_only=True).data` instead of it's custom map iterator.

`envelope_id` is wrong here, and there aren't any tests - but how this work will probably need changing in TP-396.

This is more to return some kind of base, so that when the new envelope serialization is added it's easy to see that it doesn't break anything.

### DIFF
--- a/common/renderers.py
+++ b/common/renderers.py
@@ -20,7 +20,8 @@ class TaricXMLRenderer(TemplateHTMLRenderer):
         context = super().get_template_context(*args, **kwargs)
 
         if isinstance(context, list):
-            context = {"items": context, "envelope_id": f"{counter_generator()():06}"}
+            context = {"results": context, "envelope_id": f"{counter_generator()():06}"}
+
         context["message_counter"] = counter_generator()
         context["counter_generator"] = counter_generator
         return context
--- a/common/renderers.py
+++ b/common/renderers.py
@@ -19,9 +19,7 @@ class TaricXMLRenderer(TemplateHTMLRenderer):
     def get_template_context(self, *args, **kwargs):
         context = super().get_template_context(*args, **kwargs)
 
-        if isinstance(context, list):
-            context = {"results": context, "envelope_id": f"{counter_generator()():06}"}
-
+        context["envelope_id"] = f"{counter_generator()():06}"
         context["message_counter"] = counter_generator()
         context["counter_generator"] = counter_generator
         return context
--- a/importer/management/commands/utils.py
+++ b/importer/management/commands/utils.py
@@ -696,9 +696,9 @@ class EnvelopeSerializer:
                 render_to_string(
                     template_name="workbaskets/taric/transaction.xml",
                     context={
-                        "tracked_models": map(
-                            self.serializer.to_representation, models
-                        ),
+                        "tracked_models": TrackedModelSerializer(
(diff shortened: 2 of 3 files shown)


simonwo:
I'm not really sure I understand why it's necessary to change from `results` to `items`? Does some part of not-our-code care about that key specifically?

stuaxo:
> I'm not really sure I understand why it's necessary to change from `results` to `items`? Does some part of not-our-code care about that key specifically?

It's the other way round.   Our code was putting stuff into `items`, the default for DRF is to put things into `results` - I think this might not have been the case before pagination.

Apologies if I made it sound the other way round :)

NDevox (approved):

common/renderers.py
```diff
@@ -20,7 +20,8 @@ def get_template_context(self, *args, **kwargs):
         context = super().get_template_context(*args, **kwargs)
 
         if isinstance(context, list):
```
andyhd, line 22: This test seems wrong. Isn't `context` always a `dict`?

common/renderers.py
```diff
@@ -20,7 +20,8 @@ def get_template_context(self, *args, **kwargs):
...
-            context = {"items": context, "envelope_id": f"{counter_generator()():06}"}
+            context = {"results": context, "envelope_id": f"{counter_generator()():06}"}
```
andyhd, line 23: Isn't the `results` key set by the paginator? Do we even want pagination in the XML output?

andyhd (commented):
I'm not clear on a couple of points

common/renderers.py
stuaxo: It does seem wrong, I don't know enough about how this works to know when where this list is coming from.

common/renderers.py
stuaxo: We shouldn't paginate the XML, keeping the key as `results` at least means the HTML + XML have the same structure template variables.

stuaxo: @andyhd THANKS - that `isinstance(list)` thing was really bugging me, removing it made this work.

common/renderers.py
NDevox: I can't remember why I did this... But I'm fairly sure it was for a reason 🥺. There was something weird going on in the renderers.

I'm happy to get rid of it for now. If it's causing bugs it should go, hopefully whatever it fixed had been superceded by a better solution. But we should keep it in mind in case anything else goes weird.

stuaxo: I'm hopeful this is sorted by the fix you suggested elsewhere in this PR.  
Data coming in now goes through the usual serializer infrastructure as opposed to a map statement that directly calls `model.to_representation`.

image

stuaxo: @NDevox turns out... I had a SO open about Pagination and it a similar isinstance(list)

https://stackoverflow.com/a/44371887/62709

andyhd (approved):
Nice 👍

stuaxo:
Retroactively, created an issue TP-606 for this, for tracking.

PR #176, opened 2021-02-03
TP-532 additional code list refactor

DIT-CJWaszczuk:


common/jinja2.py
```diff
@@ -62,6 +62,10 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
+def break_slashes(word):
+    return re.sub(r"\b([^\w]+)", r" \1 ", word)
```
NDevox, line 66: Can we get a docstring explaining what this regex is doing?

DIT-CJWaszczuk: Done (I think)

andyhd: Excellent bit of code 😛

andyhd: One thing I just thought of - instead of adding spaces, you could add [zero-width spaces](https://stackoverflow.com/questions/14615268/html-opposite-of-non-breaking-space)

andyhd (approved):

common/jinja2.py
```diff
@@ -62,6 +62,18 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
+def break_words(word):
+    """
+    Adds spaces at the end of words.
```
andyhd, line 67: Adds spaces around non-word characters.

common/jinja2.py
andyhd: Just to clarify, the regex says "replace 1 or more non-word characters (`[^\w]+`) at a word boundary (`\b`) with those characters surrounded by spaces". On second thought, the `\b` may be redundant.

common/jinja2.py
DIT-CJWaszczuk: I'll switch it over, this seems like a better implementation.

common/jinja2.py
```diff
@@ -62,6 +63,18 @@ def __init__(self, *args, **kwargs):
...
 
 
+def break_words(word):
+    """
+    Adds spaces around non-word characters to allow breaking lines when wrapping text.
+
+    For example:
+
+    >>> break_words("hello/goodbye")
+    "hello / goodbye"
```
andyhd, line 73: This needs changing to `"hello&8203;/&8203;goodbye"`, and the docstring above to "Adds zero-width spaces ..."

andyhd (approved):

common/jinja2.py
DIT-CJWaszczuk: Yup, works fine without `\b`

PR #179, opened 2021-02-04
TP-174 View / Search certificates

DIT-CJWaszczuk:
## Summary
Adds functionality to view / search certificates into TAP
## Screenshots
image
image

### DIFF
--- a/certificates/filters.py
+++ b/certificates/filters.py
@@ -1,0 +1,70 @@
+from datetime import datetime
+
+from crispy_forms_gds.choices import Choice
+from django import forms
+from django.contrib.postgres.aggregates import StringAgg
+from django.db.models import DateTimeField
+from django.db.models import Q
+from django.urls import reverse_lazy
+from django_filters import MultipleChoiceFilter
+
+from certificates import models
+from common.filters import ACTIVE_STATE_CHOICES
+from common.filters import LazyMultipleChoiceFilter
+from common.filters import TamatoFilter
+from common.util import TaricDateTimeRange
+
+
+def certificate_type_choices():
+    certificate_types = models.CertificateType.objects.current()
+    return [
+        Choice(
+            certificate_type.sid,
+            "{0} - {1}".format(certificate_type.sid, certificate_type.description),
+        )
+        for certificate_type in certificate_types
+    ]
+
+
+class CertificateFilter(TamatoFilter):
+    search_fields = (
+        "sid",
+        StringAgg("descriptions__description", delimiter=" "),
+    )  # XXX order is important
+
+    certificate_type = LazyMultipleChoiceFilter(
+        choices=certificate_type_choices,
+        widget=forms.CheckboxSelectMultiple,
+        field_name="certificate_type__sid",
+        label="Certificate Type",
+        help_text="Select all that apply",
+        required=False,
+    )
+
+    active_state = MultipleChoiceFilter(
+        choices=ACTIVE_STATE_CHOICES,
+        widget=forms.CheckboxSelectMultiple,
+        method="filter_active_state",
+        label="Active state",
+        help_text="Select all that apply",
+        required=False,
+    )
+
+    clear_url = reverse_lazy("certificate-ui-list")
+
+    def filter_active_state(self, queryset, name, value):
+
+        active_status_filter = Q()
+        current_date = TaricDateTimeRange(datetime.now(), datetime.now())
+        if value == ["active"]:
+            active_status_filter = Q(valid_between__upper_inf=True) | Q(
+                valid_between__contains=current_date
+            )
+        if value == ["terminated"]:
+            active_status_filter = Q(valid_between__fully_lt=current_date)
+
+        return queryset.filter(active_status_filter)
+
+    class Meta:
+        model = models.Certificate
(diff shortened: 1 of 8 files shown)


certificates/filters.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class CertificateFilter(TamatoFilter):
+    search_fields = (
+        "sid",
+        StringAgg("descriptions__description", delimiter=" "),
+    )  # XXX order is important
+
+    certificate_type = LazyMultipleChoiceFilter(
+        choices=certificate_type_choices(),
```
NDevox, line 36: No parens here

certificates/views.py
```diff
@@ -26,3 +28,23 @@ class CertificateTypeViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+
+class CertificatesList(TamatoListView):
+    """
+    UI endpoint for viewing and filtering Certificates
+    """
+
+    queryset = (
+        Certificate.objects.current()
+        # .select_related("type")
```
NDevox, line 40: should be `.select_related("certificate_type")`

NDevox (changes requested):
Couple small things, will have another check later

certificates/models.py
```diff
@@ -65,6 +65,17 @@ def get_descriptions(self, workbasket=None):
             .with_workbasket(workbasket)
         )
 
+    def get_description(self):
+        if (
+            hasattr(self, "_prefetched_objects_cache")
+            and "descriptions" in self._prefetched_objects_cache
+        ):
+            descriptions = list(self.descriptions.all())
+            print(self.descriptions.all())
```
NDevox, line 74: Remove print

certificates/jinja2/includes/certificates_list.jinja
```diff
@@ -0,0 +1,26 @@
...
+  {{ table_rows.append([
+    {"html": object_link},
+    {"text": certificate.get_description().description|default("-")},
+    {"text": certificate.certificate_type ~ " - " ~ certificate.certificate_type.description, "classes": "govuk-!-width-one-quarter"},
+    {"text": "{:%d %b %Y}".format(certificate.valid_between.lower)},
+    {"text": "{:%d %b %Y}".format(certificate.valid_between.upper) if certificate.valid_between.upper else "-"},
+    {"text": certificate.transaction.workbasket.get_status_display()},
+  ]) or "" }}
+{% endfor %}
+
```
stuaxo, line 15: I haven't seen this style of building tables before, where data is built up within the template and then passed to something like govukTable, but maybe it's a jinja2 style.

It does make the code quite readable, in other django projects the often data would be setup in the view, and then the table setup with {% block %} etc. 

I guess there isn't enough data for it to matter that there is an extra loop over the data with this.

DIT-CJWaszczuk: I see where you're going with this, I actually like the readability, but think if it was moved into the view, we could potentially move away from having a separate list file. Bit on the fence really. But yeah, the amount of data in this case is limited, so the extra loop would not cause any big issues.

andyhd: It's a side effect of using GOV.UK Frontend - their table macro wants a list of (JSON) rows, but if we want the row data to be calculated or styled, then we are forced to build the array either in the view and pass it via a context variable, or build it in the template (which has problems too - for example the ugly `]) or "" }}` which prevents the loop from outputting `None, None, None, ...`).

I think we could maybe write a better macro which uses [`call`](https://jinja.palletsprojects.com/en/2.11.x/templates/#call) to get something nicer, eg:
```jinja
{% call(certificate) tamato_list(object_list) %}
    <td><a class="govuk-link">{{ certificate.sid }}</a></td>
    ...
{% endcall %}

{% macro tamato_list(object_list) %}
  <table>
    ...
    {% for object in object_list %}
      <tr>{{ caller(object) }}</tr>
    {% endfor %}
  </table>
{% endmacro %}
```
but I'm not sure how we could integrate it with `govukTable`, or if it would be worth the effort.

certificates/filters.py
```diff
@@ -0,0 +1,70 @@
...
+    ]
+
+
+class CertificateFilter(TamatoFilter):
+    search_fields = (
+        "sid",
+        StringAgg("descriptions__description", delimiter=" "),
+    )  # XXX order is important
+
+    certificate_type = LazyMultipleChoiceFilter(
```
andyhd, line 35: So, per [my posthumous comment on the previous PR](https://github.com/uktrade/tamato/pull/169#pullrequestreview-584407685), I think you can use `django_filters.ModelMultipleChoiceFilter` here, but you may need to investigate how to override the choice label generation.

andyhd (approved):
Looks good to me, not sure if it's worth the time to generalize the list template.

certificates/jinja2/includes/certificates_list.jinja
stuaxo: Yeah, agree on both of these.

DIT-CJWaszczuk:
@NDevox Final check to make sure all your suggestions are sorted?

NDevox (approved):
LGTM

PR #180, opened 2021-02-05
TP-575: Add an Activity Stream endpoint for Data Workspace.

NDevox:
It is a requirement that the tariff data be open and as accessible as possible.
Part of this is to make the data available on Data Workspace, and the advised
route to do so is to expose an endpoint that provides all the current tariff
data in an ActivityStream format.

This commit adds an endpoint in the exporter package which formats all the
latest TrackedModels (paginated by 20) in the ActivityStream format ready
for consumption by DataWorkspace.

### DIFF
--- a/common/serializers.py
+++ b/common/serializers.py
@@ -1,6 +1,7 @@
 from django.contrib.auth.models import Group
 from django.contrib.auth.models import User
 from drf_extra_fields.fields import DateTimeRangeField
+from rest_flex_fields import FlexFieldsModelSerializer
 from rest_framework import serializers
 from rest_polymorphic.serializers import PolymorphicSerializer
 
@@ -18,7 +19,7 @@ class TARIC3DateTimeRangeField(DateTimeRangeField):
     range_type = TaricDateTimeRange
 
 
-class TrackedModelSerializerMixin(serializers.ModelSerializer):
+class TrackedModelSerializerMixin(FlexFieldsModelSerializer):
     taric_template = serializers.SerializerMethodField()
 
     formats_with_template = {"xml"}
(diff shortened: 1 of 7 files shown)


exporter/views.py
```diff
@@ -0,0 +1,125 @@
...
+    tracked_models = get_latest_tracked_models(request, per_page)
+
+    page = {
+        "@context": [
+            "https://www.w3.org/ns/activitystreams",
+            {"dit": "https://www.trade.gov.uk/ns/activitystreams/v1"},
+        ],
+        "type": "Collection",
+        "orderedItems": [
+            tracked_model_to_activity_stream_item(tracked_object)
```
NDevox, line 116: This is a right pain at the moment as up till this point we have a total of 2 queries. After this line the number of queries is 302.

Tested on fetching 20 MeasureExcludedGeographicalAreas. Starting queries would be higher than 2 if there were more types (would be n+number of types).

Could push as is, but this creates response times of around 15s. Will try hunt for a solution before coming out of draft.

stuaxo: Having a batch version of this that could take all the tracked_models would certainly be good.

NDevox: Might be hard to do as we still need to break down the data at each individual object.

NDevox: Managed to get the queries down to 3 for the specific example I'm using (which is only querying on one TrackedModel type).

common/serializers.py
```diff
@@ -95,6 +96,32 @@ def register_polymorphic_model(cls, serializer):
         cls.model_serializer_mapping[serializer.Meta.model] = serializer
         return serializer
 
+    def __init__(self, *args, child_kwargs=None, **kwargs):
+        """
+        This is a near direct copy from the base class - however it adds in
```
NDevox, line 101: I don't intend to keep this if we don't have to: https://github.com/apirobot/django-rest-polymorphic/pull/34

But for now it solves the problem we have.

stuaxo (approved):

simonwo (approved):
Can't find anything to complain about!

From memory, in the examples I saw there was some information about which user did the action defined in the stream. I guess that's optional and we can come back and put it in later if we want it?

PR #182, opened 2021-02-08
TP-172 Certificate details tab

DIT-CJWaszczuk:
## Summary
Adds the details page for certificates 
## Notes
Description does work, just a mistake when creating the records.  
## Screenshot
image

### DIFF
--- a/certificates/jinja2/includes/certificates/tabs/core_data.jinja
+++ b/certificates/jinja2/includes/certificates/tabs/core_data.jinja
@@ -1,0 +1,39 @@
+<div class="footnote__core-data">
+    <div class="footnote__core-data__content">
+        <h2 class="govuk-heading-l">Details</h2>
+        {{ govukSummaryList({
+            "rows": [
+            {
+                "key": {"text": "Certificate ID"},
+                "value": {"text": object.sid},
+                "actions": {"items": []}
+            },
+            {
+                "key": {"text": "Certificate type"},
+                "value": {"text": object.certificate_type ~ " - " ~ object.certificate_type.description},
+                "actions": {"items": []}
+            },
+            {
+                "key": {"text": "Description"},
+                "value": {"text": object.get_description().description},
+                "actions": {"items": []}
+            },
+            {
+                "key": {"text": "Start date"},
+                "value": {"text": "{:%d %b %Y}".format(object.valid_between.lower)},
+                "actions": {"items": []}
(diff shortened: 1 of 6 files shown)


DIT-CJWaszczuk:
I'm not 100% convinced the regex for the id is correct, would be good to double check(i'm terrible with regex)

certificates/jinja2/includes/certificates/tabs/core_data.jinja
```diff
@@ -0,0 +1,39 @@
...
+                "actions": {"items": []}
+            },
+            {
+                "key": {"text": "Certificate type"},
+                "value": {"text": object.certificate_type ~ " - " ~ object.certificate_type.description},
+                "actions": {"items": []}
+            },
+            {
+                "key": {"text": "Description"},
+                "value": {"text": object.get_description().description},
```
andyhd, line 18: From the screenshot, it looks like you might have to check for no descriptions here, eg:
```jinja
"value": {"text": (object.get_description()|default({})).description},
```
(I haven't tested this)

andyhd (approved):
Looks good to me, just need to catch the no-descriptions case (which probably won't happen IRL).

certificates/urls.py
```diff
@@ -14,6 +15,11 @@
         views.CertificatesList.as_view(),
         name="certificate-ui-list",
     ),
+    re_path(
+        r"^(?P<sid>[A-Z0-9]*)$",
```
andyhd, line 19: This will match an empty string. The correct regex is here: https://github.com/uktrade/tamato/blob/ae512458e3a43a51b832f3f61c064c8aa9ed9143/certificates/validators.py#L8

certificates/jinja2/includes/certificates/tabs/core_data.jinja
DIT-CJWaszczuk: Is there a scenario of a certificate / entity having no description?

DIT-CJWaszczuk: Apparently no description is against the business rules anyway, so we can make an assumption it will always be there

certificates/urls.py
DIT-CJWaszczuk: I've taken this, it is a tad messy, but the alternatives all are extra code and hardcoding results in tech debt.

PR #186, opened 2021-02-09
Fix cache issues in importer tests.

NDevox:
In some cases bits of data get stuck in the test cache when
running, this causes a cascade of all following importer tests
to fail as they try to resolve the cache.

This commit adds a clearing action to the cache and utilises it
within the tests to ensure this issue does not arise.


andyhd (approved):
👍

PR #188, opened 2021-02-09
Prevent ME32 on measures with implicit end dates

simonwo:
Previously ME32 only considered the explicit validity period of any matching measures. 

It now also considers the implicit validity period through the user of `effective_valid_between`.

Tests have also been refactored to make it easier to check all the permutations of dates and comm codes, all in their own transaction (otherwise the transactions have to be controlled manually).


NDevox (approved):
LGTM

stuaxo (approved):

PR #189, opened 2021-02-10
TP-576: Refactor validity periods to use dates

andyhd:
Currently validity periods use timestamps with time zones to record their start and end times. This overcomplicates matters, as all that is required is start and end dates.

This commit refactors validity periods for all data models to use Postgres' `daterange` type instead of `tstzrange`.

It does this with a custom migration operation `ConvertTaricDateRange`.

This is required because Postgres does not automatically cast `tstzrange` to `daterange` and vice versa. Neither does it cast from `timestamp with time zone` values to `date` values.

`ConvertTaricDateRange` is a reversible migration operation which converts the column type with `ALTER COLUMN TYPE USING` to preserve the data. There is some loss in the conversion - times and timezones are truncated when converting to dates. When converting back to `tstzrange`, start times are set to `00:00:00`, end times are set to `23:59:59` and timezones are set to UTC. Range bounds (inclusive/exclusive) are preserved.

A custom operation is used instead of `RunSQL` to facilitate reuse across all the models.

This commit also replaces `setup.cfg` with `pyproject.toml`, as the latter is now the preferred means of consolidating Python module configuration settings.

A check for missing migrations is also added to the Github workflow, which will disallow merging PRs if there are model changes not reflected in migrations.


stuaxo (approved):
LGTM :)

importer/migrations/0002_auto_20210201_1639.py
```diff
@@ -0,0 +1,21 @@
...
+
+    operations = [
+        migrations.AlterField(
+            model_name="importerxmlchunk",
+            name="status",
+            field=models.PositiveSmallIntegerField(
+                choices=[(1, "Waiting"), (2, "Running"), (3, "Done"), (4, "Errored")],
+                default=1,
+            ),
+        ),
```
simonwo, line 20: Looks like this was just never generated…

Is it possible to add a check that there are no outstanding migrations during the CI build?

simonwo (approved):
Seeing the deletion of `from datetime import timezone` is very pleasing.

common/fields.py
```diff
@@ -60,5 +66,25 @@ def deconstruct(self):
...
+        if not isinstance(value, DateRange):
+            return value
+        lower = value.lower
+        upper = value.upper
+        if not value.upper_inc and not value.upper_inf:
+            upper = upper - relativedelta(days=1)
+        return TaricDateRange(lower=lower, upper=upper)
+
+
 class TaricDateTimeRangeField(DateTimeRangeField):
```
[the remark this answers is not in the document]
andyhd: Yes, this is necessary for the migrations to work. We can remove it later.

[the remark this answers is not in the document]
andyhd: Actually, maybe we can remove it now.

importer/migrations/0002_auto_20210201_1639.py
andyhd: That's a good idea, I will see what I can do

.pre-commit-config.yaml
```diff
@@ -8,3 +8,4 @@ repos:
     rev: 20.8b1
     hooks:
       - id: black
+        exclude: ^.*\b(migrations)\b.*$
```
NDevox, line 11: Don't know how I feel about this.

On the one hand I agree it is silly to format autogenerated code.

On the other hand we've already formatted all of it so far, may as well carry on (any future commits will have new migrations - having them non formatted on first commit just poses the risk of people accidentally formatting and commiting them again)

additional_codes/migrations/0002_auto_20210201_1639.py
```diff
@@ -0,0 +1,48 @@
...
+                    ("C", "C - Anti-dumping / countervailing"),
+                    ("D", "D - Dual Use"),
+                    ("P", "P - Refund for basic products"),
+                    (
+                        "T",
+                        "T - This is the additional code type for UK trade remedies from 2021",
+                    ),
+                    ("V", "V - VAT"),
+                    ("X", "X - EXCISE"),
+                ],
```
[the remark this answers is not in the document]
NDevox: Yeah this comes from the work done with Chris on the frontend - my fault as it was my suggestion. I agree with you on this. We should get rid of these choices (and update the frontend if it is still using the choices to instead use the DB)

.pre-commit-config.yaml
andyhd: I'm also not convinced about it - all the migrations in this PR were formatted automatically by my editor anyway.

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.black]
```
NDevox, line 1: For the sake of cleanliness shall we get rid of setup.cfg in favour of this?

We can get rid of flake8 and everything else supports this format. It's only black that's too stubborn to support setup.cfg

.pre-commit-config.yaml
NDevox: I'd probably argue to keep formatting them then. Especially as we do on occasion manually write migrations.

pyproject.toml
andyhd: I feel like this is a https://xkcd.com/927/ situation.

NDevox: I agree with this in everyway

importer/migrations/0002_auto_20210201_1639.py
andyhd: Added a check that will fail the build if there are missing migrations

andyhd: ... and it has shown up a problem with my custom migration operation, so that's good

.github/workflows/check-pr.yml
```diff
@@ -51,6 +51,13 @@ jobs:
         npm run build
         python manage.py collectstatic --noinput
 
+    - name: Check for missing migrations
+      env:
+        DATABASE_URL: postgres://postgres:postgres@localhost:5432/postgres
+        DJANGO_SETTINGS_MODULE: settings.test
+      run: |
+        python manage.py makemigrations --dry-run --check
+
```
stuaxo, line 60: :+1:  :1st_place_medal:  :1st_place_medal:  :1st_place_medal:  :1st_place_medal:

stuaxo (approved):

regulations/migrations/0004_auto_20210211_1451.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('regulations', '0003_auto_20210210_1516'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='regulation',
+            name='valid_between',
+            field=common.fields.TaricDateRangeField(blank=True, null=True),
```
andyhd, line 17: For some reason, the migrations autodetector doesn't detect the `ConvertTaricDateRange` operation on this particular field. Since it is only one field, I think it is more expedient to keep this extra migration than to figure out how to avoid it.

PR #190, opened 2021-02-10
Fixes to common code from data migration work.

simonwo:
This PR contains a few tiny fixes to things that currently break when dealing with real data and migrations, particuarly around running business rules.


common/models/transactions.py
```diff
@@ -58,9 +58,9 @@ def clean(self):
         for model in self.tracked_models.all():
             for rule in model.business_rules:
                 try:
-                    rule.validate(model)
+                    rule().validate(model)
                 except BusinessRuleViolation as violation:
-                    self.errors.append((model, violation))
+                    self.errors.append(violation)
```
NDevox, line 63: Not convinced we want to change this from the tuple - I believe the idea was so that if there were errors we'd get a map of which object raised the error. Like this we'll just get a list of errors and no knowledge of which errors raised them.

common/tests/migrations/0002_auto_20210210_1534.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 3.1 on 2021-02-10 15:34
```
NDevox, line 1: I try to avoid making new migrations for the test models.

Better to just delete the 1 migration and make a new initial one. These are never loaded into the prod databases, only used in tests, so there is no danger in it. Just keeps the code cleaner.

common/models/transactions.py
simonwo: Ah yes, I think my thought on this was that the violation object was the one that should be holding the model. The `ValidationError` class has a `params` attribute that could be used to store the errored models. I think this would end up better because the error classes are designed to work well when nested whereas they don't so much understand a tuple of things. So I think my proposal here is that we would instead do something like:

    BusinessRuleViolation("message", params=model)

or even

    BusinessRuleViolation("message", params={"model": model, "related_models": …})

in cases where the rule fails due to the presence of other models. The UI could then hook these out and display actions for them. Thoughts?

NDevox: Yep I actually completely agree with this. Makes sense as we move away from human intervention to something more machine appropriate.

NDevox (approved):

PR #191, opened 2021-02-11
TP-149 Refactor geographical areas list

DIT-CJWaszczuk:
##  Summary 
Refactor geographical areas to be more  inline with  the designs 
## Screenshots 
image


geo_areas/filters.py
```diff
@@ -32,9 +36,38 @@ class GeographicalAreaFilterBackend(TamatoFilterBackend, GeographicalAreaFilterM
...
+        widget=forms.CheckboxSelectMultiple,
+        method="filter_active_state",
+        label="Active state",
+        help_text="Select all that apply",
+        required=False,
+    )
+
     clear_url = reverse_lazy("geoarea-ui-list")
 
+    def filter_active_state(self, queryset, name, value):
```
NDevox, line 58: Starting to think this should go in as a common filter mixin or something. This will be painful to maintain now if we need to make changes.

DIT-CJWaszczuk: yup, just switched it over into its own mixin, it was becoming a pain. Going to do the same for start year and then switch them all over in another cleanup PR.

DIT-CJWaszczuk:
Not quite sure what we want to do with the tests in this PR, I don't want to remove them if they provide value, because `geoarea-ui-detail` is going to re-appear in the next one anyway. They are breaking the tests currently however, so i've  ended  up commenting them out.

geo_areas/tests/bdd/test_detail_geo_areas.py
```diff
@@ -1,78 +1,56 @@
...
-
-@then("the members against the geographical_area should be presented")
-def geo_area_group_members_data(geo_area_group_detail, geographical_area_1002):
-    members = models.GeographicalMembership.objects.filter(
-        geo_group=geographical_area_1002
-    )
-    compare_members_to_html(
-        members, geo_area_group_detail.content.decode(), is_group=False
-    )
+# """Tests for browse geographical area behaviours."""
```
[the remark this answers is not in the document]
DIT-CJWaszczuk: This works, just put it in

DIT-CJWaszczuk:
@NDevox Anything else that you can see?

NDevox (approved):
Only thought is whether we want to wait on #189 for the daterange changes. I think it should be merged today... But no real harm if this is merged first.

conftest.py
```diff
@@ -1,37 +1,27 @@
...
-from typing import Any
-from typing import Callable
-from typing import Dict
-from typing import Optional
-from typing import Type
-from typing import Union
-from unittest.mock import patch
-from unittest.mock import PropertyMock
+from typing import Any, Callable, Dict, Optional, Type, Union
+from unittest.mock import PropertyMock, patch
```
andyhd, line 5: Ah, I think this is my fault

DIT-CJWaszczuk: I thought it was intentional, was it not?

andyhd: No, imports should be on individual lines

andyhd: This will be fixed by https://github.com/uktrade/tamato/pull/193

PR #192, opened 2021-02-11
TP-616: Configure ELK logging, Sentry and Pingdom

NDevox:


common/views.py
```diff
@@ -29,6 +30,18 @@ def index(request):
...
+        "<pingdom_http_custom_check><status><strong>OK</strong></status>"
+        "<response_time><strong>1</strong></response_time>"
+        "</pingdom_http_custom_check>",
+        content_type="text/xml",
+    )
+    response["Cache-Control"] = "no-cache, no-store, must-revalidate"
+
+    return response
+
+
```
simonwo, line 44: I beleive that now we have attached services (postgres, redis) the healthcheck is also meant to check that they are accessible?

At which point the response time maybe also needs to reflect truth.

simonwo (changes requested):

common/views.py
NDevox: Fun when you go looking for examples of people doing this and find examples such as https://github.com/uktrade/data-workspace/blob/97bd538617c4835a749fe166eee7e4fca0dd6acc/dataworkspace/dataworkspace/apps/core/views.py#L41

stuaxo: I guess this should try and connect to celery too.

Could be a nice excuse to use async to start all checks the subtasks then gather the results.

simonwo (approved):

common/views.py
NDevox: This is done now @simonwo

NDevox: Ha yeah not sure I'd bother with the async stuff - don't think the return is great enough - given that it's only pingdom.

Testing for celery is an interesting idea. I don't know how much I want to lean into our reliance on it before we're properly using it just yet. But once it becomes integral we should definitely put it on the list.

PR #194, opened 2021-02-12
TP-620: ADR for including preference code table.

simonwo:
It will be easiest to explain by [reading the ADR](https://github.com/uktrade/tamato/blob/5330e8ccf0ccea630b516e8001edfdd7c6a508cd/docs/adr/0011-store-and-check-preference-codes.rst).


docs/adr/0011-store-and-check-preference-codes.rst
```diff
@@ -0,0 +1,250 @@
...
+include 696 measures.
+
+It controls requirements around presence of supporting measures
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some preference codes are only valid when two measure types are found together.
+For example, the preference code "323" covers preferential tariff quotas subject
+to end-use. There are two ways for this preference code to be activated:
+
+* by the presence of a end-use preferential quota (146),
```
NDevox, line 56: *an*

docs/adr/0011-store-and-check-preference-codes.rst
```diff
@@ -0,0 +1,250 @@
...
+Decision
+--------
+
+Tables of the current preference code correlation data will be stored in TaMaTo
+using the following schema:
+
+.. code:: python
+
+    class PreferenceCode(TrackedModel, ValidityMixin):
+        identifying_fields = ("code",)
```
NDevox, line 83: Query as to whether we should just call this `SID` to homogenise the backend a bit (frontend can translate this to `code`)

docs/adr/0011-store-and-check-preference-codes.rst
```diff
@@ -0,0 +1,250 @@
...
+
+Decision
+--------
+
+Tables of the current preference code correlation data will be stored in TaMaTo
+using the following schema:
+
+.. code:: python
+
+    class PreferenceCode(TrackedModel, ValidityMixin):
```
NDevox, line 82: As this isn't really part of the tariff, but kind of an accessory to the tariff? Do we want this to be a TrackedModel?

I lean heavily to _yes_ as is done here, it helps most of the backend fall into place. But I think it may raise some added complications when it comes to things like emitting records to HMRC... (I assume we do not want to do this?)

docs/adr/0011-store-and-check-preference-codes.rst
```diff
@@ -0,0 +1,250 @@
...
+        identifying_fields = ("code",)
+
+        code = models.CharField(
+          min_length=3,
```
NDevox, line 86: There is no `min_length` option here (no such SQL constraint either). The regex covers this though.

docs/adr/0011-store-and-check-preference-codes.rst
```diff
@@ -0,0 +1,250 @@
...
+
+        description = models.TextField(
+          help_text="""A detailed description of the general treatment
+            the trader will receive through the use of this code or more
+            information on the circumstances in which this code applies.""",
+        )
+
+
+    class PreferenceCodeMapping(TrackedModel, ValidityMixin):
+        identifying_fields = ("id",)
```
NDevox, line 105: ID (i.e. the underlying pk) being the identifying field would probably cause issues. It would be impossible to ever have more than one version of this.

NDevox: Wondering if these fields can just be tagged onto `PreferenceCode`. In the end it's only adding on a `BoolField`

docs/adr/0011-store-and-check-preference-codes.rst
stuaxo: If it's the same thing, most definitely.

simonwo: I imagine there may be some other stuff in the real code also. I just found this the easiest way to show the model!

docs/adr/0011-store-and-check-preference-codes.rst
simonwo: I agree that we should track this stuff, for the ability to run business rules on past transactions using the right data at least.

Yes, it does probably have some implications e.g. we no longer want all TrackedModels to implement get_taric_template. There might be a need to have just `TrackedModel` and then a subclass `TARICModel` that does any other stuff.

docs/adr/0011-store-and-check-preference-codes.rst
simonwo: Docs [seem to think it can be done](https://docs.djangoproject.com/en/3.1/ref/forms/fields/#django.forms.CharField) but yeah as the regex has it I don't mind it not being there.

docs/adr/0011-store-and-check-preference-codes.rst
simonwo: I guess we could just give it a SID?

Basically the correlation data looks like this:
image

So one preference code can have multiple distinct mappings – and the business rule is that at least one of the mappings has to apply. Basically the mapping is just where we group the measure types together (in the above example there are 3: 146, 143+464 and [653, 654])

simonwo: This would correspond to:

    pref_code = PreferenceCode(code="323", name="Preferential tariff quotas subject to an end-use")
    mappings = [
        PreferenceCodeMapping(
            preference_code=pref_code,
            measure_types=MeasureType.objects.get(sid="146"),
            regulation_groups=Group.objects.get(group_id__in=["FTA", "PRF", "RDR"]),
        ),
        PreferenceCodeMapping(
            preference_code=pref_code,
            measure_types=MeasureType.objects.get(sid__in=["143", "464"]),
            regulation_groups=Group.objects.get(group_id__in=["FTA", "PRF", "RDR"]),
            choice_of_type=False,
        ),
        PreferenceCodeMapping(
            preference_code=pref_code,
            measure_types=MeasureType.objects.get(sid__in=["653", "654"]),
            regulation_groups=Group.objects.get(group_id__in=["FTA", "PRF"]),
        ),
    ]

docs/adr/0011-store-and-check-preference-codes.rst
simonwo: Ah, just realised I've been swindled by the Form fields instead of the Model fields!

docs/adr/0011-store-and-check-preference-codes.rst
andyhd: PreferenceCode could return a blank template, so that it renders no XML? Although it would probably be better to skip them altogether.

andyhd (approved):
👍 seems ok to me

PR #195, opened 2021-02-12
TP-609: Include objects up to a given transaction in .current

NDevox:
When checking actual current data `.current` works adequately.
However in many cases, especially the business rules this isn't
enough. In these cases the normal requirement is to check the
most current data for an object, or the data for the object in
the current transaction, or the data for the object in any
preceding transaction that is within the same workbasket as the
current transaction. Whichever has the greatest ID (so is most
recent) needs to be returned.

This commit adds a new QuerySet method called `current_as_of`
which takes a transaction and executes a query that satisfies
this requirement.

### DIFF
--- a/common/models/records.py
+++ b/common/models/records.py
@@ -14,6 +14,7 @@ from django.db import models
 from django.db.models import Case
 from django.db.models import F
 from django.db.models import Field
+from django.db.models import Max
 from django.db.models import Q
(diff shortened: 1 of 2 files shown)


stuaxo (approved):
LGTM

simonwo (approved):
Looks familiar :)

PR #199, opened 2021-02-15
TP-172 Fix non-unique certificate SID scenario

DIT-CJWaszczuk:
## Summary 
A certificate sid alone is not unique, it's the combination of the type id and the sid that is. Changed so that we now use both.

### DIFF
--- a/certificates/jinja2/certificates/detail.jinja
+++ b/certificates/jinja2/certificates/detail.jinja
@@ -2,7 +2,7 @@
 {% from "components/summary-list/macro.njk" import govukSummaryList %}
 {% from "components/table/macro.njk" import govukTable %}
 
-{% set page_title = "Certificate: " ~ object.sid %}
+{% set page_title = "Certificate: " ~ object.code %}
 {% set find_edit_url = url("certificate-ui-list") %}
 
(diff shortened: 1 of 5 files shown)


DIT-CJWaszczuk:
I think this is what we want? Not 100% sure i've implemented this correctly.

andyhd:
I think you are missing the filter stuff

DIT-CJWaszczuk:
The mixins are on the `FootnoteFilter` class slightly below

andyhd:
I mean you don't have a `CertificateFilterMixin` to handle the URL being made up of `<certificate_type_id><certificate_code>` like in footnotes. The view needs to know how to get the certificate model from the URL parameters, doesn't it?

andyhd (approved):

andyhd:
You might need to handle a missing description - not sure if that's actually allowed in the real data, though.

DIT-CJWaszczuk:
Think the description has to be there, otherwise I think this is good to go. I pretty much just pulled the search field override from the filter into the other method. Which is why it still works.

NDevox (approved):
It should be impossible to have a certificate without a description within the business rules.

Emphasis on the _should_

PR #202, opened 2021-02-16
TP-615: Update business rules query top find objects in current transaction

NDevox:
Business rules must also account for objects in the current transaction (and
unapproved transactions up to the current one in the same workbasket). This
commit updates business rules queries to take this into account.


andyhd (approved):
Looks good 👍

PR #204, opened 2021-02-16
TP-172 Fix hardcoded breadcrumb

DIT-CJWaszczuk:
## Summary 
Detail breadcrumb had hardcoded link text, made that dynamic using the same standard as list.

### DIFF
--- a/common/jinja2/layouts/detail.jinja
+++ b/common/jinja2/layouts/detail.jinja
@@ -1,10 +1,12 @@
 {% extends "layouts/layout.jinja" %}
 
+{% set object_type_plural = object_type_plural or object_type + "s" %}
+
 {% block breadcrumb %}
   {{ govukBreadcrumbs({
     "items": [
       {"text": "Home", "href": url("index")},
-      {"text": "Find and edit footnotes", "href": find_edit_url},
+      {"text": "Find and edit " ~ object_type_plural, "href": find_edit_url},
       {"text": page_title}
     ]
   }) }}
--- a/additional_codes/jinja2/additional_codes/detail.jinja
+++ b/additional_codes/jinja2/additional_codes/detail.jinja
@@ -2,6 +2,7 @@
 {% from "components/summary-list/macro.njk" import govukSummaryList %}
 {% from "components/table/macro.njk" import govukTable %}
 
+{% set object_type = "additional code" %}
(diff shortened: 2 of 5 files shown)


common/jinja2/layouts/detail.jinja
```diff
@@ -1,10 +1,12 @@
 {% extends "layouts/layout.jinja" %}
 
+{% set object_type_plural = object_type_plural or object_type + "s" %}
+
 {% block breadcrumb %}
   {{ govukBreadcrumbs({
     "items": [
       {"text": "Home", "href": url("index")},
-      {"text": "Find and edit footnotes", "href": find_edit_url},
+      {"text": "Find and edit " ~ (object_type_plural or object_type + "s"), "href": find_edit_url},
```
NDevox, line 9: As you already set `object_type_plural` to use `s` if one doesn't exist you can just replace this line with `"Find and edit " ~ object_type_plural`

DIT-CJWaszczuk: I followed the same suit as the [list](https://github.com/uktrade/tamato/blob/master/common/jinja2/layouts/list.jinja) file (I think  we spoke about the necessity of it there but I can't remember why we chose to keep it). I've switched this one over, do you remember why we kept it there?

NDevox: No idea why we kept that 🤔  looks like it should be changed over too.

NDevox (approved):

common/jinja2/layouts/detail.jinja
```diff
@@ -1,10 +1,12 @@
 {% extends "layouts/layout.jinja" %}
 
+{% set object_type_plural = object_type_plural or object_type + "s" %}
```
andyhd, line 3: I'd be tempted to make these available from the models, eg:
```python
class TrackedModel(...):
    ...
    @property
    def type_name(self) -> str:
        return self.__class__.__name__

    @property
    def type_name_plural(self) -> str:
        return f"{self.type_name}s"
```
and
```python
class GoodsNomenclature(TrackedModel, ...):
    ...
    @property
    def type_name(self) -> str:
        return "Commodity"
    
    @property
    def type_name_plural(self) -> str:
        return "Commodities"
```
Then you wouldn't need these `{% set ... %}` tags and could do
```jinja2
{{ "Find and edit " ~ object.type_name_plural }}
```

PR #205, opened 2021-02-16
Docformat all docs.

NDevox:
Docformatter has recently been introduced as a standard.
However as it has not yet been run globally it is polluting
all the current PRs.

This commit runs docformatter globally, so that it doesn't
need to interfere with a future PR again.


additional_codes/business_rules.py
```diff
@@ -88,24 +85,25 @@ def validate(self, additional_code):
...
 
 class ACN5(DescriptionsRules):
-    """At least one description is mandatory. The start date of the first description
-    period must be equal to the start date of the additional code. No two associated
-    description periods may have the same start date. The start date must be less than
-    or equal to the end date of the additional code.
+    """
+    At least one description is mandatory.
+
+    The start date of the first description period must be equal to the start
```
andyhd, line 98: This one is wrong, as the first sentence is not the "title" of the doc-comment. But I suppose the real problem is that this isn't a proper doc-comment.

additional_codes/import_parsers.py
```diff
@@ -11,16 +11,14 @@
...
-        </xs:complexType>
-    </xs:element>
+
+    <xs:complexType>         <xs:sequence>             <xs:element
+    name="additional.code.sid" type="SID"/>             <xs:element
+    name="additional.code.type.id" type="AdditionalCodeTypeId"/> <xs:element
+    name="additional.code" type="AdditionalCode"/> <xs:element
+    name="validity.start.date" type="Date"/>             <xs:element
+    name="validity.end.date" type="Date" minOccurs="0"/>         </xs:sequence>
+    </xs:complexType> </xs:element>
```
andyhd, line 21: This is a shame - should we use reST to mark this up as a literal block, eg:
```python
class AdditionalCodeParser(...):
    """
    ::

        <xs:element name="additional.code" ...>
        ...
        </xs:element>
    """

andyhd (approved):
Fine by me, but unfortunately some docstrings are reformatted badly. Maybe we can fix them as we go - they won't pollute other PRs the way the docformatter does.

additional_codes/import_parsers.py
andyhd: Although I've no idea if docformatter would respect reST 🤞

NDevox: Yep I've just pushed this in. All formatted as XML now in RST blocks.

PR #206, opened 2021-02-16
Annotation for duty sentences on Measure and MeasureCondition queries

simonwo:
A "duty sentence" is a human-readable string that we display in the UI to represent the duty a measure has associated with it. For example, `1.2% + 3.45 GBP / kg` specifies that a trader will pay 1.2% of the total order value plus £3.45 per kg of goods.

We already have a `DutySentenceParser` that can take these strings and pull them apart into measure components, and we now have an implementation that can do the reverse – take a set of components and build them into a string.

This is implemented as a queryset annotation (rather a simple python function) because we often want to output many hundreds of measures at once (on a screen, or in a spreadsheet for example) and calling a function that makes a query over the components per measure is slow. Instead we let the database compute all the duty sentences together as part of the query.

This also renames a field on `MeasureConditionComponent` at which point we get duty sentence output from those for free too.


measures/tests/test_parsers.py
```diff
@@ -17,70 +15,30 @@ def seasonal_rate_parser() -> SeasonalRateParser:
...
-    duty_sentence: str,
-    expected_results,
+    duty_sentence_data: Tuple[str, List],
 ):
-    def get_from_field(value, options, field=lambda e: e.code):
-        matches = list(filter(lambda unit: field(unit) == value, options))
-        return matches[0] if len(matches) > 0 else None
-
+    duty_sentence, expected_results = duty_sentence_data
     components = list(duty_sentence_parser.parse(duty_sentence))
```
stuaxo, line 23: Nice :) as soon as I started reading this PR, I was wondering where the round-trip testing would appear..

This give more informative errors as just one assertion:
`assert expected_results == duty_sentence_parser.parse(duty_sentence)`

At that point the helper function is only doing one thing assertion, instead it could return True on equality instead of asserting, allowing to do something like `assert duty_sentence_parser_can_roundtrip(...)`  [1]

[1] or a more sensible name.

measures/tests/test_querysets.py
```diff
@@ -0,0 +1,44 @@
...
+        models.Measure.__name__,
+        models.MeasureCondition.__name__,
+    ),
+)
+def test_duty_sentence_generation(
+    component_factory,
+    model_factory,
+    field,
+    reversible_duty_sentence_data,
+):
```
stuaxo, line 33: This could def do with a docstring here, or a comment below where the logic around component_data/component_factory is.

measures/tests/test_parsers.py
simonwo: Alas the `expected_results` is a list of dicts, whereas the parser will return real components. So I don't think they compare equally?

measures/querysets.py
```diff
@@ -1,14 +1,99 @@
...
+    def with_duty_sentence(self) -> QuerySet:
+        """
+        Annotates the query set with a human-readable string that represents the
+        aggregation of all of the linked components into a single duty sentence.
+
+        This operation relies on the `prefix` and `abbreviation` fields being
+        filled in on duty expressions and units, which are not supplied by the
+        TARIC3 XML by default.
+
+        Strings output by this annotation should be valid input to the DutySentenceParser.
```
NDevox, line 26: For docs like this where we're doing relatively niche things (not that `case/when` is niche in SQL circles, but more or less anything using DB functions in Django could be considered niche) I'd argue it's important to show what the outputted SQL is. For those who know SQL this makes it much clearer - the Django syntax at this level is really annoying.

Should be able to get the SQL by doing `print(MeasureCondition.objects.all().with_duty_sentence().query)`

measures/serializers.py
```diff
@@ -302,7 +302,7 @@ class Meta:
             "duty_expression",
             "duty_amount",
             "monetary_unit",
-            "condition_component_measurement",
+            "component_measurement",
```
NDevox, line 305: This will break some things in `measures/jinja2/taric/measure_condition_component.xml`

NDevox (changes requested):

measures/querysets.py
simonwo: I do like seeing the SQL; I do wish there was a way for it to not be hardcoded…

measures/serializers.py
simonwo: Anything more than [these things](https://github.com/uktrade/tamato/pull/206/files#diff-2dc156a9170b13d2813348c2037237c2e621e369866ac66ffa47cf0938ee7436)?

NDevox: Now I did see that as I scrolled past.

Didn't see the egg on my face though.

NDevox (approved):

PR #208, opened 2021-02-18
Only show current members for Geo Areas

NDevox:
Currently Geo Areas shows all members with all their versions in one. It should instead
only show the most current versions of each member.

This commit changes the query to only get current members.

### DIFF
--- a/geo_areas/jinja2/includes/geo_areas/tabs/memberships.jinja
+++ b/geo_areas/jinja2/includes/geo_areas/tabs/memberships.jinja
@@ -1,6 +1,6 @@
 <h2 class="govuk-heading-l">Memberships</h2>
 {% set table_rows = [] %}
-{% for membership in object.memberships.all() %}
+{% for membership in object.memberships.current() %}
   {{ table_rows.append([
     {"text": "{:%d %b %Y}".format(membership.valid_between.lower)},
     {"text": "{:%d %b %Y}".format(membership.valid_between.upper) if membership.valid_between.upper else "-"},


stuaxo (approved):
Approved, though we probably need test that covers this renders the current record at some point.

DIT-CJWaszczuk (approved):

andyhd (approved):

geo_areas/jinja2/includes/geo_areas/tabs/memberships.jinja
```diff
@@ -1,6 +1,6 @@
 <h2 class="govuk-heading-l">Memberships</h2>
 {% set table_rows = [] %}
-{% for membership in object.memberships.all() %}
+{% for membership in object.memberships.current() %}
```
simonwo, line 3: We will presumably need to come back and replace this with something workbasket-aware when we do editing?

andyhd: And we'll need to update it to something that honours the user selected date for time-machine functionality.

andyhd:
I think the missing migration is due to a change in django-treebeard? Maybe it's time to start specifying dependency versions?

stuaxo:
> I think the missing migration is due to a change in django-treebeard? Maybe it's time to start specifying dependency versions?

If we want to delay this, we can always just specify DTB for now ?

geo_areas/jinja2/includes/geo_areas/tabs/memberships.jinja
NDevox: yeah there is more work to do with filtering and how the data is handled on the backend.

For the future.

PR #209, opened 2021-02-18
Pin all requirements and add new migrations

NDevox:
As requirements update they are prone to breaking the
build pipeline. To stop this from happening this commit
pins all the current packages to their latest version.

It also introduces new migrations from the updated packages.


exporter/migrations/0002_auto_20210218_1132.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.1.6 on 2021-02-18 11:32
+
+import storages.backends.s3boto3
```
NDevox, line 3: @stuaxo this seemed to come as part of the update, but I think it may just be a missing migration?

stuaxo: It does look like it.

requirements.txt
```diff
@@ -1,53 +1,53 @@
...
+pytest-responses==0.4.0
+requests-oauthlib==1.3.0
+responses==0.12.1
+sentry-sdk==0.20.2
+Sphinx==3.5.1
 
 # This needs to be pinned below 1.26.0 for responses to work until
 # https://github.com/getsentry/responses/issues/354 is resolved.
-urllib3<1.26.0
+urllib3==1.25.11
```
andyhd, line 48: I think this issue might be fixed now

andyhd (approved):

exporter/migrations/0002_auto_20210218_1132.py
stuaxo: Do you want this migration in a separate PR ?

stuaxo (approved):
(approving, I didn't mean to hit "start a review)

exporter/migrations/0002_auto_20210218_1132.py
NDevox: Nah it's fine

PR #213, opened 2021-02-19
Remove description prefetch for Geo Areas

NDevox:
Prefetching helps save queries, however in some cases
it provides an incorrect description for Geographical
Areas. To remedy this prefetching has been removed from
Geographical Area views.

### DIFF
--- a/geo_areas/views.py
+++ b/geo_areas/views.py
@@ -19,7 +19,7 @@ class GeoAreaViewSet(viewsets.ReadOnlyModelViewSet):
 
 
 class GeographicalAreaList(TamatoListView):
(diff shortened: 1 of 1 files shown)


stuaxo (approved):
Interesting, it may be showing up an incorrect assumption somewhere else.

NDevox:
I imagine we'll be needing to apply similar solutions to this all over the place

stuaxo:
> I imagine we'll be needing to apply similar solutions to this all over the place

We really need more tests... :grimacing: including regression tests for this sort of thing, it's the kind of bug that's so easy to reintroduce.

PR #214, opened 2021-02-19
Move EnvelopeSerializer from importer to common.

stuaxo:


common/serializers.py
```diff
@@ -145,3 +153,122 @@ class Meta:
...
+        self.output = output
+        self.transaction_counter = transaction_counter
+        self.message_counter = message_counter
+        self.envelope_id = envelope_id
+        if max_envelope_size_in_mb is None:
+            self.max_envelope_size = None
+        else:
+            self.max_envelope_size = (
+                max_envelope_size_in_mb * 1024 * 1024
+            ) - self.envelope_end_size
```
andyhd, line 188: This can be negative if `max_envelope_size_in_mb` is zero (or `self.envelope_end_size` is greater than `max_envelope_size_in_mb * 1024 * 1024`)

andyhd (approved):

common/serializers.py
stuaxo: True, though `envelope_end_size` is the amount of bytes in this template:
 ```
</env:envelope>
```
So if you can't fit that, you have bigger problems (like not being able to fit the start as well).

It should probably raise a ValueError if it's < some arbitrary value.

`max_envelope_size_in_mb` .. is in MB, as I didn't want to break callers of this, but it should probably change to `max_envelope_size` and be in bytes for consistency.

stuaxo: Pushing change, with a test for this - arbitrary minimum I chose was 32k.. it could probably be 4k, but we're not optimising for small envelopes.

PR #216, opened 2021-02-23
TP-541: Prevent deletion of linked quotas.

NDevox:
Quota Associations, Quota Suspensions and Quota Blocking
Periods all demand that their linked quotas be live. I.e.
the linked Quota Definitions can not be deleted until the
relevant associations, suspensions and blocking periods
either refer to a different quota or are deleted themselves.

This commit introduces a business rule to check that any
quota being deleted cannot be referred to by an association,
suspension or blocking period.

### DIFF
--- a/quotas/business_rules.py
+++ b/quotas/business_rules.py
@@ -13,6 +13,27 @@ from quotas.validators import AdministrationMechanism
 from quotas.validators import SubQuotaType
 
 
+class PreventDeletingLinkedQuotaDefinitions(BusinessRule):
+    """A deleted Quota Definition must not be referred to by a non-deleted
+    relation."""
+
+    sid_prefix = "quota_definition__"
+
+    def get_relation_model(self, quota_definition):
+        raise NotImplementedError(
+            f"get_relation_model must be implemented on {self.__class__.__name__}.",
+        )
+
+    def validate(self, quota_definition):
+        related_model = self.get_relation_model(quota_definition)
+        if quota_definition.update_type == UpdateType.DELETE:
+            kwargs = {f"{self.sid_prefix}sid": quota_definition.sid}
+            if related_model.objects.current_as_of(
+                transaction=quota_definition.transaction,
+            ).filter(**kwargs):
+                raise self.violation(quota_definition)
+
+
 class ON1(UniqueIdentifyingFields):
     """Quota order number id + start date must be unique."""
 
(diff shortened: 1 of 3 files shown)


quotas/business_rules.py
```diff
@@ -205,6 +205,42 @@ def validate(self, quota_definition):
...
+class QuotaBlockingPeriodMustReferToANonDeletedQuotaDefinition(BusinessRule):
+    """A Quota Blocking Period must refer to a non-deleted Quota Definition."""
+
+    def validate(self, quota_definition):
+        QuotaBlocking = quota_definition.quotablocking_set.model
+        if quota_definition.update_type == UpdateType.DELETE:
+            if QuotaBlocking.objects.current_as_of(
+                transaction=quota_definition.transaction,
+            ).filter(quota_definition__sid=quota_definition.sid):
+                raise self.violation(quota_definition)
```
simonwo, line 241: Is it worth trying to introduce any genericism here on these three rules?

quotas/tests/test_business_rules.py
```diff
@@ -383,6 +383,57 @@ def test_prevent_quota_definition_deletion(unapproved_transaction, date_ranges):
...
+    Association."""
+    association = factories.QuotaAssociationFactory.create()
+    sub_quota = association.sub_quota
+    with pytest.raises(BusinessRuleViolation):
+        business_rules.QuotaAssociationMustReferToANonDeletedSubQuota().validate(
+            sub_quota.new_draft(
+                workbasket=unapproved_transaction.workbasket,
+                transaction=unapproved_transaction,
+                update_type=UpdateType.DELETE,
+            ),
```
simonwo, line 399: Would we also be able to do a positive test for a case that shouldn't raise? I know that we have other tests that should probably trigger if the rule was badly written, but there have been a few instances recently (particularly on ME32) where more positive tests would have helped.

quotas/business_rules.py
NDevox: Maybe, I was wondering this as well but it was easy to do them separate. I'll have a quick gander at it.

simonwo (approved):

quotas/business_rules.py
NDevox: Done

quotas/tests/test_business_rules.py
NDevox: Good point, done.

stuaxo (approved):

PR #217, opened 2021-02-24
TP-610 Add measures search filter

DIT-CJWaszczuk:
## Summary 
Add measures list view
## Notes 
- Had to pretty much build the form structure in the form file.
- Couldn't apply a class to the wrapper of some of the form fields, so had to do a janky `quarters` class approach to get the same style. 
## Screenshots 
image

### DIFF
--- a/footnotes/filters.py
+++ b/footnotes/filters.py
@@ -16,8 +16,8 @@ from footnotes.validators import FOOTNOTE_ID_PATTERN
 from footnotes.validators import FOOTNOTE_TYPE_ID_PATTERN
 
 COMBINED_FOOTNOTE_AND_TYPE_ID = re.compile(
-    r"^(?P<footnote_type_id>" + FOOTNOTE_TYPE_ID_PATTERN + ")"
-    r"(?P<footnote_id>" + FOOTNOTE_ID_PATTERN + ")$",
+    r"(?P<footnote_type_id>" + FOOTNOTE_TYPE_ID_PATTERN + ")"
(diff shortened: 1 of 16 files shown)


stuaxo (approved):

DIT-CJWaszczuk:
Just missing the conditions table field, but it should be reviewable now.

stuaxo (approved):
LGTM

stuaxo (approved):

footnotes/filters.py
```diff
@@ -16,8 +16,8 @@
 from footnotes.validators import FOOTNOTE_TYPE_ID_PATTERN
 
 COMBINED_FOOTNOTE_AND_TYPE_ID = re.compile(
-    r"^(?P<footnote_type_id>" + FOOTNOTE_TYPE_ID_PATTERN + ")"
-    r"(?P<footnote_id>" + FOOTNOTE_ID_PATTERN + ")$",
+    r"(?P<footnote_type_id>" + FOOTNOTE_TYPE_ID_PATTERN + ")"
+    r"(?P<footnote_id>" + FOOTNOTE_ID_PATTERN + ")",
```
andyhd, line 20: It is generally better to use f-strings than the `+` operator to concatenate strings. It won't make much difference here, though.

andyhd (approved):

footnotes/filters.py
DIT-CJWaszczuk: think that may have been to comply with `re.compile`, but we can change it up in a different PR.

PR #218, opened 2021-02-24
TP-643: Update current Tracked models.

NDevox:
Currently when CDS accepts data from TaMaTo the relevant
Tracked Models aren't flagged as current. This only happens
in the niche scenario when they are loaded into the system
pre-approved (which should only happens on initial seed
loads).

This commit adds an update to all the tracked models once
a workbasket is confirmed by CDS, flagging all the confirmed
models as the most current.


simonwo:
Code looks fine. So we seem to be settling on the idea that `current` returns "the current version according to the published transactions" of models, which is fine definition. It's certainly easy to implement that way because you don't need to worry about CDS errors or reverting stuff. 

But I think my concern is around how much utility this definition actually gives us and how well it matches what users expect to see. This might be the right time to answer the question of what we would actually expect this to be used. E.g., what queryset is appropriate for all of the situations in this table:

| # | Context                                              | Queryset        |
|---|------------------------------------------------------|-----------------|
| 1 | User interface, user not in the ops team             | ?               |
| 2 | User interface, user in ops team, no workbasket open | ?               |
| 3 | User interface, user in ops team, workbasket open    | `current_as_of(workbasket.transactions.last())` |
| 4 | Activity stream                                      | `current()`       |

Do we expect the question marks to be showing only published data? Or do we expect the question marks to be showing approved data as well? 

Certainly for 2, I expect it will be baffling to be looking at `current` data and not being able to see a change that has recently been approved, and then to open a workbasket and suddenly the data that was not even in the workbasket has changed. It would be like reading code in git and the moment you go to open it, the rest of the codebase updates to show more recent stuff. So I think from the principle of least astonishment we need to regard the data that has been approved as the system's view of what is current. I would hazard that this is the way Ops think about it too – they have put stuff "in the system" and whether or not it has been approved by CDS yet is something of a downstream concern.

For 1 there is more of an open question as these users won't be doing any editing (if they even exist at all) but it might still be baffling for e.g. an Ops person to send a link to someone else and for the content at that link to display differently to them. 

Really I think the UI needs to be consistent about what it shows – i.e. 1, 2 and 3 should all be based off the same point, with 3 only adding on the extra stuff that is in the workbasket. 

So then we're only really using `current` by this definition as part of the activity stream endpoint. Is it worth it? As I'm also worried that it's very easy for a developer without much experience on the system to assume that `current` will always show them what they want to see, when in fact it's almost never what they should be using. Maybe that's just a naming issue. Or maybe we should have the optimised `current` as regarding approved data, and then we'd have to deal with all the issues about reverting changes due to CDS errors.

NDevox:
Yeah it's good food for thought.

I do find showing non-accepted data as arguably quite weird (even if it is the current expectation) as the data in flux, even approved data is entirely subject to change based on a rejection by HMRC. So what the user is looking at isn't the actual tariff, and may not even be the future data of the tariff. It's just an assumed possibility.

I wonder if it would not be more useful to give some kind of visual sign of what stage some data is on? e.g. we show the current data but there is a visual flag to say it is currently being edited elsewhere, or is on its way to HMRC (fitting this into the current designs may be a bit of an ask I guess... Previously I would have done something similar with colour co-ordination e.g something with an orange border is currently being edited etc.). This could still show current data but let you know that some of the "current" data is liable to change.

On the other hand if the problem is simply that we really want to be showing the approval state and not the accepted state, it would be trivial to move the code to the approval state and add another transition on rejection which cycles back the versions.

On the what people expect point, is it not fair to have some expectation that ops know that they are changing tariff data? I'm not sure why data that isn't in the workbasket they are opening would change - `current_as_of` only shows current or in specific workbasket changes. But I would lean again on some sort of sign that the data they are picking may already be changing elsewhere. For developers specifically - this is why we have code reviews.

stuaxo:
Some indication is good, though if you are in a view where you can see data in this sort of "draft" state, how do you see data in that ordinary users would see ?

Should this be visible in the same views, or should there be other tabs in the various UIs to view draft vs published (using CMS terminology here for want of something better) ?

stuaxo (approved):

stuaxo (approved):

simonwo (approved):

PR #219, opened 2021-02-24
Add measure condition(s) string generation

andyhd:
When measures are listed in the UI, one of the column items is a list of conditions
associated with the measure.

This commit adds convenience methods and properties to the Measure model and associated
models to easily construct these lists.

This code was translated from the PHP prototype, which may not be complete or correct.


measures/models.py
```diff
@@ -626,6 +687,96 @@ class MeasureCondition(TrackedModel):
...
+
+        for mcc in components:
+            measures.add(mcc.condition.dependent_measure.sid)
+            measure_types.add(mcc.condition.dependent_measure.measure_type.sid)
+            if mcc.condition.dependent_measure.additional_code:
+                additional_codes.add(
+                    mcc.condition.dependent_measure.additional_code.sid,
+                )
+
+        print(measures, measure_types, additional_codes)
```
stuaxo, line 763: print

stuaxo (approved):
OK, cool - just remove the print statement I think ?

andyhd:
Django-treebeard reverted a change to the depth field on MP node models - hence the irrelevant migration.

simonwo (changes requested):
I really appreciate the effort that has gone into translating this, but perhaps the PHP wasn’t as strong as it could be here! It looks like it used a bunch of hard-coded stuff that we have made space in the db for? So we instead need to solve the problem of how we get that new stuff that isn’t in TARIC XML in to the db at the right time (e.g. after the xml load but before use…)

I’m also wondering to what extent it would be appropriate to do something like the `DutySentenceMixin` here and shove a lot of this out to the DB. The reason being that this stuff will invoke a lot of queries in a loop and might be slow, whereas the overhead observed from the db version is observed to be less (for comparison, the snapshot function in the PHP takes 2 hours to output all 250,000 measures but the sql-only version can run in 3-4 minutes). Obviously we’re not displaying anything like that number in a page but if we get to some sort of bulk download feature *cough cough* it might become necessary. Would be good to hear thoughts on that!

measures/models.py
```diff
@@ -78,6 +79,43 @@ class MeasurementUnit(TrackedModel, ValidityMixin):
...
+            "MIL": "1,000 items",
+            "MTK": "m2",
+            "MTQ": "m3",
+            "MTR": "m",
+            "MWH": "1,000 kWh",
+            "NCL": "ce/el",
+            "NPR": "pa",
+            "TJO": "TJ",
+            "TNE": "tonne",
+        }.get(self.code, self.code)
```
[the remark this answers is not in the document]
andyhd: I knew that. I was just doing a direct port of the PHP to get something ready for Chris as soon as possible to avoid blocking the Measure UI. I can just remove this property and use `abbreviation` instead.

measures/validators.py
```diff
@@ -58,6 +58,71 @@ class DutyExpressionId(models.IntegerChoices):
     DE99 = 99, "Supplementary unit"
 
 
+class MeasureConditionCode(models.TextChoices):
```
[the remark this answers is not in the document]
andyhd: I didn't check that these were the same data, sorry.

measures/models.py
```diff
@@ -668,6 +817,38 @@ class MeasureConditionComponent(TrackedModel):
...
+    @property
+    def measurement_string(self) -> str:
+        out = ""
+        measurement = self.component_measurement
+
+        if measurement:
+            if measurement.measurement_unit:
+                out = measurement.measurement_unit.label
+            if measurement.measurement_unit_qualifier:
+                out = f"{out} / {measurement.measurement_unit_qualifier.code}"
```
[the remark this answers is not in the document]
andyhd: Yep, I just didn't have any examples of them to test. I'll replace this.

andyhd:
I am aware that this code is unperformant - I was just trying to get something that works without really understanding what or why the PHP does what it does. Make it work, then make it right, then make it fast 😄

measures/models.py
```diff
@@ -626,6 +687,94 @@ class MeasureCondition(TrackedModel):
...
+            if self.condition_measurement:
+                if self.condition_measurement.measurement_unit:
+                    out.append(self.condition_measurement.measurement_unit.code)
+
+                if self.condition_measurement.measurement_unit_qualifier:
+                    out.append(
+                        self.condition_measurement.measurement_unit_qualifier.code,
+                    )
+
+        return " ".join(out)
```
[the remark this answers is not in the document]
andyhd: Unfortunately, `DutySentenceMixin` doesn't work with `MeasureConditionComponent` because it assumes the measurement column name is `component_measurement` when it is actually `condition_component_measurement`. It doesn't work with `MeasureCondition` either, for the same reason.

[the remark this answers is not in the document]
andyhd: Never mind, something weird is going on - when I directly call `MeasureCondition.objects.with_duty_sentence().first().duty_sentence` I get an error about a column `component_measurement_id` does not exist, but when I do the same thing in a test, it works.

DIT-CJWaszczuk:
Looks good to me from a FE perspective. Happy to go with it given that Simon is happy.

DIT-CJWaszczuk (approved):

measures/models.py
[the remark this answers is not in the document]
andyhd: This string wasn't already handled by `DutySentenceMixin`, but I've added a similar queryset method to generate it.

measures/models.py
```diff
@@ -626,6 +687,94 @@ class MeasureCondition(TrackedModel):
...
+    def is_certificate_required(self):
+        return self.condition_code.code in (
+            validators.MeasureConditionCode.A,
+            validators.MeasureConditionCode.B,
+            validators.MeasureConditionCode.C,
+            validators.MeasureConditionCode.H,
+            validators.MeasureConditionCode.Q,
+            validators.MeasureConditionCode.Y,
+            validators.MeasureConditionCode.Z,
+        )
```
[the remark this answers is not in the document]
andyhd: I've hard-coded this for now - let's revisit if we need it somewhere else.

measures/models.py
```diff
@@ -626,6 +626,66 @@ class MeasureCondition(TrackedModel):
...
+
+        if len(measures) == len(measure_types) == len(additional_codes) == 1:
+            out.append(self.duty_sentence)
+
+        else:
+            if len(measure_types) > 1:
+                out.append(self.duty_sentence)
+
+            elif len(additional_codes) > 1:
+                out.append(self.duty_sentence)
```
simonwo, line 685: This seems to only output `self.duty_sentence` if any of these three conditions are met… so does it need to append to `out` or just do `return self.duty_sentence if … else ""`?

measures/migrations/0007_add_duty_expression_prefixes_and_measurement_unit_abbrevs.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("measures", "0006_auto_20210225_1603"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
```
simonwo, line 12: Nice 👍 We may need to work out at some point how to run this "again" if we start from a blank database and then import all of the XML, as these expressions won't exist yet. But no worries for now.

simonwo (approved):
Looks good to me! 😎

measures/models.py
andyhd: This is a weird bit of the PHP code. I'm not sure if we want to keep it. Here's the original:
```php
if (($this->measure_count == 1) && ($this->measure_type_count == 1) && ($this->additional_code_count == 1)) {
    foreach ($this->duty_list as $d) {
        $this->combined_duty .= $d->duty_string . " ";
    }
} else {
    if ($this->measure_type_count > 1) {
        if (in_array("105", $measure_type_list_unique)) {
            foreach ($this->duty_list as $d) {
                if ($d->measure_type_id == "105") {
                    $this->combined_duty .= $d->duty_string . " ";
                }
            }
        }
    } elseif ($this->additional_code_count > 1) {
        if (in_array("500", $additional_code_list_unique)) {
            foreach ($this->duty_list as $d) {
                if ($d->additional_code_id == "500") {
                    $this->combined_duty .= $d->duty_string . " ";
                }
            }
        }
        if (in_array("500", $additional_code_list_unique)) {
            foreach ($this->duty_list as $d) {
                if ($d->additional_code_id == "500") {
                    $this->combined_duty .= $d->duty_string . " ";
                }
            }
        }
    }
}
```
What do you think?

andyhd: It seems to have a redundant `if (in_array("500", $addiditonal_code_list_unique))`, unless that is meant to appear twice. 🤷

andyhd: Oh, and yes, I should clean up the logic.

simonwo: I've just gone spelunking into the PHP to understand this and it's very odd indeed. For a start, it loads `duty_list` in one a class method but then seems to rely on attributes that are only loaded by `get_measure`… whatever. But then also the things it checks (`measure_type_id` and `additional_code_id`) don't seem to get populated anywhere.

So in short I've got no idea what this is meant to do or why it attached special meaning to measure type 105, which is to do with authorised use. 

I would advocate for dropping this logic entirely and just outputting the duty sentence if it isn't empty.

simonwo: As ever, our users have no legacy expectations and we can iterate on this later.

stuaxo:
Ugh, good work diving into the PHP !

PR #220, opened 2021-02-25
TP-367 - Split envelopes into separate files.

stuaxo:
Split envelopes into separate files.

- envelopes are split by size
- tracked models are grouped by transaction.


common/serializers.py
```diff
@@ -182,15 +185,25 @@ def __init__(
...
             raise ValueError(
                 f"Max envelope size {max_envelope_size} is too small, it should be at least {EnvelopeSerializer.MIN_ENVELOPE_SIZE}."
             )
-        self.envelope_size = 0
         self.format = format
         self.newline = newline
 
+    @cached_property
+    def envelope_start_size(self):
+        """Size in bytes of envelope end.
```
simonwo, line 202: Comment seems to not be consistent with function name

common/serializers.py
```diff
@@ -273,3 +291,89 @@ def render_transaction(self, models: List[TrackedModel]) -> None:
...
+        **kwargs,
+    ) -> None:
+        self.output_constructor = output_constructor
+        EnvelopeSerializer.__init__(self, self.output_constructor(), *args, **kwargs)
+
+    def start_next_envelope(self):
+        self.output = self.output_constructor()
+        super().start_next_envelope()
+
+    def chunked_render_transaction(self, model_groups):
```
simonwo, line 324: Took me a few minutes to get my head around this, but got there in the end!

simonwo (commented):
Looking good so far!

common/serializers.py
stuaxo: Good catch.

common/serializers.py
stuaxo: It took a couple of weeks to arrive here, hopefully it is simpler now this has changed to be about Transactions and not "model_groups".

common/tests/test_serializers.py
```diff
@@ -36,5 +46,65 @@ def test_serializer_min_envelope_size_is_large_enough():
...
+
+    output = io.BytesIO()
+    with EnvelopeSerializer(output, random.randint(2, 9999)) as env:
+        env.render_transaction(approved_workbasket.tracked_models.all())
+
+    output_xml = etree.XML(output.getvalue())
+    output_record_codes = [*taric_xml_record_codes(output_xml)]
+
+    # More items than models are output, as models for descriptions aren't kept.
+    assert expected_item_record_codes.issubset(output_record_codes)
```
simonwo, line 85: Seems sensible.

exporter/management/commands/dump_transactions.py
```diff
@@ -47,25 +53,53 @@ def get_output_file(self, filename):
...
+                        self.stdout.write(f"{i:>6}| ", line)
+                    raise
+                else:
+                    total_transactions = sum(
+                        output_transaction.tracked_models.count()
+                        for output_transaction in output_transactions
+                    )
+                    self.stdout.write(
+                        f"{output.name } ✅  XML valid.  {total_transactions} transactions in {output.tell()} bytes."
+                    )
```
simonwo, line 103: A lot of this seems to be shared with the exporter task. Can we DRY that up a bit somehow? And perhaps use a `logger` instead of writing directly to `stdout`?

exporter/management/commands/dump_transactions.py
```diff
@@ -47,25 +53,53 @@ def get_output_file(self, filename):
...
+        ):
+            if not output_transactions:
+                self.stdout.write(f"⚠️ {output.name} is empty !")
+                errors = True
+            else:
+                try:
+                    output.seek(0, os.SEEK_SET)
+                    validate_envelope(output)
+                except:
+                    output.seek(0, os.SEEK_SET)
```
simonwo, line 91: This looks slightly dangerous, in that if the `seek` fails or just the act of reading the envelope fails, we immediately try and do it agian. Would a more specific exception catch here be appropriate?

exporter/tasks.py
```diff
@@ -13,7 +13,12 @@
 from common.tests.util import TaricDataAssertionError
 from common.tests.util import validate_taric_xml_record_order
 from exporter.models import Upload
-from common.serializers import EnvelopeSerializer
+from common.serializers import (
+    EnvelopeSerializer,
+    MultiFileEnvelopeTransactionSerializer,
+    TransactionsTooLarge,
+)
```
simonwo, line 20: Again, looks like the pre-commit hooks need to be run over this?

exporter/tasks.py
```diff
@@ -68,54 +73,72 @@ def upload_workbaskets() -> Tuple[bool, str]:
...
         msg = f"Nothing to upload:  {workbaskets.count()} Workbaskets READY_FOR_EXPORT but none contain any transactions."
         logger.info(msg)
         return False, msg
 
-    # Write to a temporary file so that output can be validated afterwards.
-    with tempfile.TemporaryFile() as envelope_file:
-        envelope = Envelope.new_envelope()
+    # # Write files to a temporary, so they can all be validated before uploading.
+    with tempfile.TemporaryDirectory(prefix="dit-tamato_") as tmpdirname:
+        output_file_constructor = dit_file_generator(tmpdirname)
```
simonwo, line 86: So do we write these out and then read them back in? Not saying that's bad, but would we be better off with just a `StringIO`?

exporter/util.py
```diff
@@ -0,0 +1,22 @@
...
+
+def dit_filename_generator(start=1):
+    """Generate incrementing DIT filenames."""
+    counter = count(start)
+    while True:
+        yield f"DIT{next(counter):06}.xml"
+
+
+def dit_file_generator(directory: str):
+    """Generate iterating DIT named files in the supplied directory for writing."""
```
simonwo, line 13: Is this just for temporary files? If so, why go to all the bother with prefixing them with `DIT`, and not just let Python pick a random file? 

And if it's not just for temporary files, then how do we actually name them correctly, as it looks like you can't seed the `dit_filename_generator` with the current value when calling the `dit_file_generator`?

simonwo (commented):
We seem to have functionality splattered a bit over `taric`, `exporter` and `common` modules. Are we clear on why that is? In particular I'm a bit confused about what the difference is between `taric` and `exporter`.

exporter/tasks.py
```diff
@@ -68,54 +73,72 @@ def upload_workbaskets() -> Tuple[bool, str]:
...
 
-        with EnvelopeSerializer(
-            envelope_file, envelope_id=int(envelope.envelope_id)
-        ) as env:
-            env.render_transaction(tracked_models)
+        serializer = MultiFileEnvelopeTransactionSerializer(
+            output_file_constructor, 1, max_envelope_size=39 * 1024 * 1024
```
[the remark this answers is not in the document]
stuaxo: That would be a bug, good catch.

exporter/tasks.py
stuaxo: I did this when coding this, however, part of EnvelopeSerializers idea, which this builds on, is to be memory efficient by streaming everything.

If we use buffers, then everything will sit in memory, negating the point of streaming everything in the first place.

This would would mean when we choose VMs we would need RAM + swapsize as large a all the envelopes we might want to serialize at any time.

I don't think temporary files in Linux won't end up written to disk under most circumstances so we're good here.

stuaxo: We're really only using seek() around a file that will probably never leave memory.. but we will still run on a small server if we need to write out a large backlog of envelopes.

exporter/management/commands/dump_transactions.py
stuaxo: I'll have a better look at this, have to admit to transplanting this dump out of my notebook a bit late in the day :)

exporter/tasks.py
```diff
@@ -68,54 +73,72 @@ def upload_workbaskets() -> Tuple[bool, str]:
...
 
-        envelope_file.seek(os.SEEK_SET)
-        upload.checksum = md5(envelope_file.read()).hexdigest()
+            envelope_file.seek(0, os.SEEK_SET)
+            upload.checksum = md5(envelope_file.read()).hexdigest()
 
-        upload.file.save(upload.filename, content_file)
-        upload.notify_hmrc()  # sets notification_sent
+            upload.file.save(upload.filename, content_file)
+            upload.notify_hmrc()  # sets notification_sent
```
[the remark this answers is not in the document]
stuaxo: It does actually do the notifying.

It makes sense to split this up when we add retrying, since that will involve some splitting up of this task + not add more complexity to this PR.

exporter/management/commands/dump_transactions.py
stuaxo: Sure, the output may end up a little more terse though.

exporter/tasks.py
simonwo: Yes, fair enough.

common/tests/test_serializers.py
```diff
@@ -1,18 +1,165 @@
...
+from lxml import etree
 
 import pytest
+import random
 
-from common.serializers import EnvelopeSerializer
+from common.serializers import (
+    EnvelopeSerializer,
+    MultiFileEnvelopeTransactionSerializer,
+)
```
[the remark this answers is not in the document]
stuaxo: Took me to realise until reading the next comment this is about git pre commit hooks and not some database thing !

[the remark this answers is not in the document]
stuaxo: TIL that `reorder-python-imports **.py` doesn't do what I thought it did, and that my pre-commit wasn't setup.

exporter/tasks.py
stuaxo: Sorted now

exporter/management/commands/dump_transactions.py
stuaxo: Removed the seek and the extra output - while it was useful in the notebook, probably not so much here.

exporter/util.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def generate_files():
+        filenames = dit_filename_generator()
+
+        while True:
+            yield open(Path(directory) / next(filenames), "wb+")
```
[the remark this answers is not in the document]
stuaxo: Sorted

stuaxo:
> We seem to have functionality splattered a bit over `taric`, `exporter` and `common` modules. Are we clear on why that is? In particular I'm a bit confused about what the difference is between `taric` and `exporter`.

Maybe this could be revisited..  I did ask if I should include more things in exporter at the beginning ACAICT:

`common`
 code changes to envelope serialization to allow it to be extended [1].

`exporter`
 celery and mangement tasks to orchestrate everything as well as output files.
multi envelope serializer: that splits output, building on envelope serializer [1]

`taric`
 envelope data model.

`hmrc_sdes`
  code that calls the hmrc api

exporter/serializers.py
```diff
@@ -0,0 +1,127 @@
...
+from lxml import etree
+
+from common.serializers import EnvelopeSerializer
+from common.serializers import TaricDataAssertionError
+from common.serializers import validate_envelope
+
+logger = logging.getLogger(__name__)
+
+RenderedTransaction = namedtuple(
+    "TransactionResult",
```
andyhd, line 16: The namedtuple has the name `TransactionResult` but is assigned to (and only ever referred to as) `RenderedTransaction`. Is this intended?

stuaxo: No, it really isn't thanks for the extra eyes on this.

common/tests/test_serializers.py
```diff
@@ -1,18 +1,162 @@
...
+    assert rendered_envelope.transactions == [tx2]
+    assert output_record_codes.issuperset(tx2_expected_record_codes)
+
+    rendered_envelope = next(split_output)
+    output_xml = etree.XML(rendered_envelope.output.getvalue())
+    output_record_codes = {*taric_xml_record_codes(output_xml)}
+
+    assert rendered_envelope.output == outputs[2]
+    assert rendered_envelope.transactions == [tx3]
+    assert output_record_codes.issuperset(tx3_expected_record_codes)
```
andyhd, line 162: A bit repetitious. You could do something like:
```python
for tx, expected_record_codes in [(tx1, tx1_expected_record_codes), (tx2, ...)]:
```

It makes me think this test should be parametrized.

exporter/management/commands/dump_transactions.py
```diff
@@ -1,71 +1,89 @@
...
+        serializer = MultiFileEnvelopeTransactionSerializer(
+            output_file_constructor,
+            envelope_id=envelope_id,
+            max_envelope_size=settings.EXPORTER_MAXIMUM_ENVELOPE_SIZE,
+        )
+        errors = False
+        for rendered_envelope in serializer.split_render_transactions(transactions):
+            envelope_file = rendered_envelope.output
+            if not rendered_envelope.transactions:
+                self.stdout.write(f"{envelope_file.name} ⚠  is empty !")
```
andyhd, line 75: How is this going to render on a non-Unicode terminal? It renders badly here in Github.

exporter/serializers.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    Streaming EnvelopeSerializer that with multiple streams.
+
+    On reaching a size boundary, the current envelope is completed, the next
+    stream from outputs is selected new envelope output starts.
+    """
+
+    def __init__(
+        self,
+        output_constructor: callable,
```
andyhd, line 31: I think this is a typo - should it be `Callable`? I'm not sure callable objects are being passed in this position, though. They are instances of `BytesIO` in the test.

exporter/serializers.py
```diff
@@ -0,0 +1,127 @@
...
+        envelope_id=1,
+        *args,
+        **kwargs,
+    ) -> None:
+        self.output_constructor = output_constructor
+        EnvelopeSerializer.__init__(
+            self, self.output_constructor(), envelope_id=envelope_id, *args, **kwargs
```
andyhd, line 38: `envelope_id` doesn't need to be a keyword argument here - could it potentially clash with `kwargs`?

exporter/tasks.py
```diff
@@ -2,120 +2,142 @@
...
-            raise
+@atomic
+def upload_and_create_envelopes(
+    workbaskets: QuerySet,
+    rendered_envelopes: Sequence[RenderedTransaction],
+    first_envelope_id,
+) -> Dict[int, RenderedTransaction]:
+    user_messages = (
+        {}
+    )  # {envelope_id: message} User messages can be returned to the caller of the task.
```
andyhd, line 42: This commenting style is odd, couldn't this be
```python
# User messages can be returned to the caller of the task (is this comment necessary?)
user_messages: dict[int, RenderedTransaction] = {}
```

exporter/tasks.py
```diff
@@ -2,120 +2,142 @@
...
+            logger.info("HMRC notification disabled.")
+        else:
+            logger.info("Notify HMRC of upload, %s", upload.filename)
+            upload.notify_hmrc()  # sets notification_sent
+
+        logger.info("Workbasket sent to CDS")
+        workbaskets.update(status=WorkflowStatus.SENT_TO_CDS)
+
+        logger.debug("Uploaded: %s", upload.filename)
+        user_messages[envelope.envelope_id] = f"Uploaded {upload.filename}"
```
andyhd, line 84: Shouldn't the message be a `RenderedTransaction`?

exporter/tasks.py
```diff
@@ -2,120 +2,142 @@
...
+        invalid_envelopes = validate_rendered_envelopes(rendered_envelopes)
+        error_messages = [
+            f"Envelope {envelope_id:06} was invalid {exception}"
+            for envelope_id, exception in invalid_envelopes.items()
         ]
 
-        EnvelopeTransaction.objects.bulk_create(envelope_transactions)
-        envelope.save()
+        if error_messages:
+            return False, error_messages
```
andyhd, line 134: `error_messages` is a `list[str]`, but the return type wants a `dict[str, str]`, or `dict[None, str]`

exporter/tasks.py
```diff
@@ -2,120 +2,142 @@
...
-        upload.checksum = md5(envelope_file.read()).hexdigest()
-
-        upload.file.save(upload.filename, content_file)
-        upload.notify_hmrc()  # sets notification_sent
-
-        workbaskets.update(status=WorkflowStatus.SENT_TO_CDS)
-
-        logger.debug("Uploaded: %s", upload.filename)
-        return True, upload.filename
+        return True, user_messages
```
andyhd, line 143: `user_messages` is a `dict[int, RenderedTransaction]`, not a `dict[str, str]`

taric/models.py
```diff
@@ -22,7 +22,9 @@ def envelopes_by_year(self, year: Optional[int] = None):
         else:
             now = date(year, 1, 1)
 
-        return self.order_by("envelope_id").filter(envelope_id__startswith=f"{now:%y}")
+        return self.filter(envelope_id__regex=f"{now:%y}" + r"\d{4}").order_by(
```
andyhd, line 25: This should be `fr"{now:%y}\d{{4}}"`

andyhd (approved):
Some type errors, but nothing blocking 👍

andyhd:
@simonwo the `taric` app is my fault and I meant for it to be deleted, but doing so would have messed up some migrations. The only thing it does is contain the `Envelope` model, which is shared by the importer and the exporter. It possibly makes more sense to have it in `common`.

common/tests/test_serializers.py
```diff
@@ -1,18 +1,162 @@
 import io
+import random
 
 import pytest
+from lxml import etree
+from pytest_django.asserts import (
+    assertQuerysetEqual,
+)  # noqa pycharm can't see this as it's autogenerated.
```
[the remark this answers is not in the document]
stuaxo: I think that's from Black, but I'll see if I can persuade it to go away.

taric/models.py
stuaxo: TIL `fr""`

exporter/tasks.py
stuaxo: The OG type was wrong, but will fix this throughout.    The real preference is for simple types like dict[Union[int, None], str] so as these have to travel over celery.

exporter/tasks.py
stuaxo: No, but my typespec neeed fixing.

common/serializers.py
```diff
@@ -273,3 +301,56 @@ def render_transaction(self, models: List[TrackedModel]) -> None:
...
+            last_code = full_code
+
+
+def validate_envelope(envelope_file, skip_declaration=False):
+    """
+    Validate envelope content for XML issues and data order issues.
+
+    raises DocumentInvalid | TaricDataAssertionError
+    """
+    XML_DECLARATION = '<?xml version="1.0" encoding="UTF-8"?>\n'
```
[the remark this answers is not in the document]
stuaxo: For now, no - so will YAGNI this + maybe add a comment.

common/tests/test_serializers.py
stuaxo: Will revisit this in a following ticket, it looks yucky enough that I will be compelled to change it.

I don't like using for loops in the test itself, and this needs a bit more thought than I have energy for.

Probably can structure the data so that  one of the assertions against list content can work.

common/tests/test_serializers.py
```diff
@@ -1,18 +1,162 @@
...
+
+    assert EnvelopeSerializer.MIN_ENVELOPE_SIZE > templates_size + min_content_size
+
+
+def test_envelope_serializer_outputs_expected_items(approved_workbasket):
+    """EnvelopeSerializer should output all the models passed to it
+    and generated records for descriptions."""
+    tx = approved_workbasket.new_transaction(
+        composite_key=f"{approved_workbasket.pk}-tx1"
+    )  # Transaction context manager is not used as it creates phantom workbaskets and transactions which will cause XSD failures later.
```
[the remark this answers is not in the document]
stuaxo: Done, and reorganized.

PR #221, opened 2021-02-26
TP-571: ADR-0012: Ordering of transactions.

simonwo:
An ADR that was previously a Google doc, talking about how we order transactions.


stuaxo:
Does everything we need to do fit into a transaction ?

Some of the example operations take more than one transaction to achieve, does that mean the business rules are still valid if only one of those transactions was acted on ?

simonwo:
> 
> 
> Does everything we need to do fit into a transaction ?
> 
> Some of the example operations take more than one transaction to achieve, does that mean the business rules are still valid if only one of those transactions was acted on, or do they have to go a via a state where they are invalid ?

No, all transactions must be valid. If an operation takes multiple transactions to achieve, the state at the end of each must be valid as well. This is the thing that makes it so tricky – in the quota order number update example, you can't issue any ordering of transactions (using only `UPDATE`s) that is valid, so you have to go via anothe rmechanism (`DELETE` then `CREATE`).

docs/adr/0012-ordering-of-tariff-transactions.rst
```diff
@@ -0,0 +1,274 @@
...
+referencing a missing order number and would no longer be valid. If one was to
+update the measures first, they would be now referencing an order number that
+doesn’t exist yet and they wouldn’t be valid.
+
+The only solution here is to first delete all of the measures referencing the
+order number, issue a transaction changing the order number, and then recreate
+all the measures referencing the new number. This is one of many examples where
+the “single record per transaction” model breaks down and we are forced to
+delete measures to cope with an update to something they reference (there are
+similar problems with additional codes and goods nomenclature).
```
andyhd, line 187: Could you create a new order number and create new copies of the definitions, origins and exclusions?

andyhd: Nm, that would still be creates and not updates.

stuaxo (approved):

stuaxo:
It feels analogous to trying to find a route from one part of a map to another (and probably similar in the explosion of combinations).

I wonder if it's easier starting from every  update being  DELETE and CREATE, then trying to turn some into updates or going the other way and attempting to find a combination where the business rules are satisfied by converting UPDATES into DELETE and CREATES?

andyhd (approved):
Great stuff, very clear

simonwo:
> I wonder if it's easier starting from every update being DELETE and CREATE, then trying to turn some into updates or going the other way and attempting to find a combination where the business rules are satisfied by converting UPDATES into DELETE and CREATES?

That's interesting, I will think on that further.

It's ultimately a finite space so really it should be possible to define all the possible combinations. It's just hard to get right and will take a long time!

stuaxo:
I'm reminded of calendar scheduling algorithms (during the change from one state to another, one thing blocks another from being in a valid state).

There's an interesting graph based approach to a scheduler algorithm
https://stackoverflow.com/questions/3235306/calendar-scheduler-algorithm

Maybe some of these approaches might help.

PR #224, opened 2021-03-02
TP-640: Replace inappropriate parent nodes.

NDevox:
In some cases new Goods Nomenclature Indents are inserted in between
old indents and their children. When this happens it is important to
update the tree by migrating these children under the new indent.

This commit introduces an addition to the GoodsNomenclatureIndentHandler
post_save method, finding inappropriate parents of possible child nodes
and replacing them with the new indent.

### DIFF
--- a/commodities/import_handlers.py
+++ b/commodities/import_handlers.py
@@ -7,6 +7,7 @@ from typing import Optional
 
 from dateutil.relativedelta import relativedelta
 from django.db import transaction
+from django.db.models import Q
 
 from commodities import import_parsers as parsers
 from commodities import models
@@ -284,7 +285,6 @@ class GoodsNomenclatureIndentHandler(BaseHandler):
             )
 
             node_data["valid_between"] = TaricDateRange(indent_start, indent_end)
-
             next_parent.add_child(**node_data)
 
             start_date = (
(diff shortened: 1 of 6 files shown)


commodities/import_handlers.py
```diff
@@ -292,26 +293,179 @@ def save(self, data: dict):
...
+                        # No need to split the child node further, it fits completely
+                        # into the new parent node.
+                        new_parent = new_parents.get()
+                        previous_node = new_parent.add_child(**node_data)
+                    else:
+                        # The new child node does not fit neatly into a single parent node.
+                        # Split the child node so it fits under each new parent node.
+                        start_date = node_data["valid_between"].lower
+                        end_date = node_data["valid_between"].upper
+                        for parent in new_parents:
```
andyhd, line 426: I haven't got into the detail of this method yet, but we're 3 loops deep at this point - maybe it needs splitting into several methods/functions?

NDevox: Yep it's a reasonable request.

My only thought was that it wouldn't really replace any functionality and they'll all be one time use anyway.

stuaxo: This looks like it should be atomic.

As it's quite large, I can't work out if splitting it into a section that collects the operations to do and a part that acts on them would help here.

NDevox: It most certainly should be atomic, good point.

NDevox:
Tests fixed.

But realised there's a glaring hole in the logic where I'm not updating the rest of the tree below the children.

simonwo:
Brain is gearing up to review this one… (it takes a couple o' days…)

commodities/import_handlers.py
```diff
@@ -284,34 +285,190 @@ def save(self, data: dict):
...
+            for path in models.GoodsNomenclatureIndentNode.objects.filter(
+                indent__in=replaceable_parents,
+            ).values_list("path", flat=True):
+                has_parent_path = has_parent_path | Q(path__startswith=path)
+
+            child_nodes = models.GoodsNomenclatureIndentNode.objects.filter(
+                has_parent_path,
+                valid_between__overlap=obj.valid_between,
+                indent=child,
+            ).order_by("valid_between")
```
simonwo, line 445: Do we need to filter on the `transaction` field ever with any of these?

commodities/import_handlers.py
```diff
@@ -284,34 +285,190 @@ def save(self, data: dict):
...
-        TODO: first scenario)
-        """
         if self.data["update_type"] == UpdateType.CREATE:
             return super(GoodsNomenclatureIndentHandler, self).post_save(obj)
 
-        # TODO: Should only use approved versions - but imported objects don't have approvers.
-        previous_version = obj.version_group.versions.exclude(pk=obj.pk).last()
+        previous_version = (
+            obj.version_group.versions.exclude(pk=obj.pk).has_approved_state().last()
+        )
```
simonwo, line 471: Does this require an ordering that isn't actually specified?

simonwo (approved):
I think this looks good! I can't spot any places where it's not going to cover all the bases. Comments are pretty minor.

commodities/import_handlers.py
```diff
@@ -284,34 +285,190 @@ def save(self, data: dict):
...
+        else:
+            # The new child node does not fit neatly into a single parent node.
+            # Split the child node so it fits under each new parent node.
+            start_date = new_valid_between.lower
+            end_date = new_valid_between.upper
+            for parent in new_parents:
+                new_valid_between = TaricDateRange(
+                    parent.valid_between.lower
+                    if parent.valid_between.lower > start_date
+                    else start_date,
```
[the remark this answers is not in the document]
NDevox: I think `None` possibilities screws that up sadly.

commodities/import_handlers.py
```diff
@@ -284,34 +285,190 @@ def save(self, data: dict):
...
+                    parent.valid_between.upper
+                    if parent.valid_between.upper < end_date
+                    else end_date,
```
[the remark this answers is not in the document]
NDevox: I think `None` possibilities screws that up sadly.

commodities/import_handlers.py
```diff
@@ -284,34 +285,190 @@ def save(self, data: dict):
...
+                new_valid_between.lower,
+                parent_indent.valid_between.upper,
+            )
+
+        if new_valid_between.lower < parent_indent.valid_between.lower:
+            # There is overlap on the lower end of the indent.
+            # Split the child tree so it is a child of both the old
+            # parent and the new parent indent.
+            old_child = child_node
+            child_node = child_node.copy_tree(
```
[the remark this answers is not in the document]
NDevox: Keeping the naming consistent means we can apply common changes to `child_node`. Sadly at times the `child_node` is not always new. It is a bit awkward and generic though.

commodities/import_handlers.py
NDevox: Not sure why we would. `transaction` just gives us a reference in time really. But at this point, because it's all brand new, every `transaction` should be relevant.

commodities/import_handlers.py
NDevox: I believe it relies on the default table ordering (which will be PK). Which works. But probably a good idea to formalise this.

PR #225, opened 2021-03-02
TP-634 Architectural documentation.

simonwo:
An architecture primer. Best viewed in the rendered documentation, so would recommend checking out this branch and building it.

### DIFF
--- a/certificates/__init__.py
+++ b/certificates/__init__.py
@@ -1,0 +1,8 @@
+"""
+Certificates, licenses, and non-paper conditions are sometimes required to bring
+goods through customs.
+
+A certificate is not a representation of the actual document itself – the
+database stores only a certificate code (4 digits) and a description of that
+code.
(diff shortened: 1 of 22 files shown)


NDevox (commented):
Overall looks pretty good. Just a couple of comments.

certificates/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""
+Certificates, licenses, and non-paper conditions are sometimes required to bring
+goods through customs.
+
+A certificate is not a representation of the actual document itself – our
```
andyhd, line 5: Did we say we were avoiding "we" and "our"? Should this be "the database stores"?

andyhd (approved):
This is so good 👍

stuaxo:
It looks good, definitely a fan of the way this partially comes from the docstrings too.

PR #226, opened 2021-03-04
TP-652 Find and view quotas

DIT-CJWaszczuk:
## Summary 
Added quotas list page with search and filter 
## Screenshots
image

### DIFF
--- a/common/filters.py
+++ b/common/filters.py
@@ -24,6 +24,7 @@ from django.db.models.functions import Lower
 from django_filters import CharFilter
 from django_filters import FilterSet
 from django_filters import MultipleChoiceFilter
+from django_filters.constants import EMPTY_VALUES
 from rest_framework import filters
 from rest_framework.settings import api_settings
 
@@ -44,6 +45,17 @@ def field_to_layout(field_name, field):
         return Field.checkboxes(field_name, legend_size=Size.SMALL)
 
     return field_name
+
+
+class MultiValueCharFilter(CharFilter):
+    def filter(self, qs, value):
+        if value in EMPTY_VALUES:
+            return qs
+        if self.distinct:
+            qs = qs.distinct()
+        lookup = "%s__%s" % (self.field_name, "in")
+        qs = self.get_method(qs)(**{lookup: value.split()})
+        return qs
 
 
 class LazyMultipleChoiceFilter(MultipleChoiceFilter):
--- a/common/static/common/scss/_table.scss
+++ b/common/static/common/scss/_table.scss
@@ -7,4 +7,11 @@
       font-size: inherit;
     }
   }
+  .govuk-table-list {
+    @extend .govuk-list;
+    margin-bottom: 0;
+    li:last-child {
+      margin-bottom: 0;
+    }
+  }
 }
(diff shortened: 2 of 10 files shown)


quotas/filters.py
```diff
@@ -0,0 +1,53 @@
...
+        choices=validators.QuotaCategory.choices,
+    )
+
+    clear_url = reverse_lazy("quota-ui-list")
+
+    def filter_origins(self, queryset, name, value):
+        if value:
+            value_array = value.split()
+            return queryset.filter(origins__area_id__in=value_array)
+        return queryset
```
simonwo, line 47: It feels like we might be wanting to use this in multiple places. Maybe we could do with subclassing `Filter` and providing this as a more general component.

DIT-CJWaszczuk: I've made `MultiValueCharFilter`, which should be reusable wherever needed now.

quotas/filters.py
```diff
@@ -0,0 +1,48 @@
...
+
+    order_number = CharFilter(
+        label="Order number",
+        field_name="order_number",
+    )
+
+    origin = MultiValueCharFilter(
+        label="Origin(s)",
+        field_name="origins__area_id",
+        method="filter_origins",
```
simonwo, line 23: Do we still need `method` for now?

simonwo (approved):

quotas/filters.py
DIT-CJWaszczuk: Nope, removed

common/filters.py
```diff
@@ -46,6 +47,17 @@ def field_to_layout(field_name, field):
     return field_name
 
 
+class MultiValueCharFilter(CharFilter):
+    def filter(self, qs, value):
+        if value in EMPTY_VALUES:
+            return qs
+        if self.distinct:
+            qs = qs.distinct()
+        lookup = "%s__%s" % (self.field_name, "in")
```
andyhd, line 56: The `%` operator should be treated as deprecated. Use `f"{self.field_name}__in"` instead.

common/filters.py
```diff
@@ -46,6 +47,17 @@ def field_to_layout(field_name, field):
...
+        qs = self.get_method(qs)(**{lookup: value.split()})
+        return qs
```
andyhd, line 58: Just `return self.get_method...`

common/static/common/scss/_table.scss
```diff
@@ -7,4 +7,11 @@
       font-size: inherit;
     }
   }
+  .govuk-table-list {
+    @extend .govuk-list;
```
andyhd, line 11: This is currently giving me
```
ERROR in ./common/static/common/scss/application.scss
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: From line 30, column 3 of node_modules/govuk-frontend/govuk/core/_lists.scss: 
   ╷
30 │   .govuk-list {
   │   ^^^^^^^^^^^^
   ╵
You may not @extend selectors across media queries.
   ╷
11 │     @extend .govuk-list;
   │     ^^^^^^^^^^^^^^^^^^^
   ╵
```

PR #227, opened 2021-03-04
TP-581: Linked model updates check business rules

andyhd:
Many business rules check that a condition holds only if the main model is modified, but
those rules could be violated if a linked model is updated. For example, ACN2 asserts:

> The referenced additional code type must exist and have as application code
> “non-Meursing” or “Export Refund for Processed Agricultural Goods".

But if the AdditionalCodeType is updated after the AdditionalCode is created and the
application code is changed, the rule is not checked again.

This commit adds a concept of "indirect business rules". A model may list business rules
that should be checked against any linked models that they apply to when it is updated.

The method of checking business rules is also modified to handle these indirect rule
triggers.

The business rule violation exception class has also been modified - each business rule
now automatically contains a specific violation subclass, eg ACN2.Violation. A violation
stores a reference to the violating model.


stuaxo:
Does this check one level of business rules, or does it go through all the indirect business rules for the next level + if so, could this ever end up circularly checking the business rules ?

andyhd:
Good point, I don't think there are any loops. I will check.

common/business_rules.py
```diff
@@ -19,42 +18,118 @@
...
+
+        return new_class
+
+
+class BusinessRule(metaclass=BusinessRuleBase):
     """Represents a TARIC business rule."""
 
+    @classmethod
+    def get_linked_models(cls, model: TrackedModel) -> Iterable[TrackedModel]:
+        for link in model._meta.related_objects:
```
NDevox, line 91: we have `TrackedModel.get_relations` already which returns `List[Tuple[Field, Type[TrackedModel]]]`

So should be able to just use that.

NDevox (commented):
Quite nice how little code this takes nowadays.

Couple minor thoughts, I don't necessarily think it makes sense to move away from the metaclass. But feels like we probably should be using `TrackedModel.get_relations` here.

common/business_rules.py
```diff
@@ -19,42 +18,118 @@
...
+        parents = [parent for parent in bases if isinstance(parent, BusinessRuleBase)]
+
+        # do not initialize base class
+        if not parents:
+            return super().__new__(cls, name, bases, attrs)
+
+        new_class = super().__new__(cls, name, bases, attrs, **kwargs)
+
+        # add BusinessRuleViolation subclass
+        setattr(
```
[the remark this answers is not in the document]
andyhd: This would mean that while we could, for example, do `v2 = ACN2.Violation(...)` it would be an instance of the same class as `ACN5.Violation`. I don't know if we will need to catch different types of violation and handle them differently, so maybe you're right and YAGNI applies here.

common/business_rules.py
andyhd: This is getting the reverse relationship. `get_relations()` would work if we were getting the link to `AdditionalCodeType` from `AdditionalCode`, but we want to get all `AdditionalCode`s linked to the specified `AdditionalCodeType`.

NDevox: Ok good point

NDevox (approved):

stuaxo (approved):

common/business_rules.py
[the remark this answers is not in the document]
andyhd: Also, I don't know if subclassing all parent violations is necessary - I copied this code from Django's `Model.DoesNotExist`. This is probably overkill.

PR #232, opened 2021-03-05
TP-652 Cleanup multifilter class

DIT-CJWaszczuk:


simonwo (approved):

common/filters.py
```diff
@@ -53,9 +53,8 @@ def filter(self, qs, value):
             return qs
         if self.distinct:
             qs = qs.distinct()
-        lookup = "%s__%s" % (self.field_name, "in")
-        qs = self.get_method(qs)(**{lookup: value.split()})
-        return qs
+        lookup = f"{self.field_name}__in"
+        return self.get_method(qs)(**{lookup: value.split()})
```
stuaxo, line 57: This method could do with a docstring, with an example or two - it's possible to see what's going on but takes a bit of looking at the mo.

stuaxo: I think it's just a tiny bit more noisy to look at because we have the **{lookup: value.split()} bit...

Though I wonder if that could be simplified to 
`self.get_method(qs)(lookup=value.split())`

Then it's probably clear enough ?

simonwo: Alas no, because `lookup` is a variable.

DIT-CJWaszczuk: As Simon has said, no can do unfortunately, i've added a docstring to explain the use case, can probably merge from there.

stuaxo: Awesome, was going to approve this last night but didn't want to generate a notification.

stuaxo (approved):

stuaxo (approved):

PR #233, opened 2021-03-08
Ignore same model version in uniqueness check

simonwo:
The `NoOverlapping` rule correctly removes the model being checked from the query but the `UniqueIdentifyingFields` rule does not. This is a fix for that and tests for both rules.

### DIFF
--- a/common/business_rules.py
+++ b/common/business_rules.py
@@ -6,6 +6,7 @@ from functools import wraps
 from typing import Iterable
 from typing import Mapping
 from typing import Optional
+from typing import Type
 from typing import Union
 
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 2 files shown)


stuaxo (approved):

common/business_rules.py
```diff
@@ -174,6 +177,7 @@ def validate(self, model):
         if (
             model.__class__.objects.filter(**query)
             .approved_up_to_transaction(model.transaction)
+            .exclude(id=model.id)
```
NDevox, line 180: I'd probably use `pk=model.pk` as it is a special property to return the same thing. Only real reason for this is because the models don't really have an `id` field - they inherit it from the parent class. They instead have a `trackedmodel_ptr` field which is the actual table pk.

Whilst `id` does get the same thing (through the join on `trackedmodel`) it;s a _tiny_ bit misleading and I find `pk` is clearer.

NDevox (commented):
Minor thing, but I wouldn't consider it blocking.

PR #238, opened 2021-03-11
TP-662 Add certificate update view

andyhd:
NOTE: Skipping descriptions update as @CJWaszczuk is working on a generic solution

image

This change adds an edit page for certificates.

The edit page template and confirm-update template are generic and located in `common/jinja2/common/`. It should be possible to reuse them for all components, only the form needs to be custom.
To enable this, the templates make use of `._meta.verbose_name` and `._meta.verbose_name_plural`. These attributes can be overridden in the model Meta class (see `commodities/models.py`).
The `verbose_name` is also used in the `TrackedModel.get_url()` method.


certificates/views.py
```diff
@@ -44,4 +49,36 @@ class CertificatesList(TamatoListView):
 class CertificateDetail(TrackedModelDetailView):
     model = Certificate
     template_name = "certificates/detail.jinja"
-    queryset = Certificate.objects.latest_approved().select_related("certificate_type")
+
+    def get_queryset(self):
+        return Certificate.objects.approved_up_to_transaction(
+            WorkBasket.current(self.request).transactions.order_by("pk").last(),
```
NDevox, line 55: transactions have `updated_at` and `created_at` which may be better to sort by. Although not sure if that fits the usecase.

certificates/views.py
```diff
@@ -44,4 +49,36 @@ class CertificatesList(TamatoListView):
...
+        ).select_related("certificate_type")
+
+
+class CertificateUpdate(TrackedModelDetailMixin, DraftUpdateView):
```
NDevox, line 59: Needs permissions required.

simonwo (approved):
LGTM.

I have a question about whether we should be having tests of this stuff, but let's take that offline.

certificates/jinja2/certificates/edit.jinja
```diff
@@ -0,0 +1,36 @@
...
+      No current workbasket
+    {%- endif %}
+  {%- endwith %}
+  </p>
+{% endblock %}
+
+{% block form %}
+  {% call django_form(action=object.get_url("edit")) %}
+    {{ crispy(form) }}
+  {% endcall %}
```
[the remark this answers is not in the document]
andyhd: I actually am about to move these templates into `common` :)

certificates/views.py
andyhd: Thanks, I will try to make it generic 👍

andyhd:
Tests are coming, sorry

certificates/views.py
andyhd: I've put this into a mixin that is used by all Certificate views, but I haven't made it generic to all TrackedModels

PR #239, opened 2021-03-11
Tp 59 view regulations

stuaxo:
Screenshots
===
image
image
image


DIT-CJWaszczuk:
Mind if we get a screenshot of the list page also?

stuaxo:
WIP, need to make the last changes to columns we want.

stuaxo:
Need to check the URL is correct.

regulations/filters.py
```diff
@@ -18,9 +21,16 @@ class RegulationFilterBackend(TamatoFilterBackend, RegulationFilterMixin):
...
 
 
-class RegulationFilter(TamatoFilter, RegulationFilterMixin):
+class RegulationFilter(
+    TamatoFilter,
+    RegulationFilterMixin,
+    StartYearMixin,
+    ActiveStateMixin,
+):
     role_type = filters.TypedMultipleChoiceFilter(choices=RoleType.choices, coerce=int)
```
DIT-CJWaszczuk, line 30: I think the role_type isn't rendering with govuk styling, think you just need to add a widget to it.
https://github.com/uktrade/tamato/blob/d9a4b216a87fbb9fa2bb7221758ff39935a2cbfa/certificates/filters.py#L38

DIT-CJWaszczuk:
Not quite sure the fields in the details tab align with what's in the final comment of https://uktrade.atlassian.net/browse/TP-613. Feel free to correct me here however if there's some info i've glossed over

regulations/filters.py
DIT-CJWaszczuk: Actually looking at this again, if this is regulation_type, and they're all meant to only ever be 'base', there may just be a case to remove it all together. It won't bring any value if there's only ever one real option.

regulations/filters.py
```diff
@@ -18,9 +21,16 @@ class RegulationFilterBackend(TamatoFilterBackend, RegulationFilterMixin):
...
 
+    clear_url = reverse_lazy("regulation-ui-list")
+
     class Meta:
         model = Regulation
-        fields = ["regulation_id", "role_type"]
+        fields = ["regulation_id", "role_type", "start_year", "active_state"]
```
DIT-CJWaszczuk, line 36: I think if we're searching using the fuzzy search, rather than an id search. We can change this to 
`fields = ["search", "start_year", "active_state"]`
to display them in the correct order

DIT-CJWaszczuk:
@simonwo Just to note. Seems there's a lot of different ways of displaying the same information, is this a data issue? 
mesu 103 (chap 10)
MESU 103 (Chap 10)
MESU 103 (Chap. 10)
image

DIT-CJWaszczuk:
One small thing with the fields, otherwise I think this is good to go. Thanks for getting stuck in @stuaxo

regulations/jinja2/includes/regulations/list.jinja
```diff
@@ -0,0 +1,31 @@
+{% set table_rows = [] %}
+{% for object in object_list %}
+  {% set object_link -%}
+    <a class="govuk-link govuk-!-font-weight-bold" href="{{ url("regulation-ui-detail", kwargs={"regulation_id": object.regulation_id, "role_type": object.role_type}) }}">{{ object.regulation_id }}</a>
```
DIT-CJWaszczuk, line 4: You should just be able to use `regulation.get_url()` here instead of `url()`

regulations/filters.py
andyhd: They are all going to be Base Regulations going forward, but we might want to filter historical data by regulation type - might be a question for BAs or Product

simonwo:
> @simonwo Just to note. Seems there's a lot of different versions of same information, is this a data issue?

I don't think the system is displaying anything incorrect, I think the data is just messy like that. The EU wasn't particularly great at making this data meaningful back in the 90s, so there's a lot of this stuff unfortunately.

stuaxo:
```
> > @simonwo Just to note. Seems there's a lot of different versions of same information, is this a data issue?
> 
> I don't think the system is displaying anything incorrect, I think the data is just messy like that. The EU wasn't particularly great at making this data meaningful back in the 90s, so there's a lot of this stuff unfortunately.
```

If we ever want to make it more meaningful I guess we would need to have a way of feeding back recommended changes to them ?

regulations/filters.py
stuaxo: Yep, removing - based on Teams chat with @simonwo last week
image

stuaxo:
For some bikeshedding fun, here are some different ways of displaying the role type, since it makes up part of the ID

image

regulations/jinja2/includes/regulations/list.jinja
```diff
@@ -0,0 +1,31 @@
+{% set table_rows = [] %}
+{% for object in object_list %}
+  {% set object_link -%}
+    <a class="govuk-link govuk-!-font-weight-bold" href="{{ object.get_url() }}">{{ object.regulation_id }}</a>
+  {%- endset %}
+  {{ table_rows.append([
+    {"html": object_link},
+    {"text": object.information_text if object.information_text else "-"},
+    {"html": ("<a href=\"" ~ object.url ~ "\">" ~ object.public_identifier ~ "</a>") if object.public_identifier else "-"},
```
DIT-CJWaszczuk, line 9: Should be able to use `create_link()` here to save you having to hack together a tags. See https://github.com/uktrade/tamato/blob/fdd62bf2ef7836771a75e0ea40400d3019c96025/measures/jinja2/includes/measures/tabs/conditions.jinja#L11

regulations/jinja2/includes/regulations/tabs/core_data.jinja
```diff
@@ -0,0 +1,52 @@
...
+				"actions": {"items": []}
+			},
+			{
+				"key": {"text": "Information text"},
+				"value": {"text": object.information_text if object.information_text else "-"},
+				"actions": {"items": []}
+			},
+			{
+				"key": {"text": "Public-facing identifier"},
+				"value": {"html": ("<a href=\"" ~ object.url ~ "\">" ~ object.public_identifier ~ "</a>") if object.public_identifier else "-"},
```
DIT-CJWaszczuk, line 19: same `create_link` comment as mentioned above

regulations/jinja2/includes/regulations/list.jinja
stuaxo: :+1: :+1: :+1:

DIT-CJWaszczuk (approved):
LGTM

regulations/models.py
```diff
@@ -196,6 +198,9 @@ class Regulation(TrackedModel):
         business_rules.ROIMB47,
     )
 
+    def get_role_type_label(self):
+        return RoleType(self.role_type).label
```
NDevox, line 202: Could this not just be `self.get_role_type_display()`?

stuaxo: Sounds like a plan.. though even better; I can remove this as I was using it in protyping but am not any more.

stuaxo: Done

NDevox (approved):
LGTM

PR #246, opened 2021-03-17
Check validation rules on import and log them as errors.

simonwo:
Previously the clean method on workbaskets was not implemented, so there was no validation performed on import.

But also, running against the entire workbasket is not what we want – as per ADR 12, validations need to be performed after every transaction.

This commit changes our import validation to run business rules after every transaction, but also implements a clean method for a workbasket as well (it will be useful later).

Notably we also no longer halt an import run when an error is found and instead just print an error. The reason for this is that when importing we are normally importing well-formed XML and business rule violations are (at the moment) normally an indication that our implementations are wrong. Plus, it's much easier to debug data when it is in the database rather than in the XML files. So this gives us the option to do that more easily.

When we come around to implementing receipt parsing and marking individual transactions as errored in TP-452, we can change this behaviour and as well as logging an error we can mark the individual transaction as being in an error state.

### DIFF
--- a/common/models/mixins.py
+++ b/common/models/mixins.py
@@ -32,5 +32,22 @@ class ValidityMixin(models.Model):
 
     valid_between = TaricDateRangeField(db_index=True)
 
+    validity_field_name: str = "valid_between"
+    """The name of the field that should be used for validity date checking."""
+
+    @classmethod
+    def objects_with_validity_field(cls):
+        """
+        Returns a QuerySet which will have this model's validity date field (as
+        specified by :attr:`validity_field_name`) present on the returned
+        models.
+
+        The need for this is that some models (e.g.
+        :class:`~measures.models.Measure`) use a validity date that is computed
+        on demand as part of a database query and hence is not present on the
+        default queryset.
+        """
+        return cls.objects
+
     class Meta:
         abstract = True
--- a/importer/management/commands/import_taric.py
+++ b/importer/management/commands/import_taric.py
@@ -7,15 +7,22 @@ from workbaskets.validators import WorkflowStatus
 
 
 def import_taric(
-    taric3_file, username, status, name, split_codes: bool = False, dependencies=None
+    taric3_file,
+    username,
+    status,
+    name,
+    split_codes: bool = False,
+    dependencies=None,
 ):
     batch = setup_batch(
-        batch_name=name, dependencies=dependencies, split_on_code=split_codes
+        batch_name=name,
+        dependencies=dependencies,
+        split_on_code=split_codes,
     )
     with open(taric3_file, "rb") as seed_file:
         batch = chunk_taric(seed_file, batch)
 
-    run_batch(batch.name, username, status)
+    run_batch(batch.name, status, username)
 
 
 class Command(BaseCommand):
--- a/importer/taric.py
+++ b/importer/taric.py
@@ -9,6 +9,7 @@ from typing import Optional
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.exceptions import ValidationError
 from django.db import IntegrityError
 from django.db.transaction import atomic
 from lxml import etree
@@ -16,8 +17,8 @@ from lxml import etree
 from common import models
 from common.validators import UpdateType
 from importer.namespaces import ENVELOPE
-from importer.namespaces import nsmap
 from importer.namespaces import Tag
+from importer.namespaces import nsmap
 from importer.nursery import get_nursery
 from importer.parsers import ElementParser
 from importer.parsers import ParserError
@@ -124,7 +125,12 @@ class TransactionParser(ElementParser):
 
         for message_data in data["message"]:
             self.message.save(message_data, transaction.id)
-        self.parent.workbasket.clean()
+
+        try:
+            transaction.clean()
+        except ValidationError as e:
+            for message in e.messages:
+                logger.error(message)
 
     def end(self, element: etree.Element):
         super().end(element)
--- a/importer/taric.py
+++ b/importer/taric.py
@@ -28,6 +28,9 @@ from taric.models import EnvelopeTransaction
 from workbaskets.models import WorkBasket
 from workbaskets.validators import WorkflowStatus
 
+if settings.SENTRY_ENABLED:
+    from sentry_sdk import capture_exception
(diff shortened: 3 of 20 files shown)


importer/management/commands/import_taric.py
```diff
@@ -7,15 +7,22 @@
...
-        batch_name=name, dependencies=dependencies, split_on_code=split_codes
+        batch_name=name,
+        dependencies=dependencies,
+        split_on_code=split_codes,
     )
     with open(taric3_file, "rb") as seed_file:
         batch = chunk_taric(seed_file, batch)
 
-    run_batch(batch.name, username, status)
+    run_batch(batch.name, status, username)
```
NDevox, line 25: How on earth this made it through I'll never know.

importer/taric.py
```diff
@@ -124,7 +125,12 @@ def save(
 
         for message_data in data["message"]:
             self.message.save(message_data, transaction.id)
-        self.parent.workbasket.clean()
+
+        try:
+            transaction.clean()
+        except ValidationError as e:
+            for message in e.messages:
+                logger.error(message)
```
NDevox, line 133: Surely we should raise the error as well. No reason to suppress it.

simonwo: The reasoning is that we don't actually want to halt the import job. I guess we could handle it upwards instead? But ultimately  we won't be able to keep prod up to date as a result of our business rule bugs, whereas at least this way we can have data in the system that has been validated elsewhere whilst we fix the bugs.

NDevox: It may have been a good idea for me to read the PR summary!

So if we're assuming it is a bug on our side this makes sense. But I'm not convinced we should always be thinking that (plenty of times we've found silly data from those providing the data - admittedly we've had to adapt to it most times anyway). And once the data is in the system it's a pain to revert an import (not impossible, but a pain).

stuaxo (approved):

importer/taric.py
simonwo: Yeah I do agree, it's clearly will not always be a bug necessarily, but if we're bringing data in via this route it will basically either be EU data that we pretty much have to accept verbatim, or manually prepared data. In either case (unless something fails as the first element) we're always going to have to handle reverting the previous data that has been loaded if we want to go back? The only real difference is that this way you a) have the option of not reverting it, if you're sure it's what you want depsite the errors, and you b) have all the data in the database so you can do a more thorough analysis.

I was going to suggest if there's somewhere we can put a flag to toggle behaviour, but then I'm not sure if the explosion behaviour is desired in any environment due to a). Is there instead someway we could signal that an import job contains errors without making it impossible to proceed? Because the thing I would be worried about is not having any feedback from the import UI that something threw an error.

NDevox: What we do get right now if an error occurs is sentry telling us what is going on. I'm not convinced we get that with this change. And I don't think the job will be marked as errorred either. So there is some difficulty here (may not realise we have an error until months after the fact), finding the exact error messages etc.

I can see the reasoning behind it. I'm just wondering if this will in fact make it harder to detect problems (despite it being easier to investigate once we know we have one).

simonwo: I agree that we shouldn't be making it harder to detect problems… can you see any ways of achieving that? Could we catch the errors and then rethrow them all at the end of the job?

simonwo:
Now that we actually run the business rules when we do importer tests, our tests found a bunch of places where our business rules were wrong or didn't work. So I had to fix those 😤

importer/taric.py
stuaxo: We could use a different logger for this, that routes messages to the ordinary handler, as well as a handler that collects and outputs the errors for this.

NDevox: Here try this https://docs.sentry.io/platforms/python/usage/#capturing-errors

NDevox:
Thank god we have those tests then 😉

common/models/mixins.py
```diff
@@ -32,5 +32,11 @@ class ValidityMixin(models.Model):
 
     valid_between = TaricDateRangeField(db_index=True)
 
+    validity_field_name = "valid_between"
+
+    @classmethod
+    def objects_with_validity_field(cls):
+        return cls.objects
```
NDevox, line 39: Wait are we using this for?

simonwo: It's overriden by `Measure` to allow it to return a queryset that has `.with_effective_valid_between()` called on it.

importer/taric.py
simonwo: So if I did:

    import settings
    if settings.SENTRY_ENABLED:
        from sentry_sdk import capture_exception

And subsequently:

            transaction.clean()
        except ValidationError as e:
            for message in e.messages:
                logger.error(message)
            if settings.SENTRY_ENABLED:
                capture_exception(e)

Does that integrate it in the way we would like/be comfortable with?

common/models/mixins.py
stuaxo: Maybe this instance wants a docstring as it's a little bit odd to read in isolation.

measures/parsers.py
```diff
@@ -271,21 +282,220 @@ class SeasonalRateParser:
...
                 yield (
                     rate,
                     validity_start.day,
                     validity_start.month,
                     validity_end.day,
                     validity_end.month,
                 )
+
+
+class ConditionSentenceParser:
```
NDevox, line 300: This looks like it belongs in a different PR?

I appreciate I have no right to say this given some of my PRs...

importer/taric.py
```diff
@@ -9,15 +9,17 @@
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.exceptions import ValidationError
 from django.db import IntegrityError
 from django.db.transaction import atomic
 from lxml import etree
 
+import settings
```
NDevox, line 17: `from django.conf import settings` otherwise this is all good.

measures/parsers.py
NDevox: Looks like this is just a refactor. Ignore me.

simonwo: Yeah it does. The way I use git means stuff like this happens if I take my eye off the ball…

NDevox (approved):

PR #247, opened 2021-03-17
TP-682 Adjust conditions tab content

DIT-CJWaszczuk:
It's exactly what it looks like


andyhd (approved):
I think it's ok to merge stuff like this without a PR 😁

NDevox (approved):

PR #248, opened 2021-03-17
Replace object type with verbose name

DIT-CJWaszczuk:
## Summary
Apart from Quotas (we use `QuotaOrderNumber` for the data), we can remove the need for object_type and replace it with the verbose names.


common/jinja2/layouts/detail.jinja
```diff
@@ -1,12 +1,10 @@
 {% extends "layouts/layout.jinja" %}
 
-{% set object_type_plural = object_type_plural or object_type + "s" %}
-
 {% block breadcrumb %}
   {{ govukBreadcrumbs({
     "items": [
       {"text": "Home", "href": url("index")},
-      {"text": "Find and edit " ~ object_type_plural, "href": find_edit_url},
+      {"text": "Find and edit " ~ object._meta.verbose_name_plural, "href": find_edit_url},
```
andyhd, line 7: I have a bit in my PR that allows you to do `object.get_url(action="list")` so you can get rid of `find_edit_url` too

andyhd (approved):
Looks good to me

stuaxo (approved):
Nice, there's so much handy stuff in _meta.

Good how this removes a bunch of variables.

DIT-CJWaszczuk:
This actually would not work, not in its current state anyway. If there are no results, there's no object to get the verbose name  from. Hmmm

common/jinja2/layouts/list.jinja
```diff
@@ -4,10 +4,10 @@
 {% from "components/input/macro.njk" import govukInput %}
 {% from "components/table/macro.njk" import govukTable %}
 
-{% set object_type_plural = object_type_plural or object_type + "s" %}
 {% set create_url = create_url or "create" %}
+{% set object = object_list|first %}
 
-{% set page_title = "Find and edit " + (object_type_plural or object_type + "s") %}
+{% set page_title = "Find and edit " ~ object._meta.verbose_name_plural %}
```
andyhd, line 10: This needs to be `object_list.model._meta.verbose_name_plural`

andyhd (changes requested):

simonwo (approved):

PR #251, opened 2021-03-18
Fix type vs Type typo.

stuaxo:
There was a typing typo, somebody typed "type" instead of "Type" which made the type inferator sad when it couldn't find the type and the site fell over.

### DIFF
--- a/certificates/views.py
+++ b/certificates/views.py
@@ -1,4 +1,5 @@
 from typing import Optional
+from typing import Type
 
 from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.db import models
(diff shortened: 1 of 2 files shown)


certificates/views.py
```diff
@@ -36,7 +37,7 @@ class CertificateTypeViewSet(viewsets.ReadOnlyModelViewSet):
 
 
 class CertificateMixin:
-    model: type[models.Model] = Certificate
+    model: Type[models.Model] = Certificate
```
NDevox, line 40: tbf if we can import `TrackedModel` here I'd change it to that.

stuaxo: Should that look like `model: Type[TrackedModel] = Certificate` ?

common/views.py
```diff
@@ -150,7 +151,7 @@ class TamatoListView(WithCurrentWorkBasket, WithPaginationListView):
 class TrackedModelDetailMixin:
     """Allows detail URLs to use <Identifying-Fields> instead of <pk>"""
 
-    model: type[models.Model]
+    model: Type[models.Model]
```
NDevox, line 154: Can do it here too

NDevox (approved):

PR #253, opened 2021-03-19
Fix certificate views when no current workbasket

andyhd:
If there is no current workbasket set, then certificate UI views will error.

### DIFF
--- a/certificates/views.py
+++ b/certificates/views.py
@@ -41,9 +41,14 @@ class CertificateMixin:
     model: Type[TrackedModel] = Certificate
 
     def get_queryset(self):
-        return Certificate.objects.approved_up_to_transaction(
-            WorkBasket.current(self.request).transactions.order_by("order").last(),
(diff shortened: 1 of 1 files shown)


certificates/views.py
```diff
@@ -41,9 +41,14 @@ class CertificateMixin:
     model: Type[TrackedModel] = Certificate
 
     def get_queryset(self):
-        return Certificate.objects.approved_up_to_transaction(
-            WorkBasket.current(self.request).transactions.order_by("order").last(),
-        ).select_related("certificate_type")
+        workbasket = WorkBasket.current(self.request)
+        tx = None
```
stuaxo, line 45: The earlier version with `null` was unusual !

stuaxo (approved):

certificates/views.py
andyhd: I must be losing my marbles

DIT-CJWaszczuk (approved):

PR #254, opened 2021-03-19
TP-653 Edit additional codes

DIT-CJWaszczuk:
## Sumary 
Allows a user with permissions to edit additional codes
### Details page 
image

### Edit page 
image

### Confirmation page
image

### DIFF
--- a/additional_codes/models.py
+++ b/additional_codes/models.py
@@ -83,6 +83,9 @@ class AdditionalCode(TrackedModel, ValidityMixin):
         business_rules.ACN17,
     )
 
+    def __str__(self):
+        return f"{self.type.sid}{self.code}"
+
     def in_use(self):
         return (
             self.measure_set.model.objects.filter(
--- a/additional_codes/views.py
+++ b/additional_codes/views.py
@@ -1,15 +1,24 @@
+from typing import Optional
+from typing import Type
+
+from django.contrib.auth.mixins import PermissionRequiredMixin
+from django.db import models
 from rest_framework import permissions
 from rest_framework import viewsets
 
-from additional_codes import models
 from additional_codes.filters import AdditionalCodeFilter
 from additional_codes.filters import AdditionalCodeFilterBackend
+from additional_codes.forms import AdditionalCodeForm
 from additional_codes.models import AdditionalCode
 from additional_codes.models import AdditionalCodeType
 from additional_codes.serializers import AdditionalCodeSerializer
 from additional_codes.serializers import AdditionalCodeTypeSerializer
+from common.models import TrackedModel
 from common.views import TamatoListView
+from common.views import TrackedModelDetailMixin
 from common.views import TrackedModelDetailView
+from workbaskets.models import WorkBasket
+from workbaskets.views.generic import DraftUpdateView
 
 
 class AdditionalCodeViewSet(viewsets.ReadOnlyModelViewSet):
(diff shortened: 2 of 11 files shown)


andyhd:
This build error is fixed in the PR I just merged, so you should be able to just rebase and it will work

simonwo (approved):

additional_codes/models.py
```diff
@@ -83,6 +83,9 @@ class AdditionalCode(TrackedModel, ValidityMixin):
         business_rules.ACN17,
     )
 
+    def __str__(self):
+        return self.type.sid + self.code
```
andyhd, line 87: It's better to use `f"{self.type.sid}{self.code}"`

additional_codes/views.py
```diff
@@ -30,14 +39,31 @@ class AdditionalCodeViewSet(viewsets.ReadOnlyModelViewSet):
     ]
 
 
-class AdditionalCodeList(TamatoListView):
+class AdditionalCodeTypeViewSet(viewsets.ReadOnlyModelViewSet):
+    """API endpoint that allows additional code types to be viewed."""
+
+    queryset = AdditionalCodeType.objects.latest_approved()
+    serializer_class = AdditionalCodeTypeSerializer
+    permission_classes = [permissions.IsAuthenticated]
```
andyhd, line 47: I'm not sure you need to add this - this is the API viewset and not relevant to this ticket. Also, we probably don't want the queryset to be `latest_approved`, but I'm not sure.

DIT-CJWaszczuk: This already existed, I just moved it in the file so it went API classes -> UI classes. Was previously bunched in between

andyhd: Never mind, I see this is just moved

andyhd (approved):

additional_codes/views.py
```diff
@@ -48,19 +74,30 @@ class AdditionalCodeList(TamatoListView):
...
+    def get_object(self, queryset: Optional[models.QuerySet] = None) -> models.Model:
+        obj = super().get_object(queryset)
+
+        if self.request.method == "POST":
+            obj = obj.new_draft(
+                WorkBasket.current(self.request),
+                save=False,
+            )
+
+        return obj
```
andyhd, line 99: Looks like we can pull this out into `DraftUpdateView`

andyhd (approved):

common/jinja2/common/edit.jinja
```diff
@@ -1,34 +1,18 @@
...
     "items": [
       {"text": "Home", "href": url("index")},
-      {"text": object._meta.verbose_name_plural|capitalize, "href": object.get_url("list")},
-      {"text": object._meta.verbose_name|capitalize ~ " " ~ object|string, "href": object.get_url()},
+      {"text": "Find and edit " ~ object._meta.verbose_name_plural, "href": object.get_url("list")},
+      {"text": object._meta.verbose_name|capitalize ~ ": " ~ object|string, "href": object.get_url()},
       {"text": page_title}
     ]
   }) }}
 {% endblock %}
```
andyhd, line 14: This block is the same as the detail page and the confirm-update page, so maybe they can all extend a common base template. Not at all urgent, though.

additional_codes/views.py
andyhd: I was wrong 😁

common/jinja2/common/edit.jinja
DIT-CJWaszczuk: I have a feeling the confirm may change when we have create pages, so will wait till it potentially gets generalised then.

PR #256, opened 2021-03-22
TP-669: Show workbasket changes on index and add submit button.

NDevox:
Instead of showing a list of workbaskets on the index page it has been
decided to show the single active workbasket and its list of changes instead.
This is coupled with a button to submit the workbasket.

This commit updates the index view to provide a single unapproved workbasket,
if one does not exist it will create one. the changed objects within the workbasket
are shown with a "structure code" and "structure description" which are assumed to
be their identifying fields and descriptions (if existing).

Example:

Screenshot 2021-03-22 at 10 35 48


common/models/records.py
```diff
@@ -582,6 +582,22 @@ def get_identifying_fields(
...
+        description = None
+        if hasattr(self, "descriptions"):
+            description = self.get_descriptions().last()
+            if description:
+                # Get the actual description, not just the object
+                description = description.description
+        if hasattr(self, "description"):
+            description = self.description
+        return description or "-"
+
```
NDevox, line 600: Need to confirm what is actually expected here. @simonwo ?

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
```diff
@@ -0,0 +1,48 @@
...
+            <p class="govuk-body align-left">{{ change_count }} change{{ pluralize(change_count) }}</p>
+            <p class="govuk-body align-right">
+                <a href="{{ url('workbaskets:submit_workbasket', args=[workbasket.pk] ) }}" class="govuk-link govuk-!-margin-right-5">Mark "Done" the selected changes</a>
+                <a href="#" class="govuk-link">Download this data</a>
+            </p>
+        </div>
+        {% set table_rows = [] %}
+        {% for obj in page_obj.object_list %}
+          {% set object_link -%}
+            <a class="govuk-link govuk-!-font-weight-bold" href="#">{{ obj.pk }}</a>
```
NDevox, line 20: Doubt we want to use the PK here.

Can't actually provide a URL until all the models have detail URLs to give.

andyhd: Would it make sense to have `TrackedModel.get_url()` return `None` instead of raising an exception if the URL doesn't exist?

andyhd (approved):
Looks good to me

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
NDevox: This does make sense to me tbh. Will implement.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
```diff
@@ -0,0 +1,48 @@
...
+            <p class="govuk-body align-left">{{ change_count }} change{{ pluralize(change_count) }}</p>
+            <p class="govuk-body align-right">
+                <a href="{{ url('workbaskets:submit_workbasket', args=[workbasket.pk] ) }}" class="govuk-link govuk-!-margin-right-5">Mark "Done" the selected changes</a>
+                <a href="#" class="govuk-link">Download this data</a>
+            </p>
+        </div>
+        {% set table_rows = [] %}
+        {% for obj in page_obj.object_list %}
+          {% set object_link -%}
+            <a class="govuk-link govuk-!-font-weight-bold" href="{{ obj.get_url() or "#" }}">{{ obj.pk }}</a>
```
NDevox, line 20: Q remains as to what we actually want to show here. I don't think it should be PK

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
```diff
@@ -0,0 +1,48 @@
...
+          {%- endset %}
+          {{ table_rows.append([
+            {"html": object_link},
+            {"text": obj.__class__.__name__},
```
NDevox, line 24: Bit of an odd one, doesn't perfectly match the designs. Could try to humanize a bit but (split on capital letter)... I'm unconvinced.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
andyhd: I would remove the ID column and make the Structure and/or Structure Code values the link to the detail page, or the whole row. But that's a design/product question.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
andyhd: Could use `obj._meta.verbose_name|capitalize` like the page titles do

common/models/records.py
simonwo: I think this a fine first pass! 

What I think we probably need though is to bring the various decisions that have been made about what gets displayed for the "code" and the "description" on the forms back into the model and then we can use it everywhere. E.g. for additional code the code is [`object.type.sid ~ object.code`](https://github.com/uktrade/tamato/blob/52af7500bfc1b7a5f6c7c99f404ffc92a0d05aed/additional_codes/jinja2/includes/additional_codes/list.jinja#L4) so perhaps that should be a property on the model. (If that is just `__str__` already then fine, but I had some notion that `__str__` displays `field=id` for the all the identifying fields?)

In any case lets improve in a next PR.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
simonwo: Agree with @andyhd, we should use that.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
simonwo: Agree with @andyhd.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
```diff
@@ -0,0 +1,48 @@
...
+  <h2 class="govuk-heading-l">
+    Your tariff changes
+  </h2>
+    {% set change_count = workbasket.tracked_models.count() %}
+    <div class="govuk-grid-column-full govuk-!-margin-top-3">
+        <div class="govuk-grid-column-full">
+            <p class="govuk-body align-left">{{ change_count }} change{{ pluralize(change_count) }}</p>
+            <p class="govuk-body align-right">
+                <a href="{{ url('workbaskets:submit_workbasket', args=[workbasket.pk] ) }}" class="govuk-link govuk-!-margin-right-5">Mark "Done" the selected changes</a>
+                <a href="#" class="govuk-link">Download this data</a>
```
simonwo, line 14: Let's remove this for now until we make it do something :)

workbaskets/models.py
```diff
@@ -43,6 +43,18 @@ def ordered_transactions(self):
             workbasket__pk__in=Subquery(workbasket_pks),
         ).order_by("order")
 
+    def is_approved(self):
+        return self.filter(
+            status__in=WorkflowStatus.approved_statuses(),
+            approver__isnull=False,
```
simonwo, line 49: Is this definitely always right? What about for workbaskets uploaded from the importer? Do they get tagged with the correct approver?

workbaskets/views/__init__.py
```diff
@@ -88,3 +92,20 @@ def set_current_workbasket(self, request):
...
+@transaction.atomic
+def submit_workbasket_view(request, workbasket_pk):
+    workbasket = get_object_or_404(WorkBasket, pk=workbasket_pk)
+
+    workbasket.full_clean()
+
+    workbasket.submit_for_approval()
+    workbasket.approve(request.user)
+
+    workbasket.export_to_cds()
```
simonwo, line 108: Does this move it through workflow? Or actually trigger a job?

common/models/records.py
andyhd: I made the default `__str__` method return `field=value` for all identifying fields, but it was just for debug printlining convenience. Unless any other code relies on `__str__` doing that, I think it makes sense to have it return something useful for displaying the users.

NDevox: I used `__str__` for convenience as there are a few models which give exactly what we want via that method (footnotes, certificates, goods nomenclature maybe - you can see in the screenshot). This can be a slow burn, either just overriding `structure_code` itself or `__str__` methods as we go.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
NDevox: Done

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
NDevox: Done, shifted the structure code to replace the ID column.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
NDevox: Done

workbaskets/views/__init__.py
NDevox: Right now it just shifts the status. I haven't followed Stu's work closely, but if we want to trigger a job it should be done within that method.

stuaxo: This discussion hadn't been had, but if that's where a job should be triggered, I'm happy to add that info the relevant ticket (and then implement it).

common/models/records.py
stuaxo: We always have `__repr__` if we want something for developers.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
```diff
@@ -0,0 +1,45 @@
+{% from "components/table/macro.njk" import govukTable %}
+
+<div class="govuk-grid-row">
+<div class="govuk-grid-column-full">
+  <h2 class="govuk-heading-l">
+    Your tariff changes
+  </h2>
+    {% set change_count = workbasket.tracked_models.count() %}
+    <div class="govuk-grid-column-full govuk-!-margin-top-3">
+        <div class="govuk-grid-column-full">
```
DIT-CJWaszczuk, line 10: You can scrap these two, that should align the table and pagination with the title again. The margin is also collapsing against the heading anyway so it's currently moot.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
```diff
@@ -0,0 +1,45 @@
...
+          "head": [
+            {"text": "Structure Code"},
+            {"text": "Structure"},
+            {"text": "Structure Description"},
+            {"text": "Activity Date"},
+          ],
+          "rows": table_rows
+        }) }}
+
+        <div class="filter-layout__content govuk-!-margin-top-5">
```
DIT-CJWaszczuk, line 39: Don't need `filter-layout__content` here, it's reducing the width to 75%. Margin top is also collapsing here, can probably just remove the entire div.

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
```diff
@@ -0,0 +1,45 @@
...
+<div class="govuk-grid-column-full">
+  <h2 class="govuk-heading-l">
+    Your tariff changes
+  </h2>
+    {% set change_count = workbasket.tracked_models.count() %}
+    <div class="govuk-grid-column-full govuk-!-margin-top-3">
+        <div class="govuk-grid-column-full">
+            <p class="govuk-body align-left">{{ change_count }} change{{ pluralize(change_count) }}</p>
+            <p class="govuk-body align-right">
+                <a href="{{ url('workbaskets:submit_workbasket', args=[workbasket.pk] ) }}" class="govuk-link govuk-!-margin-right-5">Mark "Done" the selected changes</a>
```
DIT-CJWaszczuk, line 13: Gonna want to pull the margin-right off this to right align it, think that was only used for every link except the last.

DIT-CJWaszczuk (commented):
Bit late to the party here sorry, should've reviewed sooner.

common/jinja2/includes/common/pagination.jinja
```diff
@@ -0,0 +1,46 @@
+{% if page_obj.has_other_pages() %}
+    <nav class="pagination" role="navigation" aria-label="Pagination Navigation">
+      <div class="govuk-body">
+        Showing {{ page_obj|length }} of {{'{0:,}'.format(paginator.count)}} {{ object_list.model._meta.verbose_name_plural if paginator.count > 1 else object_list.model._meta.verbose_name }}
```
DIT-CJWaszczuk, line 4: I'm torn here, because we have a situation of the top saying 'changes', the bottom saying 'tracked models' and Bo's design saying 'works' (this one is probably outdated however). May want to change the verbose name to something like 'tariff changes', but that's a question for Madeline / Bo / Mashuk really

workbaskets/models.py
NDevox: Forgot to respond to this. Yes they get tagged with the user uploading the files (if it is set to approved status on import)

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
NDevox: Done

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
NDevox: Done

PR #261, opened 2021-03-23
TP-669: Remove manual WorkBasket assignment when editing.

NDevox:
Currently when editing a TrackedModel the user is asked which WorkBasket
to add the object to, and are offered to make a new WorkBasket if needed.
The intended usecase for now is to only allow a single WorkBasket at any
one time, in which case there is no reason to allow a choice or manual
creation.

This commit removes the interstitial page and defaults to using the current
active WorkBasket.


DIT-CJWaszczuk:
Are we going to manually create the initial workbasket for each environment?

NDevox:
Currently it makes itself when you go to the index page. Although we probably need something a bit more formative (e.g. create new workbasket on old workbasket approval). Could be part of a future PR.

NDevox:
https://github.com/uktrade/tamato/blob/0bff2ea/common/views.py#L31

DIT-CJWaszczuk:
Ah I see, I guess this is fine for now then. Might just be worth creating it in `check_for_current_workbasket` if there is currently none? I'm assuming that is called on the index and also before the edit page.

DIT-CJWaszczuk (approved):

PR #266, opened 2021-03-25
TP-709 Fix messages on wrapped business rule classes

simonwo:
It turns out that making a new subclass doesn't
cause most of the things we need to be inherited
by the new class, such as the Violation object and
the class' docstring. This results in cryptic
errors when the business rule fires. This new
method prevents that from happening by just
overriding the method.

As a bonus wrapped classes will now appear in
generated documentation.

### DIFF
--- a/common/business_rules.py
+++ b/common/business_rules.py
@@ -2,6 +2,7 @@
 import logging
 from datetime import date
 from datetime import datetime
+from functools import wraps
 from typing import Iterable
 from typing import Mapping
 from typing import Optional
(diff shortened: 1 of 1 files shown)


common/business_rules.py
```diff
@@ -146,15 +147,17 @@ def only_applicable_after(cutoff: Union[date, datetime, str]):
...
-        class OnlyApplicableAfter(cls):
-            def validate(self, model):
-                if model.valid_between.lower > cutoff:
-                    super().validate(model)
-                else:
-                    log.debug("Skipping %s: Start date before cutoff", cls.__name__)
-
-        OnlyApplicableAfter.__name__ = cls.__name__
-        return OnlyApplicableAfter
+        _original_validate = cls.validate
```
NDevox, line 150: I think this needs documentation

simonwo: Sure, is the docstring any more helpful now?

NDevox (approved):

PR #267, opened 2021-03-30
TP-662 Edit additional code descriptions

DIT-CJWaszczuk:
## Summary 
Allow people to edit additional code descriptions 
### Description list
image
## Edit page 
image
## Confirmation page 
image

### DIFF
--- a/additional_codes/models.py
+++ b/additional_codes/models.py
@@ -1,9 +1,12 @@
 from django.db import models
+from django.urls import NoReverseMatch
+from django.urls import reverse
 
 from additional_codes import business_rules
 from additional_codes import validators
 from common.fields import ShortDescription
 from common.fields import SignedIntSID
+from common.models import DescriptionMixin
 from common.models import TrackedModel
 from common.models import ValidityMixin
 from footnotes import business_rules as footnotes_business_rules
@@ -96,7 +99,7 @@ class AdditionalCode(TrackedModel, ValidityMixin):
         )
 
 
-class AdditionalCodeDescription(TrackedModel, ValidityMixin):
+class AdditionalCodeDescription(TrackedModel, ValidityMixin, DescriptionMixin):
     """
     The additional code description contains the description of the additional
     code for a particular period.
@@ -129,6 +132,18 @@ class AdditionalCodeDescription(TrackedModel, ValidityMixin):
         return self.identifying_fields_to_string(
             identifying_fields=("described_additional_code", "valid_between"),
         )
+
+    def get_url(self, **kwargs):
+        try:
+            return reverse(
+                f"{self.get_url_pattern_name_prefix()}-ui-edit",
+                kwargs={
+                    "described_additional_code__sid": self.described_additional_code.sid,
+                    "description_period_sid": self.description_period_sid,
+                },
+            )
+        except NoReverseMatch:
+            return
 
     class Meta:
         ordering = ("valid_between",)
--- a/additional_codes/views.py
+++ b/additional_codes/views.py
@@ -8,8 +8,10 @@ from rest_framework import viewsets
 
 from additional_codes.filters import AdditionalCodeFilter
 from additional_codes.filters import AdditionalCodeFilterBackend
+from additional_codes.forms import AdditionalCodeDescriptionForm
 from additional_codes.forms import AdditionalCodeForm
 from additional_codes.models import AdditionalCode
+from additional_codes.models import AdditionalCodeDescription
 from additional_codes.models import AdditionalCodeType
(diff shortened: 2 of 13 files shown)


additional_codes/models.py
```diff
@@ -121,7 +122,7 @@ class AdditionalCodeDescription(TrackedModel, ValidityMixin):
     )
     description = models.TextField()
 
-    identifying_fields = ("description_period_sid",)
+    identifying_fields = ("description_period_sid", "described_additional_code__sid")
```
simonwo, line 125: Hm, are you sure? The `description_period_sid` should be enough to identify description models?

common/models/mixins.py
```diff
@@ -51,3 +51,11 @@ def objects_with_validity_field(cls):
 
     class Meta:
         abstract = True
+
+
+class DescriptionMixin:
+    def get_described_object(self):
+        for attr in self.__dict__:
```
NDevox, line 58: Let's not use `__dict__`, it's pretty funky.

Best to do something like:

```python
for rel, _ in self.get_relations():
    if rel.name.startswith("described_"):
        return getattr(self, rel.name)
```

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
```diff
@@ -13,7 +13,7 @@
     {% set table_rows = [] %}
     {% for obj in page_obj.object_list %}
       {% set object_link -%}
-        <a class="govuk-link govuk-!-font-weight-bold" href="{{ obj.get_url() or "#" }}">{{ obj.structure_code }}</a>
+        <a class="govuk-link govuk-!-font-weight-bold" href="{{ obj.get_url() or "#" }}">{{ obj.get_described_object() if obj.get_described_object else obj.structure_code }}</a>
```
NDevox, line 16: Is this right? And does this work when there isn't a `get_described_object`

additional_codes/views.py
```diff
@@ -99,5 +113,34 @@ def get_object(self, queryset: Optional[models.QuerySet] = None) -> models.Model
...
+    DraftUpdateView,
+):
+    form_class = AdditionalCodeDescriptionForm
+    permission_required = "common.change_trackedmodel"
+    template_name = "common/edit_description.jinja"
+
+    def get_object(self, queryset: Optional[models.QuerySet] = None) -> models.Model:
+        obj = super().get_object(queryset)
+
+        if self.request.method == "POST":
```
NDevox, line 129: This looks like it should be something we do for all update views? Maybe a generic mixin?

workbaskets/jinja2/includes/workbaskets/workbasket-detail.jinja
DIT-CJWaszczuk: Tried it with a list of all possible changes and it seems to hold up. I think if I tried to call the attribute it would break but the reference seems to be fine.

andyhd (changes requested):
Looking good, just a couple of issues

additional_codes/views.py
andyhd: We tried that before, but ran into some problems with inheritance. We should take another look at some point.

common/models/mixins.py
andyhd: Much better!

additional_codes/models.py
andyhd: I think this is because `get_url` tries to build a URL using `identifying_fields`. The correct solution here is not to add to `identifying_fields`, but to override `get_url` in `AdditionalCode`.

additional_codes/views.py
DIT-CJWaszczuk: Think we had a crack at this before and it just doesn't override get_object properly, ends up breaking

additional_codes/models.py
andyhd: This is the change that is making the tests fail, btw

andyhd (approved):

additional_codes/models.py
```diff
@@ -130,6 +133,18 @@ def __str__(self):
...
+        try:
+            return reverse(
+                f"{self.get_url_pattern_name_prefix()}-ui-edit",
+                kwargs={
+                    "described_additional_code__sid": self.described_additional_code.sid,
+                    "description_period_sid": self.description_period_sid,
+                },
+            )
+        except NoReverseMatch:
+            return
```
stuaxo, line 146: Can't work out what I think about this return, normally it would be a `pass` as it's not really short-cutting any code.
Not a blocker though.

andyhd: Yeah, it should possibly be an explicit `return None`

PR #268, opened 2021-03-31
TP-655: Added more tests for the importer system.

NDevox:
Added tests for the nursery caching interface as well as the tasks.

There's still elements of the importer untested, and lots of edge cases which could have done with testing. But no time to do it in!

I'll document some of the key bits that are missing on the ticket.

### DIFF
--- a/common/tests/factories.py
+++ b/common/tests/factories.py
@@ -15,6 +15,7 @@ from common.tests.models import TestModelDescription1
 from common.tests.util import Dates
 from common.validators import ApplicabilityCode
 from common.validators import UpdateType
+from importer.models import ImporterChunkStatus
 from measures.validators import DutyExpressionId
 from measures.validators import ImportExportCode
 from measures.validators import MeasureTypeCombination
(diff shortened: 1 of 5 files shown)


importer/tests/conftest.py
```diff
@@ -107,8 +111,33 @@ def prepped_handler_with_dependencies2(
...
+def batch_dependency() -> models.BatchDependencies:
+    dependencies = factories.BatchDependenciesFactory.create()
+    factories.ImporterXMLChunkFactory.create(batch=dependencies.depends_on)
+    factories.ImporterXMLChunkFactory.create(batch=dependencies.dependent_batch)
+    return dependencies
+
+
+@pytest.fixture
+def transaction_element() -> etree.Element:
+    pass
```
simonwo, line 143: What's this for?

simonwo (commented):
Looks like a good set.

importer/tests/conftest.py
NDevox: Ah was me starting on the chunker tests before being mildly overwhelmed by them,

common/tests/factories.py
```diff
@@ -933,8 +934,22 @@ class Meta:
...
     class Meta:
         model = "importer.ImporterXMLChunk"
 
     batch = factory.SubFactory(ImportBatchFactory)
+    chunk_number = 1
+    status = ImporterChunkStatus.WAITING
+    chunk_text = """\
+<?xml version="1.0" encoding="UTF-8"?>
+<env:envelope xmlns="urn:publicid:-:DGTAXUD:TARIC:MESSAGE:1.0" xmlns:env="urn:publicid:-:DGTAXUD:GENERAL:ENVELOPE:1.0" id="1">
+</env:envelope>"""
```
stuaxo, line 947: I tend to use `textwrap.dedent()` to help with this, but I guess there is a little overhead.

simonwo (approved):

PR #275, opened 2021-04-07
TAMATO-19: Fix validity check for geo memberships

simonwo:
The rule as written specifies that the group must
span the *membership* period of all of its members,
but our code checked that the group spanned the
*validity* period of its members.

This commit fixes that and introduces more tests.


stuaxo (approved):

stuaxo:
LGTM

PR #276, opened 2021-04-08
Calculate code coverage correctly

simonwo:
At the moment our code coverage is not reported correctly. For some reason `pytest-cov` seems to initialise after Django has already loaded all of the code, meaning that our imports and definitions are reported as not hit. This is unhelpful because it clouds the real areas in the code that do not have adequate coverage.

This commit changes to calling `coverage` directly which correctly captures all of the statements.


stuaxo (approved):
Good catch.

PR #277, opened 2021-04-08
Spring clean

simonwo:
Fixing up the coverage has revealed a few areas where we have unused or untested code. This commit addresses the lowest of the low-hanging fruit from these.

- Clean-up of dead or duplicated code ed6d329
  * Remove dead or unused code
  * Remove instances of raises_if duplication
  * Improve type hints in some places
- Remove use of hard-coded additional code type list 90926fb
  * Also add a generic helper for generating a list of type choices.

### DIFF
--- a/common/models/records.py
+++ b/common/models/records.py
@@ -12,7 +12,6 @@ from django.db.models import F
 from django.db.models import Field
 from django.db.models import Max
 from django.db.models import Q
-from django.db.models import QuerySet
 from django.db.models import Value
 from django.db.models import When
 from django.db.models.query_utils import DeferredAttribute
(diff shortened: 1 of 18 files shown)


common/models/records.py
```diff
@@ -34,7 +33,7 @@
 
 
 class TrackedModelQuerySet(PolymorphicQuerySet):
-    def latest_approved(self) -> QuerySet:
+    def latest_approved(self) -> TrackedModelQuerySet:
```
andyhd, line 36: I think you may have to quote the `TrackedModelQuerySet` return type when in the `TrackedModelQuerySet` class definition, or mypy will complain.

andyhd (approved):
Always nice to delete stuff!

common/models/records.py
simonwo: So apparently if you use `from __future__ import annotations` that seems to be enough to allow self-referential typing. It certainly works on Pylance, not tested on mypy.

Given that we’ve not really settled on using a specific type checker, it’s all a bit “works on my machine”. For now let’s merge this and if it’s causing people issues we can revisit?

PR #279, opened 2021-04-12
Move or remove the Transition Period importers

simonwo:
The time has come. These served their purpose but if we were to do this kind of exercise again, most likely we'd do it in a very different way. We'll always have these in git if we need to refer back to them.

We've extracted most of the useful code from these into other parts of the app already – these commits do a final shift of the `MeasureEndingPattern` which becomes a simple model method when all the importer stuff is stripped away.

There are some bits that are still sort of useful that we should look to replace, but these will take more concentrated work and don't need to block this. They are:
- the `SheetImporter`, which would be useful to have back for adding things to prod that aren't in XML – but this isn't even the latest version as there is a better one on the fta branch.
- the `GoodsNomenclatureTreeCollector` which will come in handy for avoiding ME32 automatically, but again which will need some work.


stuaxo (approved):

stuaxo:
I almost want a double-approval, when it comes to removing this amount of code.

simonwo:
Happy to wait and see what others think.

andyhd (approved):
💥

PR #282, opened 2021-04-14
TP-647: Check users have permission to edit

andyhd:
This commit modifies our generic Create and Update class-based views to inherit from PermissionRequiredMixin to ensure only users with permission to edit the tariff are allowed to do so. This saves having to use the mixin with every component create and update view class.

Also, I've added a first pass at a generic BDD test for edit views - this currently only tests that a 403 response is returned if the user does not have permission to edit. It would be nice to reduce the copy-pasta by making a generic "I edit \<model> <unique-id>" step, but it's a little out of scope for this PR.

As a bonus, I've added some BDD tests for additional codes and certificates edit views. I haven't touched the footnotes BDD tests for now.

### DIFF
--- a/certificates/tests/bdd/conftest.py
+++ b/certificates/tests/bdd/conftest.py
@@ -1,0 +1,20 @@
+import pytest
+from pytest_bdd import given
+
+from common.tests import factories
+
+pytestmark = pytest.mark.django_db
+
+
+@given("certificate X000", target_fixture="certificate_X000")
+def certificate_X000(date_ranges):
+    desc = factories.CertificateDescriptionFactory.create(
+        described_certificate=factories.CertificateFactory.create(
+            certificate_type=factories.CertificateTypeFactory.create(
+                sid="X",
+            ),
+            sid="000",
+        ),
+        description="This is X000",
+    )
+    return desc.described_certificate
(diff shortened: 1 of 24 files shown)


certificates/tests/bdd/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+    desc = factories.CertificateDescriptionFactory.create(
+        described_certificate=factories.CertificateFactory.create(
+            certificate_type=factories.CertificateTypeFactory.create(
+                sid="X",
+            ),
+            sid="000",
+        ),
+        description="This is X000",
+    )
+    return desc.described_certificate
```
simonwo, line 20: Mm yes, I can also taste some potential genericism here 😃

simonwo:
Do the BDD tests actually get run? I can't spot them in the test output. And the coverage would seem to suggest they do not: https://codecov.io/gh/uktrade/tamato/src/master/footnotes/tests/bdd/test_edit_footnote.py

andyhd:
The BDD tests are run, I see them in the test output. Codecov is having trouble linking to the PR for some reason, but I can see the expected coverage in the branch eg: https://codecov.io/gh/uktrade/tamato/src/TP-647-all-views-check-permissions/additional_codes/views.py

simonwo:
Ah yes, you're quite right. The ones I was looking at are marked `@todo` and hence aren't run, and then the _other_ ones I was expecting to work are currently marked `@xfail`.

stuaxo (approved):

PR #285, opened 2021-04-16
Quick refactor of the healthcheck endpoint

andyhd:
* Removed unnecessary \<strong> elements in XML payload
* A bit DRYer
* Added some test coverage


simonwo (approved):
Noice!

PR #286, opened 2021-04-19
TP-702 Edit certificate descriptions

DIT-CJWaszczuk:
## Summary 
Allow the user to edit certificate descriptions 
### List page
image
### Edit page
image
### Confirmation page 
image
### Tariff changes page
image


additional_codes/models.py
```diff
@@ -124,7 +124,7 @@ class AdditionalCodeDescription(TrackedModel, ValidityMixin, DescriptionMixin):
     )
     description = models.TextField()
 
-    identifying_fields = ("description_period_sid",)
+    identifying_fields = ("sid",)
```
simonwo, line 128: You can actually just delete this line as `sid` is the implied default.

footnotes/models.py
```diff
@@ -144,9 +144,9 @@ class FootnoteDescription(TrackedModel, ValidityMixin, DescriptionMixin):
         related_name="descriptions",
     )
     description = models.TextField()
-    description_period_sid = SignedIntSID(db_index=True)
+    sid = SignedIntSID(db_index=True)
 
-    identifying_fields = ("description_period_sid",)
+    identifying_fields = ("sid",)
```
simonwo, line 149: As per prev comment.

certificates/views.py
```diff
@@ -88,3 +88,42 @@ def get_object(self, queryset: Optional[models.QuerySet] = None) -> models.Model
...
+    def get_object(self, queryset: Optional[QuerySet] = None) -> Model:
+        obj = super().get_object(queryset)
+
+        if self.request.method == "POST":
+            obj = obj.new_draft(
+                WorkBasket.current(self.request),
+                save=False,
+            )
+
+        return obj
```
simonwo, line 122: This looks pretty generic… can we D.R.Y. it up somehow?

simonwo (commented):
LGTM, but I'm not very familiar with Django front-end stuff so will let someone else approve.

certificates/models.py
```diff
@@ -106,6 +109,24 @@ class CertificateDescription(TrackedModel, ValidityMixin):
         business_rules.ContiguousDescriptions,
     )
 
+    def get_url(self, action="detail"):
+        kwargs = {}
+        if action != "list":
+            kwargs = self.get_identifying_fields()
+        if action == "edit" or "confirm-update":
```
andyhd, line 116: These lines can be removed - if it's not a list view, then we need the kwargs. (Not that we have detail or list views of descriptions).

certificates/models.py
```diff
@@ -106,6 +109,24 @@ class CertificateDescription(TrackedModel, ValidityMixin):
...
+            kwargs = {
+                "described_certificate__certificate_type__sid": self.described_certificate.certificate_type.sid,
+                "described_certificate__sid": self.described_certificate.sid,
+                "sid": self.sid,
+            }
```
andyhd, line 121: You could potentially make this generic by doing something like
```python
kwargs = self.get_identifying_fields()  # this will be "sid"
rel = next(rel for rel in self.get_relations() if rel.name.startswith("described_"))
for field, value in self.get_described_object().get_identifying_fields().items():
    kwargs[f"{rel.name}__{field}"] = value
```
This could be DRYer, as `get_described_object()` does the same `for rel in self.get_relations() if ...` part. It could be factored out into a separate method and reused.

certificates/views.py
andyhd: I agree, can this be pulled into `DraftUpdateView`?

andyhd: Actually, can lines 116-120 be added to `TrackedModelDetailMixin`? I think that would solve the problem with inheritance.

DIT-CJWaszczuk: This seems to work, have pushed it up

DIT-CJWaszczuk:
@andyhd @simonwo Pushed up the changes, think this is good for a final look

additional_codes/models.py
```diff
@@ -99,7 +97,7 @@ def in_use(self):
         )
 
 
-class AdditionalCodeDescription(TrackedModel, ValidityMixin, DescriptionMixin):
+class AdditionalCodeDescription(DescriptionMixin, TrackedModel, ValidityMixin):
```
stuaxo, line 100: Seems weird having the concrete TrackedModel in the middle of the two mixins.   I'm sure the Mixins here are benign, but in other cases this sort of thing can cause strange behaviour.

andyhd: Yes, the order usually matters, so `TrackedModel` should probably be last.
The method resolution order is left-to-right, so if you were accessing `additionalcodedescription.foo()`, Python would try to find the `foo` method in `AdditionalCodeDescription` first, then `DescriptionMixin`, then `TrackedModel` and then `ValidityMixin`.
I think in this particular case, the mixins and base model don't interact with each other, so it doesn't make any difference.

common/models/mixins.py
```diff
@@ -59,3 +61,23 @@ def get_described_object(self):
...
+        if action != "list":
+            kwargs = self.get_identifying_fields()  # this will be "sid"
+            described_object_name = (
+                self.get_described_object()._meta.verbose_name.replace(" ", "_")
+            )
+            for field, value in (
+                self.get_described_object().get_identifying_fields().items()
+            ):
+                kwargs[f"described_{described_object_name}__{field}"] = value
+            print(kwargs)
```
andyhd, line 76: Print

common/models/mixins.py
```diff
@@ -59,3 +61,23 @@ def get_described_object(self):
             if rel.name.startswith("described_"):
                 return getattr(self, rel.name)
         raise AttributeError()
+
+    def get_url(self, action="detail"):
+        kwargs = {}
+        if action != "list":
+            kwargs = self.get_identifying_fields()  # this will be "sid"
```
andyhd, line 68: Don't need this comment

common/models/mixins.py
```diff
@@ -59,3 +61,23 @@ def get_described_object(self):
...
+            described_object_name = (
+                self.get_described_object()._meta.verbose_name.replace(" ", "_")
+            )
+            for field, value in (
+                self.get_described_object().get_identifying_fields().items()
```
andyhd, line 73: Don't call `get_described_object` twice

footnotes/models.py
```diff
@@ -121,7 +119,7 @@ class Meta:
         ordering = ["footnote_type__footnote_type_id", "footnote_id"]
 
 
-class FootnoteDescription(TrackedModel, ValidityMixin, DescriptionMixin):
+class FootnoteDescription(DescriptionMixin, TrackedModel, ValidityMixin):
```
andyhd, line 122: Is this the output of a linter? Seems incorrect behaviour if so.

workbaskets/views/generic.py
```diff
@@ -15,3 +19,14 @@ def get_transaction(self):
...
+    def get_object(self, queryset: Optional[QuerySet] = None) -> Model:
+        obj = super().get_object(queryset)
+
+        if self.request.method == "POST":
+            obj = obj.new_draft(
+                WorkBasket.current(self.request),
+                save=False,
+            )
+
+        return obj
```
andyhd, line 32: Is this meant to be here?

andyhd (commented):
Looks good, but some small things to tidy up still

common/models/mixins.py
DIT-CJWaszczuk: I've slightly tweaked this, the issue was predominantly that you could not perform the replace operation in the string format, so now there's a single call but two variables. Think it still retains the clean-ness without making a call for`get_described_object` twice

common/models/mixins.py
```diff
@@ -59,3 +61,22 @@ def get_described_object(self):
...
+
+    def get_url(self, action="detail"):
+        kwargs = {}
+        if action != "list":
+            kwargs = self.get_identifying_fields()
+            described_object = self.get_described_object()
+            described_object_name = described_object._meta.verbose_name.replace(
+                " ",
+                "_",
+            )
```
andyhd, line 73: I don't want to hold up this PR any more, so I think we should put a ticket on the backlog to use `described_object._meta.model_name` here instead, and change the `described_{model}` fields in the `{model}Description` classes to match.

andyhd (approved):
Nice, nearly deleted more than you added 👍

simonwo (approved):

additional_codes/import_parsers.py
```diff
@@ -88,7 +88,7 @@ class AdditionalCodeDescriptionParser(Writable, ElementParser):
 
     tag = Tag("additional.code.description")
 
-    description_period_sid = TextElement(Tag("additional.code.description.period.sid"))
+    sid = TextElement(Tag("additional.code.description.period.sid"))
     described_additional_code__sid = TextElement(Tag("additional.code.sid"))
     described_additional_code__type__sid = TextElement(Tag("additional.code.type.id"))
     described_additional_code__code = TextElement(Tag("additional.code"))
```
andyhd, line 94: these need changing to `described_additionalcode`

additional_codes/jinja2/taric/additional_code_description.xml
```diff
@@ -3,7 +3,7 @@
 
 {% call base_record(message_counter(), record.period_record_code, record.period_subrecord_code, sequence(), transaction_id, record.update_type) %}
 <oub:additional.code.description.period>
-    <oub:additional.code.description.period.sid>{{ record.description_period_sid }}</oub:additional.code.description.period.sid>
+    <oub:additional.code.description.period.sid>{{ record.sid }}</oub:additional.code.description.period.sid>
     <oub:additional.code.sid>{{ record.described_additional_code.sid }}</oub:additional.code.sid>
     <oub:additional.code.type.id>{{ record.described_additional_code.type.sid }}</oub:additional.code.type.id>
     <oub:additional.code>{{ record.described_additional_code.code }}</oub:additional.code>
```
andyhd, line 9: and these

additional_codes/jinja2/taric/additional_code_description.xml
```diff
@@ -12,7 +12,7 @@
 {% endcall %}
 {% call standard_record(message_counter(), sequence(), transaction_id, record) %}
 <oub:additional.code.description>
-    <oub:additional.code.description.period.sid>{{ record.description_period_sid }}</oub:additional.code.description.period.sid>
+    <oub:additional.code.description.period.sid>{{ record.sid }}</oub:additional.code.description.period.sid>
     <oub:language.id>EN</oub:language.id>
     <oub:additional.code.sid>{{ record.described_additional_code.sid }}</oub:additional.code.sid>
     <oub:additional.code.type.id>{{ record.described_additional_code.type.sid }}</oub:additional.code.type.id>
```
andyhd, line 18: same

additional_codes/models.py
```diff
@@ -115,41 +113,22 @@ class AdditionalCodeDescription(TrackedModel, ValidityMixin, DescriptionMixin):
...
     )
     description = models.TextField()
 
-    identifying_fields = ("description_period_sid",)
-
     indirect_business_rules = (business_rules.ACN5,)
 
     def __str__(self):
         return self.identifying_fields_to_string(
             identifying_fields=("described_additional_code", "valid_between"),
```
andyhd, line 129: same here

additional_codes/serializers.py
```diff
@@ -90,7 +93,8 @@ class Meta:
 
 @TrackedModelSerializer.register_polymorphic_model
 class AdditionalCodeDescriptionSerializer(
-    ValiditySerializerMixin, TrackedModelSerializerMixin
+    ValiditySerializerMixin,
+    TrackedModelSerializerMixin,
 ):
     described_additional_code = AdditionalCodeSerializer(read_only=True)
```
andyhd, line 99: and here

additional_codes/serializers.py
```diff
@@ -99,7 +103,7 @@ class Meta:
         fields = [
             "described_additional_code",
```
andyhd, line 104: and here

additional_codes/serializers.py
```diff
@@ -113,17 +117,18 @@ class Meta:
 
 
 class AdditionalCodeDescriptionImporterSerializer(
-    ValiditySerializerMixin, TrackedModelSerializerMixin
+    ValiditySerializerMixin,
+    TrackedModelSerializerMixin,
 ):
     described_additional_code = AdditionalCodeSerializer(required=False)
```
andyhd, line 123: ditto

additional_codes/serializers.py
```diff
@@ -113,17 +117,18 @@ class Meta:
...
-    description_period_sid = serializers.IntegerField()
+    sid = serializers.IntegerField()
 
     class Meta:
         model = models.AdditionalCodeDescription
         fields = [
             "described_additional_code",
```
andyhd, line 129: ditto

common/tests/factories.py
```diff
@@ -367,7 +367,7 @@ class AdditionalCodeDescriptionFactory(TrackedModelMixin, ValidityFactoryMixin):
     class Meta:
         model = "additional_codes.AdditionalCodeDescription"
 
-    description_period_sid = numeric_sid()
+    sid = numeric_sid()
     described_additional_code = factory.SubFactory(AdditionalCodeFactory)
```
andyhd, line 371: and finally, this one

andyhd (changes requested):

andyhd (approved):

stuaxo (approved):

PR #287, opened 2021-04-20
TP-683 Test for loading indents out of order

simonwo:
We have seen an issue where nodes in our tree appear to be out of order and the possible explanation is that they were out of order in the source data. This should have been fixed in 8282d84b but this test makes doubly sure that what was in the data really should not result in the observed behaviour.

The next likely explanation is legacy errors in the data left over after the bug has been fixed.


andyhd (approved):
Looks good to me

PR #288, opened 2021-04-21
TP-690: Model in use rules should respect deletes

andyhd:
Some business rules (particularly ones which prevent deleting a record) apply if a
record is currently "in use".

For example, if a `FootnoteType` is associated with at least one `Footnote` record, then
that `FootnoteType` is considered "in use".

But if all `Footnote`s that had been associated with the type were then deleted, the 
`FootnoteType` would still consider itself "in use".

Similarly, if a `Footnote` was subsequently associated with the type, but the transaction
containing that association was not yet approved, the `FootnoteType` would still consider
itself "in use".

This change corrects this class of business rules to correctly exclude deleted records
and those in "future" transactions from consideration of whether a record is "in use".

### DIFF
--- a/additional_codes/models.py
+++ b/additional_codes/models.py
@@ -92,7 +92,7 @@ class AdditionalCode(TrackedModel, ValidityMixin):
             self.measure_set.model.objects.filter(
                 additional_code__sid=self.sid,
             )
-            .latest_approved()
+            .approved_up_to_transaction(self.transaction)
             .exists()
         )
 
--- a/additional_codes/tests/test_models.py
+++ b/additional_codes/tests/test_models.py
@@ -1,0 +1,14 @@
+import pytest
+
+from common.tests import factories
+
+pytestmark = pytest.mark.django_db
+
+
+def test_additional_code_in_use(in_use_check_respects_deletes):
+    assert in_use_check_respects_deletes(
+        factories.AdditionalCodeFactory,
+        "in_use",
+        factories.MeasureFactory,
+        "additional_code",
+    )
--- a/certificates/models.py
+++ b/certificates/models.py
@@ -31,8 +31,11 @@ class CertificateType(TrackedModel, ValidityMixin):
     )
 
     def in_use(self):
-        # TODO handle deletes
-        return Certificate.objects.filter(certificate_type__sid=self.sid).exists()
+        return (
+            Certificate.objects.filter(certificate_type__sid=self.sid)
+            .approved_up_to_transaction(self.transaction)
+            .exists()
+        )
 
     def __str__(self):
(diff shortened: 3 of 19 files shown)


simonwo:
I think this looks good!

I'm trying to work out if using `self.transaction` will always work, or whether we need to be able to pass in a transaction. I'm imagining the following:

1. Create a measure type in transaction 1. `in_use()` returns `False` because there are no measures using it yet.
2. Add a measure in transaction 2. `in_use()` still returns `False` because the measure was added in a later transaction.
3. Try to delete the measure type in transaction 3. `in_use()` returns `True` because there is a measure.

Clearly 3 is the whole point we have this rule, but does it matter about `in_use` not being correct in step 2? Certainly for direct business rules in shouldn't be a problem because you will always be running it on a new version at a later transaction. Could it ever cause a problem for _indirect_ business rules? I can't think of/find any business rules that will use `in_use` that will also be run indirectly – am I right about that?

Do we think we will use `in_use` for anything else where 2 would be a problem? Is it possible we'll use it for searching? But maybe at that point we'll need to express it using a queryset anyway to have fast enough searches…

andyhd:
`MC4` and `MA2` are used in indirect rules - I'll test them

stuaxo (approved):

andyhd:
I think MA2 and MC4 don't make sense as indirect business rules and should be replaced or removed from MeasureCondition.

For example, if we have a MeasureCondition `cond` with a MeasureAction `action` and a MeasureConditionComponent `comp`, then MA2 says you cannot delete `action`, because it is in use by `comp`.

Deleting `comp` is fine and we can subsequently delete `action` as it is no longer in use.

If we delete `cond`, we trigger MA2 indirectly, iterating through all current MeasureActions with a relation to `cond` and running MA2 against them, as though we were trying to delete all of them. But deleting a MeasureCondition implies deleting all the associated MeasureConditionComponents, not deleting all the associated MeasureActions, so MA2 should not apply.

simonwo (approved):
Agreed – I don't think `MA2` or `MC4` should be indirect business rules on `MeasureCondition`, as changes to that model shouldn't ever cause them to be broken. 

What I would say though is that there isn't anything actually ensuring the `MeasureConditionComponent`s are deleted when the `MeasureCondition` is deleted, and the same for `MeasureComponent`s, so I'd expect there to be some orphaned components which would incorrectly trigger MA2. 

But we will almost never delete MeasureConditionCodes or MeasureActions anyway… so we can definitely deal with this as a low-priority task, either by ensuring orphans don't get created or perhaps by fixing `in_use` methods to ignore condition components that are attached to deleted conditions.

PR #289, opened 2021-04-26
Fix importer test shared state

andyhd:
The object_nursery fixture was maintaining state across tests, leading to failures.

This commit erases the nursery cache between each test.


DIT-CJWaszczuk (approved):

simonwo (approved):
So state in the cache wasn't being cleared between tests? Makes sense, nice one. 👍

PR #290, opened 2021-04-29
TP-666 Store origin quota certificate info alongside quotas

rachaelcodes:
Marks quotas as origin quotas based on the existence of required cetificates:

- Adds a `required_certificates` field to `QuotaOrderNumber`
- Adds an `is_origin_quota` property and a QuerySet annotation based on the presence of required certificates
- Updates required business rules
- Updates MeasureCreationPattern with required certificate information
- Displays required certificate data with Quotas
Screenshot 2021-04-29 at 09 27 18

_Note: there are a couple of places where I've ended up not using the default test structures, as I chose not to use a through model for the ManyToManyField. Let me know if you think this is an antipattern for this repo, and I'll change it._


quotas/serializers.py
```diff
@@ -30,6 +30,7 @@ class Meta:
             "taric_template",
             "start_date",
             "end_date",
+            "required_certificates",
```
andyhd, line 33: Is there a requirement to add this?

andyhd (commented):
Looks great, but a couple of questions

common/business_rules.py
```diff
@@ -300,6 +300,26 @@ def validate(self, model):
             raise self.violation(model)
 
 
+class ValidityPeriodSpansContainer(BusinessRule):
+    """Rule enforcing validity period of each contained value
+    spans the validity period of the container."""
```
simonwo, line 305: I'm suprised we don't already have something that does this… does the `ValidityPeriodContained` definitely not work for this? I'm sure you've worked it out, but I'm struggling to understand what this does that the other doesn't. But kudos for making it generic!

measures/patterns.py
```diff
@@ -393,9 +392,12 @@ def create(
             yield from self.get_authorised_use_measure_conditions(new_measure)
 
         # If this is a measure for an origin quota, we need to add
-        # some measure conditions with the passed proof of origin.
-        if any(proofs_of_origin):
-            yield from self.get_proof_of_origin_condition(new_measure, proofs_of_origin)
+        # some measure conditions with the origin quota required certificates.
+        if order_number and any(order_number.required_certificates.all()):
```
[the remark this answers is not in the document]
simonwo: Django definitely caches the result _within_ the queryset, so will definitely get the intended behaviour that way.

quotas/models.py
```diff
@@ -43,20 +45,28 @@ class QuotaOrderNumber(TrackedModel, ValidityMixin):
...
+        "certificates.Certificate", blank=True, related_name="quotas"
+    )
+
     indirect_business_rules = (
         business_rules.ON7,
         business_rules.ON8,
         business_rules.QBP2,
         business_rules.QD1,
         business_rules.QD7,
+        business_rules.CertificateValidityPeriodMustSpanQuotaOrderNumber,
```
simonwo, line 58: So if I understand correctly, the indirect business rules get called _only_ when a related model is updated… am I wrong about that? Otherwise it would be possible to add certificates that don't meet the validity requirement?

quotas/serializers.py
simonwo: No, I don't think we need to include this in the serializer.

quotas/querysets.py
```diff
@@ -0,0 +1,18 @@
+from django.db.models import BooleanField, Case, When
```
simonwo, line 1: The pre-commit hooks should have something to say about this… have you got them set up yet?

quotas/tests/test_querysets.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from common.tests import factories
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.mark.parametrize(
+    "number_of_certificates, is_origin_quota",
```
simonwo, line 9: I tend to prefer the split out form `("number_of_certificates", "is_origin_quota")` but it's not a big deal.

quotas/tests/test_importer.py
```diff
@@ -63,6 +63,23 @@ def test_quota_order_number_origin_exclusion_importer_update(
...
+
+
+def test_quota_order_number_required_certificate_update(update_imported_fields_match):
+    assert update_imported_fields_match(
+        factories.QuotaOrderNumberFactory,
+        serializers.QuotaOrderNumberSerializer,
+        dependencies={"required_certificates": [factories.CertificateFactory.create()]},
+    )
+
+
```
simonwo, line 82: I don't think these are actually harmful, but this might not be doing what you expect: these tests ensure that the XML produced by the serializer, when imported with the importer, results in the same fields. We don't output the certificates in the XML and we don't account for them in the importer… so the tests pass. More tests are good and I guess this ensures that we couldn't do one without the other, but it might be worth a comment here explaining that we don't actually output these things.

simonwo (commented):
Lots of good work here! Just some questions.

quotas/models.py
andyhd: Indirect business rules are called for every linked model - so if we create, update or delete a `QuotaOrderNumber`, the `CertificateValidityPeriodMustSpanQuotaOrderNumber` rule will be checked against every latest approved `Certificate` associated with the `QuotaOrderNumber`.

common/business_rules.py
andyhd: I think this is because `required_certificates` is a many-to-many field. But that makes me wonder if it should be - are certificates associated with more than one QuotaOrderNumber? I know order numbers are reused, but those would be different versions of the number, right?

simonwo: Unfortunately yes, the three EU origin quota all use [the same certificate](https://www.trade-tariff.service.gov.uk/certificate_search?type=U&code=178), for example.

rachaelcodes: Unfortunately, using `ValidityPeriodContained` throws an error: `AttributeError: type object 'ManyRelatedManager' has no attribute 'objects'` which I think is due to `required_certificates` being a many-to-many field.

quotas/tests/test_importer.py
rachaelcodes: Hm. I think I might take these out entirely then, so we're not giving a misleading sense of what is happening in the project.

quotas/models.py
```diff
@@ -43,20 +45,28 @@ class QuotaOrderNumber(TrackedModel, ValidityMixin):
         related_name="quotas",
     )
 
+    required_certificates = models.ManyToManyField(
+        "certificates.Certificate", blank=True, related_name="quotas"
```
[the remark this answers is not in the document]
rachaelcodes: Hmm when I try this, I get the warning: `null has no effect on ManyToManyField`: https://docs.djangoproject.com/en/3.2/ref/models/fields/#manytomany-arguments . Do we leave it as 'blank' so that users can add quotas without required certificates? Or leave it until an add/edit form is created?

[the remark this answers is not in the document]
andyhd: Ah yes, good point. I think you can remove `blank=True` anyway, it doesn't make sense to me to add an empty value to the list of required certificates.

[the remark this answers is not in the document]
simonwo: > users can add quotas without required certificates

However it’s achieved, yes this should be possible. Most quotas don’t need any certificates.

stuaxo (approved):

quotas/models.py
```diff
@@ -43,20 +45,30 @@ class QuotaOrderNumber(TrackedModel, ValidityMixin):
         related_name="quotas",
     )
 
+    required_certificates = models.ManyToManyField(
+        "certificates.Certificate",
+        related_name="quotas",
+        blank=True,
```
andyhd, line 51: I still think this `blank=True` is redundant. If a quota has no required certificates, then there will be no relations. But I guess it won't hurt to leave it.

quotas/tests/test_business_rules.py
```diff
@@ -247,6 +247,38 @@ def test_ON14():
...
+        required_certificates=[factories.CertificateFactory.create()],
+    )
+
+    for certificate in quota_order_number.required_certificates.all():
+        certificate.delete()
+
+        with pytest.raises(BusinessRuleViolation):
+            business_rules.CertificatesMustExist(certificate.transaction).validate(
+                certificate,
+            )
```
andyhd, line 262: `quota_order_number` is created with only 1 certificate. If it has more than one, this loop is masking a problem with shared context between tests.

andyhd (approved):
Looks good to me! A couple of comments, but nothing blocking.

andyhd (approved):

PR #291, opened 2021-04-29
Refactor URLs to use path converters

andyhd:
We make heavy use of `re_path` URL routes, and these have become a little messy and hard to read.

This change replaces those with custom path converters, which encapsulate the regular
expressions to match URL parameters and handle converting them to view parameters.

For example, we go from this:
```python
description_detail = r"^(?P<described_additionalcode__sid>\d*)"
...
    re_path(
        fr"{description_detail}/description/{DESCRIPTION_ID_PATTERN}/edit/$",
        ...
```
to this
```python
description_detail = "<sid:described_additionalcode__sid>/description/<sid:sid>"
...
    path(
        f"{description_detail}/edit/",
        ...
```


regulations/validators.py
```diff
@@ -62,19 +62,22 @@ class CommunityCode(models.IntegerChoices):
...
+    )|(
+      (Z|V)                           # national prefix
+      (\d{4})                         # national year
+      ([A-Z]{3})                      # national suffix
+    )|(
+      (IYY)                           # dummy prefix
+      (\d{5})                         # dummy suffix
+    )
+"""
+regulation_id_validator = RegexValidator(fr"^{REGULATION_ID_REGEX}$")
```
simonwo, line 80: The comments are included within the string – isn't this a problem? Looks like we don't have any tests for this validator… can we write some?

simonwo (commented):
Looks great, I support this!

regulations/validators.py
andyhd: The comments are not a problem because the regex is in free-spacing mode, which allows newlines and spaces and comments. But changing the regex string caused Django to insist on a new migration 🙄 - I can move the comments out, it will probably be better for it.
I will write some tests, good point!

simonwo: Ah, I never knew that was a thing. Cool!

andyhd: I decided to split the regex up and remove the inline comments as there's already a comprehensive comment above.

simonwo (approved):

DIT-CJWaszczuk (approved):

PR #292, opened 2021-04-29
TP-739 Fix certificate search to find type and code together

DIT-CJWaszczuk:
## Summary
Before, if you searched on type and sid, it would fail. Now you can search on type, sid or both.
## Screenshots
image

### DIFF
--- a/certificates/filters.py
+++ b/certificates/filters.py
@@ -3,6 +3,7 @@ from django.contrib.postgres.aggregates import StringAgg
 from django.urls import reverse_lazy
 
 from certificates import models
+from certificates.validators import COMBINED_CERTIFICATE_AND_TYPE_ID
 from common.filters import ActiveStateMixin
 from common.filters import LazyMultipleChoiceFilter
 from common.filters import TamatoFilter
@@ -11,9 +12,12 @@ from common.filters import type_choices
 
 class CertificateFilter(TamatoFilter, ActiveStateMixin):
     search_fields = (
+        StringAgg("certificate_type__sid", delimiter=" "),
         "sid",
         StringAgg("descriptions__description", delimiter=" "),
-    )  # XXX order is important
+    )
+
+    search_regex = COMBINED_CERTIFICATE_AND_TYPE_ID
 
     certificate_type = LazyMultipleChoiceFilter(
         choices=type_choices(models.CertificateType.objects.latest_approved()),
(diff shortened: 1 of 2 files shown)


certificates/filters.py
```diff
@@ -8,12 +10,19 @@
 from common.filters import TamatoFilter
 from common.filters import type_choices
 
+COMBINED_CERTIFICATE_AND_TYPE_ID = re.compile(
+    r"(?P<certificate_type__sid>[A-Z0-9])(?P<sid>[A-Z0-9]{3})",
```
andyhd, line 14: These regexes should be imported from `certificates.validators`

andyhd (changes requested):

certificates/validators.py
```diff
@@ -1,7 +1,13 @@
...
 
 CERTIFICATE_TYPE_SID_REGEX = r"^[A-Z0-9]{1}$"
 certificate_type_sid_validator = RegexValidator(CERTIFICATE_TYPE_SID_REGEX)
 
 CERTIFICATE_SID_REGEX = r"^[A-Z0-9]{3}$"
 certificate_sid_validator = RegexValidator(CERTIFICATE_SID_REGEX)
+
+COMBINED_CERTIFICATE_AND_TYPE_ID = re.compile(
+    r"(?P<certificate_type__sid>[A-Z0-9])(?P<sid>[A-Z0-9]{3})",
+)
```
andyhd, line 13: No, I mean you should reuse the existing regexes, eg:
```python
CERTIFICATE_TYPE_SID_REGEX = r"[A-Z0-9]{1}"
certificate_type_sid_validator = RegexValidator(fr"^{CERTIFICATE_TYPE_SID_REGEX}$")

CERTIFICATE_SID_REGEX = r"[A-Z0-9]{3}"
certificate_sid_validator = RegexValidator(fr"^{CERTIFICATE_SID_REGEX}$")

COMBINED_CERTIFICATE_AND_TYPE_ID = re.compile(
    fr"(?P<certificate_type__sid>{CERTIFICATE_TYPE_SID_REGEX})(?P<sid>{CERTIFICATE_SID_REGEX})",
)
```

That way, if we ever had to change them, we would only have to change them in one place.

andyhd (changes requested):

certificates/validators.py
DIT-CJWaszczuk: My head was not switched on!

simonwo (approved):
LGTM now.

PR #294, opened 2021-05-11
TP-695 Edit Measure detail

DIT-CJWaszczuk:
## Summary
Adds the measures edit page, with an autocomplete for most fields. 
## Caveats
This PR does not currently include the duties or geographical area fields, they'll be added after this. 
### Detail page 
image
### Edit page
image
### Confirmation page 
image
### Workbasket view
image

### DIFF
--- a/common/forms.py
+++ b/common/forms.py
@@ -9,8 +9,37 @@ from crispy_forms_gds.layout import Submit
 from django import forms
 from django.contrib.postgres.forms.ranges import DateRangeField
 from django.core.exceptions import ValidationError
+from django.forms.widgets import Widget
+from django.template import loader
+from django.utils.safestring import mark_safe
 
 from common.util import TaricDateRange
+
+
+class AutocompleteWidget(Widget):
+    template_name = "components/autocomplete.jinja"
+
+    def get_context(self, name, value, attrs=None):
+        display_string = ""
+        if value:
+            display_string = value.structure_code
+            if value.structure_description:
+                display_string = f"{display_string} - {value.structure_description}"
+
+        return {
+            "widget": {
+                "name": name,
+                "value": value.pk if value else None,
+                "label": self.attrs["label"],
+                "help_text": self.attrs["help_text"],
+                "display_value": display_string,
+            },
+        }
+
+    def render(self, name, value, attrs=None, renderer=None):
+        context = self.get_context(name, value, attrs)
+        template = loader.get_template(self.template_name).render(context)
+        return mark_safe(template)
 
 
 class DateInputFieldFixed(DateInputField):
--- a/common/static/common/js/autocomplete.js
+++ b/common/static/common/js/autocomplete.js
@@ -1,0 +1,75 @@
+import accessibleAutocomplete from 'accessible-autocomplete'
+
+const getAdditionalCodeResults = (query, populateResults) => {
+    return fetch(`/api/additional_codes/?search=${query}&format=json`)
+    .then(response => response.json())
+    .then(data => populateResults(data.results.map(result => ({value: result.id, string: `${result.type.sid}${result.code} - ${result.descriptions[0].description}`}))));
+}
+
+const getMeasureTypeResults = (query, populateResults) => {
+    return fetch(`/api/measure_types/?search=${query}&format=json`)
+    .then(response => response.json())
+    .then(data => populateResults(data.results.map(result =>({value: result.id, string: `${result.sid} - ${result.description}`}))));
+}
+
+const getCommodityCodeResults = (query, populateResults) => {
+    return fetch(`/api/goods_nomenclature/?search=${query}&format=json`)
+    .then(response => response.json())
+    .then(data => populateResults(data.results.map(result => ({value: result.id, string: `${result.item_id} - ${result.descriptions[0].description}`}))));
+}
+
+const getOrderNumberResults = (query, populateResults) => {
+    return fetch(`/api/quota_order_numbers/?search=${query}&format=json`)
+    .then(response => response.json())
+    .then(data => populateResults(data.results.map(result => ({value: result.id, string:`${result.order_number}`}))));
+}
+
+const getRegulationResults = (query, populateResults) => {
+    return fetch(`/api/regulations/?search=${query}&format=json`)
+    .then(response => response.json())
+    .then(data => populateResults(data.results.map(result => ({value: result.id, string: `${result.regulation_id} - ${result.information_text}`}))));
+}
+
+
+const getResults = {
+    "AdditionalCode": getAdditionalCodeResults,
+    "MeasureType": getMeasureTypeResults,
+    "GoodsNomenclature": getCommodityCodeResults,
+    "OrderNumber": getOrderNumberResults,
+    "GeneratingRegulation": getRegulationResults,
+}
+
+const template = (result) => result && result.string
+
+const createAutocomplete = (elementName) => { 
+    let camelcaseElement = elementName.replace(/[A-Z]/g, letter => `_${letter.toLowerCase()}`).substring(1);
+    let element = document.querySelector(`#${camelcaseElement}_view_container`);
+    let realInput = document.querySelector(`#${camelcaseElement}`)
+    if (element) {
+        accessibleAutocomplete({
+            element: element,
+            id: `${camelcaseElement}_autocomplete`,
+            source: getResults[elementName],
+            defaultValue: element.dataset.originalValue,
+            name: `${camelcaseElement}`, 
+            templates: {
+                inputValue: template,
+                suggestion: template
+              },
+            onConfirm: (val) => {
+                let autoCompleteElement = document.querySelector(`#${camelcaseElement}_autocomplete`)
+                if (val)
+                    return realInput.value = val.value
+                if (!autoCompleteElement.value) 
+                    return realInput.value = ""
+            },
+        })
+    }
+}
+
+const initAutocomplete = () => { 
+    const elements = ['AdditionalCode', 'MeasureType', 'GoodsNomenclature', 'OrderNumber', 'GeneratingRegulation']
+    elements.forEach((element) => createAutocomplete(element))
+}
+
+export default initAutocomplete;
--- a/common/static/common/scss/_accordion.scss
+++ b/common/static/common/scss/_accordion.scss
@@ -1,0 +1,7 @@
+/*
+  By default, GOV.UK accordions remove the margin on the last child in
+  an accordion section. This is an override so that we can keep it.
+*/
+.govuk-accordion__section-content > :last-child {
+  @extend .govuk-\!-margin-bottom-6;
+}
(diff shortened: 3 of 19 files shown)


common/static/common/js/autocomplete.js
```diff
@@ -0,0 +1,93 @@
+import accessibleAutocomplete from 'accessible-autocomplete'
+
+const getAdditionalCodeResults = (query, populateResults) => {
+    return fetch(`http://localhost:8000/api/additional_codes/?search=${query}&format=json`)
```
stuaxo, line 4: Do you need a way to get the url for the site to the javascript ?

No idea if how "correct" it is, but one way is outputting these in a django template somewhere that generates a JS file, or just copying it from the django HTML template you include.

common/static/common/scss/_accordion.scss
```diff
@@ -0,0 +1,4 @@
+// Override GOV.UK accordion having 0 padding on the last child
```
stuaxo, line 1: Probably nit-picking, but it's always nice to include something in the blerb about why or where this is used...

`... on the last child so this can be used inline`  <-- (reason I made up)

measures/forms.py
```diff
@@ -1,13 +1,226 @@
...
+
+        if not measure_type and self.instance and self.instance.measure_type:
+            measure_type = self.instance.measure_type
+
+        # combine start and end dates into date range
+        start_date = cleaned_data.pop("start_date", None)
+        end_date = cleaned_data.pop("end_date", None)
+        cleaned_data["valid_between"] = TaricDateRange(start_date, end_date)
+
+        return cleaned_data
```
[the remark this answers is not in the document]
DIT-CJWaszczuk: Yeah I haven't started on the clean method yet, I am probably going to need a BE dev to help me with this however!

common/filters.py
```diff
@@ -34,7 +34,10 @@
 from common.models.records import TrackedModelQuerySet
 from common.util import TaricDateRange
 
-ACTIVE_STATE_CHOICES = [Choice("active", "Active"), Choice("terminated", "Terminated")]
+ACTIVE_STATE_CHOICES = [
+    Choice("active", "Active"),
+    Choice("terminated", "Terminated"),
+]
```
andyhd, line 40: I expect this is just your linter doing this automatically, but I try to avoid including just formatting changes in a PR unless I have changed the code as it gives reviewers more to read unnecessarily.

measures/forms.py
```diff
@@ -1,13 +1,226 @@
...
+            f"{self.instance.additional_code} - {self.instance.additional_code.get_description().description}"
+            if self.instance.additional_code
+            else ""
+        )
+
+        self.fields["duties"].initial = (
+            self.instance.duty_sentence if self.instance.duty_sentence else ""
+        )
+        self.fields["start_date"].initial = self.instance.valid_between.lower or ""
+        self.fields["end_date"].initial = self.instance.valid_between.upper or ""
```
andyhd, line 98: The initial values of `start_date` and `end_date` should be a `datetime.date` or `None`, not an empty string. This would be better written as
```python
self.fields["start_date"].initial = self.instance.valid_between.lower
```

measures/forms.py
```diff
@@ -1,13 +1,226 @@
...
+    valid_between = GovukDateRangeField(required=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        initial_goods_nomenclature = (
+            f"{self.instance.goods_nomenclature} - {self.instance.goods_nomenclature.get_description().description}"
+            if self.instance.goods_nomenclature
+            else ""
+        )
```
andyhd, line 77: These ternary expressions are not necessary. It's more straightforward to write them like
```python
initial_goods_nomenclature = ""
if self.instance.goods_nomenclature:
    initial_goods_nomenclature = f"{self.instance.goods_nomenclature} - {self.instance...description}"
```

measures/forms.py
```diff
@@ -1,13 +1,226 @@
...
+
+        self.helper = FormHelper(self)
+        self.helper.layout = Layout(
+            Accordion(
+                AccordionSection(
+                    "Measure type",
+                    HTML(
+                        "<label for='measure_type' class='govuk-label'>Measure type</label>"
+                        "<div id='id_measure_type_hint' class='govuk-hint'>Select the appropriate measure type.</div>",
```
andyhd, line 112: I think this form is stretching the usefulness of Crispy forms. The MVC separation of concerns is getting lost. The accordion layout is irrelevant to the form class, it is purely presentational and belongs in templates. Using Crispy is not providing a benefit in this case.

andyhd (changes requested):
I think this case is not workable with Crispy forms. Crispy is good when you can replace a long template with a single call to `{% crispy form %}`, but it is IMO making things worse when you have a massive `Layout` object structure in the form class.

common/static/common/js/autocomplete.js
stuaxo: @andyhd  ^ yep, this :)   There's always `//` if you really need an absolute url.

common/jinja2.py
```diff
@@ -85,11 +85,16 @@ def query_transform(request, **kwargs):
...
+def transform_snake_case(word):
+    return word.replace("_", " ").capitalize()
+
+
 def environment(**kwargs):
     env = GovukFrontendEnvironment(**kwargs)
 
     env.globals.update(
         {
+            "transform_snake_case": transform_snake_case,
```
andyhd, line 97: This function name is only half self-documenting - it should be something like `snake_case_to_capwords` or similar, but the name is already long enough that the function doesn't save much typing over using the built-in Jinja filters: `{{ my_snake_case_string|replace("_", " ")|capitalize }}` or using the string methods inline: `{{ my_snake_case_string.replace("_", " ").capitalize() }}`.

stuaxo: snake_case_to_capwords is not too bad, if this is used in more than a couple of places.

andyhd:
Github Actions seems to be wedged. Maybe another push will kickstart it?

measures/forms.py
```diff
@@ -8,6 +8,102 @@
...
+            "generating_regulation"
+        ].initial = self.instance.generating_regulation.regulation_id
+        self.fields["start_date"].initial = self.instance.valid_between.lower
+        self.fields["end_date"].initial = self.instance.valid_between.upper
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        cleaned_data["sid"] = self.instance.sid
+        print("CLEANED DATA", cleaned_data)
```
stuaxo, line 92: Print statement.

measures/models.py
```diff
@@ -268,6 +274,12 @@ class MeasureType(TrackedModel, ValidityMixin):
         business_rules.MT10,
     )
 
+    def __str__(self):
+        return f"{self.sid}"
+
+    def __unicode__(self):
+        return f"{self.sid}"
```
stuaxo, line 281: Not sure __unicode__ is used any more in python3.

common/forms.py
```diff
@@ -9,10 +9,44 @@
...
 
+class AutocompleteWidget(Widget):
+    template_name = "components/autocomplete.jinja"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.attrs = kwargs["attrs"]
+
+    def get_context(self, name, value, attrs=None):
+        if value:
```
stuaxo, line 27: Only creating a variable on one branch of an `if` statement and then using it later makes me nervous someone might accidentally introduce a bug here later.

Could we initialise display_string to "" above the if, then always use display_string further down ?

common/static/common/js/autocomplete.js
```diff
@@ -0,0 +1,74 @@
+import accessibleAutocomplete from 'accessible-autocomplete'
+
+const getAdditionalCodeResults = (query, populateResults) => {
+    return fetch(`/api/additional_codes/?search=${query}&format=json`)
+    .then(response => response.json())
+    .then(data => populateResults(data.results.map(result => ({value: result.id, string: `${result.type.sid}${result.code} - ${result.descriptions[0].description}`}))));
+}
```
andyhd, line 7: These getXResults functions are all very similar. You could DRY this up a bit like:
```javascript
const autocompletes = {
  "additional_code": {
    "url": "/api/additional_codes/",
    "resultFormatter": (result) => {
      value: result.id,
      string: `${result.type.sid}${result.code} - ${result.descriptions[0].description}`
    },
  },
  ...
}
function autocompleteFetcher(url, resultFormatter) {
  return (query, populateResults) => {
    return fetch(`${url}?search=${query}&format=json`)
      .then(response => populateResults(response.json().results.map(resultFormatter)));
  };
}
function template(result) {
  return result && result.string;
}
const initAutocomplete = () => {
  for (const [elementName, params] of autocompletes) {
    const {url, resultFormatter} = params;
    let element = document.querySelector(`#${elementName}_view_container`);
    let realElement = document.querySelector(`#${elementName}`);
    if (element) {
      accessibleAutocomplete({
        element: element,
        id: elementName,
        source: autocompleteFetcher(url, resultFormatter),
        defaultValue: element.dataset.originalValue,
        name: elementName, 
        templates: {
          inputValue: template,
          suggestion: template
        },
        onConfirm: (val) => {
          realElement.value = val.value
        }
      });
    }
  }
}
```

measures/forms.py
```diff
@@ -8,6 +8,97 @@
...
+    )
+    start_date = DateInputFieldFixed(
+        label="Measure start date",
+    )
+    end_date = DateInputFieldFixed(
+        label="Measure end date",
+        help_text="Leave empty if the measure is needed for an unlimited time",
+        required=False,
+    )
+    valid_between = GovukDateRangeField(required=False)
```
andyhd, line 70: You can use `common.forms.ValidityPeriodForm` as the superclass and remove these lines and the validity period initial values and validation

measures/models.py
```diff
@@ -270,6 +270,9 @@ class MeasureType(TrackedModel, ValidityMixin):
         business_rules.MT10,
     )
 
+    def __str__(self):
+        return f"{self.sid}"
```
andyhd, line 274: This is probably better as `return str(self.sid)`

measures/views.py
```diff
@@ -1,30 +1,73 @@
+from typing import Type
```
andyhd, line 1: With python 3.9, you can use the builtin `type` for type hints and remove this import, eg:
`model: type[TrackedModel] = Measure`

regulations/models.py
```diff
@@ -208,7 +212,7 @@ def is_draft_regulation(self):
         return self.regulation_id.startswith("C")
 
     def __str__(self):
-        return f"{self.regulation_id} ({self.get_role_type_display()})"
+        return f"{self.regulation_id}"
```
andyhd, line 215: `return str(self.regulation_id)`

andyhd (changes requested):
Only one blocking issue, I think - one of the autocomplete URLs is `localhost:8000`

common/static/common/js/autocomplete.js
andyhd: Also, why are some functions defined with `function name() {}` and others with `const name = () => {}`?

andyhd (approved):
👍

PR #295, opened 2021-05-11
Refactor models that only use a validity start date

simonwo:
There are several models in the TARIC schema that model some time-varying property of a parent model. For example, the Description models all track the change in description for their parent object over time. These models only use a validity start date instead of a full validity range because at least one of them must be valid at any time.

So far, we have just contained these dates in a normal validity range field, and made sure that the end date is always blank. This is problematic for two good reasons: one, the fact that the end date is always blank is not obvious from the model so it's likely to lead to programming mistakes, and two, it's difficult to work out what the real date range is.

This commit removes the full validity range and replaces it with just the validity start date, and then provides some methods on a queryset to recover the full validity range from the data that is present. Users can now call `with_end_date()` on models with just a start date, and the database will compute the implicit end date by itself.

(I did most of this as a result of trying to analyse indents in TP-683, so I thought I might as well finish it off.)


andyhd:
Sorry, I did look at this! I posthumously approve this PR.

PR #296, opened 2021-05-13
Add footnote applicability check to commodity codes

simonwo:
Only footnotes with a footnote type that have one of three possible application codes are allowed to be applied to commodity codes. This commit implements the business rule NIG18 that ensures only footnotes with the right type can be associated.

It also DRYs up some similar logic for measures, and makes what footnote codes should apply a property of the models themselves.

### DIFF
--- a/common/business_rules.py
+++ b/common/business_rules.py
@@ -388,3 +388,26 @@ class DescriptionsRules(BusinessRule):
             filter(lambda x: x > model.valid_between.upper, valid_froms),
         ):
             raise self.generate_violation(model, "start after end")
+
+
+class FootnoteApplicability(BusinessRule):
+    """Check that a footnote type can be applied to a certain model, based on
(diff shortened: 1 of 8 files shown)


common/business_rules.py
```diff
@@ -388,3 +388,26 @@ def validate(self, model):
...
+    """Check that a footnote type can be applied to a certain model, based on
+    the set of :class:`~footnote.validators.ApplicationCode` that the model
+    reports is valid for itself."""
+
+    applicable_field: str
+    """The field containing a model to check applicability for. It must have a
+    property or attribute called :attr:`footnote_application_codes`."""
+
+    footnote_type_field: str = "associated_footnote__footnote_type"
+    """The field containing a footnote type who's applicability is being checked."""
```
andyhd, line 403: ~who's~ whose

But maybe this should be "The field containing a footnote type to check for applicability"?

andyhd (approved):
Looks good to me

PR #297, opened 2021-05-13
TP-730: Algorithm for automated commodity code extraction.

simonwo:
This ADR proposes that we automatically extract EU commodity codes from XML files by adding new logic to the importer to only save goods nomenclature-related models.

That's the easy bit. Unfortunately we also have to deal with the fact that many commodity code changes can break our own business rules. We will need to handle these rule violations by automatically editing or deleting models as the goods nomenclature updates are imported. This ADR goes into a lot of detail on all of the business rule violations we can expect and how to deal with them correctly.

I'd be greatful if everyone could have a read because someone is going to have to implement it!

image


docs/adr/0013-import-taric-nomenclature-updates.rst
```diff
@@ -0,0 +1,485 @@
...
+goods nomenclature is now a parent of something where before it was not, and
+the measures defined on that nomenclature now overlap with another measure.
+
+The only way in which this can cause a problem is if a measure defined on a
+goods nomenclature that has just moved now overlaps with child measures. This is
+because when a goods nomenclature moves via the indent mechanism it can only
+move up or down the same subtree – moving down wouldn't create any more overlaps
+that already existed, and moving up can only bring into view measures that are
+defined on siblings.
+
```
kristof-bb, line 319: Not sure if i understand this paragraph. What do you mean with "moving down wouldn't create any more overlaps that already existed". In scenario 6 for example the code moves down and measure now overlaps with parent?

docs/adr/0013-import-taric-nomenclature-updates.rst
```diff
@@ -0,0 +1,485 @@
...
+As above, the only course of action is to delete the remaining associations.
+
+ME32
+~~~~
+
+.. autoclass:: measures.business_rules.ME32
+  :noindex:
+
+In this case an indent of a goods nomenclature has been changed such that a
+goods nomenclature is now a parent of something where before it was not, and
```
kristof-bb, line 310: Or a code deletion (scenario 3)?

docs/adr/0013-import-taric-nomenclature-updates.rst
```diff
@@ -0,0 +1,485 @@
...
+~~~~~~~~~~
+
+.. autoclass:: measures.business_rules.ME1
+  :noindex:
+.. autoclass:: measures.business_rules.ME7
+  :noindex:
+
+It is technically possible to have either of the product code or suffix of a
+goods nomenclature updated, which would either cause a duplicate or turn it from
+real to intermediate code where measures cannot be created.
```
kristof-bb, line 352: maybe better to group ME1 with ME32 in the title as they very similar but quite different from ME7?

kristof-bb (approved):
Looks great!

docs/adr/0013-import-taric-nomenclature-updates.rst
simonwo: Yeah I wrote this before you showed me scenario 6. I think this paragraph is erroneous and I was assuming that changing an indent couldn't put you underneath a sibling, but it can.

stuaxo:
When it comes to the explanations of the business rules, particularly stuff like {outer thing has validity dates} {child things validity dates must: span {outer things validity dates}} it feels like some diagrams could help.

(Since the sentences kind of hurt the brain to read even though they are fairly straightforward)

Maybe we can come with a small diagram that shows the particular part of the hierarchy, that the business rule deals with, with relevant fields highlighted could be useful for visualisation.

PR #298, opened 2021-05-17
Clean up XML and importer tests

simonwo:
Here's some self-contained work that has fallen out of the XML performance branch – updates and deduplication of lots of test logic around the import and export features. Merging this now will give a much smaller PR for the XML work later.

> Our importer tests are quite verbose – in particular the same tests are repeated once for creation and again for delete/update. The two tests also have different styles but are functionally equivalent.
>
> There is no real need for this, so this commit removes this duplication. We now instead can simply call `imported_fields_match` once and this will test the importer for all variations of update type.

> Our XML tests validate that the XML we output is valid according to the TARIC3 spec and contain the elements we expect. They do this by checking for tags named correctly. They should also check that the namespace is correct – ultimately, tags that are named correctly but in the wrong namespace are not what we want. 
>
> This commit is a find/replace to add the correct namespaces to all of these checks.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -6,7 +6,6 @@ from typing import Callable
 from typing import Dict
 from typing import Optional
 from typing import Type
-from typing import Union
 from unittest.mock import PropertyMock
 from unittest.mock import patch
 
@@ -26,8 +25,6 @@ from pytest_bdd import parsers
 from rest_framework.test import APIClient
 
 from common.models import TrackedModel
-from common.models.mixins.validity import ValidityMixin
-from common.models.mixins.validity import ValidityStartMixin
 from common.serializers import TrackedModelSerializer
 from common.tests import factories
 from common.tests.util import Dates
@@ -312,9 +309,9 @@ def validity_period_contained(date_ranges):
 
 
 @pytest.fixture
-def imported_fields_match(valid_user, settings):
+def run_xml_import(valid_user, settings):
     """
-    Provides a function for checking a model can be imported correctly.
+    Returns a function for checking a model can be imported correctly.
 
     The function takes the following parameters:
         model: A model instance, or a factory class used to build the model.
@@ -331,18 +328,17 @@ def imported_fields_match(valid_user, settings):
     """
 
     def check(
-        model: Union[TrackedModel, Type[DjangoModelFactory]],
+        factory: Callable[[], TrackedModel],
         serializer: Type[TrackedModelSerializer],
     ) -> TrackedModel:
         get_nursery().cache.clear()
         settings.SKIP_WORKBASKET_VALIDATION = True
-        if isinstance(model, type) and issubclass(model, DjangoModelFactory):
-            model = model.build(update_type=UpdateType.CREATE)
 
+        model = factory()
         assert isinstance(
             model,
             TrackedModel,
-        ), "Either a factory or an object instance needs to be provided"
+        ), "A factory that returns an object instance needs to be provided"
 
         xml = generate_test_import_xml(
             serializer(model, context={"format": "xml"}).data,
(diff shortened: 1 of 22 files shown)


common/tests/factories.py
```diff
@@ -710,6 +710,13 @@ def data(self):
                 "target.quota.definition.sid": "1",
             }
 
+    @classmethod
+    def for_subrecord_code(cls, sub_code: str):
+        class SpecialisedFactory(cls):
```
stuaxo, line 715: This could use a docstring to say what it is / what it is for.

conftest.py
```diff
@@ -312,7 +310,7 @@ def check(dependency_name, dependency_factory, dependent_factory):
 
 
 @pytest.fixture
-def imported_fields_match(valid_user, settings):
+def run_xml_import(valid_user, settings):
     """
     Provides a function for checking a model can be imported correctly.
```
stuaxo, line 315: Possibly "Return a function that verifies a ..." may be less ambiguous - though maybe "Provides a .." is a pytest thing ?

conftest.py
```diff
@@ -375,89 +372,93 @@ def check(
...
-    Provides much the same functionality as imported_fields_match, however makes
-    some adjustments for updates and deletes.
+    Provides much the same functionality as run_xml_import, however makes some
+    adjustments for updates and deletes.
 
-    In addition to imported_fields_match a previously created object is
-    generated. The data around version groups is also tested.
+    In addition to run_xml_import a previously created object is generated. The
+    data around version groups is also tested.
     """
```
stuaxo, line 395: If the language here is "previously created object", there should be a comment further down with similar language so it's easy to see where the code that creates the "previously created object".

A hint about the functionality provided that is similar to run_xml_import would be nice here so the user didn't have to immediately go there to check that docstring.

Creates an existing object, then import

common/tests/factories.py
simonwo: Actually, there's a far simpler way of doing this so I've just outright deleted this.

conftest.py
simonwo: I have renamed `parent_model` to `previous_version` to reflect this.

conftest.py
andyhd: No, that's just the way I wrote it - writing code is often easier than writing comments 😆

conftest.py
```diff
@@ -375,89 +372,99 @@ def check(
...
-                ):
-                    kwargs["validity_start"] = validity[1].lower
-
-            model = model.build(
-                update_type=update_type,
-                **kwargs,
-            )
-        elif not parent_model:
-            raise ValueError("parent_model must be defined if an instance is provided")
+        if validity:
```
andyhd, line 439: Should there be an `else:` here? Otherwise, the `valid_between` or `validity_start` kwargs will always be set from `validity[1]`, regardless of `update_type`.

conftest.py
```diff
@@ -375,89 +372,99 @@ def check(
...
+            if issubclass(factory, factories.ValidityFactoryMixin):
+                kwargs["valid_between"] = validity[1]
+            elif issubclass(
+                factory,
+                factories.ValidityStartFactoryMixin,
+            ):
+                kwargs["validity_start"] = validity[1].lower
```
andyhd, line 446: A chunk of similar code here, could be factored out?
```python
def get_validity_kwargs(factory, valid_between):
    kwargs = {}
    if issubclass(factory, factories.ValidityFactoryMixin):
        kwargs["valid_between"] = valid_between
    # don't think we have any factories that subclass both mixins, so don't need elif
    if issubclass(factory, factories.ValidityStartFactoryMixin):
        kwargs["validity_start"] = valid_between.lower
    return kwargs
```
and
```python
if update_type in (...):
    if validity:
        kwargs.update(get_validity_kwargs(factory, validity[0]))
...
if validity:
    kwargs.update(get_validity_kwargs(factory, validity[1]))
```

andyhd (changes requested):
I like how much code this removes 👍 I think there might be a logic issue

conftest.py
```diff
@@ -375,89 +372,99 @@ def check(
...
+    expected_versions = 1 if update_type == UpdateType.CREATE else 2
 
     def check(
-        model: Union[TrackedModel, Type[DjangoModelFactory]],
+        factory: Type[DjangoModelFactory],
         serializer: Type[TrackedModelSerializer],
-        parent_model: TrackedModel = None,
-        dependencies: Dict[str, Union[TrackedModel, Type[DjangoModelFactory]]] = None,
-        kwargs: Dict[str, Any] = None,
+        dependencies: Dict[str, Union[Any, Type[DjangoModelFactory]]] = None,
```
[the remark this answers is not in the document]
simonwo: Yeah fair enough. I was trying to make clear that something special will happen if you pass in a factory, but I guess this isn't the mechanism for that.

conftest.py
simonwo: So the logic above line 439 is all about setting up a previous version of the model, which we only need to do if we are running an update or a delete. We still want to set up a current version of the model in all three cases. So `validity[0]` is only ever used for the previous version and `validity[1]` is only ever used for the version to be imported. You're right in that this is a change from the old behaviour, as the CREATE case did not even accept validity ranges at all (and just uses whatever is on the factory). I don't think is necessarily wrong, but it is different and a bit confusing… 

I think perhaps I will spend a bit more time looking at why this logic is even necessary; I think it's only used in 1 place to do with quotas, but to me it doesn't make sense that the previous version and the next version of the model necessarily have to have different date ranges.

simonwo: So I’ve checked this out. For a start, the tests continue to pass without any specific logic around validity. So no surprises there – it shouldn’t be a requirement to change validity dates when updating or deleting.

This would imply that this logic was only here to prevent some future regression. I can’t find any good reason for what that would be though. If we’re worried about a future bug discarding the content of updates, the sequences and fuzzy choices on the factories should account for that.

So I think we’re justified in removing this validity logic. I’ve factored it as a separate commit so that if we do need to revert then it’ll be easier.

andyhd (approved):
👍 👍

PR #299, opened 2021-05-18
TP-742: Basic error handling

andyhd:
This PR addresses both TP-705 and TP-742

Edit forms currently only provide simple field validation (eg, SID matches a regular expression).

We need to prevent users accidentally adding invalid data to the database which would violate the business rules. To prevent this at the earliest opportunity, this commit adds checks for the following:

* When start date is after end date on models which have a validity period
* Selected business rules - some business rules can only be checked across all models in a transaction, some do not make sense in a create/edit form (eg PreventDeleteIfInUse rules)

If these checks fail, the form is redisplayed with error messages, allowing the user to correct the mistakes:
image
image

### DIFF
--- a/additional_codes/views.py
+++ b/additional_codes/views.py
@@ -3,6 +3,7 @@ from typing import Type
 from rest_framework import permissions
 from rest_framework import viewsets
 
+from additional_codes import business_rules
 from additional_codes.filters import AdditionalCodeFilter
 from additional_codes.filters import AdditionalCodeFilterBackend
 from additional_codes.forms import AdditionalCodeDescriptionForm
@@ -13,6 +14,7 @@ from additional_codes.models import AdditionalCodeType
 from additional_codes.serializers import AdditionalCodeSerializer
 from additional_codes.serializers import AdditionalCodeTypeSerializer
 from common.models import TrackedModel
+from common.views import BusinessRulesMixin
 from common.views import TamatoListView
 from common.views import TrackedModelDetailMixin
 from common.views import TrackedModelDetailView
@@ -91,10 +93,20 @@ class AdditionalCodeDetail(AdditionalCodeMixin, TrackedModelDetailView):
 
 class AdditionalCodeUpdate(
     AdditionalCodeMixin,
+    BusinessRulesMixin,
     TrackedModelDetailMixin,
     DraftUpdateView,
 ):
     form_class = AdditionalCodeForm
+
+    validate_business_rules = (
+        business_rules.ACN1,
+        business_rules.ACN2,
+        business_rules.ACN4,
+        business_rules.ACN13,
+        business_rules.ACN17,
+        # business_rules.ACN5,  # XXX should it be checked here?
+    )
 
 
 class AdditionalCodeUpdateDescription(
--- a/common/views.py
+++ b/common/views.py
@@ -20,6 +20,7 @@ from django.views import generic
 from django_filters.views import FilterView
 from redis.exceptions import TimeoutError as RedisTimeoutError
 
+from common.business_rules import BusinessRuleViolation
 from common.models import TrackedModel
 from common.models import Transaction
 from common.pagination import build_pagination_list
@@ -212,3 +213,26 @@ class TrackedModelDetailView(
     generic.DetailView,
 ):
     pass
+
+
+class BusinessRulesMixin:
+    """Check business rules on form_submission."""
+
+    validate_business_rules = []
+
+    def form_valid(self, form):
+        violations = False
+        workbasket = WorkBasket.current(self.request)
+        transaction = workbasket.transactions.last()
+
+        for rule in self.validate_business_rules:
+            try:
+                rule(transaction).validate(form.instance)
+            except BusinessRuleViolation as v:
+                form.add_error(None, v.args[0])
+                violations = True
+
+        if violations:
+            return self.form_invalid(form)
+
+        return super().form_valid(form)
--- a/conftest.py
+++ b/conftest.py
@@ -17,11 +17,13 @@ from django.contrib.auth.models import Group
 from django.contrib.auth.models import Permission
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.exceptions import ValidationError
+from django.test.html import parse_html
 from factory.django import DjangoModelFactory
 from lxml import etree
 from moto import mock_s3
 from pytest_bdd import given
 from pytest_bdd import parsers
+from pytest_bdd import then
 from rest_framework.test import APIClient
 
 from common.models import TrackedModel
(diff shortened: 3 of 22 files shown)


stuaxo (approved):

common/views.py
```diff
@@ -212,3 +213,24 @@ class TrackedModelDetailView(
...
+
+    def form_valid(self, form):
+        violations = False
+
+        for rule in self.validate_business_rules:
+            try:
+                rule().validate(form.instance)
+            except BusinessRuleViolation as v:
+                form.add_error(None, v.args[0])
+                violations = True
```
simonwo, line 231: As we're not passing a transaction or workbasket, what does this check against? Will it, for example, identify business rule violations based on data in the same workbasket? E.g. if I change the validity dates on a footnote, and then edit the dates on an associated measure, will it be able to "see" the changes to the footnote?

conftest.py
```diff
@@ -644,3 +646,23 @@ def check(
...
+
+@pytest.fixture
+def response():
+    """Hacky fixture to enable passing the client response between BDD steps."""
+    return {}
+
+
+@then(parsers.parse('I see the form error message "{error_message}"'))
+def form_error_shown(response, error_message):
+    soup = BeautifulSoup(response["response"].content.decode(), features="lxml")
```
simonwo, line 659: Could we use [`assertInHTML`](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#django.test.SimpleTestCase.assertInHTML) rather than introducing a new dependency for this?

simonwo (commented):
So I think the code looks pretty great!

I still have the lingering concern about having to specify business rules manually on each view. I understand why we're doing this, I think – which is that if we are spreading object creation over multiple pages there will be some rules that will break. The classic example is everything with a description must have one created in the same transaction, so if we create the object on one page and then create the description on another, the business rule will fail after the first page.

But the concern I have is that it's really hard to know _which_ business rules to apply _when_. I think it's easy to get wrong, and we're going to have to do it on every view.

As long as we check the business rules properly before we submit I suppose the worst that can happen is an error in the UI later. We were trying to avoid that because in the current UI users won't be able to fix them.

So I just want to quickly explore, what would we have to do to be able to check all the business rules automatically always, and would that be better?

We could just make it a design constraint that all of the necessary fields have to be on one page. So the description field could be forced to be on the same page as the other fields, and we can create them at the same time in the backend, and avoid any business rule errors. I don't think that's an unreasonable constraint, but it may not scale to things like measure creation where at the moment it's spread very thinly across multiple pages.

So what we maybe would have to do as well is store unsaved data from a form in some sort of session-specific location, and then use it on a subsequent screen to do all the model creation. Would that be hard? And how much more complex would it be than setting the business rule list manually? Do we think it would be worth it if it allowed us to run all the rules and not worry about getting that wrong?

additional_codes/views.py
```diff
@@ -91,11 +93,21 @@ class AdditionalCodeDetail(AdditionalCodeMixin, TrackedModelDetailView):
...
+        business_rules.ACN2,
+        business_rules.ACN4,
+        business_rules.ACN13,
+        business_rules.ACN17,
+        # business_rules.ACN5,  # XXX should it be checked here?
+    )
+
 
 class AdditionalCodeUpdateDescription(
     AdditionalCodeDescriptionMixin,
```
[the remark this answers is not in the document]
andyhd: Oops, forgot

common/views.py
andyhd: I did think about this, but didn't address it because I had this code in the form which isn't aware of the request. Now I have moved the business rule checks into the view, so I can use the last transaction in the current workbasket. 👍

stuaxo (commented):
[snipped]
> .... maybe would have to do as well is store unsaved data from a form in some sort of session-specific location, and then use it on a subsequent screen to do all the model creation. Would that be hard? And how much more complex would it be than setting the business rule list manually? Do we think it would be worth it if it allowed us to run all the rules and not worry about getting that wrong?

I think it could be a good idea.

In WagTail (the django based CMS) the user can have draft versions of pages, that contain many fields with validation rules.

The data in draft form is stored in JSON, this allows the fields on the page to be filled even if the data isn't valid yet.

This allows the possibly-invalid data to be stored not just in the session, but in the database too.

The data must be valid if to be saved into the fields its destined for (e.g. move a page out of draft status).

additional_codes/views.py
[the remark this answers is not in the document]
andyhd: The description business rules are all currently indirect rules on the described object. To check them on form submit would require a refactor. I think it does make sense to be able to check that you aren't trying to modify a description such that it starts at the same time as another, or is outside the validity period of the described object, so perhaps another look at how indirect business rules work is needed.

andyhd:
The problem with running some of the cross-relationship business rules and indirect business rules in form validation is that whichever object you edit first, you will get a violation and the form submit will be prevented. For example, if you want to change the Certificate validity start date, when you will get an error because the first description must have a matching start date. But if you edit the first description, you get an error for the same reason. If you create a new first description, it may fail because the validity period is not contained by the Certificate.
Perhaps this is a contrived example and we should not allow changing the Certificate start date? But then the only thing you can edit on the form is the end date.
When creating a Certificate, you will never be able to submit the form successfully, because there will be no description. But you can't create a description if there is no Certificate.
These business rules can only be checked once all the related objects are in the workbasket, after form submission.

andyhd:
I'm not sure I see the need for storing the form data in the session - isn't that what the workbasket is for? Until a workbasket is approved, the models inside it can violate all the business rules they like, no?

simonwo (approved):
Ah hm, I hadn't thought of anything quite as complex as that example. Yeah, the only alternative for something like that case would be to detect the new start date and pull/push the start of the description to match, which would work… but it's getting to the same level of complexity around needing to understand each and every business rule and how it interacts. So in terms of work to do, I guess it's harder than just specifying what rules to check.

What this does mean though is that we'll have to bring forward more work on being able to edit and reorder workbaskets, so that if we do break a rule there is at least someway to go back and fix the workbasket. I thought we'd be able to do that later but I guess we'll have to do it sooner.

The example is also double worrying because in that specific case the certificate and it's description need to be edited _in the same transaction_ – you're correct that no ordering of them will work… at the moment I'd assumed any edits we do would be in separate transactions. Making the system understand if edits should be in the same transaction or not is "hard" as per ADR 12. 

I don't know how we're going to get around that. I really hope we won't have to delegate that option to the user. Perhaps we will have to introduce sooner rather than later a more advanced thing so that if the system sees a start date being edited it knows, "aha! I have to do something special with this!" and can go and update a bunch of other stuff automatically…?

Ah well, onwards and upwards!

conftest.py
```diff
@@ -644,3 +646,15 @@ def check(
         return True
 
     return check
+
+
+@pytest.fixture
+def response():
+    """Hacky fixture to enable passing the client response between BDD steps."""
+    return {}
```
stuaxo, line 654: ✅😀

PR #302, opened 2021-05-21
TP-695 Fix api calls for edit measures

DIT-CJWaszczuk:
## Summary 
Just cleaning up the APIs to search on the parameters related to what we're looking for in the autocompletes, given that they aren't used elsewhere? 
## Notes
The search criteria is meant to match how the search suggestion strings display currently, this is likely to change after some UR and we learn how they use the search.

### DIFF
--- a/commodities/serializers.py
+++ b/commodities/serializers.py
@@ -25,13 +25,38 @@ class SimpleGoodsNomenclatureSerializer(
         ]
 
 
+class SimpleGoodsNomenclatureDescriptionSerializer(
+    TrackedModelSerializerMixin,
+    ValidityStartSerializerMixin,
+):
+    class Meta:
+        model = models.GoodsNomenclatureDescription
+        fields = [
+            "sid",
+            "description",
+            "update_type",
+            "record_code",
+            "subrecord_code",
+            "period_record_code",
+            "period_subrecord_code",
+            "taric_template",
+            "start_date",
+            "validity_start",
+        ]
+
+
 @TrackedModelSerializer.register_polymorphic_model
 class GoodsNomenclatureSerializer(TrackedModelSerializerMixin, ValiditySerializerMixin):
     sid = serializers.IntegerField()
+    descriptions = SimpleGoodsNomenclatureDescriptionSerializer(
+        many=True,
+        required=False,
+    )
 
     class Meta:
         model = models.GoodsNomenclature
         fields = [
+            "id",
             "sid",
             "item_id",
             "suffix",
@@ -40,6 +65,7 @@ class GoodsNomenclatureSerializer(TrackedModelSerializerMixin, ValiditySerialize
             "record_code",
             "subrecord_code",
             "taric_template",
+            "descriptions",
             "start_date",
             "end_date",
             "valid_between",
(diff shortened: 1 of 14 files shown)


common/tests/factories.py
```diff
@@ -127,6 +127,7 @@ class Meta:
 
 
 class TrackedModelMixin(factory.django.DjangoModelFactory):
+    id = numeric_sid()
```
andyhd, line 130: This shouldn't be here. This will interfere with the model->trackedmodel one-to-one relation which is how django-polymorphic models work. You can remove this, the factory created objects should have an `id` field anyway.

andyhd (approved):
Looks ok except for one line.

commodities/serializers.py
```diff
@@ -25,13 +25,36 @@ class Meta:
         ]
 
 
+class SimpleGoodsNomenclatureDescriptionSerializer(
+    TrackedModelSerializerMixin,
+    ValiditySerializerMixin,
```
andyhd, line 30: I think this needs to be `ValidityStartSerializerMixin`

andyhd (changes requested):

commodities/serializers.py
```diff
@@ -25,13 +25,36 @@ class Meta:
...
+            "sid",
+            "description",
+            "update_type",
+            "record_code",
+            "subrecord_code",
+            "period_record_code",
+            "period_subrecord_code",
+            "taric_template",
+            "start_date",
+            "end_date",
```
andyhd, line 44: You also need to remove this - descriptions don't have end dates

commodities/serializers.py
```diff
@@ -25,13 +25,36 @@ class Meta:
...
+            "valid_between",
```
andyhd, line 45: ... and this should be `"validity_start",`

commodities/serializers.py
```diff
@@ -25,13 +25,35 @@ class Meta:
...
+            "taric_template",
+            "start_date",
+            "validity_start",
+        ]
+
+
 @TrackedModelSerializer.register_polymorphic_model
 class GoodsNomenclatureSerializer(TrackedModelSerializerMixin, ValiditySerializerMixin):
     sid = serializers.IntegerField()
+    descriptions = SimpleGoodsNomenclatureDescriptionSerializer(many=True)
```
andyhd, line 51: I think if you add `required=False` to the arguments here, your tests may pass.

commodities/serializers.py
```diff
@@ -25,13 +25,38 @@ class Meta:
...
+        model = models.GoodsNomenclatureDescription
+        fields = [
+            "sid",
+            "description",
+            "update_type",
+            "record_code",
+            "subrecord_code",
+            "period_record_code",
+            "period_subrecord_code",
+            "taric_template",
```
andyhd, line 42: You probably don't need these fields for a "Simple" serializer

commodities/serializers.py
```diff
@@ -25,13 +25,38 @@ class Meta:
...
 
 
+class SimpleGoodsNomenclatureDescriptionSerializer(
+    TrackedModelSerializerMixin,
+    ValidityStartSerializerMixin,
+):
+    class Meta:
+        model = models.GoodsNomenclatureDescription
+        fields = [
+            "sid",
```
andyhd, line 35: should this be `id`?

andyhd (approved):
Looks good 👍

PR #304, opened 2021-05-26
TP-788: Fix edits after publishing workbasket

andyhd:
After successfully editing and publishing a change, subsequent edits do not appear in
the list of tariff changes on the home page.

This is due to the workbasket not being updated in the user session. Subsequent edits
are appended to the already-published workbasket. The home page only lists the contents
of the current workbasket, which is a new and empty one after publishing the last one.

This change ensures that the session is updated on publishing the workbasket, and that
subsequent edits are correctly appended to the new unpublished workbasket.

Before testing make sure to clear your session by logging out and back in again.

### DIFF
--- a/common/forms.py
+++ b/common/forms.py
@@ -46,7 +46,10 @@ class DateInputFieldFixed(DateInputField):
     def compress(self, data_list):
         day, month, year = data_list or [None, None, None]
         if day and month and year:
-            return date(day=int(day), month=int(month), year=int(year))
+            try:
+                return date(day=int(day), month=int(month), year=int(year))
+            except ValueError as e:
+                raise ValidationError(str(e).capitalize()) from e
         else:
             return None
 
--- a/workbaskets/views/__init__.py
+++ b/workbaskets/views/__init__.py
@@ -1,10 +1,10 @@
-from django.contrib.auth.decorators import permission_required
+from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.db import transaction
-from django.shortcuts import get_object_or_404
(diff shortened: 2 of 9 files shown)


common/forms.py
```diff
@@ -17,7 +17,10 @@ class DateInputFieldFixed(DateInputField):
     def compress(self, data_list):
         day, month, year = data_list or [None, None, None]
         if day and month and year:
-            return date(day=int(day), month=int(month), year=int(year))
+            try:
+                return date(day=int(day), month=int(month), year=int(year))
+            except ValueError as e:
+                raise ValidationError(str(e).capitalize()) from e
```
andyhd, line 23: This is an unrelated fix for validating date inputs - previously, you could enter invalid day and month values (eg: day > 31 or month > 12) and you would get a 500 error instead of a form error.

workbaskets/views/__init__.py
```diff
@@ -26,55 +25,41 @@ class WorkBasketViewSet(viewsets.ModelViewSet):
...
+    permission_required = "workbaskets.change_workbasket"
 
-    workbasket.export_to_cds()
-    workbasket.save()
+    @transaction.atomic
+    def get_redirect_url(self, *args, **kwargs):
+        workbasket: WorkBasket = self.get_object()
 
-    return redirect("index")
+        workbasket.submit_for_approval()
```
andyhd, line 59: `submit_for_approval` calls `workbasket.full_clean`

stuaxo (approved):
Nice view changes in this too.

PR #305, opened 2021-05-27
XML performance refactor: part 1

simonwo:
This is the first part of the work (started but not finished in firebreak!) to make better use of the database to generate XML envelope output. I'm trying to get as much stuff out of the way which can be merged early to make the later PR more manageable.

This first set of commits is focused around more accurately representing the XML datamodel in the import parsers. This lets us get rid of some custom logic and provides natural extension points which we will use in a later PR to provide serialisation. On the whole though, not much _behaviour_ has changed here – we are just making the import parsers into more accurate data models.

There is some stuff in here which at first glance seems a bit weird (I'd encourage you to read the commit messages) and that we don't use yet e.g. adding the record and subrecord codes. This all gets used later and I can explain further if the commit messages are not clear enough.

### DIFF
--- a/importer/parsers.py
+++ b/importer/parsers.py
@@ -361,10 +361,8 @@ class CompoundElement(ValueElementMixin, ElementParser):
 class ValidityMixin:
     """Parse validity start and end dates."""
 
-    _additional_components = {
-        RangeLowerElement(Tag("validity.start.date")): "valid_between_lower",
-        RangeUpperElement(Tag("validity.end.date")): "valid_between_upper",
-    }
+    valid_between_lower = RangeLowerElement(Tag("validity.start.date"))
+    valid_between_upper = RangeUpperElement(Tag("validity.end.date"))
 
     def clean(self):
         super().clean()
@@ -383,12 +381,7 @@ class ValidityMixin:
 class ValidityStartMixin:
     """Parse validity start date."""
 
-    _additional_components = {
-        TextElement(Tag("validity.start.date")): "validity_start",
-    }
-
-    def clean(self):
-        super().clean()
+    validity_start = TextElement(Tag("validity.start.date"))
 
 
 class Writable:
--- a/importer/parsers.py
+++ b/importer/parsers.py
@@ -282,6 +282,39 @@ class RangeUpperElement(TextElement):
     """Represents an element that is the upper part of a range."""
 
 
+class CompoundElement(ValueElementMixin, ElementParser):
+    """
+    Represents an element in XML that is actually a concatenation of one or more
+    logical values and separators.
+
+    The separator by default is assumed to be a pipe character. The parsed data
+    will always contain a tuple that is the size of the number of expected
+    fields (the original field and any extras) – if less than the specified
+    number of separators occur the rightmost fields will have value ``None``.
+
+    .. code-block:: XML
+
+        <msg:some.value>one|two|three</msg:some.value>
+    """
+
+    native_type = tuple
+
+    def __init__(
+        self,
+        tag: Tag,
+        *extra_fields: str,
+        separator: str = "|",
+    ):
+        super().__init__(tag)
+        self.extra_fields = extra_fields
+        self.separator = separator
+
+    def clean(self):
+        parts = self.text.split(self.separator, len(self.extra_fields))
+        missing = len(self.extra_fields) - len(parts) + 1
+        self.data = self.native_type([*parts, *([None] * missing)])
+
+
 class ValidityMixin:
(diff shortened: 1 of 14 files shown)


importer/parsers.py
```diff
@@ -202,43 +226,131 @@ def wraps(parser):
...
+
+
+class CompoundElement(ValueElementMixin, ElementParser):
+    """
+    Represents an element in XML that is actually a concatenation of one or more
+    logical values and separators.
+
+    The separator by default is assumed to be a pipe character. The parsed data
+    will always contain a tuple that is the size of the number of fields – if
+    less than the specified number of separators occur some values will be `None`.
```
stuaxo, line 323: Maybe specify which values will be None, I'm assuming it's the ones on the end ?

stuaxo:
Before I start reading this properly - is it worth separating some of the more generic XML parts into their own library, or would that be better done afterwards once this has matured ?

simonwo:
Yeah, I think it’ll probably be better after, or at least the stuff in the next PR. This is all quite domain specific.

importer/parsers.py
```diff
@@ -202,43 +226,131 @@ def wraps(parser):
...
+class ConstantElement(ValueElementMixin, ElementParser):
+    """
+    Represents an element that is always a constant value in the XML.
+
+    The actual value is ignored and not put into the database. The value
+    specified in the constructor will be put back into the XML.
+    """
+
+    def __init__(self, tag: Tag, value: str) -> None:
+        super().__init__(tag)
```
andyhd, line 250: Pylint doesn't like the unused argument. You can suppress the warning in several ways:

```python
# option 1
def __init__(
    self,
    tag: Tag,
    value: str,  # pylint: disable=unused-argument
) -> None:
    ...

# option 2
def __init__(self, tag: Tag, _value: str) -> None:
    ...

# option 3
def __init__(self, tag: Tag, value: str) -> None:
    _ = value
    ...

# option 4
def __init__(self, tag: Tag, **kwargs) -> None:
```

I quite like option 1 because it is very clear what the intention is. But it's just for pylint, so no big deal either way.

importer/parsers.py
```diff
@@ -202,43 +226,131 @@ def wraps(parser):
...
-    only an integer value and have no attributes or children, eg:
+    .. code-block:: XML
 
         <msg:record.code>430</msg:record.code>
     """
 
+    native_type = int
+
+    def __init__(self, *args, format: str = "FM99999999999999999999"):
+        super().__init__(*args)
```
andyhd, line 280: Is the `format` string meant to be stored?

importer/parsers.py
```diff
@@ -202,43 +226,131 @@ def wraps(parser):
...
+    ):
+        super().__init__(tag)
+        self.children = fields
+        self.separator = separator
+
+    def clean(self):
+        if self.separator in self.text:
+            self.data = self.native_type(self.text.split(self.separator))
+        else:
+            self.data = self.native_type([self.text, *(None for _ in self.children)])
```
andyhd, line 346: If `self.children` has length 3 and `self.text` is `foo|bar`, then `self.data` will be `("foo", "bar")` and not `("foo", "bar", None)`, as described by the doc comment - unless I misunderstood it?

Also, if there are no separators in `self.text` then `self.data` will have length `len(self.children) + 1`, which I think also contradicts the doc comment.

I think this does what the comment says:
```python
parts = self.text.split(self.separator, len(self.children) - 1)
self.data = self.native_type([*parts, *([None] * len(self.children) - len(parts))])
```
Or you could use `itertools` and do something like:
```python
self.data = islice(
    chain(
        self.text.split(self.separator, len(self.children) - 1),
        repeat(None),
    ),
    len(self.children)
)
```

andyhd (changes requested):
I think there is either a misleading doc comment or some incorrect code in `CompoundElement`. Otherwise, looks great 👍

importer/parsers.py
simonwo: I have also marked this as unused. It is used later in the serialization commits but it was just easier to leave it unused and get all the data on the parsers in now.

importer/parsers.py
simonwo: You're right, the doc string and the code don't really match. I've added some tests and used your implementation. Thanks!

andyhd (approved):
🎩

importer/parsers.py
stuaxo: Glad you covered this @andyhd I spent a while staring at the part that constructs the data trying to work out if it should work or not.

PR #306, opened 2021-05-27
TP-695 Edit measure geographical area

DIT-CJWaszczuk:
## Summary
Adds ability to edit the geographical area of a measure 
## Screenshots 
image

### DIFF
--- a/geo_areas/models.py
+++ b/geo_areas/models.py
@@ -10,6 +10,7 @@ from common.models.mixins.description import DescriptionMixin
 from common.models.mixins.validity import ValidityMixin
 from geo_areas import business_rules
 from geo_areas import validators
+from geo_areas.validators import AreaCode
 from measures import business_rules as measures_business_rules
 from quotas import business_rules as quotas_business_rules
 
@@ -98,6 +99,15 @@ class GeographicalArea(TrackedModel, ValidityMixin):
             .approved_up_to_transaction(self.transaction)
             .exists()
         )
+
+    def is_single_region_or_country(self):
+        return self.area_code == AreaCode.COUNTRY or self.area_code == AreaCode.REGION
+
+    def is_all_countries(self):
+        return self.area_code == AreaCode.GROUP and self.area_id == "1011"
+
+    def is_group(self):
+        return self.area_code == AreaCode.GROUP
 
     def __str__(self):
         return f"{self.get_area_code_display()} {self.area_id}"
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -12,9 +12,11 @@ from additional_codes.models import AdditionalCode
 from commodities.models import GoodsNomenclature
 from common.forms import AutocompleteWidget
 from common.forms import ValidityPeriodForm
+from geo_areas.models import GeographicalArea
 from measures import models
 from quotas.models import QuotaOrderNumber
 from regulations.models import Regulation
+from workbaskets.models import WorkBasket
 
 
 class AutoCompleteField(forms.ModelChoiceField):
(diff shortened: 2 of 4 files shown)


geo_areas/models.py
```diff
@@ -98,6 +98,15 @@ def is_a_parent(self):
             .exists()
         )
 
+    def is_single_region_or_country(self):
+        return self.area_code == 0 or self.area_code == 2
+
+    def is_erga_omnes(self):
```
stuaxo, line 104: Would be great to have a short docstring saying what erga_omnes is !

measures/forms.py
```diff
@@ -57,10 +58,67 @@ class MeasureForm(ValidityPeriodForm):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.initial_geographical_area = self.instance.geographical_area
+
+        self.fields[
+            "geographical_area_group"
+        ].label_from_instance = lambda obj: obj.structure_description
+        self.fields[
+            "geographical_area_country_or_region"
+        ].label_from_instance = lambda obj: obj.structure_description
```
stuaxo, line 91: I assume these two are meant to be the same ?

geo_areas/models.py
DIT-CJWaszczuk: I'd rather we just remove the damn phrase all together!

measures/forms.py
```diff
@@ -57,10 +58,67 @@ class MeasureForm(ValidityPeriodForm):
...
+                "geographical_area_country_or_region"
+            ].initial = self.instance.geographical_area
 
     def clean(self):
         cleaned_data = super().clean()
 
+        erga_omnes_instance = GeographicalArea.objects.latest_approved().filter(
+            area_code=1,
+            area_id=1011,
+        )[0]
```
stuaxo, line 109: Can the query return multiple objects ?

If it is always one, then `.get(area_code=1, area_id=1011)` may be a better fit than `.filter(....)[0]`.

geo_areas/models.py
stuaxo: Sounds good, I have no idea what it is... yet another cryptic taric thing.

simonwo: Feel free to use `all_countries` instead.

geo_areas/models.py
```diff
@@ -98,6 +98,15 @@ def is_a_parent(self):
             .exists()
         )
 
+    def is_single_region_or_country(self):
+        return self.area_code == 0 or self.area_code == 2
```
simonwo, line 102: The constants in [`geo_areas.validators.AreaCode`](https://github.com/uktrade/tamato/blob/master/geo_areas/validators.py#L10) would help avoid the magic numbers.

measures/forms.py
simonwo: It's technically possible for this [area code to be reused](https://uktrade.github.io/tamato/docs/business_rules/geo_areas.html#geo_areas.business_rules.GA1) which would result in multiple areas with the area ID 1011, so I would advise adding a call to [`as_at(instance.valid_between.lower)`](https://github.com/uktrade/tamato/blob/master/common/models/records.py#L136) here, and then as Stuart said, calling `get`.

simonwo (changes requested):
Can't see anything controversial! 😁 Would be good to land a few tweaks though.

Also did we figure out how to write tests for these things yet?

stuaxo:
> Can't see anything controversial! grin 
Me neither - thought I'd need to reread it as I couldn't see any.

> And I heartily approved of new fields on the model (maybe they could be `@property`?). Would be good to land a few tweaks though.
Had the same thought on property, but wasn't sure I knew the code enough here to write the comment.

> 
> Also did we figure out how to write tests for these things yet?

For the FE ?   Yes, though I need to come back and implement them.

stuaxo:
@andyhd would this make most sense as part of the QuerySet ?

measures/forms.py
DIT-CJWaszczuk: yeah i'm concerned of the select populating with an item that shouldn't be in, i.e you previously had turkey as the area, you didn't want Turkey showing in the groups, can probably clean it up

measures/forms.py
stuaxo: @andyhd would erga_omnes make most sense as part of the QuerySet ?

So this ends up with something like:
GeographicalArea.objects.erga_omnes.latest_approved()

measures/forms.py
```diff
@@ -57,10 +59,108 @@ class MeasureForm(ValidityPeriodForm):
...
+        self.fields[
+            "additional_code"
+        ].queryset = AdditionalCode.objects.approved_up_to_transaction(tx)
+        self.fields[
+            "geographical_area"
+        ].queryset = GeographicalArea.objects.approved_up_to_transaction(tx)
+        self.fields[
+            "geographical_area_group"
+        ].queryset = GeographicalArea.objects.approved_up_to_transaction(tx).filter(
+            area_code=1,
```
andyhd, line 116: I wonder about the order of `approved_up_to_transaction` and `filter` - does it matter which way round they go? I haven't looked into the details, but I feel like filtering by `area_code` first will mean the `approved_up_to_transaction` will have fewer rows to operate on and will therefore be quicker? @stuaxo any thoughts?

andyhd (approved):
Looks good to me 👍

PR #307, opened 2021-06-01
TP-738 create new descriptions

rachaelcodes:
Allows users to create new descriptions for:
* Additional codes
* Certificates
* Footnotes
* Geographical Areas

Also updates the GeographicalAreaDescription model to fix some bugs with descriptions.


certificates/forms.py
```diff
@@ -83,3 +83,23 @@ class CertificateDescriptionForm(DescriptionForm):
...
+            self.fields["described_certificate"].disabled = True
+            self.fields["described_certificate"].help_text = "You can't edit this"
+
+        self.helper.layout.insert(
+            0,
+            Field(
+                "described_certificate",
+                context={"label_size": "govuk-label--s"},
+            ),
+        )
```
simonwo, line 101: It seems like it might be possible to genericise this through use of `DescriptionMixin.described_object_field` to get the right field name on a per-model basis?

certificates/models.py
```diff
@@ -110,5 +111,14 @@ class CertificateDescription(DescriptionMixin, TrackedModel):
...
 
+    def save(self, *args, **kwargs):
+        if getattr(self, "sid") is None:
+            highest_sid = CertificateDescription.objects.aggregate(Max("sid"))[
+                "sid__max"
+            ]
+            self.sid = highest_sid + 1
+
+        return super().save(*args, **kwargs)
+
```
simonwo, line 122: Could this be genericised as well? Or would it make sense to override something on the `SignedIntSID` field type?

certificates/views.py
```diff
@@ -98,6 +99,35 @@ def get_queryset(self):
...
 
+class CertificateCreateDescriptionMixin:
+    model: Type[TrackedModel] = models.CertificateDescription
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["described_object"] = models.Certificate.objects.get(
+            certificate_type__sid=(self.kwargs.get("certificate_type__sid")),
+            sid=(self.kwargs.get("sid")),
+        )
```
simonwo, line 110: It might also be possible to do this generically using `identifying_fields`?

geo_areas/migrations/0005_rename_described_area.py
```diff
@@ -0,0 +1,18 @@
...
+        ("geo_areas", "0004_validity_start"),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name="geographicalareadescription",
+            old_name="area",
+            new_name="described_geographicalarea",
+        ),
+    ]
```
simonwo, line 18: Looks good to me – you could also override `described_object_field` or something like that, but no reason to not do this really.

simonwo (commented):
Looks good – my comments are just about whether this could be DRYer. But it's up to you whether this is the appropriate time to do that or not.

geo_areas/forms.py
```diff
@@ -0,0 +1,27 @@
...
+
+class GeographicalAreaCreateDescriptionForm(CreateDescriptionForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.instance:
+            self.fields["described_geographicalarea"].disabled = True
+            self.fields["described_geographicalarea"].help_text = "You can't edit this"
+            self.fields[
+                "described_geographicalarea"
+            ].label = "Described additional code"
```
simonwo, line 15: Not an additional code

stuaxo (approved):

simonwo (approved):
LGTM. 

Only final thought is: should we do a test for the `sid` automatic setting?

andyhd:
Sorry that this is so long after the fact, but did we want to do any sanitation of HTML in the description text? We tell the user that they can enter HTML, but don't prevent them from entering any malicious code. We should probably only allow a subset of HTML tags (`b`, `strong`, `i`, `em`, `u` (really?), and `a` - but links probably need further restrictions). We certainly shouldn't allow `<script>`, or we've just added a huge security hole.

PR #308, opened 2021-06-01
TP-193: Create Measure wizard

andyhd:
Creating new measures in the tariff is a complex task, and is broken down into multiple
steps. Measures are made up of a number of linked database models, and so creating a new
measure and all its dependant models must be carefully orchestrated.

This change adds a "wizard", a sequence of connected forms that gather the input data to
construct a new measure and its associated models.

The wizard stores the state of the sequence of forms in the session until the final form
is submitted, and then the models are created, written to the database and linked
together. The user is able to backtrack or even jump around the steps of the sequence,
and is shown a final review page before the measure creation is confirmed.

The sequence of forms is visualized on the right hand column of each step page, using a
step-by-step navigation component.

Some of the steps are forms which allow the user to build up a list of entries, with an
"Add another" button and "Delete" buttons to add and remove entries from the list before
submitting the list and proceeding to the next step.

The geographical area form input is a composite input, where the user can select one of
three options (All countries, a specific country group, or a specific country).
Selecting the group or country option allows the user to make a further selection. There
is currently no way to specify country exclusions from a country group (or all
countries).


measures/jinja2/measures/create-start.jinja
```diff
@@ -0,0 +1,13 @@
+{% extends "measures/create-wizard-step.jinja" %}
+
+{% block form %}
+  <p class="govuk-body">
+    It is essential to plan your activities in advance of using this measure creation functionality.
```
stuaxo, line 5: Hopefully in future we can have a link here to the tariff manual, or something like a screencast that elaborates on this.

stuaxo (approved):

simonwo:
tfw you see minus 11,000 lines and then the relief when it's all in package-lock.json 😂

andyhd:
Yeah, +3477 -11182 in package-lock.json 🙄 and another +1k for step-by-step-nav css and js, which is just copied from a non-packaged gov.uk source.

geo_areas/forms.py
```diff
@@ -17,5 +22,121 @@ def __init__(self, *args, **kwargs):
...
+
+    def __init__(self, *args, **kwargs):
+        tx = kwargs.pop("transaction", None)
+        self.transaction = tx
+        super().__init__(*args, **kwargs)
+
+        # self.fields[
+        #     "erga_omnes_exclusions"
+        # ].queryset = GeographicalArea.objects.approved_up_to_transaction(tx)
+        self.fields["geo_group"].queryset = (
```
stuaxo, line 79: ^ some commented bits here ^^

geo_areas/forms.py
```diff
@@ -17,5 +22,121 @@ def __init__(self, *args, **kwargs):
...
+            .approved_up_to_transaction(tx)
+            .with_latest_links("descriptions")
+            .prefetch_related("descriptions")
+            .order_by("descriptions__description")
+            # .distinct("descriptions__description")
+        )
+        # self.fields[
+        #     "geo_group_exclusions"
+        # ].queryset = GeographicalArea.objects.approved_up_to_transaction(tx)
+        self.fields["geo_area"].queryset = (
```
stuaxo, line 93: ^ some commented bits here ^^

geo_areas/forms.py
```diff
@@ -17,5 +22,121 @@ def __init__(self, *args, **kwargs):
...
+            GeographicalArea.objects.exclude(
+                area_code=AreaCode.GROUP,
+            )
+            .active()
+            .approved_up_to_transaction(tx)
+            .with_latest_links("descriptions")
+            .prefetch_related("descriptions")
+            .order_by("descriptions__description")
+            # .distinct("descriptions__description")
+        )
```
stuaxo, line 103: ^ I guess the distinct is commented here as not sure if it's needed ?

regulations/tests/bdd/test_regulations.py
```diff
@@ -31,52 +31,44 @@ def regulations_search(client):
...
+        for value in [
+            reg.regulation_id,
+            f"{reg.regulation_group.group_id}: {reg.regulation_group.description}",
+            reg.information_text,
+            reg.public_identifier,
+            reg.url,
+            f"{reg.valid_between.lower:%d %b %Y}",
+            reg.transaction.workbasket.get_status_display(),
+        ]
+    )
```
stuaxo, line 74: Not a blocker - if this does fail, there isn't a clue about what went wrong in the assertion.   It seems a little tricky to come up with a good way of doing this while keeping the test small.

geo_areas/forms.py
andyhd: Yeah, I had left the geo exclusion stuff for later, but it looks like I need to tackle it now for both create and edit measures.

geo_areas/forms.py
andyhd: I found it didn't actually make a difference, so left this here as an invitation to tell me what I'm doing wrong :)

regulations/tests/bdd/test_regulations.py
andyhd: I should make an assertion for each value inside the loop, rather than using `any`

andyhd:
@stuaxo I'm afraid the comments you left are on code changes in [the other PR TP-703](https://github.com/uktrade/tamato/pull/329) - this PR just builds on top of that one

regulations/tests/bdd/test_regulations.py
andyhd: @stuaxo I'm afraid this code change is in [the other PR TP-703](https://github.com/uktrade/tamato/pull/329) - this PR builds on top of that one

measures/views.py
```diff
@@ -49,6 +61,189 @@ class MeasureDetail(MeasureMixin, TrackedModelDetailView):
...
+                    condition_code=condition_data["condition_code"],
+                    action=condition_data.get("action"),
+                    required_certificate=condition_data.get("required_certificate"),
+                )
+                condition.save()
+
+                # XXX the design doesn't show whether the condition duty_amount field
+                # should handle duty_expression, monetary_unit or measurements, so this
+                # code assumes some sensible(?) defaults
+                if condition.duty_amount:
```
simonwo, line 183: Hm, they can sometimes be more complex than just a single amount… how possible would it be to run this field through the DutySentenceParser instead? I appreciate at the moment it only returns MeasureComponents but IIRC it should be reasonably easy to make it return MeasureConditonComponents instead?

measures/views.py
```diff
@@ -49,6 +61,189 @@ class MeasureDetail(MeasureMixin, TrackedModelDetailView):
...
+                item["footnote"]
+                for item in cleaned_data.get("formset-footnotes", [])
+                if not item["DELETE"]
+            ],
+            # condition_sentence here, or handle separately and duty_sentence after?
+            "duty_sentence": cleaned_data["duties"],
+        }
+
+        try:
+            measure, *_ = list(measure_creation_pattern.create(**measure_data))
```
simonwo, line 163: Yeah this interface is a bit sucky. I think we can change this to just return the measure (and not as a generator) if that would be better? It was only ever like this so we could get all the models we needed to output for the XML, but we do that in a different way now.

simonwo (approved):
Looks good! Some comments about scope of data we return from autocompletes and choices, but won't stop it from working in the immediate term.

Dare I ask about tests? Because I can see a future refactor to e.g. the `MeasureCreationPattern` could break this accidentally…

common/jinja2/components/geographical_area_select/template.jinja
```diff
@@ -0,0 +1,60 @@
+{% from "components/radios/macro.njk" import govukRadios %}
+{% from "components/details/macro.njk" import govukDetails %}
+
+{% set option_html = namespace() %}
+{% set option_html.ERGA_OMNES -%}
+  {# form.erga_omnes_exclusions #}
+{%- endset %}
```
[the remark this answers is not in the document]
andyhd: I will remove these - I'm going to implement exclusions next anyway

geo_areas/forms.py
```diff
@@ -17,5 +22,121 @@ def __init__(self, *args, **kwargs):
...
+    """
+
+    class GeoAreaType(models.TextChoices):
+        ERGA_OMNES = "ERGA_OMNES", "All countries (erga omnes)"
+        GROUP = "GROUP", "A group of countries"
+        COUNTRY = "COUNTRY", "A single country or region"
+
+    geo_area_type = forms.ChoiceField(choices=GeoAreaType.choices, required=False)
+    erga_omnes_exclusions = forms.ModelMultipleChoiceField(
+        queryset=GeographicalArea.objects.all(),
```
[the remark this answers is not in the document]
andyhd: The querysets all depend on the current request, which we only have access to in the view. We pass the request into the form during initialization, but that was a problem for the formsets on the conditions and footnotes forms, so I've resorted to overriding the querysets from outside the form. I don't like it.

geo_areas/models.py
```diff
@@ -9,13 +9,19 @@
...
 from geo_areas import business_rules
 from geo_areas.validators import AreaCode
 from geo_areas.validators import area_id_validator
 from measures import business_rules as measures_business_rules
 from quotas import business_rules as quotas_business_rules
 
 
+class GeographicalAreaQuerySet(TrackedModelQuerySet):
+    def erga_omnes(self):
+        return self.filter(area_code=AreaCode.GROUP, area_id=1011)
```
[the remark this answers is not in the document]
stuaxo: @simonwo do you have a pointer to what the significance of area_id=1011 is here?
I'll probably add a comment.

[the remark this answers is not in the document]
simonwo: Area ID 1011 is reasonably well defined to be "all countries" aka "erga omnes".

measures/views.py
simonwo: Well it would still need to call all the separate functions, just not `yield from` them. So I think it would be worth keeping the functions just for modularity. But anyway, this is just refactoring so let's not let that get in the way of merging this.

paulpepper-trade (changes requested):
Labels and help text are repeated on some of the forms.
image

stuaxo (approved):
> [16:52] stuart.axon (Guest)
    Nice !   And it doesn't effect any of the other screens ?

[16:53] Paul Pepper (Guest)
    The only place that we use the AutoCompleteField is in Create Measures forms, so no further changes necessary.

measures/views.py
stuaxo: Pushed this up - (between Paul and myself we have divided this so I am going from the bottom up and him, the top-down).

measures/forms.py
```diff
@@ -204,3 +212,311 @@ def __init__(self, *args, **kwargs):
...
+            "generating_regulation",
+            "geographical_area",
+            "order_number",
+            "valid_between",
+        ]
+
+    measure_type = AutoCompleteField(
+        label="Measure type",
+        help_text="Select the appropriate measure type.",
+        queryset=models.MeasureType.objects.all(),
```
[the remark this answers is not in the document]
stuaxo: Does this mean `as_at(measure_start_date)` as above ?

paulpepper-trade:
> Labels and help text are repeated on some of the forms:
> 
> image

Fixed.

measures/views.py
stuaxo: Note: from our conversation earlier that this refactor can be implemented later.

geo_areas/forms.py
paulpepper-trade: In this first version of create measures it's been decided to delay geographical exclusions. We have a Jira ticket in our backlog to deal with adding the missing features: https://uktrade.atlassian.net/browse/TP-891

geo_areas/forms.py
paulpepper-trade: In this first version of create measures it's been decided to delay geographical exclusions. We have a Jira ticket in our backlog to deal with adding the missing features: https://uktrade.atlassian.net/browse/TP-891

geo_areas/forms.py
paulpepper-trade: Commented `.distinct()` now removed: https://github.com/uktrade/tamato/pull/308/commits/2b98bb1eea37bbd12891da3ff37e21efca284bb2

simonwo: Let's create a backlog item for changing this to be `as_at(measure_start_date)`, unless that's trivial now.

paulpepper-trade: Jira ticket created for adding support for "geographic area activation at a future point in time" added, splitting this off from the main create measures requirement: https://uktrade.atlassian.net/browse/TP-893

measures/forms.py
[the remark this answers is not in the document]
simonwo: Yes, pretty much. But this will probably suffice for an initial deploy.

common/jinja2/components/geographical_area_select/template.jinja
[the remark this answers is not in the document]
paulpepper-trade: The geographic exclusions requirement has been split out of the Create measures ticket and therefore this PR. See https://uktrade.atlassian.net/browse/TP-891 for the requirement.

geo_areas/forms.py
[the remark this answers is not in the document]
paulpepper-trade: Similar to previous geographic exclusion, split out on to a new Jira https://uktrade.atlassian.net/browse/TP-891. I've also added a note to that ticket to referencing Simon's remark about filtering with `approved_up_to_transaction()`.

measures/forms.py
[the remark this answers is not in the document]
paulpepper-trade: It looks as though Andy has scoped down by transaction for all fields that use a queryset within `MeasureCreateWizard.get_form()` https://github.com/uktrade/tamato/blob/8d6a05b8a93e9d2e9dd2b78f2ddbe2ca0078230c/measures/views.py#L226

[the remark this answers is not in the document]
paulpepper-trade: Scoping down available measure types by date within the form may not be possible with the current UI arrangement. If we want to scope down by the measure's start and/or end dates, then they would need entering before the measure type is selected. As it stands, the same form first presents the measure type select then the measure start and end dates.

A couple of possibilities:

- Split the form into two (an extra wizard step) so that the measure start and end dates are collected first. Then present the request for measure type, filtering based upon the dates collected at the previous step.
- Don't filter down the measure type by date, but perform a check date filtered check on measure type when the user completes this step, returning an error message that the measure type is unavailable for the selected measure start and end dates. (It seems likely that this approach would be a bit annoying and potentially mysterious to the user though.)

Any thoughts on this @simonwo and @stuaxo ?

measures/views.py
paulpepper-trade: Simon's suggestion now split out into a separate Jira ticket: https://uktrade.atlassian.net/browse/TP-897

measures/forms.py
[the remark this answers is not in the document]
stuaxo: (based conclusions from our chat on teams):

This is a good call, we don't know the valid_between of the measure_type until a user has chosen one from the dropdown.

User experience wise:   I don't think it would be a good experience to have to enter dates before you enter the measure type (though of course I may be wrong here).

We definitely want some validation, the MVP version is probably to implement this when the user hits Continue.

A more useful version would update some information on the validity/invalidity of the dates when the user chooses a measure type with the dropdown.
There are few enough measure types that the info on dates could be embedded into the page, saving the implementation of a new API.

geo_areas/forms.py
stuaxo: Clarified the comments on these, so someone will know what's up if it breaks in the future.

paulpepper-trade (changes requested):
The current step number in the page title lags by one (-1) from the correct value displayed in the sidebar.

paulpepper-trade:
> The current step number in the page title lags by one (-1) from the correct value displayed in the sidebar.

Fixed: https://github.com/uktrade/tamato/pull/308/commits/cbc3dc74ed62229b373be66b7e34e02927de8a77

measures/forms.py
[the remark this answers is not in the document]
stuaxo: This turned out to be the autocomplete adding entries (it calls an API to get them).

To expedite this I scoped to `as_at(today)` but after this is up will make it call the API with the extra scoping.

stuaxo:
We definitely want some better test coverage of this once the intial version lands.

commodities/views.py
```diff
@@ -15,7 +17,17 @@ class GoodsNomenclatureViewset(viewsets.ReadOnlyModelViewSet):
...
         tx = WorkBasket.get_current_transaction(self.request)
-        return GoodsNomenclature.objects.approved_up_to_transaction(
-            tx,
-        ).prefetch_related("descriptions")
+        return (
+            GoodsNomenclature.objects.approved_up_to_transaction(
+                tx,
+            )
+            .prefetch_related("descriptions")
+            .as_at(date.today())
```
simonwo, line 31: As discussed in the other comments, this is not really a final solution because it's possible we are making a measure to start quite a way into the future, so it's still possible to select a commodity that will have expired. However, mostly commodity validity dates are open-ended until right before they're terminated, so this won't actually cause a problem in 99.9% of cases.

But we should revisit this and fix all the places where we've used `.as_at(date.today())` when we should probably be using `.filter(valid_between__contains=TaricDateRange(<measure start date>, <measure end date>)`. I've extended TP-893 to include that.

geo_areas/models.py
```diff
@@ -49,6 +55,8 @@ class GeographicalArea(TrackedModel, ValidityMixin):
     # This deals with subgroups of other groups
     parent = models.ForeignKey("self", on_delete=models.PROTECT, null=True, blank=True)
 
+    objects = GeographicalAreaQuerySet.as_manager()
```
simonwo, line 58: Is this right? Normally it's something like `objects = PolymorphicManager.from_queryset(GeographicalAreaQuerySet)()`?

measures/forms.py
```diff
@@ -204,3 +214,321 @@ def __init__(self, *args, **kwargs):
...
+    class Meta:
+        model = models.Measure
+        fields = [
+            "goods_nomenclature",
+            "additional_code",
+        ]
+
+    goods_nomenclature = AutoCompleteField(
+        label="Commodity code",
+        help_text="Select the 10-digit commodity code to which the measure applies.",
```
simonwo, line 307: I can imagine users asking the question "how do i add measures to a higher level commodity code" – they don't understand that all codes are 10 digit from TAP's perspective. We may need to revisit this content.

simonwo (approved):
Let's do it.

paulpepper-trade (approved):

geo_areas/models.py
stuaxo: You can do it either way, this way will still have the methods of GeographicalQueryAreaQuerySet available.

Read this *after* hitting merge, so will definitely come back to this.

measures/forms.py
stuaxo: This can be part of revisiting automcomplete in general.

paulpepper-trade: Jira ticket created to deal with this in a next or future iteration: https://uktrade.atlassian.net/browse/TP-903

geo_areas/models.py
stuaxo: Looks like there was a reason here: this leaves the .erga_omnes() method available on the created manager, so there is nothing to fix.

paulpepper-trade: Stuart and I just did a quick hotfix: https://github.com/uktrade/tamato/pull/355

PR #310, opened 2021-06-02
Add TARIC3 principles to documentation

simonwo:
There is a wealth of information in the TARIC3 principles document that is currently locked away in our wiki and in PDFs. This commit introduces some of this to the documentation area in the codebase so that it's more accessible and specific parts can be sent around as links.

As this is a document originally written by the EU, there are some references which do not make sense in a UK context (such as references to "Member States" or "the Commission"). These have been retained as-is to most accurately describe the original intention of the specification.

There is also documentation that is not relevant to the UK Tariff and its ystems (such as the documentation on EDIFACT). These are retained for completeness and potential future interest.


andyhd (approved):

stuaxo:
Just a post-merge approval message to say "yes, this is really useful".

PR #312, opened 2021-06-07
TP 650 - Schedule workbasket upload to bucket

rachaelcodes:
Triggers workbasket upload on setting workbasket to 'READY_TO_EXPORT'.

### DIFF
--- a/workbaskets/models.py
+++ b/workbaskets/models.py
@@ -144,6 +144,11 @@ class WorkBasket(TimestampedMixin):
             version_group.current_version = obj
             version_group.save()
 
+        # imported lower down to avoid circular import error
+        from exporter.tasks import upload_workbaskets
+
+        upload_workbaskets.delay()
+
     @transition(
         field=status,
         source=WorkflowStatus.READY_FOR_EXPORT,
--- a/workbaskets/tests/test_models.py
+++ b/workbaskets/tests/test_models.py
@@ -1,4 +1,5 @@
 from contextlib import nullcontext as does_not_raise
+from unittest import mock
 
 import pytest
 from django_fsm import TransitionNotAllowed
@@ -34,7 +35,7 @@ def test_workbasket_transactions():
 
 
 @pytest.mark.parametrize(
-    "start_status,transition,target_status,expect_error",
+    "start_status,transition,target_status,expect_error,expect_upload",
     [
         # Submission
         (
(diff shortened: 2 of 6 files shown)


rachaelcodes:
After discussion of the team, have decided to leave the mid-file import to avoid circular import problems for now. We may need to look more into how we queue tasks if we get many more of them.

workbaskets/tests/test_models.py
```diff
@@ -580,28 +674,40 @@ def test_workbasket_accepted_updates_current_tracked_models(new_workbasket, vali
...
-    new_workbasket.export_to_cds()
-    new_footnote.refresh_from_db()
-    assert new_footnote.version_group.current_version.pk == new_footnote.pk
-    new_workbasket.cds_error()
-    new_footnote.refresh_from_db()
-    assert new_footnote.version_group.current_version.pk == original_footnote.pk
+
+    with mock.patch(
+        "exporter.tasks.upload_workbaskets",
+    ) as mock_save:
```
stuaxo, line 699: Patching `exporter.tasks.upload_worbaskets.delay` might be better as `upload_workbaskets` is still our own code, but `delay` is the point where 3rd party code (Celery).

stuaxo: If you ever want to call the task itself, then thats safe too - (always_eager=True and the broker backend is set to memory://).
The only caveat is saving to s3; some other tests use moto to mock this.

workbaskets/models.py
```diff
@@ -16,6 +16,9 @@
 from common.models import TrackedModel
 from common.models import Transaction
 from common.models.records import TrackedModelQuerySet
+
+# imported lower down to avoid circular import error
+# from exporter.tasks import upload_workbaskets
```
simonwo, line 21: I don't think you really need this comment, or move the line 20 to replace line 150?

simonwo (approved):
Great!

workbaskets/models.py
```diff
@@ -144,6 +147,11 @@ def approve(self, user):
             version_group.current_version = obj
             version_group.save()
 
+        # hack to avoid circular import
+        from exporter.tasks import upload_workbaskets
```
stuaxo, line 151: I had a bit of a look at the importer and its use of `tasks` (after remembering that it also uses celery).
This is fine ... if it points at anything, it's a much bigger refactor so this is probably fine :)

But yeah, there's always importer to stare at :man_shrugging:

PR #317, opened 2021-06-10
__str__ don't die if a field has validity_start instead of valid_between

stuaxo:
Description `__str__` should not raise an exception a field has validity_between instead of validity_start.

Fixes an error that was occuring as one of the fields still referenced valid_between.

AttributeError: 'AdditionalCodeDescription' object has no attribute 'valid_between'

Add very basic test that checks every TrackedModel subclass, via @simonwo

### DIFF
--- a/common/models/mixins/description.py
+++ b/common/models/mixins/description.py
@@ -48,7 +48,7 @@ class DescriptionMixin(ValidityStartMixin):
         return self.identifying_fields_to_string(
             identifying_fields=(
                 self.described_object_field.name,
-                "valid_between",
+                "validity_start",
             ),
         )
 
--- a/common/tests/test_models.py
+++ b/common/tests/test_models.py
@@ -368,3 +368,21 @@ def test_get_descriptions_with_update(sample_model, valid_user):
 
     assert new_description in description_queryset
     assert description not in description_queryset
+
+
+@pytest.mark.parametrize(
+    "factory",
+    factories.TrackedModelMixin.__subclasses__(),
+    ids=[
+        factory._meta.model.__name__
+        for factory in factories.TrackedModelMixin.__subclasses__()
+    ],
+)
+def test_trackedmodel_str(factory):
+    """Verify no __str__ methods of TrackedModel classes crash or return non-
+    strings."""
+    instance = factory.create()
+    result = instance.__str__()
+
+    assert isinstance(result, str)
+    assert len(result.strip())


common/models/mixins/description.py
```diff
@@ -45,10 +45,14 @@ def get_url(self, action="detail"):
...
     def __str__(self):
+        validity_field = "validity_start"
+        if hasattr("self", "valid_between"):
+            validity_field = "valid_between"
+
         return self.identifying_fields_to_string(
             identifying_fields=(
                 self.described_object_field.name,
-                "valid_between",
+                validity_field,
```
andyhd, line 55: I'd be tempted to use a ternary condition expression here

simonwo (changes requested):
Hm, there shouldn’t be any descriptions with a valid_between… so you should be able to just change it without the condition?

Also the ValidityMixin defines a validity_field I think, so maybe you can use `self.validity_field’?

stuaxo:
> Hm, there shouldn’t be any descriptions with a valid_between… so you should be able to just change it without the condition?
> 
I'll check if this works, I was hoping for a reply like this as maybe it was something that got missed in the refactor ?

> Also the ValidityMixin defines a validity_field I think, so maybe you can use `self.validity_field’?
Will give this a try too...

simonwo:
> I'll check if this works, I was hoping for a reply like this as maybe it was something that got missed in the refactor ?

Yeah, I think it just got missed.

> TODO: need to add tests for this.

Perhaps you could implement [this test](https://teams.microsoft.com/l/message/19:80f6ef27ef454a87a8c4844d3aa5b3f7@thread.tacv2/1622129986648?groupId=7dec05eb-83aa-4ca6-8bff-bf9b86f579e7)?

stuaxo:
> > I'll check if this works, I was hoping for a reply like this as maybe it was something that got missed in the refactor ?
> 
> Yeah, I think it just got missed.
> 
> > TODO: need to add tests for this.
> 
> Perhaps you could implement [this test](https://teams.microsoft.com/l/message/19:80f6ef27ef454a87a8c4844d3aa5b3f7@thread.tacv2/1622129986648?groupId=7dec05eb-83aa-4ca6-8bff-bf9b86f579e7)?

That's the one, this ticket is basically adding some code you already wrote and picking your brain !

Managed to get away without it up to now as I had the enclosed patch running locally but needed to come back and do it properly once it came to getting Kristof on an up to date codebase.

simonwo (approved):

andyhd (approved):
Nice test!

stuaxo:
> Nice test!

It should be... Simon wrote the initial version in a teams a conversation... I just tweaked it. 
As usual with code from @simonwo I had to go and learn how some part of it worked (I hadn't used `ids=` in pytest)

PR #319, opened 2021-06-11
Positive cases for "fuzzy" business rule tests

simonwo:
Some of our business rules are only tested fully by fuzzy test cases that sometimes exercise all code paths and sometimes do not. This makes the code coverage vibrate around the true value as some lines are only randomly covered.

So this commit adds positive test cases to:

- ON6
- ME70
- QA5
- ME40
- all uniqueness business rules

### DIFF
--- a/common/tests/util.py
+++ b/common/tests/util.py
@@ -18,6 +18,7 @@ from lxml import etree
 from common.renderers import counter_generator
 from common.serializers import validate_taric_xml_record_order
 from common.util import TaricDateRange
+from common.util import get_field_tuple
 
 INTERDEPENDENT_IMPORT_IMPLEMENTED = True
 UPDATE_IMPORTER_IMPLEMENTED = True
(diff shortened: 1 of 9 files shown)


common/tests/util.py
```diff
@@ -83,6 +84,39 @@ def check_validator(validate, value, expected_valid):
...
+
+
+def make_non_duplicate_record(factory, identifying_fields=None):
+    """Creates two records using the passed factory that are not duplicates of
+    each other and returns the record created last."""
+    existing = factory.create()
+    not_duplicate = factory.create()
+    identifying_fields = identifying_fields or list(
+        factory._meta.model.identifying_fields,
+    )
```
andyhd, line 111: Tiny quibble - I think it is clearer to write
```python
if identifying_fields is None:
    identifying_fields = list(factory._meta.model.identifying_fields)
```

common/tests/util.py
```diff
@@ -83,6 +84,39 @@ def check_validator(validate, value, expected_valid):
             pytest.fail(f'Expected validation error for value "{value}"')
 
 
+def make_duplicate_record(factory, identifying_fields=None):
+    """Creates two records using the passed factory that are duplicates of each
+    other and returns the record created last."""
+    existing = factory.create()
+
+    # allow overriding identifying_fields
+    if identifying_fields is None:
```
andyhd, line 93: Just checking - should this be `if not identifying_fields:` in case the value is an empty list/tuple?

common/tests/util.py
andyhd, line 111: Tiny quibble - I think it is clearer to write 
```python
if not identifying_fields:
    identifying_fields = list(factory._meta.model.identifying_fields)
```
(and consistent with `make_duplicate_record` too)

common/tests/util.py
```diff
@@ -83,6 +84,39 @@ def check_validator(validate, value, expected_valid):
...
+        identifying_fields = list(factory._meta.model.identifying_fields)
+
+        if hasattr(existing, "valid_between"):
```
andyhd, line 96: Do we need something similar for `validity_start`?

conftest.py
```diff
@@ -542,26 +546,28 @@ def get_bucket():
...
+    ids=(
+        "duplicate",
+        "not_duplicate",
+    ),
+)
+def assert_handles_duplicates(request):
+    def do_assert(
+        factory: Type[factories.TrackedModelMixin],
+        business_rule: Type[BusinessRule],
+        identifying_fields: Dict[str, Any] = None,
```
andyhd, line 563: The type hint should be `Optional[dict[str, Any]]` because the default value is `None`

andyhd (approved):
Nice to have more rigorous tests! A couple of comments, but nothing blocking.

andyhd (approved):

common/tests/util.py
simonwo: I think the empty list/tuple is a slightly different case… maybe you'd want to specify explicitly that a factory should already be producing duplicate records? Don't think this happens anywhere but I think it is a different semantic meaning?

common/tests/util.py
simonwo: Actually I think this is a legacy from when we used datetimes, and hence the time could change from moment to moment? So I've removed.

PR #320, opened 2021-06-11
TP-822: SQLite export and upload

simonwo:
This is a first pass at an export system that will output SQLite and upload it to some S3 bucket. The __init__.py docstring explains how the system works.

There are still a number of operational questions that I have, which would be great to get some review feedback on:

- How do we feel about this practice of using the `psql` and `sqlite3` CLI tools? This is the simplest and fastest way of getting what we want, but are we able to run these in a PaaS envrionment?
- How do we feel about using the `settings.SQLITE` switch to control which fields are in the schema? And about using the Django migration system and then cleaning it up afterwards? Again, this was the simplest way of achieving what we need.

TODO:
- Use a router to not add system tables (e.g. users) to the schema, even if we don't populate those tables.
- Schedule the task from somewhere – either on export, or on a schedule.


settings/common.py
```diff
@@ -102,7 +105,7 @@
     "crispy_forms_gds",
 ]
 
-INSTALLED_APPS = [*DJANGO_CORE_APPS, *THIRD_PARTY_APPS, *TAMATO_APPS]
+INSTALLED_APPS = [*DJANGO_CORE_APPS, *THIRD_PARTY_APPS, *TAMATO_APPS, *DOMAIN_APPS]
```
stuaxo, line 108: I guess there may be some environment variable to run this in different modes in future ?

stuaxo:
Like the rest of this project there's stuff in here that makes you think "hang on a second..." that then makes sense with the explanation.

Not being able to use date-range natively is a bit of a trap for future developers.. it's a shame there isn't automation for this -- but since that's hard*, it would be nice to have some sort of custom linter for this.

*this would probably involve using the AST, or a library like Astor to do the transformation, is also a little on the magical side.

stuaxo:
Not sure if I missed it, but the article below mentioned you need to convert any boolean variables: 'true' -> 't', 'false' -> 'f'

The tip about putting everything into a single transaction for speed may help too (though not sure if this comes with some memory implications).
https://manuelvanrijn.nl/blog/2012/01/18/convert-postgresql-to-sqlite/

simonwo:
> Not sure if I missed it, but the article below mentioned you need to convert any boolean variables: 'true' -> 't', 'false' -> 'f'

Yeah, I went with 0 and 1 which also works, here: https://github.com/uktrade/tamato/pull/320/files#diff-49fbc00ea871f2d2fe98eec1ec7696f7f0874422e9bece3ab1e1a7bb71a4990eR39

stuaxo:
> > Not sure if I missed it, but the article below mentioned you need to convert any boolean variables: 'true' -> 't', 'false' -> 'f'
> 
> Yeah, I went with 0 and 1 which also works, here: https://github.com/uktrade/tamato/pull/320/files#diff-49fbc00ea871f2d2fe98eec1ec7696f7f0874422e9bece3ab1e1a7bb71a4990eR39

Maybe this should be named `column_to_expr` to match the function underneath it that doesn't have `name` in it's name ?

exporter/sqlite/script.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # Run an SQLite import statement, which will call the passed command and
+        # load a database row for every line in the CSV. The columns need to be
+        # supplied in the order they are defined in the SQLite schema.
+        print(
+            ".import '|psql -f \"{0}\"' {1}".format(
+                str(sql_filename.absolute()),
+                model._meta.db_table,
+            ),
+        )
```
stuaxo, line 127: Elsewhere in the code we have been using `sys.stdout` instead of print statements.

exporter/sqlite/script.py
```diff
@@ -0,0 +1,136 @@
...
+        queryset = queryset.values(*output_columns)
+
+        # Generate the SQL necessary to generate the queryset for psql to run later.
+        compiler = queryset.query.get_compiler(using=queryset.db)
+        sql: bytes = compiler.connection.cursor().mogrify(*compiler.as_sql())
+
+        # Put the SQL that Postgres should run into a temporary file. This is
+        # necessary because SQLite cannot handle escaping of quotes, so
+        # including the SQL directly in the import script is not possible.
+        sql_filename = self.output_directory.joinpath(model._meta.db_table + ".sql")
```
stuaxo, line 113: Since `output_directory` is a Path, this may as well be:

sql_filename = self.output_directory / "{model._meta.db_table}.sql"

exporter/sqlite/script.py
```diff
@@ -0,0 +1,136 @@
...
+            output_columns.append(output_column)
+
+        if hasattr(queryset, "has_approved_state"):
+            queryset = queryset.has_approved_state()
+
+        queryset = queryset.values(*output_columns)
+
+        # Generate the SQL necessary to generate the queryset for psql to run later.
+        compiler = queryset.query.get_compiler(using=queryset.db)
+        sql: bytes = compiler.connection.cursor().mogrify(*compiler.as_sql())
```
stuaxo, line 108: Magic !  (not really - but this looks like a fun bit of django I'll have to go and read up on).

exporter/sqlite/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+logger = logging.getLogger(__name__)
+
+
+@app.task
+def export_and_upload_sqlite():
+    storage = S3Boto3Storage(bucket_name=settings.SQLITE_STORAGE_BUCKET_NAME)
+    latest_envelope = Envelope.objects.order_by("envelope_id").last()
+    export_filename = storage.generate_filename(
+        f"{settings.SQLITE_STORAGE_DIRECTORY}{latest_envelope.envelope_id}.db",
+    )
```
stuaxo, line 21: `path.join` might be good here in case someone omits a slash from the end of `setting.SQLITE_STORAGE_DIRECTORY`.

exporter/sqlite/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+
+    logger.debug("Checking for need to upload tariff database %s", export_filename)
+    if storage.exists(export_filename):
+        logger.debug("Database %s already present", export_filename)
+        return
```
stuaxo, line 26: Is it worth returning something here (and then later the caller could return an errorlevel on the commandline ?

exporter/sqlite/script.py
stuaxo: Makes more sense seeing this is psycopg2 API.

kintisheff:
I wonder if we should also ensure that sqlite3 is installed in the containerised environment, in case people want to run the `dump_sqlite` command from there; the base image we use `node:lts-buster-slim-bash` doesn't include that package by default. So we might add the following after line 4 in `Dockerfile`:
`apt-get install sqlite3 libsqlite3-dev`

exporter/sqlite/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+
+from django.apps import apps
+from django.conf import settings
+
+from exporter.sqlite import runner
+from exporter.sqlite import script
+
+
+def make_export_script(sqlite: runner.Runner, directory: Path = None):
+    directory = directory or Path(TemporaryDirectory().name)
```
kintisheff, line 47: There's a known issue with TemporaryDirectory() (explained [here](https://bugs.python.org/issue31959)), which may cause the tmpdir to not be found (as indeed happened a couple of times when I ran the script).
Following the issue page comments, a solution might be:
```
if directory is None: # explicitly avoiding the case of directory=""
    td = TemporaryDirectory()
    directory = Path(td.name)
```

simonwo: Oh good catch, this is some eager garbage-collect in action.

simonwo: I am going to set `PYTHONWARNINGS=default` in my local environment to see more stuff like this in the future.

simonwo:
From @michalc:

>>How do we feel about this practice of using the psql and sqlite3 CLI tools? This is the simplest and fastest way of getting what we want, but are we able to run these in a PaaS envrionment?
>
>We run psql fine in data-flow (and then pg_restore) which is all in PaaS, and have had no problems. It does seem a bit odd admittedly to be firing up processes...

stuaxo:
> From @michalc:
> 
> > > How do we feel about this practice of using the psql and sqlite3 CLI tools? This is the simplest and fastest way of getting what we want, but are we able to run these in a PaaS envrionment?
> > 
> > 
> > We run psql fine in data-flow (and then pg_restore) which is all in PaaS, and have had no problems. It does seem a bit odd admittedly to be firing up processes...

I guess `psql` is increasing our attack surface a little bit, but I have no idea how much - by the time someone has access to the container django is running in they have a python interpreter with django, sqlite and access to postgres, so I guess not having `psql` would mostly slow people down a bit (I'm probably wrong here).

As @kintisheff mentioned, it would be a nice-to-have to have this in a different container.

> ... the base image we use node:lts-buster-slim-bash doesn't include that package by default.
@kintisheff the idea is a good one, though the node container is just for building the node parts of the site, not the base when django runs.

If we were running a more paranoid architecture, we would make the celery worker use a container that extends the one django runs in it, adding psql and sqlite.

It doesn't add too much since python already contains the sqlite shared library, and postgres bindings, so even if psql isn't there the libraries are.   It probably provides a little protection against script kiddie tools that are looking for psql.

common/models/mixins/validity.py
```diff
@@ -26,7 +28,11 @@ class ValidityMixin(models.Model):
...
     """
 
-    valid_between = TaricDateRangeField(db_index=True)
+    if settings.SQLITE:
+        validity_start = DateField(db_index=True, null=True, blank=True)
+        validity_end = DateField(db_index=True, null=True, blank=True)
+    else:
+        valid_between = TaricDateRangeField(db_index=True)
 
     validity_field_name: str = "valid_between"
```
stuaxo, line 37: ^ what happens to this in SQLITE mode ?

stuaxo:
@simonwo so the big thing in here, is having to toggle between the `SQLITE` mode or not... it would be interesting to know what other options you looked at and how evil they were (all the things I thought of were fairly magic - like using Astor to transform the code).

This article is interesting -
https://blog.elsdoerfer.name/2008/01/08/fuzzydates-or-one-django-model-field-multiple-database-columns/

Using a hack, he managed to create a date field, that actually has another hidden field it uses behind the scenes (precision)
I wonder if the same technique could be used to swap out the underlying fields - or does it just get too complex.

I'm assuming it is too tricky ... I guess integration testing will catch that the sqlite exporter failing when someone adds a field that is a DateRange without putting it behind a toggle ?

common/models/mixins/validity.py
simonwo: Hm, nothing as yet…

simonwo:
> This article is interesting -
> https://blog.elsdoerfer.name/2008/01/08/fuzzydates-or-one-django-model-field-multiple-database-columns/
> 
> Using a hack, he managed to create a date field, that actually has another hidden field it uses behind the scenes (precision)
> I wonder if the same technique could be used to swap out the underlying fields - or does it just get too complex.

Hm, interesting, I never found anything quite as advanced as this. I did read the code of `contribute_to_class` and thought it could work, but yeah wasn't sure what other methods I would need to override. So maybe this could work? (But it maybe also that his method won't work anymore, as Django seems to move quite quickly underneath it's "public" API). 

I was interested in whether we could define views in the SQlite using just Django querysets without having to write manual SQL. E.g. we could turn `Measures.objects.as_at("now")` into a view, and store that querying logic in the SQLite for users to use (else they'll basically have to reimplement all our queryset methods themselves). I think one of the reasons I stopped persuing this method was that it wasn't clear to me how to transform things like `valid_between__contains` to query the two different fields individually, although I guess that probably could be made to work… But I suppose with the current method that's still an open problem anyway.

So maybe there's some more exploring to do here then. One thing I will say is that this method doesn't introduce a lot of "magic" – it's pretty clear what the if-statement is doing for the most part.
 
> I'm assuming it is too tricky ... I guess integration testing will catch that the sqlite exporter failing when someone adds a field that is a DateRange without putting it behind a toggle ?

It won't currently, but yes, we could make that a test failure.

simonwo:
Ok, one improvement I have made is that we don't rely on the command-line tools anymore. I thought about how to test this and having everything out of process made it very difficult. The good news is that whilst it's slower, it's still acceptably fast to generate the SQLite export this way.

exporter/sqlite/__init__.py
```diff
@@ -43,15 +34,15 @@
...
-
-    with script.ImportScript(directory) as import_script:
-        names = (name.split(".")[0] for name in settings.DOMAIN_APPS)
-        models = chain(*[apps.get_app_config(name).get_models() for name in names])
-        for model in models:
-            columns = list(sqlite.read_column_order(model._meta.db_table))
-            import_script.add_table(model, columns)
+def make_export_script(sqlite: runner.Runner) -> script.ImportScript:
+    import_script = script.ImportScript()
+    names = (name.split(".")[0] for name in settings.DOMAIN_APPS)
```
stuaxo, line 39: I guess `name.partition(".")[0]` would be a little more efficient, but not a blocker :man_shrugging:

exporter/sqlite/script.py
```diff
@@ -64,34 +49,55 @@ def add_column_to_queryset(column: str, queryset: QuerySet) -> Tuple[QuerySet, s
...
+    executed.
 
-    This class will just `print` its output. Callers who want a string or to
-    redirect the output should make use of `contextlib.redirect_stdout`.
+    Once the script is finished, access the operations using the ``operations``
+    property.
     """
 
-    def __init__(self, directory: Path) -> None:
-        self.output_directory = directory
```
stuaxo: Micro ORMs are fun to write :)

stuaxo (approved):

simonwo:
Ok, I think this is ready for a review.

Here's what's changed:

- I've moved away from using the CLI tools which actually simplifies things a lot. It's slower, but much easier to test.
- So there are now a bunch of tests that check that the data we expect gets put into the SQLite, and that the entire export and upload process works.
- I've configured the export to operate every 30 minutes using celery. It will check to see if any new data has been published and will upload a new Sqlite export if there is something new.

The first TODO above is still relevant but not unsafe so we can revisit that later – pushing this to prod will be enough to get all our tariff data seamlessly flowing into Data Workspace!

common/models/transactions.py
```diff
@@ -107,6 +109,23 @@ def add_to_transaction(self, instance, **kwargs):
...
+        return (
+            cls.objects.exclude(
+                workbasket=WorkBasket.objects.first(),
+            )
+            .filter(
+                workbasket__status__in=WorkflowStatus.approved_statuses(),
+            )
+            .order_by("order")
+            .last()
+        )
```
stuaxo, line 127: Does this work as expected when app is initialised and there are no Workbaskets yet ?

exporter/sqlite/tasks.py
```diff
@@ -6,24 +6,26 @@
...
+from common.models.transactions import Transaction
 from exporter import sqlite
-from taric.models import Envelope
 
 logger = logging.getLogger(__name__)
 
 
 @app.task
-def export_and_upload_sqlite():
+def export_and_upload_sqlite() -> bool:
```
stuaxo, line 16: Needs __doc__ mentioning the return values.

exporter/tests/test_sqlite.py
```diff
@@ -107,3 +109,54 @@ def test_valid_between_export(factory, date_ranges, date_range, sqlite_database:
         assert validity_end == object.valid_between.upper.strftime(r"%Y-%m-%d")
     else:
         assert validity_end is None
+
+
+def test_export_task_does_not_reupload():
```
stuaxo, line 114: Needs a tiny bit of doc on the "why" of what this guards against.

common/models/transactions.py
simonwo: Hm. Last will return `None` so that at least it doesn't explode. But this does make an assumption that the first workbasket loaded is a seed file and should be ignored. So if you run this locally without the production data loaded it will ignore the first workbasket and only start returning something from the second.

Ultimately this is the correct bahaviour for the production data… but maybe our data model needs to reflect that? Maybe all of the seed file transactions should have a flag that they should be ignored somehow?

simonwo: We could renumber the seed file transaction IDs to be negative, and then they would order correctly by magic?

stuaxo: I was imagining something like a `Coalesce(.... ,Value(0))`  or `or 0` ?

I guess having the seed file end at 0 makes it easy to find out where our updates start, but maybe not sure if there are disadvantages to making the field signed... 

It could be surprise to a new developer if it started from some arbitrary number.

kintisheff: ^ the same questions occurs in cases where you run tamato webserver locally with no workbaskets in the database.

paulpepper-trade (commented):
Some initial, general questions:

1. For context really; how do we expect consumers to use a (partial) export of TaMaTo in sqlite format and why not dump a sanitised version of our Postgresql DB (sorry if I've missed discussion around this already).
2. I suppose that ^ might inform us on what workarounds and technologies we might apply (e.g. the `daterange` conversion)?
3. The sqlite export capability appears to have broader use beyond TaMaTo data. Perhaps useful to place the changes in a separate repo? That might help with isolating data transformation from TaMaTo-specific changes (separation of concerns) and may draw-in contribution and application from the broader open source community.

exporter/sqlite/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+from django.apps import apps
+from django.conf import settings
+
+from exporter.sqlite import runner
+from exporter.sqlite import script
+from exporter.sqlite import tasks  # noqa
+
+
+def make_export_script(sqlite: runner.Runner) -> script.ImportScript:
+    import_script = script.ImportScript()
```
paulpepper-trade, line 39: It's not immediately obvious (to me at least) why `make_export_script()` returns an instance of `ImportScript`. If the naming is correct, then a small explanatory comment may be useful.

stuaxo:
Something I forgot, that may already be in here: does this sanitise the user tables ?

simonwo:
So the main reason we've been looking at SQLite rather than another kind of database dump is that the barrier to entry for open data users is much lower. Almost every language has excellent support for connecting to SQLite databases on disk (or even remote over HTTP) and there's no loading/restoring step required (e.g. you don't have to set up Postgres, you don't have to work out how to configure it, and you don't need to restore the dump somewhere). So if you want to open the database and just start analysing there's nothing much quicker. It's also amenable to us adding more user-facing value without massively increasing our own support footprint using off-the-shelf tools like [Datasette](https://github.com/simonw/datasette).

I'd definitely be up for isolating this and separating it out – I feel like it would be useful elsewhere, but also as it's such an immature part of the code right now I'm thinking it might be good to keep it around a little longer in this main repo where it's easier to make changes?

Yes, only the models contained in `settings.DOMAIN_APPS` are exported.

exporter/sqlite/__init__.py
simonwo: Yeah fair enough – I guess it's a script to export data from postgres and import it to sqlite but even just picking one would be better than having both! I'll come up with something better.

paulpepper-trade (approved):

simonwo:
Ok, having added some docstrings and renames to address the remaining comments, I'm going to merge this (@kintisheff also let me know verbally that he is happy).

There are still some obvious places to improve, but it seems like this works well, is a solid base to start from, and will deliver value from being deployed in prod.

PR #321, opened 2021-06-14
TP-825 - Add Update Type business rules

rachaelcodes:
Adds business rules to Update Types:

- The first update must be of type Create.
- Subsequent updates must not be of type Create. 
- After an update of type Delete, there must be no further updates. 
- Only one version of the object may be updated in a single transaction.

### DIFF
--- a/additional_codes/tests/test_models.py
+++ b/additional_codes/tests/test_models.py
@@ -12,3 +12,12 @@ def test_additional_code_in_use(in_use_check_respects_deletes):
         factories.MeasureFactory,
         "additional_code",
     )
+
+
+def test_additional_code_update_types(check_update_validation):
+    assert check_update_validation(factories.AdditionalCodeTypeFactory)
+    assert check_update_validation(
+        factories.AdditionalCodeFactory,
+        description_factory=factories.AdditionalCodeDescriptionFactory,
+    )
+    assert check_update_validation(factories.AdditionalCodeDescriptionFactory)
--- a/additional_codes/tests/test_models.py
+++ b/additional_codes/tests/test_models.py
@@ -15,19 +15,17 @@ def test_additional_code_in_use(in_use_check_respects_deletes):
 
 
 @pytest.mark.parametrize(
-    ["factory", "description_factory"],
+    "factory",
     [
-        (factories.AdditionalCodeTypeFactory, None),
-        (factories.AdditionalCodeFactory, factories.AdditionalCodeDescriptionFactory),
-        (factories.AdditionalCodeDescriptionFactory, None),
+        factories.AdditionalCodeTypeFactory,
+        factories.AdditionalCodeFactory,
+        factories.AdditionalCodeDescriptionFactory,
     ],
 )
 def test_additional_code_update_types(
     factory,
-    description_factory,
     check_update_validation,
 ):
     assert check_update_validation(
         factory,
-        description_factory=description_factory,
     )
--- a/additional_codes/tests/test_models.py
+++ b/additional_codes/tests/test_models.py
@@ -20,6 +20,7 @@ def test_additional_code_in_use(in_use_check_respects_deletes):
         factories.AdditionalCodeTypeFactory,
         factories.AdditionalCodeFactory,
         factories.AdditionalCodeDescriptionFactory,
+        factories.FootnoteAssociationAdditionalCodeFactory,
     ],
 )
 def test_additional_code_update_types(
--- a/additional_codes/tests/test_models.py
+++ b/additional_codes/tests/test_models.py
@@ -14,10 +14,20 @@ def test_additional_code_in_use(in_use_check_respects_deletes):
     )
 
 
-def test_additional_code_update_types(check_update_validation):
-    assert check_update_validation(factories.AdditionalCodeTypeFactory)
+@pytest.mark.parametrize(
+    ["factory", "description_factory"],
+    [
+        (factories.AdditionalCodeTypeFactory, None),
+        (factories.AdditionalCodeFactory, factories.AdditionalCodeDescriptionFactory),
+        (factories.AdditionalCodeDescriptionFactory, None),
+    ],
+)
+def test_additional_code_update_types(
+    factory,
+    description_factory,
+    check_update_validation,
+):
     assert check_update_validation(
-        factories.AdditionalCodeFactory,
-        description_factory=factories.AdditionalCodeDescriptionFactory,
+        factory,
+        description_factory=description_factory,
     )
-    assert check_update_validation(factories.AdditionalCodeDescriptionFactory)
--- a/certificates/tests/test_business_rules.py
+++ b/certificates/tests/test_business_rules.py
@@ -4,6 +4,7 @@ from django.db import DataError
 from certificates import business_rules
 from common.business_rules import BusinessRuleViolation
 from common.tests import factories
+from common.validators import UpdateType
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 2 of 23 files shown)


common/models/mixins/description.py
```diff
@@ -46,10 +46,7 @@ def get_url(self, action="detail"):
 
     def __str__(self):
         return self.identifying_fields_to_string(
-            identifying_fields=(
-                self.described_object_field.name,
-                "valid_between",
-            ),
+            identifying_fields=(self.described_object_field.name,),
```
rachaelcodes, line 49: Small bug fix to allow you to view objects in the shell again. Let me know if you'd prefer this go in its own PR.

andyhd: Shouldn't this be `identifying_fields=(self.described_object_field.name, "validity_start")`?

certificates/business_rules.py
```diff
@@ -51,3 +52,14 @@ class CE7(ValidityPeriodContained):
     of the certificate."""
 
     container_field_name = "certificate_type"
+
+
+class CertificateUpdateValidity(UpdateValidity):
```
andyhd, line 57: I don't think you need to add a class - just add `UpdateValidity` to the `business_rules` list in each model.

certificates/tests/test_business_rules.py
```diff
@@ -146,3 +147,59 @@ def test_CE7(date_ranges):
             valid_between=date_ranges.overlap_normal,
         )
         business_rules.CE7(certificate.transaction).validate(certificate)
+
+
+def test_CertificateUpdateValidity_first_update_must_be_Create():
```
andyhd, line 152: I'd be inclined to put these tests in `common/tests/test_business_rules.py`, as the code under test will be the same for each model, but I am lazy.

There is value in testing each model has the business rule correctly applied, so I recommend making a few helper functions you can call from these tests, passing the model factory class as a parameter, similar to `in_use_check_respects_deletes()` in `conftest.py`.

andyhd (commented):
Looks good 👍  Just a couple of suggestions

certificates/tests/test_business_rules.py
rachaelcodes: Yes, particularly now I'm thinking that each app has a number of models to be tracked (it's not just the Certificates, but the CertificateTypes and CertificateDescriptions etc).

common/business_rules.py
```diff
@@ -411,3 +411,33 @@ def validate(self, model):
...
+    def validate(self, model):
+        existing_objects = model.__class__.objects.filter(
+            version_group=model.version_group,
+        ).exclude(sid=model.sid)
+
+        if existing_objects.exists():
+            if model.update_type == UpdateType.CREATE:
+                raise self.violation(model)
+
+            if existing_objects.filter(update_type=UpdateType.DELETE).exists():
```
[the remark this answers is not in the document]
simonwo: I think it’s still valid. Measures search is currently quite slow because of many repeated queries… I guess the overhead of each query is still significant?

common/models/mixins/description.py
simonwo: Also check out @stuaxo PR which also fixes this

certificates/tests/test_business_rules.py
```diff
@@ -146,3 +147,59 @@ def test_CE7(date_ranges):
...
+
+def test_CertificateUpdateValidity_later_updates_must_not_be_Create():
+    """Updates to a Certificate after the first update must not be of type
+    Create."""
+    first_certificate = factories.CertificateFactory.create()
+    second_certificate = factories.CertificateFactory.create(
+        update_type=UpdateType.CREATE,
+        version_group=first_certificate.version_group,
+    )
+
```
simonwo, line 168: It would be good if you could add some positive test cases for these, I.e. checking that the rule doesn’t fire in good conditions. Judicious use of `raises_if` and `pytest.mark.parametrize` should make that easier.

common/business_rules.py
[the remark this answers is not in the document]
simonwo: …but the implementation logic looks good to me!

common/models/mixins/description.py
stuaxo: Will fix that up when I get back, this morning..

additional_codes/tests/test_models.py
```diff
@@ -12,3 +12,12 @@ def test_additional_code_in_use(in_use_check_respects_deletes):
...
     )
+
+
+def test_additional_code_update_types(check_update_validation):
+    assert check_update_validation(factories.AdditionalCodeTypeFactory)
+    assert check_update_validation(
+        factories.AdditionalCodeFactory,
+        description_factory=factories.AdditionalCodeDescriptionFactory,
+    )
+    assert check_update_validation(factories.AdditionalCodeDescriptionFactory)
```
rachaelcodes, line 23: What's happening with the `FootnoteAssociationAdditionalCode` model - there's a comment in `additional_codes/models.py` about it not being used. Is that about the whole model, and if so, does that mean it's not included in these validations?

simonwo: So we don't really use them in the UK tariff because they don't appear anywhere online. We might use them in the future though… so if it's not any extra work, I would add these validations to them too.

stuaxo: @rachaelcodes merged this earlier - try rebasing against master and see if things are OK now.

workbaskets/tests/test_models.py
```diff
@@ -689,9 +693,11 @@ def test_workbasket_accepted_updates_current_tracked_models(new_workbasket, vali
...
 
 def test_workbasket_errored_updates_tracked_models(new_workbasket, valid_user):
-    original_footnote = factories.FootnoteFactory.create()
-    new_footnote = original_footnote.new_version(workbasket=new_workbasket)
-
+    original_footnote = factories.FootnoteFactory.create(update_type=UpdateType.CREATE)
+    new_footnote = original_footnote.new_version(
+        workbasket=new_workbasket,
+        update_type=UpdateType.UPDATE,
+    )
```
simonwo, line 700: Hm, I'm suprised this change is necessary… don't they all have update_type=CREATE by default?

conftest.py
```diff
@@ -657,6 +658,164 @@ def check(
...
+        transaction = factories.TransactionFactory.create()
+        first_instance = factory.create(update_type=UpdateType.DELETE)
+        second_instance = factory.create(
+            transaction=transaction,
+            update_type=UpdateType.UPDATE,
+            version_group=first_instance.version_group,
+        )
+
+        with pytest.raises(
+            BusinessRuleViolation,
```
simonwo, line 747: You might also be able to supply `UpdateValidity.Violation` instead of using a match – each business rule gets its own Violation class.

conftest.py
```diff
@@ -657,6 +658,164 @@ def check(
...
+        except BusinessRuleViolation:
+            pytest.fail()
+
+        return True
+
+    return check
+
+
+@pytest.fixture
+def check_later_update_validation():
```
simonwo, line 693: If you wanted to be uber terse, you might even be able to replace these four test methods with one `parametrize` which provides `("first_update_type", "second_update_type", "error_expected")`, where the `first_update_type` being `None` would mean the test doesn't create a first instance, and `error_expected` gets passed to `raises_if`…

Terse does not imply better though…!

rachaelcodes: Hmm I was having some difficulty getting parametrize to play nicely with `@pytest.fixture` - will have another go.

workbaskets/tests/test_models.py
rachaelcodes: Yes, you're right. I'd left that in after trying it to fix the test (and needing to change something else instead).

conftest.py
simonwo: The `fxiture` method takes it's own `params=` argument, which you might need to use instead :)

rachaelcodes: Coming back to this, I'm not sure that this would be a helpful abstraction, as it's not just update_type that's being changed: (for example, in the final rule, we're testing a case where the version_group is also explicitly set). I think that the abstraction that took this into account could easily end up less readable than this version.

simonwo: Yep, fair enough!

additional_codes/tests/test_models.py
```diff
@@ -12,3 +12,21 @@ def test_additional_code_in_use(in_use_check_respects_deletes):
...
     )
+
+
+@pytest.mark.parametrize(
+    "factory",
+    [
+        factories.AdditionalCodeTypeFactory,
+        factories.AdditionalCodeFactory,
+        factories.AdditionalCodeDescriptionFactory,
+        factories.FootnoteAssociationAdditionalCodeFactory,
```
simonwo, line 23: Hm, how does this work? I can't spot anywhere that the business rule is added to this model?

simonwo: Ah I see, the test makes sure that the business rule fires correctly when passed an instance of this model, but it doesn't check that the business rule is actually defined on it.

To be fair, none of our other business rule tests do that either… but should we add something like `assert UpdateValidity in type(model).business_rules` or something? Given that we are testing each model individually, we can also check that the business rule is actually defined on it?

simonwo (approved):
Looks great!

simonwo (approved):

PR #322, opened 2021-06-21
Stop commodity codes being parents of themselves

simonwo:
There was a bug in the indents importer which meant that a commodity code could end up as a parent of itself if an indent is updated without changing any of its fields. This was because the parent search didn't correctly only allow the item id to be equal in the case that the suffix is also less.

This also fixes up some tests to use the import infrastructure which makes it easy to swap out our XML export mechanism later.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -341,6 +341,7 @@ def run_xml_import(valid_user, settings):
         settings.SKIP_WORKBASKET_VALIDATION = True
 
         model = factory()
+        model_class = model.__class__
         assert isinstance(
             model,
             TrackedModel,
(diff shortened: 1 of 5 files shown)


conftest.py
```diff
@@ -341,6 +341,7 @@ def check(
         settings.SKIP_WORKBASKET_VALIDATION = True
 
         model = factory()
+        model_class = model.__class__
```
stuaxo, line 344: Elsewhere in Tamato we've been using `type(model)`

stuaxo (approved):

PR #324, opened 2021-06-22
TP-92 Create a new foonote

DIT-CJWaszczuk:
## Summary 
Allow the user to create footnotes 
### Create screen
image
### Front page
image

### DIFF
--- a/common/jinja2/common/confirm_create.jinja
+++ b/common/jinja2/common/confirm_create.jinja
@@ -1,0 +1,20 @@
+{% extends "layouts/confirm.jinja" %}
+
+{% set page_title = "Create a new " ~ object._meta.verbose_name %}
+
+{% block beforeContent %}
+  {{ govukBreadcrumbs({
+    "items": [
+      {"text": "Home", "href": url("index")},
+      {"text": page_title}
+    ]
+  }) }}
+{% endblock %}
+
+{% block panel %}
+  {{ govukPanel({
+    "titleText": object._meta.verbose_name|title ~ " " ~ object|string ~ " has been created",
+    "text": "This new " ~ object._meta.verbose_name ~ " has been added to your tariff changes",
+    "classes": "govuk-!-margin-bottom-7"
+  }) }}
+{% endblock %}
--- a/common/jinja2/components/description_help.jinja
+++ b/common/jinja2/components/description_help.jinja
@@ -1,0 +1,43 @@
+{% from "components/table/macro.njk" import govukTable %}
+{% from "components/details/macro.njk" import govukDetails %}
+
+{% set description_html %}
+  <p class="govuk-body">
+    It is possible to use HTML when defining descriptions. The table below shows 
+    a few examples of useful codes that you may want to use. Please use HTML sparingly and only where 
+    absolutely necessary.
+  </p>
+  {{ govukTable({
+    "head": [
+      {"text": "Code(s)"},
+      {"text": "Usage"},
+      {"text": "Example", "classes": "govuk-!-width-one-third"},
+    ],
+    "rows": [[
+      {"text": "<b></b>"},
+      {"text": "Make a piece of content bold. It is preferable to use the tag pair <strong></strong> instead of <b></b>, however many of the Taric descriptions already use standard bold tags."},
+      {"html": "This is " ~ '<b>bold</b>'|e ~ " content <br> becomes <br> This is <b>bold<b> content"},
+    ], 
+    [
+      {"text": "<em></em>"},
+      {"text": "Italicise a piece of content"},
+      {"html": "This is " ~ '<em>italicised</em>'|e ~ " content <br> becomes <br> This is <em>italicised</em> content"},
+    ],
+    [
+      {"text": "<u></u>"},
+      {"text": "Underline a piece of content"},
+      {"html": "This is " ~ '<u>underlined</u>'|e ~ " content <br> becomes <br> This is <u>underlined</u> content"},
+    ],
+    [
+      {"text": "<a href=\'\'></a>"},
+      {"text": "Add a hyperlink"},
+      {"html": 'This is <a href="https://www.gov.uk">a hyperlink to gov.uk</a>'|e ~ '<br>becomes<br> This is <a href="https://www.gov.uk">a hyperlink to gov.uk</a>'},
+    ]
+    ]
+  }) }}
+{% endset%}
+
+{{ govukDetails({
+    "summaryText": "Help with formatting description",
+    "html": description_html
+  }) }}
--- a/footnotes/forms.py
+++ b/footnotes/forms.py
@@ -8,8 +8,11 @@ from django import forms
 
 from common.forms import CreateDescriptionForm
 from common.forms import DescriptionForm
+from common.forms import DescriptionHelpBox
 from common.forms import ValidityPeriodForm
+from common.util import get_next_id
 from footnotes import models
+from workbaskets.models import WorkBasket
 
 
 class FootnoteForm(ValidityPeriodForm):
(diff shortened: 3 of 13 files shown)


footnotes/models.py
```diff
@@ -93,6 +93,29 @@ class Footnote(TrackedModel, ValidityMixin):
...
+            )
+            new_id = max_id + 1
+            if new_id <= 999:
+                new_id = f"{new_id:03}"
+            elif new_id <= 99999:
+                new_id = f"{new_id:05}"
+            else:
+                raise ValueError("No more footnote ids available")
+
+            self.footnote_id = new_id
```
simonwo, line 115: Hm, I think this should actually be done on form load and we should give the user the option to override it. The IDs are public-facing, and whilst they don't normally have any meaning, I can think of a few instances where we might want to specifically pick a certain ID. E.g. if we have a certificate C123 we might want to pick FT123 to match.

simonwo (commented):
To answer your question in standup @DIT-CJWaszczuk, the problem with items not appearing anywhere is probably that all our views use `latest_approved` which will ignore draft data in workbaskets. We'll need to change them to use `approved_up_to_transaction(…)`.

DIT-CJWaszczuk:
Yeah I thought it would be more so to do with `get_descriptions()` not picking up the draft descriptions currently

footnotes/models.py
DIT-CJWaszczuk: I'm not quite sure how we would do this, given that we don't know the type beforehand, and given we generate the id  based on the last of each type, we wouldn't be able to generate it correctly on one page.

common/jinja2/common/confirm_create.jinja
```diff
@@ -0,0 +1,43 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l">Confirmation</span>
+      <h1 class="govuk-heading-xl">{{ page_title }}</h1>
+      
+      {{ govukPanel({
+        "titleText": object._meta.verbose_name|title ~ " " ~ object|string ~ " has been created",
+        "text": "This new " ~ object._meta.verbose_name ~ " has been added to your tariff changes",
+        "classes": "govuk-!-margin-bottom-7"
+      }) }}
```
andyhd, line 27: This template is essentially the same as `common/confirm_update.jinja`, so they should both extend from the same base template. Move the common code into `layouts/confirm.jinja` and add a new block (maybe `{% block panel %}{% endblock %}`) to override the panel message.

Then `confirm_create.jinja` would look something like
```jinja
{% extends "layouts/confirm.jinja" %}

{% set page_title = "Create a new " ~ object._meta.verbose_name %}

{% block beforeContent %}
  ...
{% endblock %}

{% block panel %}
  {{ govukPanel({
    "titleText": ... ~ " has been created",
    ...
  }) }}
{% endblock %}
```

common/jinja2/common/confirm_create.jinja
```diff
@@ -0,0 +1,43 @@
...
+{% from "components/panel/macro.njk" import govukPanel %}
+{% from "components/button/macro.njk" import govukButton %}
+
+{% set page_title = "Create a new " ~ object._meta.verbose_name %}
+
+{% block beforeContent %}
+  {{ govukBreadcrumbs({
+    "items": [
+      {"text": "Home", "href": url("index")},
+      {"text": page_title}
```
andyhd, line 12: I think there's a missing breadcrumb link here - shouldn't there be a `{"text": "Find and edit " ~ object._meta.verbose_name_plural, "href": object.get_url("list")},` link?

common/jinja2/components/description_help.jinja
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    [
+      {"text": "<u></u>"},
+      {"text": "Make a piece of content underlined"},
+      {"html": "This is " ~ '<u>underlined</u>'|e ~ " content <br> becomes <br> This is <u>underlined</u> content"},
+    ],
+    [
+      {"text": "<a href=\'\'></a>"},
+      {"text": "Add a hyperlink"},
+      {"html": "This is " ~ "<a href=:\'https://www.gov.uk\'>a hyperlink to gov.uk</a>"|e ~ "<br> becomes <br> <a href=:'https://www.gov.uk'>a hyperlink to gov.uk</a>"},
```
andyhd, line 32: There seem to be extraneous `:`s in here - `<a href=:` should be `<a href=`

footnotes/forms.py
```diff
@@ -74,6 +74,51 @@ class Meta:
...
+            "footnote_type"
+        ].label_from_instance = (
+            lambda obj: f"{obj.footnote_type_id} - {obj.description}"
+        )
+
+        self.helper = FormHelper(self)
+        self.helper.layout = Layout(
+            Field("footnote_type", context={"label_size": "govuk-label--s"}),
+            Field("start_date", context={"legend_size": "govuk-label--s"}),
+            Field.textarea("description", label_size=Size.SMALL, rows=5),
```
andyhd, line 105: If you do `self.helper.label_size = Size.SMALL` and `self.helper.legend_size = Size.SMALL`, you can reduce this to
```python
self.helper.layout = Layout(
    "footnote_type",
    "start_date",
    Field.textarea("description", rows=5),
    ...
```

footnotes/jinja2/footnotes/create.jinja
```diff
@@ -0,0 +1,14 @@
...
+{% from "components/details/macro.njk" import govukDetails %}
+
+{% set page_title = 'Create a new footnote' %}
+
+{% set description_html %}{% include 'components/description_help.jinja' %}{% endset %}
+
+{% block below_form %}
+    {{ govukDetails({
+        "summaryText": "Help with formatting description",
+        "html": description_html
```
andyhd, line 12: There is a [`HTML.details` layout object in crispy-forms-gds](https://crispy-forms-gds.readthedocs.io/en/latest/components/details.html), which could remove the need for this template.

footnotes/models.py
```diff
@@ -93,6 +94,20 @@ class Footnote(TrackedModel, ValidityMixin):
...
+        if getattr(self, "footnote_id") == "":
+            self.footnote_id = get_next_id(
+                Footnote.objects.filter(
+                    footnote_type__footnote_type_id=self.footnote_type.footnote_type_id,
+                )
+                .as_at(self.valid_between.lower)
+                .approved_up_to_transaction(self.transaction),
+                self._meta.get_field("footnote_id"),
+                max_len=3,
+            )
```
andyhd, line 107: This should go in the form save method, because that's the only place where trying to save with an empty footnote_id should happen without an error being raised.

footnotes/views.py
```diff
@@ -100,6 +102,30 @@ class FootnoteList(FootnoteMixin, TamatoListView):
...
+    template_name = "footnotes/create.jinja"
+    form_class = forms.FootnoteCreateForm
+
+    def form_valid(self, form):
+        transaction = self.get_transaction()
+        transaction.save()
+        self.object = form.save(commit=False)
+        self.object.update_type = UpdateType.CREATE
+        self.object.transaction = transaction
+        self.object.save()
```
andyhd, line 115: This should be:
```python
form.instance.update_type = UpdateType.CREATE
with transaction:
    form.save()
```

footnotes/views.py
```diff
@@ -100,6 +102,30 @@ class FootnoteList(FootnoteMixin, TamatoListView):
...
+
+        self.object_description = form.cleaned_data["footnote_description"]
+        self.object_description.described_footnote = self.object
+        self.object_description.update_type = UpdateType.CREATE
+        self.object_description.transaction = transaction
+        self.object_description.save()
```
andyhd, line 121: And you can put this into the same `with transaction:` block, eg:
```python
with transaction:
    self.object = form.save()

    description = form.cleaned_data["footnote_description"]
    description.described_footnote = self.object
    description.update_type = UpdateType.CREATE
    description.save()
```
I don't see any need to add the description object to the view as an attribute `self.object_description` - is there one?

andyhd (changes requested):
Looks good, just a typo needs fixing and the auto id increment needs moving. Plus some tidying up if you like.

footnotes/jinja2/footnotes/create.jinja
DIT-CJWaszczuk: I want to do this, but the content of that details is fairly complex, and i'm not sure the method will handle it properly. 
image

footnotes/models.py
andyhd: Ah, yeah, that would be better in a template

footnotes/jinja2/footnotes/create.jinja
andyhd: Ah yeah, that is better off in a template

common/jinja2/common/confirm_create.jinja
DIT-CJWaszczuk: Going off the design, this breadcrumb has been removed for 'creates'

footnotes/models.py
stuaxo: Tangent:  I've been thinking about ids, and their generation in the database.

The basic idea, is that one of the identifying keys is labelled as the "sequential key" and this is incremented, while all the other keys are not.

While this could be done for character based keys, for now it's just implemented for integers to keep things simple.

The code above shows some of the complications, other interesting ones are that we have fields with the same name, but different types (worked around in the first version by just checking integers).

stuaxo (approved):

common/jinja2/components/description_help.jinja
```diff
@@ -0,0 +1,35 @@
...
+      {"html": "This is " ~ '<b>bold</b>'|e ~ " content <br> becomes <br> This is <b>bold<b> content"},
+    ], 
+    [
+      {"text": "<em></em>"},
+      {"text": "Make a piece of content italicised"},
+      {"html": "This is " ~ '<em>italicised</em>'|e ~ " content <br> becomes <br> This is <em>italicised</em> content"},
+    ],
+    [
+      {"text": "<u></u>"},
+      {"text": "Make a piece of content underlined"},
```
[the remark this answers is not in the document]
DIT-CJWaszczuk: This is what happens when we don't have a dedicated content designer!

PR #326, opened 2021-06-24
Two tiny fixes to __str__ and quota origin XML

simonwo:
-  Show remote fields in model `__str__` output
- Fix validity end date position in quota origin xml

### DIFF
--- a/common/tests/models.py
+++ b/common/tests/models.py
@@ -42,6 +42,11 @@ class TestModelDescription1(TrackedModel, ValidityMixin):
     record_code = "01"
     subrecord_code = "02"
 
+    identifying_fields = (
+        "described_record__sid",
+        "valid_between",
+    )
+
     described_record = models.ForeignKey(
         TestModel1,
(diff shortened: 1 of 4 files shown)


common/tests/models.py
```diff
@@ -42,6 +42,11 @@ class TestModelDescription1(TrackedModel, ValidityMixin):
     record_code = "01"
     subrecord_code = "02"
 
+    identifying_fields = (
+        "described_record__sid",
+        "valid_between",
```
andyhd, line 47: This is only a test model, so perhaps this doesn't matter, but don't all description models have `validity_start` instead of `valid_between`?

andyhd (approved):

common/tests/models.py
stuaxo: Yeah, I think so.

simonwo: Yeah… it should probably inherit from DescriptionMixin. I did think about changing it and will happily go back and update if you think it's worth it.

PR #327, opened 2021-06-29
Allow descriptions to be blank in the database

simonwo:
Our XSD allows description fields to be blank
but these are rejected by the importer. This
commit changes that behaviour to allow blank
descriptions but ensures they are not blank using
a (runtime configurable) business rule.

This allows us to import valid XML whilst still
ensuring our own data does not contain blank
descriptions.


stuaxo:
This ticket names me wonder if there is value in an integration test that generates test data using the XSD?
https://www.davekuhlman.org/generateDS.html

andyhd (approved):
Nice

PR #331, opened 2021-07-13
Update nodejs buildpack to 1.7.56

rachaelcodes:
Specifies a particular [NodeJS buildpack](https://github.com/cloudfoundry/nodejs-buildpack/releases/tag/v1.7.56) for Cloud Foundry, as the current default does not include our version of Node.

Also updates Node version to 16.4.1 (for security fixes).

Note: this should be reverted once the Node version is included in the default nodejs buildpack.


simonwo (approved):
LGTM… do we need to peg the nodejs-buildback to a certain version, or should we always use the latest?

rachaelcodes:
> LGTM… do we need to peg the nodejs-buildback to a certain version, or should we always use the latest?

The advice from WebOps was to just change it back to `nodejs_buildpack` once this supports our Node version.

PR #332, opened 2021-07-13
Run webpack with npx

rachaelcodes:
After #331 did not fix all our Jenkins pipeline problems, this is WebOps' other suggestion.


simonwo (approved):
Happy to try it…

PR #334, opened 2021-07-14
TP-841 - Order and partition transactions into partitions: SEED_FILE, REVISION, UPDATE.

stuaxo:
This PR introduces partitioning to Transactions.

Global ordering is achieved by ordering on partition, order.

Partitions are:

SEED_FILE:   Transactions imported from the initial seed file.
REVISION:   Transactions imported or updated later.
DRAFT:   Transactions in a workbasket.

During seed file update transactions may transitions are in the SEED_FILE partition.
During ordinary operation transactions are created as DRAFT, then when the WorkBasket is approved the transactions are changed to UPDATE and the order is rewritten to start after the highest REVISION transaction.


common/models/transactions.py
```diff
@@ -41,6 +48,35 @@ def ordered_tracked_models(self):
...
+                workbasket__in=workbaskets,
+            ).last_finalised_transaction()
+            last_order = last_tx.order if last_tx is not None else 0
+            for global_id, txn in enumerate(
+                self.order_by("order"),
+                last_order + 1,
+            ):
+                txn.status = TransactionStatus.FINALISED
+                txn.order = global_id
+                txn.save()
```
simonwo, line 78: Would it be better to use [`QuerySet.update`](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.query.QuerySet.update)?

```suggestion
            self.update(
                order=F("order") + Value(last_order),
                status=TransactionStatus.FINALISED,
            )
```

common/models/transactions.py
```diff
@@ -53,20 +89,35 @@ class Transaction(TimestampedMixin):
...
     objects = TransactionManager.from_queryset(TransactionQueryset)()
 
+    @transition(
+        field=status,
+        source=TransactionStatus.DRAFT,
+        target=TransactionStatus.FINALISED,
+    )
+    def finalise_transaction(self):
+        pass
+
```
simonwo, line 120: I appreciate that the "correct" thing to do is to call `finalise_transactions` on a `TransactionQueryset` because it's obviously more performant, but shouldn't this at least let the user know that nothing has actually occured? I.e. if I call `txn.finalise_transaction()` I would expect that the transaction is now finalised.

Should the method have a one-at-a-time implementation for those cases? Or, do we even need to define this method?

I think it would be better called just `finalise`.

simonwo (commented):
I like where this is going – will need some tests ofc.

To what extent should we be using the DRAFT/FINALISED flag elsewhere? Would it be more efficient to use it in `TrackedModelQuerySet.approved_up_to_transaction(…)` for example, instead of referring to the workbasket status?

common/models/transactions.py
stuaxo: Thanks, I was blind to how this would work with an update :)

common/models/transactions.py
stuaxo: (this may be me misunderstanding DjangoFSM) - there is only a method here to tell django_fsm that this is a valid state transition: if anonymous methods were legal python I would have done something like 

```
    @transition(
        field=status,
        source=TransactionStatus.DRAFT,
        target=TransactionStatus.FINALISED,
    )
    def (self):
        pass
```

I'll probably rename this to finalise(), along with a docstring pointing out that the QS way is usually preferred, unless there is some DjangoFSM API I can call to add the valid state change.

common/models/transactions.py
```diff
@@ -41,6 +48,35 @@ def ordered_tracked_models(self):
...
+        """Set transaction status to FINALISED and ordering to global."""
+        statuses = self.distinct("status").values_list("status", flat=True)
+
+        if statuses == [TransactionStatus.FINALISED.value]:
+            logger.info("Transactions are already finalised.")
+            return
+
+        if statuses != [TransactionStatus.DRAFT.value]:
+            logger.warning("Transactions have non-DRAFT statuses: %s", statuses)
+            return
```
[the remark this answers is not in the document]
stuaxo: Yes, this using this instead of workbasket status would seem to make sense here, it definitely feels like it would be more efficient.

I think some of these questions were why I had to look at some other things for a little bit and come back to this :-)

common/models/transactions.py
stuaxo: OK, sorted.

common/models/transactions.py
stuaxo: Understand django_fsm more and made this protected.

I should also update the querymanager to prevent un-finalising Transactions in `update`, can you think of any reasons this might be a bad idea ?

common/validators.py
```diff
@@ -71,3 +71,10 @@ class ApplicabilityCode(models.IntegerChoices):
 
 
 EnvelopeIdValidator = RegexValidator(r"^(?P<year>\d\d)(?P<counter>\d{4})$")
+
+
+class TransactionStatus(models.IntegerChoices):
+    # Newly created, order is relative to other transactions in the Workbasket
+    DRAFT = 1, "Draft"
+    # Finalised, order is global and immutable.
+    FINALISED = 2, "Finalised"
```
simonwo, line 80: I had some thoughts about this. Could we make it in this order:

1. SEED
2. FINALISED
3. DRAFT

And then add:

```python
class Meta:
    ordering = ("status", "order")
```

The default ordering would then put all seed transactions before finalised transactions, and draft transactions always come at the end? This would remove the need to renumber anything because simply doing `Transaction.objects.exclude(stauts=DRAFT).last()` would always get you the latest finalised transaction?

stuaxo: Yes, perfect - status partitions things, and order is just within a status.

stuaxo: If 'global' order is always "status", "order", then we can do  

unique_together = ("status", "order")

(^ not exactly this, as this is deprecated, but the new version)

I'll try adding this to this PR.

simonwo:
I'm worried that `UPDATE` already has another use in our system as one of `UPDATE`, `CREATE` and `DELETE` update types. Is there another word?

Maybe `REVISION`?

stuaxo:
@simonwo one (non essential) thing I wanted to get working on this, was using RowNumber in the update(...) statement that corrects the order this would mean we could write transactions without any gaps, in the end I couldn't get this working, probably/hopefully missing something obvious there.

I guess gaps shouldn't be a problem - (though it opens a question about deletions: can transactions in a draft workbasket be deleted (probably) - and in that case should we consider removing them before making putting them in the REVISION partition ?

common/models/records.py
```diff
@@ -113,7 +172,7 @@ def approved_up_to_transaction(self, transaction=None) -> TrackedModelQuerySet:
                             version_group__versions__transaction__workbasket=transaction.workbasket,
                             version_group__versions__transaction__order__lt=transaction.order,
                         )
-                        | self.approved_query_filter("version_group__versions__")
+                        | self.approved_query_filter_new("version_group__versions__")
```
stuaxo, line 175: TODO

common/models/records.py
```diff
@@ -280,14 +339,27 @@ def with_latest_links(self, *lookups, recurse_level=0) -> TrackedModelQuerySet:
     def get_queryset(self):
         return self.annotate_record_codes().order_by("record_code", "subrecord_code")
 
-    def approved_query_filter(self, prefix=""):
+    def approved_query_filter_old(self, prefix=""):
```
stuaxo, line 342: TODO

stuaxo:
This still needs a few bits and pieces, but the difficult pieces are there:  transaction partition is now set from near the beginning of the importer and goes all the way through the system.

There are some small cleanups needed and the code to default to SEED_FILE on first launch, as well as some tests for the new code, but this should be straightforward.

common/models/transactions.py
```diff
@@ -51,22 +106,59 @@ class Transaction(TimestampedMixin):
...
+    def save_draft(self):
+        """
+        Finalise a single draft transaction as an update.
+
+        Note, TransactionQueryset.finalise_transactions is more efficient for
+        finalising many transactions in a queryset.
+        """
+        self.partition = TransactionPartition.REVISION
+        self.save()
+
```
simonwo, line 161: I'm a bit nervous about this because it doesn't handle updating the `order` field. Is it possible to do that here too?

workbaskets/models.py
```diff
@@ -128,15 +130,23 @@ def withdraw(self):
...
+
+        approved_partition = self.status_as_transaction_partition(
+            **{**transaction_options, "status": WorkflowStatus.APPROVED.value}
+        )
+
         self.approver = user
         for obj in self.tracked_models.order_by("pk"):
             version_group = obj.version_group
             version_group.current_version = obj
             version_group.save()
```
simonwo, line 145: This code should potentially be part of `save_drafts` too?

workbaskets/models.py
```diff
@@ -194,6 +204,37 @@ def load_from_session(cls, session):
             raise KeyError("WorkBasket not in session")
         return WorkBasket.objects.get(pk=session["workbasket"]["id"])
 
+    def status_as_transaction_partition(
+        self,
+        approved_partition: int = TransactionPartition.REVISION,
+        first_partition_is_seed: bool = True,
```
simonwo, line 210: These two have seeminly infected a lot of higher-level code … is it worth some kind of abstraction over this? E.g.

```python
@dataclass
class PartitionScheme:
    approved_partition: TransactionPartition
    first_partition_is_seed: bool

    def __post_init__(self):
        if approved_partition not in TransactionPartition.approved_partitions():
            raise ValueError(
                f"approved_partition {approved_partition} is not one of {TransactionPartition.approved_partitions()}",
            )

    def get_partition(status: WorkflowStatus) -> TransactionPartition:
        if self.first_partition_is_seed and not WorkBasket.objects.exists():
            approved_partition = TransactionPartition.SEED_FILE
        else:
            approved_partiton = self.approved_partition

        if status in WorkflowStatus.approved_statuses():
            return approved_partition
        return TransactionPartition.DRAFT

SEED_FIRST = PartitionScheme(TransactionPartition.REVISION, True)
REVISION_ONLY = PartitionScheme(TransactionPartition.REVISION, False)
# … etc
```

And then you could just pass around `PartitionScheme` objects rather than multiple variables? WDYT?

common/migrations/0002_transaction_partition_1_of_3.py
```diff
@@ -0,0 +1,42 @@
...
+class Migration(migrations.Migration):
+    """
+    Transaction partition field 1 of 3.
+
+    1/3
+    Create Partition field, with partition defaulting to 1, SEED, as at the time of writing most of Transactions are from the seed file.
+
+    2/3
+    Data migration to set Transaction partitions to UPDATE and DRAFT.
+    UPDATE:  At the time of writing UPDATE transactions are in the first Workbasket, this may not be true in the future.
```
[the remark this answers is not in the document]
stuaxo: Good catch, also: REVISION, not UPDATE.

common/models/records.py
```diff
@@ -63,6 +63,65 @@ def latest_approved(self) -> TrackedModelQuerySet:
...
+                        )
+                        | self.approved_query_filter("version_group__versions__")
+                    ),
+                ),
+            )
+            .filter(latest=F("id"))
+            .exclude(update_type=UpdateType.DELETE)
+        )
+
+    def approved_up_to_transaction_new(self, transaction=None) -> TrackedModelQuerySet:
```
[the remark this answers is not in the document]
stuaxo: Ah, that's a temporary edit - so I could sanity check while I updated it to take the partition into account.

common/models/records.py
```diff
@@ -281,9 +340,11 @@ def get_queryset(self):
         return self.annotate_record_codes().order_by("record_code", "subrecord_code")
 
     def approved_query_filter(self, prefix=""):
+        from common.models.transactions import TransactionPartition
+
```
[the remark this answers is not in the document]
stuaxo: I think there's a circular import happened because of an import in the business rules - this still needs fixing in a different way though.

common/models/transactions.py
```diff
@@ -30,17 +54,48 @@ def get_queryset(self):
...
         ).order_by(
+            "transaction__partition",
             "transaction__order",
         )  # order_by record_code, subrecord_code already happened in get_queryset
         return tracked_models
 
+    def approved(self):
+        """Currently approved Transactions are SEED_FILE and REVISION this can
+        be more efficiently expressed as != DRAFT."""
+        return self.exclude(partition=TransactionPartition.DRAFT)
```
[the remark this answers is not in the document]
stuaxo: This is how I had it originally, at a guess .exclude() should be a little more efficient as it is only comparing with one value.

common/models/transactions.py
stuaxo: That's definitely a bug !

Thanks, will look into adding a failing test as well.

workbaskets/models.py
stuaxo: Sounds good, I need to stare at this as haven't seen this pattern before.

common/migrations/0004_transaction_partition_3_of_3.py
```diff
@@ -0,0 +1,23 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="transaction",
+            name="partition",
+            field=django_fsm.FSMIntegerField(
+                choices=[(1, "Seed"), (2, "Revision"), (3, "Draft")],
+                db_index=True,
+                default=3,
```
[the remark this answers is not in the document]
stuaxo: I'll have another look at this, I think there are a couple of things 
[1] There are many many transactions, so this tries to update the least amount at each stage.
[2] Separating data migrations and schema migrations tends to be a good thing.

The alternative is setting the default to DRAFT straight away, but we then incur and update for every single transaction (at the time of writing no transactions were DRAFT).

So this sets the default to SEED (most transactions at time of writing were SEED)
Then it does the data updates.
It should probably default things to DRAFT though, not REVISION.

common/migrations/0002_transaction_partition_1_of_3.py
```diff
@@ -0,0 +1,42 @@
...
+            name="transaction",
+            options={"ordering": ("partition", "order")},
+        ),
+        migrations.AddField(
+            model_name="transaction",
+            name="partition",
+            field=django_fsm.FSMIntegerField(
+                choices=[(1, "Seed"), (2, "Revision"), (3, "Draft")],
+                db_index=True,
+                default=1,
```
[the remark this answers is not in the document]
simonwo: It's been changed in the next migration to detect REVISION rather than SEED updates. Resolved!

simonwo (approved):

PR #339, opened 2021-07-16
Backport 3.9 typehints to 3.8 so tests can run under Pyston (faster p…

stuaxo:
Backport typehints to 3.8, which allows Pyston to run.

It turns out there wasn't much to do to port this, and we didn't even need strip-hints :)

This also adds info to the readme about running tests in parallel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Browse to http://localhost:8000/
 Skip this section if running under docker.
 
 The following dependencies are required to run this app:
- * Python 3.9.x
+ * Python 3.8.x
  * Node 14.16.x
  * Postgres 12.x
  * Redis 5.x
@@ -134,6 +134,40 @@ To run with coverage use the following:
     ./manage.py test -- --cov
 
 When running tests the settings module defaults to settings.test
+
+
+### Tips to run tests faster
+
+#### Run tests in parallel:
+
+When running locally it's possible to run tests in parallel using pytest-xdist.
+pytest-rerunfailures is also needed as a small number of tests clash when running in parallel and will fail.
+As re-running failing tests is a workaround, parallelization is undesirable under CI.
+
+Install dependencies:
+
+    pip install pytest-xdist pytest-rerunfailures
+
+Run the tests:
+
+    pytest -n=8 --reruns 8 --reruns-delay 4
+
+The example above is for a CPU with 8 threads, set "n" to a number less than or equal to the number of threads on the test machine.
+
+
+#### Run tests in Pyston instead of CPython:
+
+Pyston is a faster python implementation that aims for compatibility with the default CPython implementation.  
+Ad-hoc testing on one laptop showed tests completed in 6 minutes in CPython and 4 with Pyston. 
+
+Download and install a release from here: https://github.com/pyston/pyston/releases
+
+ - Create a python environment using venv[1]
+-  Install Tamato and it's dependencies to it.
+ - Run tests as usual.
+
+
+[1] The version of virtualenv on Ubuntu 20.04 is old and incompatible, it is advisable to use venv instead here: `$ pyston -mvenv`
 
 
 ## Environment Variables
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.9]
+        python-version: [3.8]
 
     services:
       postgres:
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.9]
+        python-version: [3.8]
 
     services:
       postgres:
--- a/common/business_rules.py
+++ b/common/business_rules.py
@@ -6,6 +6,7 @@ from functools import wraps
 from typing import Iterable
 from typing import Mapping
 from typing import Optional
+from typing import Tuple
 from typing import Type
 from typing import Union
 
@@ -121,7 +122,7 @@ class BusinessRule(metaclass=BusinessRuleBase):
 
 class BusinessRuleChecker:
     def __init__(self, models: Iterable[TrackedModel], transaction):
-        self.checks: set[tuple[type[BusinessRule], TrackedModel]] = set()
+        self.checks: set[Tuple[Type[BusinessRule], TrackedModel]] = set()
 
         self.transaction = transaction
 
--- a/common/tests/util.py
+++ b/common/tests/util.py
@@ -5,6 +5,7 @@ from datetime import timezone
 from functools import wraps
 from io import BytesIO
 from itertools import count
+from typing import Dict
 
 import pytest
 from dateutil.parser import parse as parse_date
@@ -342,7 +343,7 @@ def only_applicable_after(cutoff):
     return decorator
 
 
-def validity_period_post_data(start: date, end: date) -> dict[str, int]:
+def validity_period_post_data(start: date, end: date) -> Dict[str, int]:
     """
     Construct a POST data fragment for the validity period start and end dates
     of a ValidityPeriodForm from the given date objects, eg:
--- a/importer/tests/test_nursery.py
+++ b/importer/tests/test_nursery.py
@@ -71,7 +71,13 @@ def test_nursery_caches_object(object_nursery, handler_class):
 
 
 @pytest.mark.django_db
-def test_nursery_gets_object_from_cache(object_nursery):
+def test_nursery_gets_object_from_cache(settings, object_nursery):
+    settings.CACHES = {
+        "default": {
+            "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        },
+    }
+
     instance = factories.FootnoteFactory.create()
     object_nursery.cache_object(instance)
 
--- a/measures/querysets.py
+++ b/measures/querysets.py
@@ -143,6 +143,7 @@ class DutySentenceMixin(QuerySet):
                     ),
                 ),
                 delimiter=" ",
+                ordering="components__duty_expression__sid",
             ),
         )
 
--- a/measures/tests/test_querysets.py
+++ b/measures/tests/test_querysets.py
@@ -1,5 +1,8 @@
+from typing import Any
+from typing import Callable
 from typing import Dict
 from typing import List
+from typing import Sequence
 from typing import Tuple
 
 import factory
@@ -29,11 +32,23 @@ pytestmark = pytest.mark.django_db
         factories.MeasureConditionFactory._meta.model.__name__,
     ),
 )
+@pytest.mark.parametrize(
+    ("reorder"),
+    (
+        lambda x: x,
+        lambda x: reversed(x),
+    ),
+    ids=(
+        "normal",
+        "reversed",
+    ),
+)
 def test_duty_sentence_generation(
     component_factory: factory.django.DjangoModelFactory,
     model_factory: factory.django.DjangoModelFactory,
     field: str,
     reversible_duty_sentence_data: Tuple[str, List[Dict]],
+    reorder: Callable[[Sequence[Any]], Sequence[Any]],
 ):
     """
     Links components to the same model and tests the resulting single duty
@@ -47,7 +62,7 @@ def test_duty_sentence_generation(
     expected, component_data = reversible_duty_sentence_data
 
     model = model_factory()
-    for kwargs in component_data:
+    for kwargs in reorder(component_data):
         component_factory(
             **{field: model},
             **kwargs,
--- a/measures/views.py
+++ b/measures/views.py
@@ -1,3 +1,5 @@
+from typing import Type
+
 from rest_framework import viewsets
 
 from common.models import TrackedModel
@@ -28,7 +30,7 @@ class MeasureTypeViewSet(viewsets.ReadOnlyModelViewSet):
 
 
 class MeasureMixin:
-    model: type[TrackedModel] = Measure
+    model: Type[TrackedModel] = Measure
 
     def get_queryset(self):
         tx = WorkBasket.get_current_transaction(self.request)
--- a/runtime.txt
+++ b/runtime.txt
@@ -1,1 +1,1 @@
-python-3.9.x
+python-3.8.x
--- a/settings/test.py
+++ b/settings/test.py
@@ -25,7 +25,7 @@ AWS_S3_REGION_NAME = os.environ.get("TEST_AWS_S3_REGION_NAME", "eu-west-2")
 # Cache settings - put things in memory to minimise dependencies.
 CACHES = {
     "default": {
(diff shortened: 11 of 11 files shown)


README.md
```diff
@@ -136,6 +136,40 @@ To run with coverage use the following:
 When running tests the settings module defaults to settings.test
 
 
+### Tips to run tests faster
+
+#### Run tests in parallel tests:
+
+When running locally it's possible to run tests in parallel using pytest-xdist.
+pytest-rerunfailures is also needed as a small number of tests clash when running in parallel and will fail.
```
simonwo, line 144: How hard would it be to fix the clashes? Are there any pytest marks or things that prevent them from clashing?

README.md
```diff
@@ -136,6 +136,40 @@ To run with coverage use the following:
 When running tests the settings module defaults to settings.test
 
 
+### Tips to run tests faster
+
+#### Run tests in parallel tests:
```
simonwo, line 141: Delete the final "tests"?

README.md
```diff
@@ -48,7 +48,7 @@ Browse to http://localhost:8000/
 Skip this section if running under docker.
 
 The following dependencies are required to run this app:
- * Python 3.9.x
+ * Python 3.8.x
```
simonwo, line 51: You also need to modify `/runtime.txt` and the versions defined in `.github/workflows/*.yml`?

simonwo (changes requested):

README.md
stuaxo: Some of this may be about making the tests or factories threadsafe (or the code if it's an actual bug)...

Here's one I captured just now (there may be others on subsequent runs) -
```
=================================== FAILURES ===================================
______________________ test_condition_sid_is_next_highest ______________________
[gw3] linux -- Python 3.8.8 /home/stu/.virtualenvs/dit-pyston/bin/pyston

authorised_use_measure_data = {'authorised_use': True, 'duty_sentence': '0.00 %', 'geographical_area': <GeographicalArea: Region CBAA>, 'goods_nomenclature': <GoodsNomenclature: 1510000000>, ...}
measure_creation_pattern = <measures.patterns.MeasureCreationPattern object at 0x7efd204e4b20>

    def test_condition_sid_is_next_highest(
        authorised_use_measure_data,
        measure_creation_pattern: MeasureCreationPattern,
    ):
        condition = factories.MeasureConditionFactory()
>       models = list(measure_creation_pattern.create(**authorised_use_measure_data))

measures/tests/test_patterns.py:108: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
measures/patterns.py:392: in create
    yield from self.get_authorised_use_measure_conditions(new_measure)
measures/patterns.py:160: in get_authorised_use_measure_conditions
    action=self.apply_mentioned_duty,
/usr/lib/pyston3.8/functools.py:967: in __get__
    val = self.func(instance)
measures/patterns.py:139: in apply_mentioned_duty
    return MeasureAction.objects.get(code="27")
/home/stu/.virtualenvs/dit-pyston/lib/pyston3.8/site-packages/django/db/models/manager.py:85: in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <TrackedModelQuerySet [<MeasureAction: code=27>, <MeasureAction: code=08>, <MeasureAction: code=27>]>
args = (), kwargs = {'code': '27'}
clone = <TrackedModelQuerySet [<MeasureAction: code=27>, <MeasureAction: code=27>]>
limit = 21, num = 2

    def get(self, *args, **kwargs):
        """
        Perform the query and return a single object matching the given
        keyword arguments.
        """
        clone = self._chain() if self.query.combinator else self.filter(*args, **kwargs)
        if self.query.can_filter() and not self.query.distinct_fields:
            clone = clone.order_by()
        limit = None
        if not clone.query.select_for_update or connections[clone.db].features.supports_select_for_update_with_limit:
            limit = MAX_GET_RESULTS
            clone.query.set_limits(high=limit)
        num = len(clone)
        if num == 1:
            return clone._result_cache[0]
        if not num:
            raise self.model.DoesNotExist(
                "%s matching query does not exist." %
                self.model._meta.object_name
            )
>       raise self.model.MultipleObjectsReturned(
            'get() returned more than one %s -- it returned %s!' % (
                self.model._meta.object_name,
                num if not limit or num < limit else 'more than %s' % (limit - 1),
            )
        )
E       measures.models.MeasureAction.MultipleObjectsReturned: get() returned more than one MeasureAction -- it returned 2!

/home/stu/.virtualenvs/dit-pyston/lib/pyston3.8/site-packages/django/db/models/query.py:433: MultipleObjectsReturned

```

stuaxo: This implies something interesting - different processes are interfering with each other:  BUT it this shouldn't be happening, python-xdist actually creates a different database for each DATABASE-n.

So maybe we don't need to fix the test, I just need to check this is using the expected databases.

stuaxo: And another run:  

```
____________________________ test_type_choices[10] _____________________________
[gw13] linux -- Python 3.9.5 /home/stu/.virtualenvs/dit-tamato/bin/python

choice_inputs = [<TestModel2: custom_sid=3>, <TestModel2: custom_sid=4>, <TestModel2: custom_sid=5>, <TestModel2: custom_sid=6>, <TestModel2: custom_sid=7>, <TestModel2: custom_sid=8>, ...]

    def test_type_choices(choice_inputs):
        get_choices = type_choices(TestModel2.objects.all())
        choices = get_choices()
    
        assert len(choices) == len(choice_inputs)
        for input, output in zip(choice_inputs, choices):
>           assert output.value == input.custom_sid
E           assert 7 == 3
E            +  where 7 = <crispy_forms_gds.choices.Choice object at 0x7f21a8b3cd00>.value
E            +  and   3 = <TestModel2: custom_sid=3>.custom_sid

common/tests/test_filters.py:21: AssertionError
```

stuaxo: @simonwo having poked this and re-run over night:   getting it to fail is easy, and it does in the same places.

Unfortunately, I haven't had any luck triggering the debugger when it fails, once I add `-k`  `-s`  with `ipdb.set_trace()` or just `--ipdb` those tests pass.

stuaxo: Right, there aren't that many of these... the fun thing I listed above is that they only happen in conditions where you can't use the debugger.

This next one is fixed, by fixing the test and using `key=attrgetter("pk")` in the call to `sorted`
```
    def test_transaction_envelope_serializer_splits_output():
        """
        Verify MultiFileEnvelopeTransactionSerializer outputs the tracked_models
        passed to it and generates records for descriptions.
    
        This test is a bit artificial: testing 40mb splitting would be inefficient,
        max_envelope_size is set to 7k, small enough to trigger envelope splitting
        after just one transaction.
        """
        # Add transactions with different kinds of data to the workbasket.
        approved_workbasket = ApprovedWorkBasketFactory.create()
        with ApprovedTransactionFactory.create(workbasket=approved_workbasket) as tx1:
            factories.RegulationFactory.create()
    
        with ApprovedTransactionFactory.create(workbasket=approved_workbasket) as tx2:
            factories.RegulationFactory.create(regulation_group=None),
            factories.FootnoteTypeFactory.create()
    
        with ApprovedTransactionFactory.create(workbasket=approved_workbasket) as tx3:
            factories.FootnoteTypeFactory.create()
    
        transactions = Transaction.objects.filter(pk__in=[tx1.pk, tx2.pk, tx3.pk])
        expected_transactions = [
            Transaction.objects.filter(pk=tx.pk) for tx in [tx1, tx2, tx3]
        ]
        expected_record_codes = [
            [
                (tracked_model.record_code, tracked_model.subrecord_code)
                for tracked_model in tx.tracked_models.all()
            ]
            for tx in [tx1, tx2, tx3]
        ]
    
        # Create a static buffers to output to + a function to grab each one in turn to use as the constructor.
        expected_outputs = [io.BytesIO(), io.BytesIO(), io.BytesIO()]
    
        def create_output_constructor():
            output_iter = iter(expected_outputs)
            return lambda: next(output_iter)
    
        serializer = MultiFileEnvelopeTransactionSerializer(
            create_output_constructor(),
            envelope_id=int(Envelope.next_envelope_id()),
            max_envelope_size=7000,
        )
    
        for i, rendered_envelope in enumerate(
            serializer.split_render_transactions(transactions),
        ):
            # Base assumption is that this yields RenderedTransactions
            assert isinstance(rendered_envelope, RenderedTransactions)
    
            assert rendered_envelope.output == expected_outputs[i]
            assert rendered_envelope.is_oversize is False
            assert 0 < rendered_envelope.output.tell() < 7000
    
            assert len(
                rendered_envelope.transactions,
            ), "Serializer should skip empty transactions, they cause XSD validation to fail."
    
>           assert sorted(rendered_envelope.transactions) == sorted(
                expected_transactions[i],
            )
E           TypeError: '<' not supported between instances of 'Transaction' and 'Transaction'

common/tests/test_serializers.py:154: TypeError

```

stuaxo: The bug around the test_patterns (stacktrace two up) is pretty nasty... changing all the @cached_property to @property seems to fix it so far.

There is something very weird with timing or buffering, so outputting stdout or asking for the debugger automatically stops it working.

I tried adding using a mixin to delete the cache from the cached_properties when the measure_pattern was created, but this didn't seem to help.

Switching the cached_property implementation between django and functools also didn't help.

Weirdly, the bug seems to surface more when I run my laptop in the morning, when it's cool .. either a random artifact, or time of day affecting the random number generator somehow (or even the laptop getting hotter and running slower).

stuaxo: This is one of the last ones, and kind of interesting:   I think one of the factories is outputting a date something doesn't like:

```python
[gw11] linux -- Python 3.8.8 /home/stu/.virtualenvs/dit-pyston/bin/pyston

valid_user = <User: <factory.faker.Faker object at 0x7f43e8942a00>3232>, update_type = <UpdateType.DELETE: 2>

    @pytest.mark.parametrize("update_type", [UpdateType.UPDATE, UpdateType.DELETE])
    def test_suspension_importer_update(valid_user, update_type):
        effective_end_date = date(2021, 2, 1)
>       suspension = create_and_test_m2m_regulation(
            RoleType.FULL_TEMPORARY_STOP,
            "taric/suspension.xml",
            factories.SuspensionFactory,
            valid_user,
            update_type=update_type,
            factory_kwargs={"effective_end_date": effective_end_date},
            effective_end_date=effective_end_date,
            action_record_code=models.Suspension.action_record_code,
            action_subrecord_code=models.Suspension.action_subrecord_code,
        )

regulations/tests/test_importer.py:175: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
regulations/tests/test_importer.py:71: in create_and_test_m2m_regulation
    process_taric_xml_stream(
importer/taric.py:224: in process_taric_xml_stream
    if handler.end(elem):
importer/taric.py:197: in end
    super().end(element)
importer/parsers.py:186: in end
    self.child.end(element)
importer/taric.py:153: in end
    self.save(
importer/taric.py:130: in save
    self.message.save(message_data, transaction.id)
importer/taric.py:84: in save
    self.record.save(record_data, transaction_id)
importer/taric.py:67: in save
    getattr(parser, method_name)(record_data, transaction_id)
importer/parsers.py:433: in delete
    self.nursery.submit(dispatch_object)
importer/nursery.py:79: in submit
    result = handler.build()
importer/handlers.py:442: in build
    self.dispatch()
importer/handlers.py:454: in dispatch
    data = self.clean(self.data)
regulations/import_handlers.py:82: in clean
    return super().clean(data)
importer/handlers.py:398: in clean
    serializer.is_valid(raise_exception=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = RegulationImporterSerializer(data={'role_type': 8, 'regulation_id': 'R2101190', 'target_regulation__role_type': 1, 'ta...aric_template = SerializerMethodField()
    start_date = SerializerMethodField()
    end_date = SerializerMethodField()
raise_exception = True

    def is_valid(self, raise_exception=False):
        assert hasattr(self, 'initial_data'), (
            'Cannot call `.is_valid()` as no `data=` keyword argument was '
            'passed when instantiating the serializer instance.'
        )
    
        if not hasattr(self, '_validated_data'):
            try:
                self._validated_data = self.run_validation(self.initial_data)
            except ValidationError as exc:
                self._validated_data = {}
                self._errors = exc.detail
            else:
                self._errors = {}
    
        if self._errors and raise_exception:
>           raise ValidationError(self.errors)
E           rest_framework.exceptions.ValidationError: {'valid_between': [ErrorDetail(string='Date has wrong format. Use one of these formats instead: YYYY-MM-DD, YYYY-MM-DD.', code='invalid')]}

/home/stu/.virtualenvs/dit-pyston/lib/pyston3.8/site-packages/rest_framework/serializers.py:228: ValidationError

```

stuaxo: OK, getting everything multiprocess friendly IS a bit complex, I will split this into a new ticket for the backlog - it's probably something everyone can look at as you can run these while going to lunch then check the failures. 

We need a testing epic I think (at least until we're happy with it: 20+ mins is def too long)

simonwo (approved):

simonwo:
You will need to force merge this and go and update the settings in Github to not be expecting a 3.9 build anymore.

README.md
stuaxo: I take back the @cached_property fix, that didn't work.

I'm separating out the task to make everything multiprocess safe (and thus ready for CI).

There's definitely something weird that goes on, in that enabling options to output to stdout seems to make this particular bug not occur.

I wonder if it's the act of serialization (i.e. converting some query to a string) that actually makes the test pass.

The only way to trigger it is to leave it running, like: `while pytest -n=9; do :; done` and see it fail.
The other theory I have is that something is fetched from a model in a fixture in the master process, then in the test we try and use it, but it's no longer valid.

I guess seeing if reload_from_db fixes it would be one thing to try.

stuaxo: Further guess: enabling stdout means the value is logged, that materialises it making it work.

PR #341, opened 2021-07-27
Tp 867 backport typehints to 3.8

stuaxo:
Same branch as before, just renamed:

Backport typehints to 3.8 so we can use pyston instead of CPython, which completes our tests 30% faster.

Didn't add Pyston to CI, since we still need to make sure things run in CPython as the main runtime, though could add it as an additional interpreter to run at the same time.

This branch, also sets the cache backend to "dummy" during testing, which removes most of the state shared between tests.

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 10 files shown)


paulpepper-trade (approved):
LGTM.

kintisheff (approved):

PR #342, opened 2021-07-27
TP-875 Running tests in random order exposes more failing tests.

stuaxo:
**DRAFT PR**:  Until we are sure all the failing tests pass.

This PR is to show which tests fail with `--random-order`.

Fixes will appear in other PRs until `--random-order` can consistently without failing (e.g. for some number of hours on a dev machine).

This will help shake out bugs related to shared state as a stepping stone towards running tests with `pytest-xdist` which has a much faster runtime.

To run with `--random-order` locally on your current branch:

Install pytest random order:

`$ pip install pytest-random-order` 

Examples of running tests in a loop until a failure occurs:

`$ while pytest -s -v --random-order; do :; done`

`$ while pytest -s -v --random-order measures; do :; done`

`$ while pytest -s -v --random-order measures/tests/test_importer.py; do :; done`

Note:  
Failures are mostly due to shared state, running just one test, or even just one module will not always trigger a failure.
Running in a random order runs surfaces new combinations and failures that running in a static order does not.

Running tests in [Pyston](https://github.com/pyston/pyston) completes 30% faster, but it's also useful to verify in CPython.

Currently, Pyston implements Python 3.8, so we have temporarily backported our type hints, once 3.9 is supported we can move to 3.9 again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,12 +72,12 @@ run: collectstatic migrate
 test-fast:
 	@echo
 	@echo "> Running tests..."
-	${PYTHON} manage.py test --failfast
+	${PYTHON} manage.py test --failfast -- --random-order
 
 test:
 	@echo
 	@echo "> Running tests..."
-	@coverage run --source='.' manage.py test -- --alluredir=allure-results --nomigrations
+	@coverage run --source='.' manage.py test -- --random-order --alluredir=allure-results --nomigrations
 	@coverage xml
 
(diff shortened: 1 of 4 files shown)


stuaxo:
Using this technique we have proved all the intermittent failures can happen on a single process and so were able to enable pytest-xdist to run tests in multiple processes, which in turn supercedes this PR.

Individual intermittent failures will have their own PRs.

PR #344, opened 2021-07-30
Enable python-xdist with --dist=loadfile

stuaxo:
Use python-xdist[1] to split the tests across many processes.

For a splitting strategy, we choose --dist=loadfile, which loads a file into one process and won't split the tests further, this is pretty conservative and should keep any state that we accidentally share in that file just as we do now.

As a downside `--dist-loadfile` will have slightly worse performance, if one of our files has lots of tests that are especially slow.
I'm anticipating we can move off this later.

[1] python-xdist can split tests over separate machines or processes, using different strategies - in this case we use it to spin up different processes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,13 +72,12 @@ run: collectstatic migrate
 test-fast:
 	@echo
 	@echo "> Running tests..."
-	${PYTHON} manage.py test --failfast
+	${PYTHON} python -m pytest -x -n=auto --dist=loadfile --cov --cov-report html:htmlcov --cov-report=term --cov-report=xml
 
 test:
 	@echo
 	@echo "> Running tests..."
-	@coverage run --source='.' manage.py test -- --alluredir=allure-results --nomigrations
-	@coverage xml
+	@python -m pytest -n=auto --dist=loadfile --alluredir=allure-results --nomigrations --cov --cov-report html:htmlcov --cov-report=term --cov-report=xml
 
 ## docker-image: Build docker image
 docker-image:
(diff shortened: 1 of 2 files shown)


stuaxo:
Findings:  

1. We've established that test failures were not due to adding multiprocessing support, so I think we can merge this.

2. This doesn't speed up CI (this is probably due to how our usage is allocated)  pytest-split, may be a way to get some speed up in CI:  https://blog.jerrycodes.com/pytest-split-and-github-actions/

simonwo:
> [Codecov](https://codecov.io/gh/uktrade/tamato/pull/344) Report

Coverage report is in bits... how do we fix that?

simonwo:
Could you add to the commit message to explain this a bit? I don't know what xdist is or what the loadfile option does.

paulpepper-trade (approved):

simonwo:
In the past pytest-cov didn’t count the lines outside of tests that are definitely loaded on import, hence giving us a lot of false negatives, because Django initialised it after the imports. That’s why we started using coverage directly instead. Is that fixed now? Else we’ll end up with incorrectly reported coverage again.

simonwo (approved):
> @stuaxo:
> Ok think coverage is sorted, it isn't updating the branch since I rebased, but if you click through and make sure its looking at the commit hash it's OK

stuaxo:
@simonwo only just re-read this and saw this comment.  Will probably need to look into this.   
It seems like coverage is a bit of a tricky beast from my tour of various coverage based tickets.

PR #345, opened 2021-08-02
View-level tests for models

simonwo:
This PR suggests an approach for testing our UI-level functionality by testing views. It introduces a simple primitive that can be used to test the update form for a model. 

These test that:
- when the form is loaded, it shows the correct data
- when the form is submitted, the form knows whether the data is invalid or not
- when the form is submitted, the changes are persisted in the correct way
- any business rules are **not** triggered as part of the form submission

The use of lambdas is to decouple the tests from the factory data… in particular I didn't want the parameters to have to be set out explicitly for each model and duplicate the factory. So instead the current value is passed to each lambda and the result can be mutated (or discarded) as appropriate.

### DIFF
--- a/common/tests/util.py
+++ b/common/tests/util.py
@@ -5,11 +5,13 @@ from datetime import timezone
 from functools import wraps
 from io import BytesIO
 from itertools import count
+from typing import Any
 from typing import Dict
 
 import pytest
 from dateutil.parser import parse as parse_date
 from dateutil.relativedelta import relativedelta
+from django import forms
 from django.core.exceptions import ValidationError
 from django.template.loader import render_to_string
 from django.urls import reverse
@@ -366,3 +368,27 @@ def validity_period_post_data(start: date, end: date) -> Dict[str, int]:
         for name, date in (("start_date", start), ("end_date", end))
         for i, part in enumerate([date.day, date.month, date.year])
     }
+
+
+def get_form_data(form: forms.ModelForm) -> Dict[str, Any]:
+    """Returns a dictionary of the fields that the form will put onto a page and
+    their current values, taking account of any fields that have sub-fields and
+    hence result in multiple HTML <input> objects."""
+
+    data = {**form.initial}
+    for field in form.rendered_fields:
+        value = data[field] if field in data else form.fields[field].initial
+        if hasattr(form.fields[field].widget, "decompress"):
+            # If the widget can be decompressed, then it is not just a simple
+            # value and has some internal structure. So we need to generate one
+            # form item per decompressed value and append the name with _0, _1,
+            # etc. This mirrors the MultiValueWidget in django/forms/widgets.py.
+            if field in data:
+                del data[field]
+            value = form.fields[field].widget.decompress(value)
+            data.update(
+                **{f"{field}_{i}": v for i, v in enumerate(value) if v is not None}
+            )
+        elif value is not None:
+            data.setdefault(field, value)
+    return data
--- a/conftest.py
+++ b/conftest.py
@@ -33,6 +33,7 @@ from common.serializers import TrackedModelSerializer
 from common.tests import factories
 from common.tests.util import Dates
 from common.tests.util import generate_test_import_xml
+from common.tests.util import get_form_data
 from common.tests.util import make_duplicate_record
 from common.tests.util import make_non_duplicate_record
 from common.tests.util import raises_if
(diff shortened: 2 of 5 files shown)


conftest.py
```diff
@@ -335,6 +336,71 @@ def check(dependency_name, dependency_factory, dependent_factory):
...
+
+    The ``new_data`` dictionary should contain callable objects that when passed
+    the existing value will return a new value to be sent with the form.
+
+    Will raise :class:`~django.core.exceptions.ValidationError` if form thinks
+    that the passed data contains errors.
+    """
+
+    def use(object: TrackedModel, new_data: Dict[str, Callable[[Any], Any]]):
+        model = object.__class__
```
stuaxo, line 353: `type(object)` to fit in with our style elsewhere.

conftest.py
```diff
@@ -335,6 +336,71 @@ def check(dependency_name, dependency_factory, dependent_factory):
...
+    the object to have the passed new data.
+
+    The ``new_data`` dictionary should contain callable objects that when passed
+    the existing value will return a new value to be sent with the form.
+
+    Will raise :class:`~django.core.exceptions.ValidationError` if form thinks
+    that the passed data contains errors.
+    """
+
+    def use(object: TrackedModel, new_data: Dict[str, Callable[[Any], Any]]):
```
stuaxo, line 352: `type(object)` to fit in with our style elsewhere.

conftest.py
```diff
@@ -335,6 +336,71 @@ def check(dependency_name, dependency_factory, dependent_factory):
...
+        model = object.__class__
+        versions = model.objects.filter(**object.get_identifying_fields()).count()
```
stuaxo, line 354: `count()` is used to check if versions are the same, it's always good to check actual content, at least the pks, but maybe comparing the identifying fields, and the pk could work ?

```python
fields = object.get_identifying_fields()
versions = [*model.objects.filter(**fields).values(*(*fields.keys(), 'pk'))]
```

`current_versions` may communicate intention better here ?

Queryset equality with versioning and identifying fields seems useful enough that it could be worth pulling out into a utility function ?

(Basically I'm missing assertQuerySetEqual, but it doesn't fit in the pytest world).

common/tests/util.py
```diff
@@ -366,3 +368,23 @@ def validity_period_post_data(start: date, end: date) -> Dict[str, int]:
...
+    their current values, taking account of any fields that have sub-fields and
+    hence result in multiple HTML <input> objects."""
+
+    data = form.initial
+    for field in form.rendered_fields:
+        value = data[field] if field in data else form.fields[field].initial
+        if hasattr(form.fields[field].widget, "decompress"):
+            if field in data:
+                del data[field]
+            value = form.fields[field].widget.decompress(value)
```
stuaxo, line 384: The data structure `{field}_{number}` seems like domain knowledge (OK ,really a fancy way of saying I didn't know this) - it would be good to mention it briefly in the docstring and where it comes from so the developer can read further (whether it's django docs / our code base or whatever).

conftest.py
simonwo: I've made it look at the PKs… looking at the identifying fields wouldn't add anything because we already filter on them. If they'd changed then the model wouldn't be returned.

common/tests/util.py
```diff
@@ -366,3 +368,27 @@ def validity_period_post_data(start: date, end: date) -> Dict[str, int]:
...
+            # If the widget can be decompressed, then it is not just a simple
+            # value and has some internal structure. So we need to generate one
+            # form item per decompressed value and append the name with _0, _1,
+            # etc. This mirrors the MultiValueWidget in django/forms/widgets.py.
+            if field in data:
+                del data[field]
+            value = form.fields[field].widget.decompress(value)
+            data.update(
+                **{f"{field}_{i}": v for i, v in enumerate(value) if v is not None}
+            )
```
stuaxo, line 391: ^ This is fine as long as None is not a value that can occur here (sometimes you want to distinguish between a value being present and `== None` vs the value being absent).

stuaxo, line 391: ^ This is fine as long as there is no need to distinguish between `None` and the value being absent.

common/tests/util.py
```diff
@@ -366,3 +368,27 @@ def validity_period_post_data(start: date, end: date) -> Dict[str, int]:
...
+        elif value is not None and field not in data:
+            data[field] = value
```
stuaxo, line 393: `data.setdefault(field, value)`   can work instead of `not in data`

conftest.py
```diff
@@ -335,6 +336,81 @@ def check(dependency_name, dependency_factory, dependent_factory):
     return check
 
 
+@pytest.fixture
+def use_update_form(valid_user_api_client: APIClient):
+    """
+    Provides a function that will use the appropriate edit form and view to edit
+    the object to have the passed new data.
```
stuaxo, line 343: ^ trying to think of a way to make this sentence clearer, but it's pretty difficult, I guess because this fixture does a lot!

conftest.py
```diff
@@ -335,6 +336,81 @@ def check(dependency_name, dependency_factory, dependent_factory):
...
+        model = type(object)
+        versions = set(
+            model.objects.filter(**object.get_identifying_fields()).values_list(
+                "pk",
+                flat=True,
+            ),
+        )
+
+        # Visit the edit page and ensure it is a success
+        edit_url = object.get_url("edit")
```
stuaxo, line 362: If the URL doesn't exist here, will it be None ?

If so can there be an assert  .... `assert edit_url, "No edit page found for ...` ?

common/tests/util.py
simonwo: I think that's fine, yes.

conftest.py
simonwo: I reworded.

stuaxo (approved):

PR #346, opened 2021-08-02
Implementation for editing an existing regulation.

paulpepper-trade:


regulations/views.py
```diff
@@ -72,3 +78,46 @@ class RegulationConfirmCreate(TrackedModelDetailView):
     def get_queryset(self):
         tx = WorkBasket.get_current_transaction(self.request)
         return Regulation.objects.approved_up_to_transaction(tx)
+
+
+# TODO:
+# * Should Regulation views apply 'permission_classes' attr.
```
simonwo, line 84: Yes :)

regulations/views.py
```diff
@@ -72,3 +78,46 @@ class RegulationConfirmCreate(TrackedModelDetailView):
...
+# * RegulationUpdate inherits BusinessRulesMixin as with other edit views. Which
+#   rules should be applied, if any?
```
simonwo, line 86: Oh, good question. Based on reading https://uktrade.github.io/tamato/docs/business_rules/regulations.html, I think:

- ROIMB1
- ROIMB4
- ROIMB8
- ROIMB44
- ROIMB47

regulations/views.py
```diff
@@ -72,3 +78,46 @@ class RegulationConfirmCreate(TrackedModelDetailView):
...
+# * Should business rules be applied to RegulationCreate too?
```
simonwo, line 87: Yes, you can probably skip ROIMB8 but all the others are relevant. No harm running ROIMB8 if it's easier to make them the same.

regulations/views.py
paulpepper-trade: Ah, actually, my bad. The `permission_classes` is used on DRF's ViewSets not in Django views. Our create and edit views indirectly inherit from `PermissionrequiredMixin` and their base classes define `permission_required = "common.add_trackedmodel"`.

simonwo:
Do we now have a good way to implement some tests for this, a la [here](https://github.com/uktrade/tamato/blob/master/footnotes/tests/test_views.py#L20)?

paulpepper-trade:
> Do we now have a good way to implement some tests for this, a la [here](https://github.com/uktrade/tamato/blob/master/footnotes/tests/test_views.py#L20)?

Not yet...

common/tests/factories.py
```diff
@@ -217,6 +217,17 @@ class BaseRegulationFactory(RegulationFactory):
...
 
+class UIRegulationFactory(BaseRegulationFactory):
+    """
+    Regulation factory used by our UI form tests.
+
+    These are distinct from our other Regulation factories because the required
+    status differs between Regulation model fields and form fields.
+    """
+
+    published_at = Dates().datetime_now
```
simonwo, line 228: We should prefer use of `date_ranges("now")` if possible.

simonwo (approved):

common/tests/factories.py
paulpepper-trade: Done.

paulpepper-trade:
> > Do we now have a good way to implement some tests for this, a la [here](https://github.com/uktrade/tamato/blob/master/footnotes/tests/test_views.py#L20)?
> 
> Not yet...

Done.

paulpepper-trade:
> > Do we now have a good way to implement some tests for this, a la [here](https://github.com/uktrade/tamato/blob/master/footnotes/tests/test_views.py#L20)?
> 
> Not yet...

Done.

PR #348, opened 2021-08-03
Make sure duty expressions are ordered correctly

simonwo:
If we don't set this explicitly they will just output in the order that the components were created which is not always correct.

E.g. sometimes you will get

    + 12.00 GBP / kg 3.00%

when you should get

    3.00% + 12.00 GBP / kg


stuaxo (approved):

stuaxo:
Approved. nice to have one-less intermittent test fail.

PR #349, opened 2021-08-04
TP 589 - Configure SQLite upload

rachaelcodes:
This configures the storage settings for SQLite uploads, complete with custom AWS S3 bucket settings. 

I've also updated one of the SQLite export task tests in line with how the `HMRCStorage` is exercised in the other export task tests, which allows checking whether the different bucket name is set as expected.


rachaelcodes:
This is a work in progress - some things I'm still thinking about:
* whether I can make the storage fixtures in `conftest.py` DRYer
* how to test the other custom Storage settings (which I think will involve me digging into the mock S3 documentation)

simonwo (commented):
LGTM.

I don't think the DRYness is overly concerning. But I guess you could extract that logic to a separate method in conftest.py that is a `@contextmanager` and then 
```python
with make_storage(SQLiteStorage, settings.SQLITE_STORAGE_BUCKET_NAME) as storage:
    yield storage
```
or something?

stuaxo (approved):

conftest.py
```diff
@@ -525,11 +526,10 @@ def check(bucket_name, key):
     return check
 
 
-@pytest.fixture
-def hmrc_storage():
-    """Patch HMRCStorage with moto so that nothing is really uploaded to s3."""
+@contextlib.contextmanager
+def make_storage(storage_class, bucket_name):
```
stuaxo, line 530: Maybe we can find a name that lets the user know this is patching/faking/mocking the storage class ?

conftest.py
```diff
@@ -525,11 +526,10 @@ def check(bucket_name, key):
...
     with mock_s3():
-        storage = HMRCStorage()
+        storage = storage_class()
```
stuaxo, line 532: May be wrong but: think we can pass `bucket_name=bucket_name` in here....

In fact looks, like we can pass a dict of settings to override:

```python
def make_storage(storage_class, **override_settings)
```

And
```
storage = storage_class(**override_settings)
```

conftest.py
```diff
@@ -549,20 +549,35 @@ def get_connection():
             def get_bucket():
                 connection = get_connection()
                 connection.create_bucket(
-                    Bucket=settings.HMRC_STORAGE_BUCKET_NAME,
+                    Bucket=bucket_name,
```
stuaxo, line 552: Looks possible to grab `bucket_name` from storage.bucket_name:

`Bucket=storage.bucket_name`

stuaxo (approved):

conftest.py
stuaxo: Apologies, just realised make_storage was @simonwo suggestion, always annoying to get conflicting suggestions - totally not a blocker (maybe in this case really is maybe, so feel free not to change that)

conftest.py
stuaxo: (I realise some of these are comments on things I wrote originally, getting everything to play nicely with moto ended up a little  convoluted in the end)

settings/common.py
```diff
@@ -287,7 +287,16 @@
 HMRC_STORAGE_DIRECTORY = os.environ.get("HMRC_STORAGE_DIRECTORY", "tohmrc/staging/")
 
 # SQLite AWS settings
-SQLITE_STORAGE_BUCKET_NAME = os.environ.get("SQLITE_STORAGE_BUCKET_NAME")
+SQLITE_STORAGE_BUCKET_NAME = os.environ.get("SQLITE_STORAGE_BUCKET_NAME", "sqlite")
+SQLITE_S3_ACCESS_KEY_ID = os.environ.get(
+    "SQLITE_S3_ACCESS_KEY_ID",
+    "test_sqlite_key_id",
```
rachaelcodes, line 293: Any thoughts on whether setting these alternatives could cause any untoward consequences?

stuaxo: It shouldn't do, since these are your own keys.

I did see the test fail in the logs earlier about it not finding the bucket - but, didn't really have anything useful to add - it is a bit confusing.

I guess everything runs in the tests locally.

You could try minio locally (I use `docker-compose s3 up`) and run the s3 upload to confirm it goes to the right place (browse to the endpoint in a browser).

rachaelcodes: It was previously failing in GitHub CI until I added these fallback values - @stuaxo have you seen it failing more recently somewhere?

stuaxo: No, but I haven't been running any of the s3 stuff in a long time, not sure if we had it running on CI before this ?

simonwo (approved):
Still LGTM!

PR #351, opened 2021-08-05
Make test-fast not run coverage.

stuaxo:
Update README with info on running tests fast.


simonwo:
> /me: Looks at coverage report: "WAT?"
> 
> Will come back to this...

Same issue as #319 … thought we caught them all but clearly not.

simonwo (approved):

stuaxo:
@simonwo OK, we can find and fix these fuzzy tests in the way using the method I outline here to find intermittent tests "run them in a loop".

PR #354, opened 2021-08-10
Refactor regulations

simonwo:
This is some generic work extracted from the "XML refactor" branch.

Many of our regulation importers generate a `Regulation` object and then follow up with a through model that describes the relationship between the new regulation object and some older regulation.

The old importers are focused around the `Regulation` object, whereas it makes more sense to focus them around the through model because this means we can use the common testing infrastructure and will also makes refactoring the XML easier.

This also adds new tests for some XML that hasn't been tested before and fixes the bugs that this has found.


importer/parsers.py
```diff
@@ -193,6 +194,11 @@ def end(self, element: etree.Element):
                 field_name = self._field_lookup[self.child]
                 if self.child.many:
```
stuaxo, line 195: Not a blocker, but might be worth raising an error if both `child.many` and `child.extra_fields` are set.

simonwo: Good shout.

PR #356, opened 2021-08-12
TP-859 Make SQLite runner inherit main process env

rachaelcodes:
There is a problem with the SQLite celery task where shared Python libraries aren't being found. Having confirmed that these libraries do exist where they're expected to in the app environment, I think that what is happening is that the SQLite runner is not accessing them. This PR makes the main process environment available to the SQLite runner, which should help it find libraries located in directories listed in $LD_LIBRARY_PATH.

### DIFF
--- a/exporter/sqlite/runner.py
+++ b/exporter/sqlite/runner.py
@@ -1,5 +1,6 @@
 import decimal
 import logging
+import os
 import sqlite3
 import sys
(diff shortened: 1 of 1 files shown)


simonwo (approved):
LGTM.

stuaxo (approved):

PR #357, opened 2021-08-16
Prevent SQLite from using the default db

rachaelcodes:
Unfortunately, #356 didn't fully resolve the SQLite problem. By copying over the env variables, the SQLite runner has access to the VCAP_SERVICES env variable, which causes a DB address overwrite (specifically the postgres details it contains): https://github.com/uktrade/tamato/blob/master/settings/common.py#L220 . 

The alternative might be to set `LD_LIBRARY_PATH` explicitly, instead of copying over all the env variables, but I'm concerned this would mean missing out on other hidden ones, and we continue playing 'whack-a-mole' trying to catch the all.

### DIFF
--- a/exporter/sqlite/runner.py
+++ b/exporter/sqlite/runner.py
@@ -36,6 +36,8 @@ class Runner:
         """
         sqlite_env = os.environ.copy()
         sqlite_env["DATABASE_URL"] = f"sqlite:///{self.db}"
+        # Required to make sure the postgres default isn't set as the DB_URL
+        sqlite_env["VCAP_SERVICES"] = "{}"
 
         run(
             [sys.executable, "manage.py", *args],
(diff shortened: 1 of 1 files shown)


exporter/sqlite/runner.py
```diff
@@ -36,6 +36,8 @@ def manage(self, *args: str):
         """
         sqlite_env = os.environ.copy()
         sqlite_env["DATABASE_URL"] = f"sqlite:///{self.db}"
+        # Required to make sure the postgres default isn't set as the DB_URL
+        sqlite_env["VCAP_SERVICES"] = "{}"
```
stuaxo, line 40: Is `del sqlite_env["VCAP_SERVICES"]` an option here ?

simonwo (approved):
Are there any other services that the resulting process might need? E.g. redis? I think not in this case? Otherwise we might have dig into that dictionary :(

rachaelcodes:
Based on @simonwo 's comment, I'm now only removing the 'postgres' VCAP details.

exporter/sqlite/runner.py
rachaelcodes: I think I'll use `dict.pop()` rather than `del` to avoid KeyErrors. Though thanks to Simon's suggestion, I'm now not going to take out the entire `VCAP_SERVICES` dict.

PR #358, opened 2021-08-16
Generate envelope sequences

stuaxo:
This is a *very* draft PR about generating summaries of envelopes (currently so draft that, the mentioned feature is not in the PR currently, and needs bringing across from a notebook).

To be able create summaries that output sequences:

Here's how it could look for a fictional model "Goods" and "Certs":
```
Goods:
Updated T51-T58  T64-T65

Certs:
Updated C646-C650
Created C651
```

To find sequences, a countable field needs to be chosen from the models `identifying_fields`, currently this must be a `SignedIntSID` or `NumericIntSID`.

### DIFF
--- a/common/models/records.py
+++ b/common/models/records.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import re
 from datetime import date
 from typing import Any
+from typing import Dict
 from typing import Iterable
 from typing import Optional
 from typing import Set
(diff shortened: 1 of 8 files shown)


common/models/records.py
```diff
@@ -383,7 +506,7 @@ class TrackedModel(PolymorphicModel):
     these fields.
     """
 
-    objects: TrackedModelQuerySet = PolymorphicManager.from_queryset(
+    objects: TrackedModelQuerySet = TrackedModelManager.from_queryset(
         TrackedModelQuerySet,
     )()
```
stuaxo, line 511: Looks like this should still be PolymorphicManager ..

common/models/records.py
```diff
@@ -610,6 +733,48 @@ def get_identifying_fields(
...
+
+        Deriving classes can override this to use other fields.
+        """
+        for qualified_name, field in cls.get_class_identifying_fields().items():
+            from common.fields import SignedIntSID
+
+            # print("q, f", qualified_name, field, type(field), issubclass(type(field), SignedIntSID), isinstance(field, SignedIntSID))
+            if qualified_name == "sid" and (
+                issubclass(type(field), SignedIntSID)
+                or issubclass(type(field), NumericSID)
```
stuaxo, line 762: `issubclass(type(field), (SignedIntSID, NumericSID)`

stuaxo:
Closing this for now, I think we will need something like this for summaries - it looks like we still send this in email form, but the PR will probably need re-doing, the query in here is too slow.

PR #359, opened 2021-08-16
TP-882 Selective Taric Xml Parser

kintisheff:
NB: This ticket is part of the [CC Importer epic](https://uktrade.atlassian.net/browse/TP-911). The design approach was agreed as part of [TP-858](https://uktrade.atlassian.net/browse/TP-858)

Changes:
1. Prerequisites and Foundational:
- turn `importer.namespace.Tag` into a dataclass and consolidate `RegexTag` into it (all diffs outside `importer` in this PR are simply adapting instantiation calls to Tag to the new dataclass form)
- load all `taric3.xsd` and `envelope.xsd` elements in a special dataclass - to provide a robust link with schema and address some prior DRY violations related to hardcoding, etc.
- extend `build_dependency_tree` to optionally work at the `subrecord_code` level (this is foundational for the next tickets in the epic and may need more tweaks subsequently)
2. Ticket Functionality related:
- add support for transaction record filtering to `importer.chunker.chunk_taric` - this essentially includes plugging into this method a call to a new small filter method `filter_transaction_records`
- add a button on the `/importer/create` form to trigger selective xml parsing for the commodity records group in particular (note, this does not imply that this would be our ultimate design decision on this, we have [TP-805](https://uktrade.atlassian.net/browse/TP-805) in this epic for that)
3. Add test coverage, type annotations and docs to some existing adjacent code also

### DIFF
--- a/additional_codes/import_parsers.py
+++ b/additional_codes/import_parsers.py
@@ -72,7 +72,9 @@ class AdditionalCodeDescriptionPeriodParser(
 
     sid = TextElement(Tag(name="additional.code.description.period.sid"))
     described_additionalcode__sid = TextElement(Tag(name="additional.code.sid"))
-    described_additionalcode__type__sid = TextElement(Tag(name="additional.code.type.id"))
+    described_additionalcode__type__sid = TextElement(
+        Tag(name="additional.code.type.id"),
+    )
     described_additionalcode__code = TextElement(Tag(name="additional.code"))
     validity_start = ValidityStartMixin.validity_start
 
@@ -106,7 +108,9 @@ class AdditionalCodeDescriptionParser(Writable, ElementParser):
     sid = TextElement(Tag(name="additional.code.description.period.sid"))
     language_id = ConstantElement(Tag(name="language.id"), value="EN")
     described_additionalcode__sid = TextElement(Tag(name="additional.code.sid"))
-    described_additionalcode__type__sid = TextElement(Tag(name="additional.code.type.id"))
+    described_additionalcode__type__sid = TextElement(
+        Tag(name="additional.code.type.id"),
+    )
     described_additionalcode__code = TextElement(Tag(name="additional.code"))
     description = TextElement(Tag(name="description"))
 
--- a/additional_codes/import_parsers.py
+++ b/additional_codes/import_parsers.py
@@ -32,11 +32,11 @@ class AdditionalCodeParser(ValidityMixin, Writable, ElementParser):
     record_code = "245"
     subrecord_code = "00"
 
-    tag = Tag(name="additional.code")
+    tag = Tag("additional.code")
 
-    sid = IntElement(Tag(name="additional.code.sid"))
-    type__sid = TextElement(Tag(name="additional.code.type.id"))
-    code = TextElement(Tag(name="additional.code"))
+    sid = IntElement(Tag("additional.code.sid"))
+    type__sid = TextElement(Tag("additional.code.type.id"))
+    code = TextElement(Tag("additional.code"))
     valid_between_lower = ValidityMixin.valid_between_lower
     valid_between_upper = ValidityMixin.valid_between_upper
 
@@ -68,14 +68,14 @@ class AdditionalCodeDescriptionPeriodParser(
     record_code = "245"
     subrecord_code = "05"
 
-    tag = Tag(name="additional.code.description.period")
+    tag = Tag("additional.code.description.period")
 
-    sid = TextElement(Tag(name="additional.code.description.period.sid"))
-    described_additionalcode__sid = TextElement(Tag(name="additional.code.sid"))
+    sid = TextElement(Tag("additional.code.description.period.sid"))
+    described_additionalcode__sid = TextElement(Tag("additional.code.sid"))
     described_additionalcode__type__sid = TextElement(
-        Tag(name="additional.code.type.id"),
+        Tag("additional.code.type.id"),
     )
-    described_additionalcode__code = TextElement(Tag(name="additional.code"))
+    described_additionalcode__code = TextElement(Tag("additional.code"))
     validity_start = ValidityStartMixin.validity_start
 
 
@@ -103,16 +103,16 @@ class AdditionalCodeDescriptionParser(Writable, ElementParser):
     record_code = "245"
     subrecord_code = "10"
 
-    tag = Tag(name="additional.code.description")
+    tag = Tag("additional.code.description")
 
-    sid = TextElement(Tag(name="additional.code.description.period.sid"))
-    language_id = ConstantElement(Tag(name="language.id"), value="EN")
-    described_additionalcode__sid = TextElement(Tag(name="additional.code.sid"))
+    sid = TextElement(Tag("additional.code.description.period.sid"))
+    language_id = ConstantElement(Tag("language.id"), value="EN")
+    described_additionalcode__sid = TextElement(Tag("additional.code.sid"))
     described_additionalcode__type__sid = TextElement(
-        Tag(name="additional.code.type.id"),
+        Tag("additional.code.type.id"),
     )
-    described_additionalcode__code = TextElement(Tag(name="additional.code"))
-    description = TextElement(Tag(name="description"))
+    described_additionalcode__code = TextElement(Tag("additional.code"))
+    description = TextElement(Tag("description"))
 
 
 @RecordParser.register_child("additional_code_type")
@@ -138,12 +138,12 @@ class AdditionalCodeTypeParser(ValidityMixin, Writable, ElementParser):
     record_code = "120"
     subrecord_code = "00"
 
-    tag = Tag(name="additional.code.type")
+    tag = Tag("additional.code.type")
 
-    sid = TextElement(Tag(name="additional.code.type.id"))
+    sid = TextElement(Tag("additional.code.type.id"))
     valid_between_lower = ValidityMixin.valid_between_lower
     valid_between_upper = ValidityMixin.valid_between_upper
-    application_code = TextElement(Tag(name="application.code"))
+    application_code = TextElement(Tag("application.code"))
 
 
 @RecordParser.register_child("additional_code_type_description")
@@ -167,11 +167,11 @@ class AdditionalCodeTypeDescriptionParser(Writable, ElementParser):
     record_code = "120"
     subrecord_code = "05"
 
-    tag = Tag(name="additional.code.type.description")
+    tag = Tag("additional.code.type.description")
 
-    sid = TextElement(Tag(name="additional.code.type.id"))
-    language_id = ConstantElement(Tag(name="language.id"), value="EN")
-    description = TextElement(Tag(name="description"))
+    sid = TextElement(Tag("additional.code.type.id"))
+    language_id = ConstantElement(Tag("language.id"), value="EN")
+    description = TextElement(Tag("description"))
 
 
 @RecordParser.register_child("footnote_association_additional_code")
@@ -199,12 +199,12 @@ class FootnoteAssociationAdditionalCodeParser(ValidityMixin, Writable, ElementPa
     record_code = "245"
     subrecord_code = "15"
 
-    tag = Tag(name="footnote.association.additional.code")
+    tag = Tag("footnote.association.additional.code")
 
-    additional_code__sid = TextElement(Tag(name="additional.code.sid"))
-    associated_footnote__footnote_type__sid = TextElement(Tag(name="footnote.type.id"))
-    associated_footnote__footnote_id = TextElement(Tag(name="footnote.id"))
+    additional_code__sid = TextElement(Tag("additional.code.sid"))
+    associated_footnote__footnote_type__sid = TextElement(Tag("footnote.type.id"))
+    associated_footnote__footnote_id = TextElement(Tag("footnote.id"))
     valid_between_lower = ValidityMixin.valid_between_lower
     valid_between_upper = ValidityMixin.valid_between_upper
-    additional_code__type__sid = TextElement(Tag(name="additional.code.type.id"))
-    additional_code__code = TextElement(Tag(name="additional.code"))
+    additional_code__type__sid = TextElement(Tag("additional.code.type.id"))
+    additional_code__code = TextElement(Tag("additional.code"))
--- a/additional_codes/import_parsers.py
+++ b/additional_codes/import_parsers.py
@@ -32,11 +32,11 @@ class AdditionalCodeParser(ValidityMixin, Writable, ElementParser):
     record_code = "245"
     subrecord_code = "00"
 
-    tag = Tag("additional.code")
+    tag = Tag(name="additional.code")
 
-    sid = IntElement(Tag("additional.code.sid"))
-    type__sid = TextElement(Tag("additional.code.type.id"))
-    code = TextElement(Tag("additional.code"))
+    sid = IntElement(Tag(name="additional.code.sid"))
+    type__sid = TextElement(Tag(name="additional.code.type.id"))
+    code = TextElement(Tag(name="additional.code"))
     valid_between_lower = ValidityMixin.valid_between_lower
     valid_between_upper = ValidityMixin.valid_between_upper
 
@@ -68,12 +68,12 @@ class AdditionalCodeDescriptionPeriodParser(
     record_code = "245"
     subrecord_code = "05"
 
-    tag = Tag("additional.code.description.period")
+    tag = Tag(name="additional.code.description.period")
 
-    sid = TextElement(Tag("additional.code.description.period.sid"))
-    described_additionalcode__sid = TextElement(Tag("additional.code.sid"))
-    described_additionalcode__type__sid = TextElement(Tag("additional.code.type.id"))
-    described_additionalcode__code = TextElement(Tag("additional.code"))
+    sid = TextElement(Tag(name="additional.code.description.period.sid"))
+    described_additionalcode__sid = TextElement(Tag(name="additional.code.sid"))
+    described_additionalcode__type__sid = TextElement(Tag(name="additional.code.type.id"))
+    described_additionalcode__code = TextElement(Tag(name="additional.code"))
     validity_start = ValidityStartMixin.validity_start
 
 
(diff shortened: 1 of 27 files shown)


importer/namespaces.py
```diff
@@ -1,39 +1,136 @@
...
+PATH_ASSETS = os.path.join(settings.BASE_DIR, "common", "assets")
+PATH_XSD_ENVELOPE = os.path.join(PATH_ASSETS, "envelope.xsd")
+PATH_XSD_TARIC = os.path.join(PATH_ASSETS, "taric3.xsd")
 
+xsd_schema_paths: Dict[str, str] = (
+    ("env", PATH_XSD_ENVELOPE),
+    ("oub", PATH_XSD_TARIC)
+)
+
+TARIC_RECORD_GROUPS: Dict[str, Sequence[str]] = dict(
```
stuaxo, line 28: This needs a docstring to explain its purpose.

importer/namespaces.py
```diff
@@ -1,39 +1,136 @@
...
+    commodities = (
+        "40000", "40005", "40010", "40015",
+        "40020", "40025", "40035", "40040"
+    )
```
stuaxo, line 32: Could do with a comment to explain what these commodities are  / why they are here.

importer/namespaces.py
```diff
@@ -1,39 +1,136 @@
...
+
+        if isinstance(tag, Tag):
+            tag_qualified_name = tag.qualified_name
+        else:
+            tag_qualified_name = tag
+
+        if is_pattern is True:
+            return self.pattern.match(tag_qualified_name)
+        else:
+            return self.qualified_name == tag_qualified_name
```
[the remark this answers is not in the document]
kintisheff: well spotted :)

importer/utils.py
```diff
@@ -81,17 +84,45 @@ def build_dependency_tree() -> Dict[str, Set[str]]:
...
-        if subclass.record_code not in dependency_map:
-            dependency_map[subclass.record_code] = set()
-        for _, relation in subclass.get_relations():
-            if (
-                relation.record_code != subclass.record_code
-                and relation.record_code in record_codes
-            ):
-                dependency_map[subclass.record_code].add(relation.record_code)
+        if subclass.__name__[:4] == "Test":
+            continue
```
simonwo, line 114: This looks a little brittle… what's it all about?

importer/tests/conftest.py
```diff
@@ -142,3 +148,65 @@ def batch_dependency() -> models.BatchDependencies:
...
+                        <oub:record.code>430</oub:record.code>
+                        <oub:subrecord.code>00</oub:subrecord.code>
+                        <oub:record.sequence.number>1</oub:record.sequence.number>
+                        <oub:update.type>3</oub:update.type>
+                        <oub:measure></oub:measure>
+                    </oub:record>
+                </oub:transmission>
+            </env:app.message>
+        </env:transaction>
+    </env:envelope>""".encode()
```
simonwo, line 181: In other tests of the importer we have tended to generate XML on demand using e.g. the `get_taric_xml` fixture. Would that be appropriate here? I realise that it's somewhat marking one's own homework, but we have tests in other places that check that the XML we output is correct and means we don't have to keep blobs of XML around.

importer/chunker.py
```diff
@@ -252,7 +253,67 @@ def write_transaction_to_chunk(
...
+        if len(message) == 0:
+            elem.remove(message)
+
+    if len(elem) != 0:
+        msg = (
+            f"Transaction {transaction_id}: "
+            f"{len(elem)} out of {n} records match the record group."
+        )
+        logger.info(msg)
+        return elem
```
simonwo, line 305: One thing that the existing scripts do that I don't think we have captured here: they ignore the transaction ID on the transaction and use the next available transaction ID. They need to do this because we have our own transaction ID sequence in the UK Tariff which doesn't match the EU's, so we need to use our own numbers else things will get out of order. We are also ignoring most of their transactions and while leaving holes is fine it's not super desirable. 

This will be fine if the import happens as pre-approved because the approval step will renumber the transactions, but if it goes straight to a published state that would be bad…

importer/namespaces.py
```diff
@@ -1,39 +1,173 @@
...
 
-T = TypeVar("T", bound="Tag")
+TTag = TypeVar("TTag", bound="Tag")
+TTags = TypeVar("TTags", bound="SchemaTagsBase")
+
+RE_PATTERN_TEST = re.compile(r"[^A-Za-z\.\_]")
+
+PATH_ASSETS = os.path.join(settings.BASE_DIR, "common", "assets")
+PATH_XSD_ENVELOPE = os.path.join(PATH_ASSETS, "envelope.xsd")
+PATH_XSD_TARIC = os.path.join(PATH_ASSETS, "taric3.xsd")
```
simonwo, line 28: I think the XSD path also appears in the `settings/common.py`… does that need to be cleaned up now?

simonwo (commented):
I like this method because we are pre-filtering the XML… storing less in the DB, and probably quicker overall. 

But I am wondering… would it be better to do this as part of `TransactionParser.save`? We will already have parsed the data at that point and so instead of iterating through XML elements we can just key down into the dictionary to look at record codes? That might also be a natural place to handle the changing ID requirements.

additional_codes/import_parsers.py
```diff
@@ -32,11 +32,11 @@ class AdditionalCodeParser(ValidityMixin, Writable, ElementParser):
...
 
-    tag = Tag("additional.code")
+    tag = Tag(name="additional.code")
 
-    sid = IntElement(Tag("additional.code.sid"))
-    type__sid = TextElement(Tag("additional.code.type.id"))
-    code = TextElement(Tag("additional.code"))
+    sid = IntElement(Tag(name="additional.code.sid"))
+    type__sid = TextElement(Tag(name="additional.code.type.id"))
+    code = TextElement(Tag(name="additional.code"))
```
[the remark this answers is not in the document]
kintisheff: as a matter of fact we can in this case, even without the __init__ construct. I just went for explicit, but will revert if that's the preference from a readability pov

importer/chunker.py
```diff
@@ -252,7 +253,67 @@ def write_transaction_to_chunk(
...
+) -> ET.Element:
+    """
+    Filters the records in a transaction based on record codes in the record
+    group.
+
+    Record identifiers are concatenated record_code and subrecord_code child element values.
+
+    Returns the a copy of the transaction element with non-matching records removed.
+    Returns the untouched transaction if record_group is none.
+    Returns None if there are no matching records in the transaction.
```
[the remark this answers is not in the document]
kintisheff: indeed

importer/utils.py
kintisheff: ah, well spotted I forgot I left that in - the issue is this - this utils method did not have test coverage before; upon adding some tests, it was revealed that models we create during testing that inherit from TrackedModel are caught in, polluting these tests. I'll fix

importer/tests/conftest.py
kintisheff: I'm agnostic actually - I saw we used `get_taric_xml`, hardcoded xml in test assets in a few places (e.g. in `commodities`),  and even in non-fixture functions in tests (e.g. `get_chuck_opener`) - I suppose `get_taric_xml` is more robust so happy to go with that.

importer/chunker.py
kintisheff: yes I am mindful of the need to reapply transaction id-s (since our order id-s won't match the EU's) - but i've deferred this for the next ticket - for now this is only used in the logs we print

additional_codes/import_parsers.py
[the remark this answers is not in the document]
stuaxo: Related:  can we take the opportunity docstring in the Tag class itself ?
It's non-obvious when in that file what the `name` parameter represents in the XML when starting from there (I can see now that there is one here, but the github view handily doesn't show it).

importer/namespaces.py
kintisheff: yes I debated whether these should go here, I guess I'll move them there so we can now have both there, and we can avoid DRY violations

kintisheff: Re the question on `TransactionParser`: no, I think it's preferable to do it early on at xml level:
1. It's cheaper to fail and redo it here (we also have the opportunity for further sanity checks)
2. Doing it here also allows us to print out a modified xml envelope artefact - for logging, lineage and reproducibility purposes, that would be beneficial
3. We still preserve the option to hook into TransactionParse.save downstream for other purposes, e.g. swap-out of transaction ID-s
We have added capabilities to make our taric xml parsing a bit more robust as part of this ticket, of course

stuaxo: Not a blocker, but we use pathlib.Path here ?

```python
PATH_ASSETS = Path(settings.BASE_DIR) / "common" / "assets"
PATH_XSD_ENVELOPE = PATH_ASSETS / "envelope.xsd"
PATH_XSD_TARIC = PATH_ASSETS / "taric3.xsd"
```

additional_codes/import_parsers.py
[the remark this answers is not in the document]
kintisheff: Yes!

importer/chunker.py
```diff
@@ -252,7 +253,67 @@ def write_transaction_to_chunk(
...
+    """
+    if record_group is None:
+        return elem
+
+    transaction_id = elem.get("id")
+    n = len(elem)
+
+    for message in list(Tags.ENV_APP_MESSAGE.iter(elem)):
```
stuaxo, line 276: ⚠️
====

The list manipulation in the following loop may be a performance trap, both `in` and `.remove`.

If this code is used e.g. in the seed importer, I would make double sure its not making a difference to the import time.
(I don't know this code, maybe it's not called very much and this is fine)

If it is an issue, consider whether lists are needed at all here, and data does need to be stored and looked up, some other structure may be better.

importer/chunker.py
```diff
@@ -252,7 +253,67 @@ def write_transaction_to_chunk(
...
+                if record_identifier not in record_group:
+                    transmission.remove(record)
+
+                    msg = (
+                        f"Transaction {transaction_id}: "
+                        f"Removed record with code {record_identifier}."
+                    )
+                    logger.info(msg)
+
+            if len(transmission) == 0:
```
stuaxo, line 293: Can this be `if not transmission:` ?

importer/chunker.py
```diff
@@ -252,7 +253,67 @@ def write_transaction_to_chunk(
...
+                message.remove(transmission)
+
+        if len(message) == 0:
+            elem.remove(message)
+
+    if len(elem) != 0:
```
stuaxo, line 299: Can this be `if elem:`

importer/chunker.py
kintisheff: Good question - yes I've cprofiled - although in theory there is no `maxOccurs ` bound defined on the app.message element in the envelope schema, in practice there are only a handful of messages per transaction in actual envelopes.

importer/chunker.py
```diff
@@ -252,7 +253,67 @@ def write_transaction_to_chunk(
...
+    for message in list(Tags.ENV_APP_MESSAGE.iter(elem)):
+        for transmission in list(Tags.OUB_TRANSMISSION.iter(message)):
+            for record in list(Tags.OUB_RECORD.iter(transmission)):
+                record_code = Tags.OUB_RECORD_CODE.first(record).text
+                subrecord_code = Tags.OUB_SUBRECORD_CODE.first(record).text
+
+                record_identifier = f"{record_code}{subrecord_code}"
+
+                if record_identifier not in record_group:
+                    transmission.remove(record)
```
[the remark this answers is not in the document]
kintisheff: we are NOT removing the list element itself (it stays in memory) - we are removing the element from the parent, which does not affect the memory object itself

importer/chunker.py
kintisheff: yes it can!

importer/chunker.py
kintisheff: yes!

importer/tests/conftest.py
```diff
@@ -142,3 +148,65 @@ def batch_dependency() -> models.BatchDependencies:
...
+                        <oub:record.code>400</oub:record.code>
+                        <oub:subrecord.code>00</oub:subrecord.code>
+                        <oub:record.sequence.number>1</oub:record.sequence.number>
+                        <oub:update.type>3</oub:update.type>
+                        <oub:goods.nomenclature></oub:goods.nomenclature>
+                    </oub:record>
+                </oub:transmission>
+            </env:app.message>
+        </env:transaction>
+    </env:envelope>""".encode()
```
stuaxo, line 202: Instead of .encode(), could this use a byte string   `return b"""<?xml version`... ?

importer/tests/test_chunker.py
```diff
@@ -22,14 +27,26 @@ def get_basic_chunk_text(id: str) -> bytes:
...
+    xml: str,
+    Tags: TTags,
+    record_group: Sequence[str],
+) -> ET.Element:
+    transaction = get_snippet_transaction(xml, Tags)
+    return filter_transaction_records(transaction, record_group)
+
+
 @mock.patch("importer.chunker.TemporaryFile")
 def test_get_chunk(mock_temp_file: mock.MagicMock):
```
stuaxo, line 40: I know you didn't originate this, but if you have an idea what it does, a docstring would be nice.

importer/tests/test_chunker.py
```diff
@@ -61,3 +78,28 @@ def test_close_chunk():
         batch.chunks.get(record_code=400, chapter="01").chunk_text
         == get_basic_chunk_text("2").decode()
     )
+
+
+def test_transaction_filter_positive(
+    taric_schema_tags,
+    record_group,
+    envelope_commodity,
+):
```
stuaxo, line 87: A docstring would be nice here and for the test below.

importer/tests/test_handlers.py
```diff
@@ -118,7 +120,8 @@ def test_resolve_dependencies_returns_true(
 def test_get_generic_link(prepped_handler):
```
stuaxo, line 120: Not a blocker: I know you didn't write this, but if you have an idea what this test verifies, a short docstring would be handy.

importer/tests/test_chunker.py
kintisheff: hm we might as well on all the tests then?

importer/tests/conftest.py
kintisheff: we're switching this to  the use of get_taric_xml in this fixture instead anyway (see earlier suggestion by Simon) - but throughout tamato we use both b"" and .encode(). should we settle on a preferred one for the future?

importer/tests/test_handlers.py
kintisheff: good question, i'll check

importer/tests/test_namespaces.py
```diff
@@ -0,0 +1,32 @@
...
+def get_snippet_transaction(
+    xml: str,
+    Tags: TTags,
+) -> ET.Element:
+    envelope = ET.fromstring(xml)
+    return Tags.ENV_TRANSACTION.first(envelope)
+
+
+def test_element_tag_finder(taric_schema_tags, envelope_commodity):
+    transaction = get_snippet_transaction
```
stuaxo, line 19: Did you mean to call get_snippet_transaction here ?

If you did - the test may need a stronger assertion.

importer/tests/test_namespaces.py
```diff
@@ -0,0 +1,32 @@
...
+    assert transaction is not None
+
+
+def test_tag_pattern(tag_name):
+    assert tag_name.is_pattern == False
```
stuaxo, line 24: is False

importer/tests/test_namespaces.py
```diff
@@ -0,0 +1,32 @@
...
+    assert tag_name == quota_event
+    assert tag_name != quota_balance_event
+
+
+def test_regex_tag_pattern(tag_regex):
```
stuaxo, line 29: is True

importer/tests/test_namespaces.py
```diff
@@ -0,0 +1,32 @@
...
+    envelope = ET.fromstring(xml)
+    return Tags.ENV_TRANSACTION.first(envelope)
+
+
+def test_element_tag_finder(taric_schema_tags, envelope_commodity):
+    transaction = get_snippet_transaction
+    assert transaction is not None
+
+
+def test_tag_pattern(tag_name):
```
stuaxo, line 23: Consider adding some background in a docstring

"A tag with a pattern should XYZ verify by checking it ..."

importer/tests/test_namespaces.py
```diff
@@ -0,0 +1,32 @@
...
+
+def get_snippet_transaction(
+    xml: str,
+    Tags: TTags,
+) -> ET.Element:
+    envelope = ET.fromstring(xml)
+    return Tags.ENV_TRANSACTION.first(envelope)
+
+
+def test_element_tag_finder(taric_schema_tags, envelope_commodity):
```
stuaxo, line 18: A little background on how the tag finder should behave and how this test checks that might be good for a docstring here.

importer/tests/test_utils.py
```diff
@@ -1,6 +1,38 @@
...
+    return (key_lens, value_lens)
+
+
 def test_generate_key():
     key = utils.generate_key("test", ["1", "2"], {"1": 1, "2": 2})
     assert key == "a98ec5c5044800c88e862f007b98d89815fc40ca155d6ce7909530d792e909ce"
+
+
+def test_dependency_tree_structure_default():
+    key_lens, value_lens = get_dependency_tree(False)
```
stuaxo, line 28: Are lens's a haskel concept ?

This test needs a docstring, and probably some commetns to communicate what it's doing and why.

importer/tests/test_namespaces.py
kintisheff: hm looks like I accidentally deleted the rest of that line of code or something. thanks for spotting

importer/tests/test_utils.py
```diff
@@ -1,6 +1,38 @@
...
+
+    assert len(key_lens) == len(value_lens) == 1
+    assert list(key_lens)[0] == list(value_lens)[0] == 3
```
stuaxo, line 31: Where possible assertions should be on content not len (though sometimes it can be appropriate).

The bottom assertion isn't self-explanatory, at the very least it could do with a last param as extra info, but probably needs some line(s) of comments to explain what's being tested.

importer/tests/test_utils.py
```diff
@@ -1,6 +1,38 @@
...
+
+
+def test_dependency_tree_structure_subrecord_codes():
```
stuaxo, line 34: (same comments as above test)

importer/tests/test_utils.py
kintisheff: -ish. yes to docstrings as per above

importer/tests/test_utils.py
kintisheff: yes i'll add comments for this case

importer/chunker.py
```diff
@@ -252,7 +254,67 @@ def write_transaction_to_chunk(
...
+    if elem:
+        msg = (
+            f"Transaction {transaction_id}: "
+            f"{len(elem)} out of {n} records match the record group."
+        )
+        logger.info(msg)
+        return elem
+
+    msg = f"Transaction id {transaction_id}: " "Removed - no matching records. "
+    logger.info(msg)
```
stuaxo, line 309: Using fstrings for log messages here is fine as we probably don't care about the extra cost (and basically always have info enabled).

But if you ever want to not pay the price for string interpolation if the logging is not enabled, you can use the old style string formatting:

```python
# In this version the string will not be interpolated if log level is WARNING or ERROR:
logger.info(""Transaction id %s: " "Removed - no matching records. ", transaction_id)
```

TBH it's a shame there isn't a really need way to use fstrings together with logging .. something like a lazy fstring.

importer/tests/test_chunker.py
stuaxo: Only as we come across them, unless they are so trivial that a comment wouldn't add anything* ?

*This is in the eye of the beholder, I know.

importer/chunker.py
kintisheff: oh true that;s a good point

importer/tests/test_chunker.py
kintisheff: makes sense to me

simonwo (approved):
I think from my POV I am happy as long as we don't forget about the ID thing in the later work.

stuaxo (approved):

PR #360, opened 2021-08-17
TP859: Use json to parse env variables

rachaelcodes:
Builds on #357 , but parses the environment variables using JSON.

I'm going to ask WebOps today about us deploying non-master branches in Jenkins, to avoid all this incremental PR approach to fixing deployment related issues.

### DIFF
--- a/exporter/sqlite/runner.py
+++ b/exporter/sqlite/runner.py
@@ -1,4 +1,5 @@
 import decimal
+import json
 import logging
 import os
 import sqlite3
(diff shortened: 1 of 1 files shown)


exporter/sqlite/runner.py
```diff
@@ -38,7 +39,9 @@ def manage(self, *args: str):
         sqlite_env["DATABASE_URL"] = f"sqlite:///{self.db}"
         # Required to make sure the postgres default isn't set as the DB_URL
         if sqlite_env.get("VCAP_SERVICES"):
-            sqlite_env["VCAP_SERVICES"].pop("postgres", None)
+            vcap_env = json.loads(sqlite_env["VCAP_SERVICES"])
+            vcap_env.pop("postgres", None)
+            sqlite_env["VCAP_SERVICES"] = json.dumps(vcap_env)
```
rachaelcodes, line 44: Let me know if there's some kind of snazzy Python one-liner to shorten this stage.

simonwo (approved):

exporter/sqlite/runner.py
stuaxo: Only to make it less readable:

```python
sqlite_env["VCAP_SERVICES"] = json.dumps({k: v for k, v in json.loads(sqlite_env["VCAP_SERVICES"]) if k != 'postgres'})

You could probably make it more evil by replacing json.dumps with `str` or `repr`.

PR #365, opened 2021-08-24
Run all indirect business rules on workbasket clean

simonwo:
Previously this only detected a small subset of relations and so didn't run many indirect business rules. It now correctly returns relations and runs all of the appropriate indirect business rules.

### DIFF
--- a/common/business_rules.py
+++ b/common/business_rules.py
@@ -7,12 +7,11 @@ from typing import Iterable
 from typing import Iterator
 from typing import Mapping
 from typing import Optional
-from typing import Set
-from typing import Tuple
 from typing import Type
 from typing import Union
 
 from django.core.exceptions import ObjectDoesNotExist
+from django.core.exceptions import ValidationError
 from django.db.models import QuerySet
 
 from common.models.records import TrackedModel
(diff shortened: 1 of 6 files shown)


common/business_rules.py
```diff
@@ -102,19 +103,15 @@ def get_linked_models(
...
-                    related_instance = getattr(model, field.name)
-                    yield current_related_instances.get(
-                        version_group=related_instance.version_group,
-                    )
+                    related_instances = [getattr(model, field.name)]
+                for instance in related_instances:
+                    try:
+                        yield instance.version_at(transaction)
+                    except TrackedModel.DoesNotExist:
+                        continue
```
rachaelcodes, line 114: Not a change request, but out of interest, why are we skipping here?

common/models/records.py
```diff
@@ -638,17 +638,27 @@ def version_at(self: Cls, transaction) -> Cls:
 
     @classmethod
     def get_relations(cls) -> list[tuple[Field, type[TrackedModel]]]:
-        """Find all foreign key and one-to-one relations on an object and return
-        a list containing tuples of the field instance and the related model it
-        links to."""
+        """Find all the models that are related to this one either a link stored
+        on this model or on the related model and return a list containing
+        tuples of the field instance and the related model."""
```
rachaelcodes, line 643: I found this a bit tricky to parse. Could it be "...to this one (either a link stored on this model or on the related model) and...'?

rachaelcodes (approved):

common/business_rules.py
simonwo: Because `related_instances` will return all instances, even deleted ones, and `version_at` will return `DoesNotExist` if the item has been deleted as of a certain transaction. That's ok, because we don't need to run indirect business rules against deleted things. I'll add a comment to clarify.

common/models/records.py
```diff
@@ -638,17 +638,30 @@ def version_at(self: Cls, transaction) -> Cls:
...
-            and not f.auto_created
-            and f.concrete
+            if (f.many_to_one or f.one_to_one or f.one_to_many)
             and f.model == cls
             and issubclass(f.related_model, TrackedModel)
+            and f.related_model is not TrackedModel
+        ]
+
+    @classmethod
+    def get_objects_linked_to(cls) -> list[tuple[Field, type[TrackedModel]]]:
```
andyhd, line 658: Tiny quibble - this returns classes, so maybe the name should be `get_models_linked_to` or `get_linked_models` or something, but avoiding `objects`. And is the order important? If not, perhaps return `dict[Field, type[TrackedModel]]` instead?

andyhd (approved):

andyhd (approved):

PR #371, opened 2021-09-02
TP-803: Commodity Tree Change Handler

kintisheff:
This ticket is part of the [HS22 Epic](https://uktrade.atlassian.net/browse/TP-911?atlOrigin=eyJpIjoiODE4ZmQ5ZjhmY2FiNGQxOTg0ZDZlMzQ1NzMxMGYwNjMiLCJwIjoiaiJ9).

A previous ticket implemented a selective xml commodity importer.
This ticket is responsible for evaluating any side effects of pending changes to the commodity tree, and providing remedies
A subsequent feature (see TP-931) will provide functionality to implement the sum of commodity changes and remedies

Notes:
- Please read the docs included in `commodities/__init__.py` before you start your review - this will help put the changes in this PR in context
- Please consider reviewing the 60 or so new tests included in `commodities/tests` as a key part of your review - esp. the scenario tests, which reflect ADR13; before reviewing the tests, please read the module-level docs on fixtures in `commodities/tests/conftest.py`


conftest.py
```diff
@@ -79,6 +79,20 @@ def pytest_bdd_apply_tag(tag, function):
...
 
+@contextlib.contextmanager
+def not_raises(ExpectedException):
+    """Provides a context manager for tests that need to assert a specific
+    exception is not raised."""
+    try:
+        yield
+
+    except ExpectedException as error:
+        raise AssertionError(f"Raised exception {error} when it should not!")
```
stuaxo, line 90: I wonder if raise `AssertionError(f"Raised exception {error} when it should not!") from error` would be good here, to perserve the stacktrace?

Same with the next exception.

simonwo: I'm not sure I understand the point of `not_raises`… surely a raised exception will simply bubble up and fail the test? And why declare the `ExpectedException` specifically if we then also raise an assertion on any `Exception` below?

kintisheff: the only motivation is code readability

commodities/__init__.py
```diff
@@ -5,4 +5,120 @@
...
+  - e.g. are all parent, child, sibling and other relations as they should be
+  - the question applies for the changing commodity
+    as well as all commodities in its hierarchy prior and post change
+    (these may or may not be the same depending on the change)
+2. Have we dealt with any side effects on any related taric records
+  - related records may include measures, footnote associations, etc.
+  - affected records may be related to the changing commodity itself
+    or to any commodity in the surrounding hierarchy pre- and post-change
+  - side effects are incidental violations of business rules
+    resulting form the changes made to the commodity tree
```
rachaelcodes, line 23: Should this be 'resulting from'?

commodities/__init__.py
```diff
@@ -5,4 +5,120 @@
...
+    -- included commodities may or may not be effective at the same point in time
+    -- the may wrap any version of a good object
+  - commodity changes are applied to this collection construct
+
+3. CommodityTreeSnapshot
+  - a collection provides the ability to take "snapshots"
+  - a snapshot is a collection of commodities that are in effect
+    as of a given moment, and constitute the tree hierarchy as of that moment
+  - a snapshot can be taken based on one of two "clocks", but not both:
+    -- a calendar clock (what goods where in effect as of date x)
```
rachaelcodes, line 66: 'what goods were in effect'?

commodities/migrations/0009_auto_20210902_1002.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.1.12 on 2021-09-02 10:02
```
rachaelcodes, line 1: Are we naming our migrations files now? It looks like a lot of the recent ones have been named.

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+                    self.commodities,
+                ),
+            )
+        except StopIteration:
+            return
+
+    def _get_sanitized_code(self, code: str) -> str:
+        """Returns the commodity code in GoodsNomenclature.item_id format."""
+        code = code.replace(".", "")
+        return code + "0" * (10 - len(code))
```
andyhd, line 255: `return f"{code:0<10}"`

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
+"""Includes dataclasses used in goods nomenclature hierarchy tree management."""
+import logging
+from dataclasses import dataclass
+from datetime import date
+from hashlib import md5
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
```
andyhd, line 10: If you `from __future__ import annotations`, I think you can use `dict[]`, `list[]`, and `tuple[]` instead of these soon-to-be-deprecated types from `typing`

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+@dataclass
+class Commodity(TrackedModelWrapper):
+    """
+    Provides a wrapper of the GoodsNomenclature model.
+
+    See the docs for TrackedModelWrapper for additional context.
+    """
+
+    obj: GoodsNomenclature
+    indent: int = None
```
andyhd, line 52: Must be `Optional[int]` if the default value is `None` - `None` is not an `int`.

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+    """
+
+    commodities: List[Commodity]
+
+    def get_commodity(
+        self,
+        code: str,
+        suffix: str = None,
+        version: int = None,
+        version_group_id: int = None,
```
andyhd, line 214: These all need to be `Optional` because the default value is `None`

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+    - a calendar date (when the clock type is calendar)
+    - a transaction id (when the clock type is transaction)
+
+    For additional context on commodity trees, see the docs for CommodityTree.
+    """
+
+    commodities: List[Commodity]
+    clock_type: ClockType
+    moment: Union[date, int]
+    edges: Dict[str, Commodity] = None
```
andyhd, line 329: `Optional`

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+          -- the method keeps a list of the prior sorted commodities
+          -- this allows it to assign the correct parents for 0-indent headings.
+        """
+        chapters = {c.chapter for c in self.commodities}
+
+        try:
+            assert len(chapters) == 1
+        except AssertionError:
+            print("All commodities in a group must be from the same HS chapter.")
+            raise
```
andyhd, line 356: Should this be an `assert`? Why not
```python
if not len(chapters) == 1:
    raise CommodityTreeSnapshotException("All commodities in a group must be from the same HS chapter.")
```

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+    def is_declarable(self, commodity: Commodity) -> bool:
+        if commodity.suffix != SUFFIX_DECLARABLE:
+            return False
+
+        return len(self.get_children(commodity)) == 0
+
+    def compare_parents(
+        self,
+        commodity: Commodity,
+        snapshot: "CommodityTreeSnapshot",
```
andyhd, line 433: If you `from __future__ import annotations`, I think you can do `snapshot: CommodityTreeSnapshot`.

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+
+    def compare_descendants(
+        self,
+        commodity: Commodity,
+        snapshot: "CommodityTreeSnapshot",
+    ) -> SnapshotDiff:
+        return self._get_diff(commodity, snapshot, TreeNodeRelation.DESCENDANTS)
+
+    @property
+    def snapshot_date(self) -> date:
```
andyhd, line 466: `Optional`

andyhd (commented):
There's a lot here - I'm still reviewing, but I thought I'd submit some of my feedback so far

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+@pytest.fixture
+def transaction_pool(workbasket) -> Iterator[Transaction]:
+    """
+    Returns an iterator with transactions in descending order of id.
+
+    See the module-level docs for details on the use of the transaction_pool.
+    """
+    factory = factories.TransactionFactory
+
+    transactions = [factory.create(workbasket=workbasket) for _ in range(50)][::-1]
```
andyhd, line 171: This would be clearer (and more performant) as
```python
transactions = reversed([factory.create(workbasket=workbasket) for _ in range(50)])
```
and if you don't need to create the transactions in advance of stepping through the iterator, you could do
```python
transactions = reversed(factory.create(workbasket=workbasket) for _ in range(50))
```
which might be more memory efficient.

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+        ("9910.20.00.00", "80", 2, date_ranges.normal),
+        ("9999.00.00.00", "80", 1, date_ranges.normal),
+        ("9999.10.00.00", "80", 2, date_ranges.normal),
+        ("9999.20.00.00", "80", 2, date_ranges.normal),
+        ("9999.20.00.10", "80", 3, date_ranges.normal),
+    )
+
+    commodities = [create_commodity(transaction_pool, *args) for args in params]
+    keys = [f"{c.trimmed_dot_code}_{c.suffix}_{c.indent}" for c in commodities]
+    return dict(zip(keys, commodities))
```
andyhd, line 200: `zip` isn't especially esoteric, but this would probably be clearer as
```python
return {
    f"{c.trimmed_do_code}_{c.suffix}_{c.indent}": c
    for c in commodities
}
```

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+
+    attrs = dict(valid_between=date_ranges.normal)
+    create_dependent_measure(candidate, transaction_pool, **attrs)
+    create_footnote_association(candidate, transaction_pool, **attrs)
+
+    attrs = dict(valid_between=date_ranges.overlap_normal_earlier)
+    create_dependent_measure(candidate, transaction_pool, **attrs)
+    create_footnote_association(candidate, transaction_pool, **attrs)
+
+    print(1)
```
andyhd, line 342: 👀

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+    delayed_transaction = factories.TransactionFactory.create(
+        workbasket=workbasket,
+    )
+    attrs = dict(
+        transaction=delayed_transaction,
+        measure_type=measure_type,
+    )
+    measure = create_dependent_measure(candidate, transaction_pool, **attrs)
+
+    print(measure)
```
andyhd, line 413: `print`

commodities/tests/test_dataclasses.py
```diff
@@ -0,0 +1,699 @@
...
+
+@contextlib.contextmanager
+def not_raises(ExpectedException):
+    """Provides a context manager for tests that need to assert a specific
+    exception is not raised."""
+    try:
+        yield
+
+    except ExpectedException as error:
+        raise AssertionError(f"Raised exception {error} when it should not!")
```
andyhd, line 31: I think you could use `pytest.fail` here instead of `raise AssertionError`.

commodities/tests/test_dataclasses.py
```diff
@@ -0,0 +1,699 @@
...
+
+    except Exception as error:
+        raise AssertionError(f"An unexpected exception {error} raised.")
```
andyhd, line 34: This seems redundant - an uncaught exception is unexpected by definition.

commodities/tests/test_dataclasses.py
```diff
@@ -0,0 +1,699 @@
...
+
+        assert a == b
+
+
+def test_commodity_identifier(commodities):
+    re_identifier = re.compile(r"([0-9\.]{13})-([1-8]{1}0)-([0-9]{1,2})/([0-9]{1})")
+
+    for commodity in commodities.values():
+        with not_raises(StopIteration):
+            match = next(re_identifier.finditer(commodity.identifier))
```
andyhd, line 184: I think this might be clearer without the context manager, eg:
```python
try:
    match = next(re_indentifier.finditer(commodity.identifier))
except StopIteration:
    pytest.fail("Raised StopIteration when it should not!")
```

commodities/tests/test_dataclasses.py
```diff
@@ -0,0 +1,699 @@
...
+            groups = match.groups()
+
+            assert len(groups) == 4
+            code, suffix, indent, version = match.groups()
```
andyhd, line 188: `code, suffix, indent, version = groups`

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+        suffix=suffix,
+        valid_between=validity,
+        transaction=transaction,
+    )
+
+    return Commodity(obj=obj, indent=indent)
+
+
+def create_collection(
+    commodities: List[Commodity],
```
andyhd, line 93: Looks like this should be `Union[List[Commodity], Dict[str, Commodity]]]`

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+    return factories.GoodsNomenclatureFactory.create(
+        valid_between=date_ranges.normal,
+        transaction=next(transaction_pool),
+    )
+
+
+@pytest.fixture
+def commodities(date_ranges, transaction_pool) -> Dict[str, Commodity]:
+    params = (
+        ("9900.00.00.00", "80", 0, date_ranges.normal),
```
andyhd, line 187: You could potentially parametrize this fixture, eg:
```python
@pytest.fixture(params=(
    ("9900.00.00.00", "80", 0),
    ...
))
def commodity(date_ranges, transaction_pool, request):
    return create_commodity(transaction_pool, *request.param, *(date_ranges.normal,)) # a bit ugly
```
which would mean your test functions would be automatically parameterized and could be like
```python
def test_commodity_code(commodity):
    assert commodity.code == commodity.obj.item_id
```
but it looks like this `commodities` fixture is being passed to `create_collection` too, so perhaps a no-go.

commodities/tests/conftest.py
andyhd: Since the key is computed from the commodity attributes, you don't really need to return a `dict` here, do you?

andyhd (commented):
Still going! More comments to come, but here are a bunch for now.

commodities/migrations/0009_auto_20210902_1002.py
stuaxo: @andyhd maybe ?

They are often squashed as a result of something else - releasing, should we start releasing often, or even all the time ?

Things probably get a bit more interesting when data migrations are in the mix (though if we were to do a seed import they go away).

commodities/tests/conftest.py
stuaxo: It's also probably worth using create_batch here:

```python
... factory.create_batch(50, workbasket=workbasket) ...
```

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+            )
+        if self.current is not None:
+            current = self.collection.get_commodity(
+                self.current.code,
+                self.current.suffix,
+            )
+
+        if self.update_type == UpdateType.CREATE:
+            if self.candidate is None:
+                msg = "The operation is CREATE but no commodity was provided."
```
andyhd, line 820: These messages follow a similar pattern, perhaps you could construct them in `_handle_validation_issue` like
```python
self._handle_validation_issue("no commodity was provided")
```
and
```python
def _handle_validation_issue(self, msg: str) -> None:
    """Logs a warning message or raises an error."""
    msg = f"The operation is {self.update_type} but {msg}"
    if self.ignore_validation_rules is True:
        logger.warn(msg)
    else:
        raise ValueError(msg)
```
and perhaps this should be an `Exception` subclass, eg
```python
class CommodityChangeException(ValueError):
    def __init__(self, msg: str, change: CommodityChange) -> None:
        super().__init__(f"Commodity {change.update_type} error: {msg}")
```
which would look like `CommodityChangeException: Commodity Create error: no commodity was provided` in the traceback. Or perhaps even better -
```python
class CommodityChangeException(ValueError):
    pass

class CommodityCreateException(CommodityChangeException):
    pass

class CommodityUpdateException...
```
which would look like `CommodityCreateException: no commodity was provided` in the traceback

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+                self._handle_validation_issue(msg)
+
+            if candidate == self.candidate:
+                msg = (
+                    "The operation is UPDATE "
+                    "but there is no effective change to the commodity."
+                )
+                self._handle_validation_issue(msg)
+
+    def handle_side_effects(self) -> None:
```
andyhd, line 860: I'd be tempted to rename this something like `collect_side_effects`, or maybe `infer_side_effects`? Because "handle" is quite vague and, to me, implies that something is going to be done with the side effects, rather than just collecting a list of them.

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+            self._add_pending_delete(association)
+
+    def _handle_update_side_effects(self) -> None:
+        """
+        Preempt business rule violations due to commodity code updates.
+
+        NOTE:
+        1. NIG35 is not used as it is redundant in this context after NIG34
+        2. For some rules invoked from a measure's point of view (esp. ME7)
+          there is no need to strict need to invoke them
```
andyhd, line 909: there is no need to invoke them

andyhd (commented):
Still going!

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,214 @@
...
+    "Commodities.GoodsNomenclature": "item_id",
+    "Commodities.GoodsNomenclatureIndentNode": "depth",
+    "Geo_Areas.GeographicalArea": "area_id",
+    "Measures.MeasureAction": "code",
+    "Measures.MeasureConditionCode": "code",
+    "Measures.MeasurementUnit": "code",
+    "Measures.MeasurementUnitQualifier": "code",
+    "Measures.MonetaryUnit": "code",
+    "Quotas.QuotaOrderNumber": "order_number",
+    "Regulations.Group": "group_id",
```
andyhd, line 23: We have the concept of "identifying fields" on `TrackedModel` already. Are these keys used differently? I thought that some models could not be uniquely identified by a single field - for example a quota order number can be reused at a later date and can potentially represent an entirely different quota.

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,214 @@
...
+
+    @property
+    def envelope(self) -> str:
+        """
+        Returns the envelope with which the record was added.
+
+        TODO: Apply a more robust logic to identifying the envelope.
+        """
+        workbasket = self.obj.transaction.workbasket
+        envelope = workbasket.title.rsplit(" ", 1)[1]
```
andyhd, line 98: 👀

common/models/meta/wrappers.py
andyhd: Are these just for display?

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,214 @@
...
+        return self.get_current_version_as_of_transaction_id(transaction.id)
+
+    def is_current_as_of_transaction(self, transaction: Transaction) -> bool:
+        """Returns True if this is the current version of the record as of a
+        given transaction."""
+        return self.is_current_as_of_transaction_id(transaction.id)
+
+    def get_current_version_as_of_transaction_id(
+        self,
+        transaction_id: int = None,
```
andyhd, line 188: `Optional`

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,214 @@
...
+    ) -> int:
+        """Returns the current version of the record as of a given transaction
+        id."""
+        if transaction_id is None:
+            return self.current_version
+
+        ids = self._get_transaction_ids()
+        return len([tid for tid in ids if tid <= transaction_id])
+
+    def is_current_as_of_transaction_id(self, transaction_id: int = None) -> bool:
```
andyhd, line 198: `Optional`

andyhd (commented):
I made it to the end! Still digesting, but looks good to me on first impressions. 💥

commodities/migrations/0009_auto_20210902_1002.py
kintisheff: ^ what is the preference, naming them or not? i'm agnostic

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+        If the indent field of this dataclass has been set, this will take
+        precedence over the indent of the wrapped GoodsNomenclature object.
+        """
+        if self.indent is not None:
+            return self.indent
+
+        obj = self.obj.indents.order_by("validity_start").last()
+
+        if obj is None:
+            return
```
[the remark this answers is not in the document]
kintisheff: it's `Optional[int]`

commodities/models/dc.py
kintisheff: readibility

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+        transaction = next(transaction_pool)
+
+    return factory.create(transaction=transaction, **kwargs)
+
+
+def create_dependent_measure(
+    commodity: Commodity, transaction_pool: Iterator[Transaction], **kwargs
+) -> Measure:
+    """Returns a new measure linked to a given good."""
+    factory = factories.MeasureFactory
```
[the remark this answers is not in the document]
kintisheff: I think that's just black reformatting it this way

commodities/tests/conftest.py
kintisheff: good shout on readability/performance with reversed and on use of create batch. unfortunately, the pool is needed in advance so that will need to stay

commodities/tests/conftest.py
kintisheff: using a mapping here makes the tests code more readable elsewhere

commodities/tests/test_dataclasses.py
kintisheff: three people don't like the context manager by now. ok :) it's just for readability

commodities/tests/conftest.py
kintisheff: yep that's the reason (your latest point)

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+
+@dataclass
+class CommodityTreeBase(BaseModel):
+    """
+    Provides a base model for commodity tree and snapshots.
+
+    See the docs of CommodityTree and CommodityTreeSnapshot for details.
+    """
+
+    commodities: List[Commodity]
```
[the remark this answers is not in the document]
kintisheff: anything is allowed, the order is not significant (unless in the case of `CommodityTreeSnapshot`, where order is ensured in post_init)

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+        elif relation == TreeNodeRelation.SIBLINGS:
+            attr = "get_siblings"
+        elif relation == TreeNodeRelation.CHILDREN:
+            attr = "get_children"
+        elif relation == TreeNodeRelation.ANCESTORS:
+            attr = "get_ancestors"
+        elif relation == TreeNodeRelation.DESCENDANTS:
+            attr = "get_descendants"
+        else:
+            raise ValueError("Unknown relation")
```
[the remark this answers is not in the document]
kintisheff: ah true

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+
+        return CommodityTreeSnapshot(
+            commodities=commodities,
+            clock_type=ClockType.TRANSACTION,
+            moment=transaction_id,
+        )
+
+    def clone(self) -> "CommodityCollection":
+        """Returns an independent copy of the collection."""
+        return CommodityCollection(self.commodities[:])
```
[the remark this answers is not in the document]
kintisheff: agnostic. i'll use `copy`

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+    """
+    "Provides a model for commodity change side effects.
+
+    A side effect is simply a pending action taken to preempt a business rule
+    violation related to a commodity code change.
+    """
+
+    obj: TrackedModel
+    update_type: UpdateType
+    attrs: Dict[Any, Any] = None
```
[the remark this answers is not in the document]
kintisheff: it's `dict[str, Any]` def

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+    we will need to be in sync with the EU on any changes to the goods nomenclature,
+    and deal with any repercussions of this downstream.
+    """
+
+    collection: CommodityCollection
+    update_type: UpdateType
+    current: Commodity = None
+    candidate: Commodity = None
+    ignore_validation_rules: bool = False
+    side_effects: Dict[Union[str, int], SideEffect] = None
```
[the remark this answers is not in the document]
kintisheff: adding the type will improve readability, yes!

commodities/models/dc.py
kintisheff: ^ yes to all

kintisheff: actually i'll stop with `CommodityChangeException`

commodities/models/dc.py
kintisheff: agreed! i will go with `preempt_side_effects`

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+        """Handle conflicts related to validity span rule violations."""
+        good = self.candidate.obj
+
+        gvb = good.valid_between
+        ovb = obj.valid_between
+
+        # If obj validity does not overlap the good validity at all,
+        # mark the object for deletion
+        if (ovb.upper and gvb.lower > ovb.upper) or (
+            gvb.upper and gvb.upper < ovb.lower
```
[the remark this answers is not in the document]
kintisheff: agree in principle. Postgres' daterange supports all sorts of contains, overlaps, etc, but that can only be leveraged in django db query constructs, but the psycopg2 Range only provides __contains__ (but perhaps because overlaps can be constructed from the other dundrs of Range

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+        if gvb.upper and ovb.upper and gvb.upper < ovb.upper:
+            end_date = gvb.lower
+        if gvb.lower > ovb.lower:
+            start_date = gvb.upper
+
+        # Circuit breaker for cases where the obj is spanned by the good
+        if start_date is None and end_date is None:
+            return
+
+        # Mark the object for update with trimmed validity
```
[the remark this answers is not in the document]
kintisheff: also agree. won't touch it for now but re this and above comment, will add a ticket to attach some reusable convenience methods not just to TaricDateRange but other frequently used classes - and cleanup related DRY violations in our code

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+            """
+            return (
+                f"{m.measure_type}."
+                f"{m.geographical_area.area_id}."
+                f"{m.order_number}."
+                f"{m.dead_order_number}."
+                f"{m.additional_code}."
+                f"{m.dead_additional_code}."
+                f"{m.reduction}"
+            )
```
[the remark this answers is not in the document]
kintisheff: a bit controversial though as we might need a separate ticket to refactor and unify ME32 uses (I think i pinned a TODO somewhere there). so would defer to that separate ticket.

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+
+            NOTE: If the two measures also have the same "fingerprint"
+            (see _get_measure_key() above), this will violate ME32.
+            """
+            earlier, later = sorted((m1, m2), key=lambda x: x.valid_between.lower)
+
+            return (
+                earlier.valid_between.upper is None
+                or earlier.valid_between.upper >= later.valid_between.lower
+            )
```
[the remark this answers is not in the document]
kintisheff: ^ same remark as above, i'll punch in a separate ticket (which actually will partially revolve around wrappers and the introduction on ModelMeta from labs as previously discussed last week during deep dive)

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+
+        commodity = self.candidate or self.current
+        good = commodity.obj
+        measures = {_get_measure_key(m): m for m in good.dependent_measures}
+
+        clone = self.collection.clone()
```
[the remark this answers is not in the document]
kintisheff: neat

commodities/models/dc.py
```diff
@@ -0,0 +1,1196 @@
...
+            for relative in getattr(after, attr)(commodity):
+                related_measures = relative.obj.dependent_measures
+
+                for related_measure in related_measures:
+                    key = _get_measure_key(related_measure)
+
+                    try:
+                        measure = measures[key]
+
+                        if _validities_overlap(measure, related_measure):
```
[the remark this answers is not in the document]
kintisheff: ^ potentially yes, see reply to earlier comment re ME32

commodities/tests/test_dataclasses.py
```diff
@@ -0,0 +1,699 @@
...
+def test_commodity_get_indent(commodities):
+    for commodity in commodities.values():
+        a = commodity.indent is not None
+        b = commodity.get_indent() == commodity.indent
+
+        assert a == b
+
+
+def test_commodity_identifier(commodities):
+    re_identifier = re.compile(r"([0-9\.]{13})-([1-8]{1}0)-([0-9]{1,2})/([0-9]{1})")
```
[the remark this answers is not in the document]
kintisheff: true it's an rstring already

commodities/tests/test_dataclasses.py
```diff
@@ -0,0 +1,699 @@
...
+def test_collection_current_snapshot(collection_full):
+    snapshot = collection_full.current_snapshot
+    tx_snapshot = collection_full.get_transaction_clock_snapshot()
+    cal_snapshot = collection_full.get_calendar_clock_snapshot()
+
+    tx_commodities = set(map(lambda x: x.identifier, tx_snapshot.commodities))
+    cal_commodities = set(map(lambda x: x.identifier, cal_snapshot.commodities))
+    tx_cal_commodities = tx_commodities.intersection(cal_commodities)
+
+    commodities = set(map(lambda x: x.identifier, snapshot.commodities))
```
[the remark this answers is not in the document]
kintisheff: agree

commodities/tests/test_dataclasses.py
```diff
@@ -0,0 +1,699 @@
...
+
+    for snapshot in snapshots:
+        with not_raises(StopIteration):
+            match = next(res[snapshot.clock_type].finditer(snapshot.identifier))
+            groups = match.groups()
+
+            assert len(groups) == 2
+
+            if snapshot.clock_type == ClockType.CALENDAR:
+                assert groups[0] == str(snapshot.moment)
```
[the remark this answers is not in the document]
kintisheff: yes, except using `hash_` since hash is a reserved keyword

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,214 @@
...
+    "Measures.MeasureConditionCode": "code",
+    "Measures.MeasurementUnit": "code",
+    "Measures.MeasurementUnitQualifier": "code",
+    "Measures.MonetaryUnit": "code",
+    "Quotas.QuotaOrderNumber": "order_number",
+    "Regulations.Group": "group_id",
+}
+
+TRACKEDMODEL_IDENTIFIER_FALLBACK_KEY = "sid"
+TRACKEDMODEL_PRIMARY_KEY = "id"
```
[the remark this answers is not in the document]
kintisheff: yes, that would be safer even though as you point out by definition this wrapper only wraps models that inherit from `TrackedModel`

common/models/meta/wrappers.py
kintisheff: yes they are used differently, and these are used in viz displays primarily yes (their use will become more apparent in coming tickets, for now i've just chucked them in there as I already need to ref them in the wrapper

common/models/meta/wrappers.py
kintisheff: ah yes, we did this to ourselves :) /unless i'm missing something/

commodities/models/dc.py
andyhd: Also `assert` should generally be avoided for non-test code, as it is not executed if optimization is turned on.

But even if you want to keep this as `assert`, you can replace this with
```python
assert len(chapters) == 1, "All commodities in a group must be from the same HS chapter."
```

commodities/models/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+__all__ = [
+    "GoodsNomenclature",
+    "GoodsNomenclatureDescription",
+    "GoodsNomenclatureIndent",
+    "GoodsNomenclatureIndentNode",
+    "GoodsNomenclatureIndentQuerySet",
+    "GoodsNomenclatureOrigin",
+    "GoodsNomenclatureSuccessor",
+    "FootnoteAssociationGoodsNomenclature",
+]
```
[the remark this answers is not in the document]
kintisheff: it's related to how `manage.py` handles these imports

commodities/models/dc.py
kintisheff: ^agree with this

common/models/meta/wrappers.py
kintisheff: (therefore the TODO)

andyhd: I was just checking you meant to leave that TODO there. I thought the logic looks sketchy, but then I realised that these properties may only be used for display.

andyhd (approved):
Very nice 👍

commodities/tests/conftest.py
stuaxo: +1 on the dict comprehension @andyhd suggested above.

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+        ("9999.20.00.10", "80", 3, date_ranges.normal),
+    )
+
+    commodities = [create_commodity(transaction_pool, *args) for args in params]
+    keys = [f"{c.trimmed_dot_code}_{c.suffix}_{c.indent}" for c in commodities]
+    return dict(zip(keys, commodities))
+
+
+@pytest.fixture
+def commodities_spanned(date_ranges, transaction_pool):
```
stuaxo, line 209: A docstring would be nice here, and in the fixtures below.

commodities/tests/test_dataclasses.py
```diff
@@ -0,0 +1,709 @@
...
+from commodities.models.constants import SUFFIX_DECLARABLE
+from commodities.models.constants import ClockType
+from commodities.models.dc import CommodityChange
+from commodities.models.dc import CommodityTreeBase
+from common.util import TaricDateRange
+from common.validators import UpdateType
+
+from .conftest import copy_commodity
+
+# from conftest import not_raises
```
stuaxo, line 18: Old commented code.

common/models/meta/wrappers.py
simonwo: Agreed that this should be delegated to each individual model. But what are they used for? Is `IDENTIFIER KEYS` a good name… because as Andy points out, they are not unique.

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,223 @@
...
+                key = TRACKEDMODEL_IDENTIFIER_FALLBACK_KEY
+            else:
+                key = TRACKEDMODEL_PRIMARY_KEY
+
+        return key
+
+    @property
+    def identifier(self) -> TTrackedModelIdentifier:
+        """Returns the object identifier which is the value of the identifier
+        field."""
```
simonwo, line 90: This docstring doesn't tell me what this does or what properties these identifiers have. As above, they are not unique from model to model, so what are they for?

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,223 @@
...
+
+        if key is None:
+            model_field_names = self.obj._meta.fields
+
+            if TRACKEDMODEL_IDENTIFIER_FALLBACK_KEY in model_field_names:
+                key = TRACKEDMODEL_IDENTIFIER_FALLBACK_KEY
+            else:
+                key = TRACKEDMODEL_PRIMARY_KEY
+
+        return key
```
simonwo, line 85: Why is this not just the `__str__` of the model?

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,223 @@
...
+        try:
+            v = self.obj.valid_between
+
+            return " - ".join(
+                (
+                    v.lower.strftime(DATE_FORMAT),
+                    "inf" if v.upper is None else v.upper.strftime(DATE_FORMAT),
+                ),
+            )
+        except self.model.DoesNotExist:
```
simonwo, line 121: Shouldn't this only get raised if the object itself does not exist? Why would that be raised here?

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,223 @@
...
+            try:
+                validity_start = self.obj.validity_start
+                return f"{validity_start.strftime(DATE_FORMAT)} - inf"
```
simonwo, line 124: But this isn't correct – an object with a validity start date is not valid for unlimited time, it's valid up until the start of the next model. See the docstring on `ValidityStartMixin`.

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,223 @@
...
+        """Returns the model class for the wrapped object."""
+        return self.obj._meta.model
+
+    @property
+    def record_type(self) -> str:
+        """Returns the Taric3 record type for the wrapped model."""
+        try:
+            record_type = self.obj.record_code[0]
+        except AttributeError:
+            record_type = "0"
```
simonwo, line 51: This is surely not ok

simonwo (changes requested):
You already know I agree with the design in principle :)

I'm a little unsure if the testing is thorough enough – I feel like I've found a bunch of places where tests are needed. E.g. did you get all the test cases associated with `commodities/tests/test_import_handlers.py`? There are some tricky cases in there of e.g three levels of zero indent, and of children arriving before their parents, that we have added by observation when fixing bugs, and I this code will need to pass them.

I'm also not happy about much of the code in `common/models/meta/*` because it seems like it is repeating a lot of the things we have elsewhere or should live elsewhere (e.g. most of this could live in `common/models/records.py`). It also seems to be taking a lot of shortcuts and returning data that is incorrect in lots of places. I'm not clear on why we need much of this? "For display in a data engineer's notebook" is fine, but that doesn't mean it has to live in the main repo. Or if we're really keen on it being in the main repo, then I'm not happy committing things that return misleading results in some circumstances and not finding the right place in the code base for them to live.

Keen to hear thoughts.

kintisheff:
> You already know I agree with the design in principle :)
> 
> I'm a little unsure if the testing is thorough enough – I feel like I've found a bunch of places where tests are needed. E.g. did you get all the test cases associated with `commodities/tests/test_import_handlers.py`? There are some tricky cases in there of e.g three levels of zero indent, and of children arriving before their parents, that we have added by observation when fixing bugs, and I this code will need to pass them.
> 
> I'm also not happy about much of the code in `common/models/meta/*` because it seems like it is repeating a lot of the things we have elsewhere or should live elsewhere (e.g. most of this could live in `common/models/records.py`). It also seems to be taking a lot of shortcuts and returning data that is incorrect in lots of places. I'm not clear on why we need much of this? "For display in a data engineer's notebook" is fine, but that doesn't mean it has to live in the main repo. Or if we're really keen on it being in the main repo, then I'm not happy committing things that return misleading results in some circumstances and not finding the right place in the code base for them to live.
> 
> Keen to hear thoughts.

Yes to both:
1) I picked up a few more scenarios while doing the plumbing (TP-931) - and will add a few more tests in scenarios as a result; however we can't predict them all and ultimately my suggestion would be to ensure some extra coverage here, but then ensure more scenario coverage in separate tickets down the road as we encounter more outlier cases.
2) We also had the same chat with Stuart and agreed that `TrackedModelWrapper` is overkill and/or redundant for now (it would become more relevant when we start doing the model meta facades - so i'll clean it up from this PR as we actually don't need it here)

commodities/business_rules.py
```diff
@@ -304,16 +309,21 @@ class NIG35(BusinessRule):
...
+            .approved_up_to_transaction(
+                good.transaction,
+            )
             .exists()
-        ):
+        )
+
+    def validate(self, good):
+        if good.update_type != UpdateType.DELETE:
+            return
```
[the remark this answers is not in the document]
kintisheff: ah that's a good shout!

commodities/business_rules.py
```diff
@@ -304,16 +309,21 @@ class NIG35(BusinessRule):
...
-        if (
+    def has_violation(self, good):
+        return (
             good.measures.model.objects.filter(
                 goods_nomenclature__sid=good.sid,
                 additional_code__isnull=False,
             )
-            .approved_up_to_transaction(good.transaction)
+            .approved_up_to_transaction(
+                good.transaction,
```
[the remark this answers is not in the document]
kintisheff: happy to change

commodities/models/constants.py
```diff
@@ -0,0 +1,22 @@
...
+    CHILDREN = "children"
+    ANCESTORS = "ancestors"
+    DESCENDANTS = "descendants"
+
+
+class ClockType(Enum):
+    """Provides an enumeration for clock types used in the Taric system."""
+
+    CALENDAR = auto()
+    TRANSACTION = auto()
```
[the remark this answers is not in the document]
kintisheff: good question; we can move it there; since we're doing more work down the line on clocks, might make sense indeed.

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+
+    @property
+    def subheading(self) -> str:
+        """Returns the HS subheading for the commodity code."""
+        return self.code[:6]
+
+    @property
+    def cn_subheading(self) -> str:
+        """Returns the CN subheading for the commodity code."""
+        return self.code[:8]
```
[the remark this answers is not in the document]
kintisheff: one reason alone - wanted to see if people want them in model first (didn't want to presume and pollute), if yes, than i'm totally for it

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+
+    @property
+    def is_cn_subheading(self) -> bool:
+        """Returns true if the commodity code represents a CN subheading."""
+        return self.trimmed_code == self.cn_subheading
+
+    @property
+    def is_taric_subheading(self) -> bool:
+        """Returns true if the commodity code represents a Taric subheading."""
+        return self.trimmed_code == self.code
```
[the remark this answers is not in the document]
kintisheff: same reply as above, will do that

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+        """
+        Returns the true indent of the commodity.
+
+        If the indent field of this dataclass has been set, this will take
+        precedence over the indent of the wrapped GoodsNomenclature object.
+        """
+        if self.indent is not None:
+            return self.indent
+
+        obj = self.obj.indents.order_by("validity_start").last()
```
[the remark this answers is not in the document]
kintisheff: i've changed it! will push the changes with the commit that reflects all your other comments

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+        The two commodities are considered equal if they share the same code,
+        suffix, indent, version, and validity dates.
+        """
+        if commodity is None:
+            return False
+
+        return (
+            self.identifier == commodity.identifier
+            and self.obj.valid_between == commodity.obj.valid_between
+        )
```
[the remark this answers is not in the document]
kintisheff: We're comparing the wrapper states, not the underlying obj states here.
`Commodity.identifier` will be a function of code, suffix, indent and version
E.g. in an `UpdateType.UPDATE` scenario, the indent (e.g.) might be changed on the wrapper to simulate a new snapshot and look for side effects, but the underlying obj will continue to have the same sid. so this __eq__ needs to be more detailed than sid

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+                    self.commodities,
+                ),
+            )
+        except StopIteration:
+            return None
+
+    def _get_sanitized_code(self, code: str) -> str:
+        """Returns the commodity code in GoodsNomenclature.item_id format."""
+        code = code.replace(".", "")
+        return f"{code:0<10}"
```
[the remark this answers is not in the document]
kintisheff: wasn't aware of that method, will check it

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+      at two different points in time or as of two different transactions
+    - compare the 'before' and 'after' snapshots around a tree update
+
+    For additional context on on snapshots, see the docs for CommodityTreeSnapshot.
+    """
+
+    commodity: Commodity
+    relation: TreeNodeRelation
+    exists: dict[str, bool]
+    values: dict[str, Union[Commodity, list[Commodity]]]
```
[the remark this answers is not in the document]
kintisheff: good spot, remnants from a prior iteration:
`exists` is redundant now
`values` can be a list now indeed

commodities/models/orm.py
```diff
@@ -101,15 +102,19 @@ def __str__(self):
...
-            self.measures.model.objects.filter(
-                goods_nomenclature__sid=self.sid,
-            )
-            .approved_up_to_transaction(self.transaction)
-            .exists()
+    @property
+    def dependent_measures(self):
+        currently_effective_range = TaricDateRange(
+            lower=date.today(),
         )
```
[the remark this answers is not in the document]
kintisheff: exactly! also a remnant from a prior iteration - thanks for spotting

commodities/tests/test_dataclasses.py
```diff
@@ -0,0 +1,709 @@
...
+
+@contextlib.contextmanager
+def not_raises(ExpectedException):
+    """Provides a context manager for tests that need to assert a specific
+    exception is not raised."""
+    try:
+        yield
+
+    except ExpectedException as error:
+        pytest.fail(f"Raised exception {error} when it should not!")
```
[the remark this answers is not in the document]
kintisheff: i don't think we use it anymore actually in the tests; will remove

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,223 @@
...
+    "Measures.MeasureAction": "code",
+    "Measures.MeasureConditionCode": "code",
+    "Measures.MeasurementUnit": "code",
+    "Measures.MeasurementUnitQualifier": "code",
+    "Measures.MonetaryUnit": "code",
+    "Quotas.QuotaOrderNumber": "order_number",
+    "Regulations.Group": "group_id",
+}
+
+TRACKEDMODEL_IDENTIFIER_FALLBACK_KEY = "sid"
```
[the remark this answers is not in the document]
kintisheff: i'll check

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+                child
+                for commodity in generation
+                for child in self.get_children(commodity)
+            ]
+
+            if len(children) == 0:
+                break
+
+            descendants.extend(children)
+            generation = children
```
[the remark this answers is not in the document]
kintisheff: yes `get_descendants` is expensive at the moment. didn't want to pre-materilize it either though; i'll fish for smarter algorithms

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+        Note that since this is a commodity tree snapshot,
+        multiple versions of the same commodity are not possible.
+        """
+        self.commodities = sorted(
+            self.commodities,
+            key=lambda x: (
+                x.code,
+                x.suffix,
+                x.indent,
+            ),
```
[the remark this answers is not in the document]
kintisheff: just in case :)

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+        edges = {}
+
+        for commodity in self.commodities:
+            indent = commodity.get_indent()
+
+            if indent == 0:
+                try:
+                    parent = indents[indent][-1]
+                except IndexError:
+                    parent = None
```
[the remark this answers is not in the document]
kintisheff: YES!! noticed that as well - also fixed and I'll push in the commit that reflects your other comments.

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+                    parent = indents[indent - 1]
+                except KeyError:
+                    parent = None
+
+            edges[commodity.identifier] = parent
+
+            if indent == 0:
+                indents[indent].append(commodity)
+            else:
+                indents[indent] = commodity
```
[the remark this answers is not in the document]
kintisheff: no, just the latest commodity in the sorted order. we don't need to remember the prior ones (therefore the pre-sort). except when the indent is `0` of course

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+        - add the candidate commodity for update type create
+
+        For detials on change constructs and current vs candidate commodities in a change,
+        see the docs for the CommodityChange model.
+        """
+        for change in changes:
+            try:
+                self.commodities.remove(change.current)
+            except ValueError:
+                pass
```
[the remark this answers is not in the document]
kintisheff: `remove` rasies a `ValueError` if it receives `None` as value. this is another way of saying, you can't do an update if there is no `current` commodity in the `CommodityChange` object. 
The algo then is this:
1. always try to delete the current commodity - this will only work if the op is `DELETE` or `UPDATE` because only in those cases do you have `self.current != None` (CommodityChange validates the sanity of the op with respect to the provided attributes in `__post__init__`)
2. if op is `DELETE`, nothing else to do, leave
3. if op is `UPDATE` or `CREATE`, there will be a `candidate`, add it to the `CommodityCollection`

Does this answer it sufficiently?

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+        """
+        Return a commodity tree snapshot as of a certain transaction (based on
+        transaction id).
+
+        If the optional transaction_id argument is not provided, this method
+        will return the snapshot as of the most recent transaction in the db and
+        will assign as the snapshot moment the highest transaction id across all
+        snapshot commodities.
+        """
+        today = date.today()
```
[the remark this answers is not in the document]
kintisheff: we agree in a conversation not to allow that to happen at the moment  (taking both a transaction and a calendar snapshot at the same time) - I can bring that capability back of course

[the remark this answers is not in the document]
kintisheff: (or am I misunderstanding your point here?)

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+            if current != self.current:
+                self._handle_validation_issue(
+                    "there is no matching commodity in the commodity collection",
+                )
+
+        if self.update_type == UpdateType.UPDATE:
+            if self.current == self.candidate:
+                self._handle_validation_issue(
+                    "the current and new versions are identical",
+                )
```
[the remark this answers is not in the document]
kintisheff: fair, we can allow it then - should we not still issue a warning though at least?

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+        super().__post_init__()
+
+    def validate(self) -> None:
+        """
+        Validates the sanity of the field values for the change model instance.
+
+        There are two types of validation checks: a) ensure that the current and
+        candidate fields are provided where required by the update type b)
+        ensure that the requested changes makes sense in the context of the
+        affected commodity collection c) ensure that the requested changes do
```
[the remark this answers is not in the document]
kintisheff: via preemption of side effects

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+        if self.candidate is not None:
+            candidate = self.collection.get_commodity(
+                self.candidate.code,
+                self.candidate.suffix,
+            )
+        if self.current is not None:
+            current = self.collection.get_commodity(
+                self.current.code,
+                self.current.suffix,
+            )
```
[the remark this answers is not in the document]
kintisheff: YES! please suggest something, my muse has been obdurate on this one

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+                try:
+                    mbr.ME7().validate(measure)
+                except BusinessRuleViolation:
+                    self._add_pending_delete(measure)
+
+        # ME88: Invoked from the POV of a measure
+        # here, find all related measures and invoke the BR
+        # (inefficient for this workflow, but consistent use of BR-s)
+        if len(self.candidate.trimmed_code) > len(self.current.trimmed_code):
+            for measure in good.measures.latest_approved().all():
```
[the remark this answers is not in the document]
kintisheff: measures should be enough I think, yes

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+            """
+            return (
+                f"{m.measure_type}."
+                f"{m.geographical_area.area_id}."
+                f"{m.order_number}."
+                f"{m.dead_order_number}."
+                f"{m.additional_code}."
+                f"{m.dead_additional_code}."
+                f"{m.reduction}"
+            )
```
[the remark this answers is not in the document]
kintisheff: yes to actual values! it's changed pending commit already
re the other potential issue, need to factor that in, good shout!

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+        """Handle conflicts related to validity span rule violations."""
+        good = self.candidate.obj
+
+        gvb = good.valid_between
+        ovb = obj.valid_between
+
+        # If obj validity does not overlap the good validity at all,
+        # mark the object for deletion
+        if (ovb.upper and gvb.lower > ovb.upper) or (
+            gvb.upper and gvb.upper < ovb.lower
```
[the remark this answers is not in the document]
kintisheff: no - we're only checking if a measure end date is before the `start_date` of a commodity, we're not comparing measure end date to commodity end date

if a measure end date is before the start date of the commodity, there is no date range overlap at all
if a measure start date is after the end date of the commodity, there is no date range overlap at all

thus if a measure end date is only before the _end date_ of a commodity (and not before the _start date_ as well)
(e. g. `gvb.upper and ovb.upper and ovb.upper < gvb.upper`), 
this `or` condition will not trigger and the measure will not be deleted (at least not for this reason, it may be deleted for other reasons down the line)

I could always be going senile of course, so pls shout if this doesn't make sense

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+
+    See the module-level docs for details on the use of the transaction_pool.
+    """
+    try:
+        transaction = kwargs["transaction"]
+        del kwargs["transaction"]
+    except KeyError:
+        transaction = next(transaction_pool)
+
+    return factory.create(transaction=transaction, **kwargs)
```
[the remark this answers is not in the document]
kintisheff: yep i think so

commodities/tests/conftest.py
```diff
@@ -1,8 +1,479 @@
...
+  on related measures, footnote associations, etc.
+- In order to avoid creating duplicate logic in our app,
+  we leverage existing business rules for detecting any side effects
+- This involves a mix of goods-centric business rules (NIG-s)
+  and measure-centric business rules (ME-s)
+- Sometimes, NIG-s will look for measures approved as of a good's transaction
+  and ME-s will look for goods approved as of a mesure's transaction
+- In order to properly configure the fixtures,
+  we use a small transaction pool with transactions in descending order
+  where needed, retaining the ability to add delayed_transactions as well
```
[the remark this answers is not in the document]
kintisheff: ME type business rules will check on commodities up to their own transaction often.
NG type business rules will check on measures up to their own transaction in some cases too.
In order to avoid not triggering some rules incidentally, I needed this. (this is an issue arising in testing only)

common/models/meta/wrappers.py
kintisheff: for visialization purposes only, but as per prior comments, this is not needed here and will go away for this PR

common/models/meta/wrappers.py
kintisheff: `TrackedModelWrapper` is being removed, but yes, that could be the str dundr

common/models/meta/wrappers.py
kintisheff: good point

common/models/meta/wrappers.py
kintisheff: true; also this has been removed by now, just need to commit and push

common/models/meta/wrappers.py
kintisheff: indeed - and has been removed :)

commodities/models/dc.py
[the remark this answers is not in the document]
kintisheff: there's a broader issue here: should perhaps make these checks stand alone functions and surround them with the appropriate entourage of tests

commodities/business_rules.py
```diff
@@ -304,16 +309,21 @@ class NIG35(BusinessRule):
 
     # XXX this is redundant - NIG34 will be violated first
```
stuaxo, line 310: ^ For some future PR, these sort of things point to something we can encode in the machinery itself (either it knows and optimises them away, or we can add tests or something (hand waves ))

common/models/meta/wrappers.py
stuaxo: I remember seeing somehow rationalising it like "oh, there must be a good reason" before it even got to the front of my mind properly !???

common/models/meta/wrappers.py
```diff
@@ -0,0 +1,223 @@
...
+            )
+        except self.model.DoesNotExist:
+            try:
+                validity_start = self.obj.validity_start
+                return f"{validity_start.strftime(DATE_FORMAT)} - inf"
+            except self.model.DoesNotExist:
+                return "n/a"
+
+    @property
+    def version(self) -> int:
```
stuaxo, line 129: Transaction.id won't work as a version.

You won't be able to get one int, the best you can do once my TP-841 is merged is a tuple of `(transaction.partition, transaction.order)`

For current master things are a little more complex, though if you transactions from the seed import and non approved workbaskets you could get away with it.

commodities/models/dc.py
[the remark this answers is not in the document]
simonwo: Ahh yes, of course!

commodities/models/dc.py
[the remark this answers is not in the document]
simonwo: I do remember the conversation but can't remember why now … 🤦‍♂️

commodities/models/dc.py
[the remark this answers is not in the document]
simonwo: Ah yeah fair enough

commodities/models/dc.py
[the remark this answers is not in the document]
kintisheff: yep will use that

commodities/models/dc.py
```diff
@@ -0,0 +1,1182 @@
...
+
+    def __eq__(self, commodity: "Commodity") -> bool:
+        """
+        Returns true if the two commodities are equal.
+
+        The two commodities are considered equal if they share the same code,
+        suffix, indent, version, and validity dates.
+        """
+        if commodity is None:
+            return False
```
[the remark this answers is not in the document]
kintisheff: turns out it's better to leave it returning `False` as there are many cases where we are comparing a missing current or candidate commodity to another commodity (legitimately)

commodities/models/orm.py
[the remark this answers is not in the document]
kintisheff: i think we didn't want the devs to have too much freedom at first, before we hash clocks out better - and I continue to agree with your point there!

commodities/models/dc.py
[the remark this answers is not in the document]
kintisheff: i think we didn't want the devs to have too much freedom at first, before we hash clocks out better - and I continue to agree with your point there!

common/models/meta/wrappers.py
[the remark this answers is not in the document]
kintisheff: yep switching to that

commodities/models/dc.py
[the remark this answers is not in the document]
kintisheff: i'm changing it as a log warning only, not raising error

commodities/models/dc.py
```diff
@@ -519,6 +564,12 @@ def snapshot_transaction_id(self) -> Optional[int]:
         if self.clock_type == ClockType.TRANSACTION:
             return self.moment
 
+    @property
+    def snapshot_transaction(self) -> Optional[int]:
```
simonwo, line 568: Returns a `Transaction` object, not `int`

kintisheff:
`c1ef6ff` also fixes the handling of parent-child relations for cases with multiple 0-indent headings

commodities/models/orm.py
```diff
@@ -97,18 +181,26 @@ class Meta:
     def __str__(self):
         return self.item_id
 
+    @property
+    def code(self) -> CommodityCode:
+        return CommodityCode(code=self.item_id)
+
```
simonwo, line 187: This is defined twice?

commodities/models/orm.py
```diff
@@ -60,13 +143,14 @@ class GoodsNomenclature(TrackedModel, ValidityMixin):
...
-        return self.item_id[8:] != "00"
+    def code(self) -> CommodityCode:
+        """Returns a CommodityCode instance for the good."""
+        return CommodityCode(code=self.item_id)
 
     @property
     def footnote_application_codes(self) -> Set[ApplicationCode]:
         codes = {ApplicationCode.TARIC_NOMENCLATURE, ApplicationCode.DYNAMIC_FOOTNOTE}
-        if not self.is_taric_code:
+        if not self.code.is_taric_code:
```
simonwo, line 153: We have changed all the call sites of this but we still have a functioning `is_taric_code` property, so can we revert these back to `self.is_taric_code`?

commodities/models/orm.py
```diff
@@ -97,18 +181,26 @@ class Meta:
...
+
     @property
     def autocomplete_label(self):
         return f"{self} - {self.get_description().description}"
 
+    @property
+    def dependent_measures(self):
+        return self.measures.model.objects.filter(
+            goods_nomenclature__sid=self.sid,
+        ).approved_up_to_transaction(self.transaction)
```
simonwo, line 196: I think this is a little wrong in the same way that `in_use` is wrong, in that it should be a function of a user-supplied transaction rather than `self.transaction`. The latter will only tell you the measures that are there at the point the commodity was created or updated…

simonwo (approved):

simonwo:
Can we make sure to squash this branch when merging :)

commodities/models/orm.py
kintisheff: done

commodities/models/orm.py
kintisheff: that's just how the original `in_use` was written - happy to write a separate ticket to investigate if that needs improvement. in TP-803, we only use dependent measures in the context of a snapshot (except in some tests where it is more convenient).

kintisheff:
> Can we make sure to squash this branch when merging :)

I thought we typically squash

commodities/models/orm.py
kintisheff: fixed

PR #373, opened 2021-09-02
Update disk quota with correct syntax

rachaelcodes:
Same principle as #370 for attempting to update the disk space we have allocated in PaaS, but with correct syntax this time round.


simonwo (approved):
Ha, I did wonder, but thought it was just YAML being a bizarre langauge…

stuaxo (approved):

PR #375, opened 2021-09-07
Tp 929 remove automatic triggering of workbasket export

gabelton:
- Removing code calling upload_workbaskets task from WorkBasket model approve method
- Updating existing tests
- Adding WorkBasket to admin
- Adding custom button Workbasket list page. This button calls the upload_workbaskets task. (See screenshot below)
- Adding tests for custom button

Screenshot 2021-09-06 at 14 26 24

### DIFF
--- a/workbaskets/admin.py
+++ b/workbaskets/admin.py
@@ -1,6 +1,28 @@
 from django.contrib import admin
+from django.http import HttpResponseRedirect
+from django.http.response import HttpResponseForbidden
+from django.urls import path
+
+from exporter.tasks import upload_workbaskets
 from workbaskets.models import WorkBasket
+from workbaskets.validators import WorkflowStatus
 
 class WorkBasketAdmin(admin.ModelAdmin):
-    pass
+    change_list_template = "workbasket_change_list.html"
+
+    def get_urls(self):
+        urls = super().get_urls()
+        export_url = path('upload/', self.upload)
+        
+        return [export_url] + urls
+    
+    def upload(self, request):
+        if not request.user.is_superuser:
+            return HttpResponseForbidden('Only superusers may export workbaskets')
+        
+        upload_workbaskets.delay()
+        self.message_user(request, f"Uploading workbaskets with status of '{WorkflowStatus.READY_FOR_EXPORT.label}'")
+        return HttpResponseRedirect("../")
+
+
 admin.site.register(WorkBasket, WorkBasketAdmin)
--- a/workbaskets/admin.py
+++ b/workbaskets/admin.py
@@ -1,7 +1,9 @@
 from django.contrib import admin
+from django.contrib.admin.views.decorators import staff_member_required
 from django.http import HttpResponseRedirect
-from django.http.response import HttpResponseForbidden
 from django.urls import path
+from django.urls import reverse
+from django.utils.decorators import method_decorator
 
 from exporter.tasks import upload_workbaskets
 from workbaskets.models import WorkBasket
@@ -13,20 +15,18 @@ class WorkBasketAdmin(admin.ModelAdmin):
 
     def get_urls(self):
         urls = super().get_urls()
-        upload_url = path("upload/", self.upload)
+        upload_url = path("upload/", self.upload, name="upload")
 
         return [upload_url] + urls
 
+    @method_decorator(staff_member_required)
     def upload(self, request):
-        if not request.user.is_superuser:
-            return HttpResponseForbidden("Only superusers may upload workbaskets")
-
         upload_workbaskets.delay()
         self.message_user(
             request,
             f"Uploading workbaskets with status of '{WorkflowStatus.READY_FOR_EXPORT.label}'",
         )
-        return HttpResponseRedirect("../")
+        return HttpResponseRedirect(reverse("admin:workbaskets_workbasket_changelist"))
 
 
 admin.site.register(WorkBasket, WorkBasketAdmin)
--- a/workbaskets/admin.py
+++ b/workbaskets/admin.py
@@ -7,21 +7,25 @@ from exporter.tasks import upload_workbaskets
 from workbaskets.models import WorkBasket
 from workbaskets.validators import WorkflowStatus
 
+
 class WorkBasketAdmin(admin.ModelAdmin):
     change_list_template = "workbasket_change_list.html"
 
     def get_urls(self):
         urls = super().get_urls()
-        export_url = path('upload/', self.upload)
-        
-        return [export_url] + urls
-    
+        upload_url = path("upload/", self.upload)
+
+        return [upload_url] + urls
+
     def upload(self, request):
         if not request.user.is_superuser:
-            return HttpResponseForbidden('Only superusers may export workbaskets')
-        
+            return HttpResponseForbidden("Only superusers may upload workbaskets")
+
         upload_workbaskets.delay()
-        self.message_user(request, f"Uploading workbaskets with status of '{WorkflowStatus.READY_FOR_EXPORT.label}'")
+        self.message_user(
+            request,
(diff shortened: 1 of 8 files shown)


andyhd:
A general comment about commit messages: I think a good convention to follow is to have a commit message complete the sentence "When merged, this commit will ...".

Your commit messages are good, but the -ings are a bit superfluous.

Of course when you merge this PR you will squash all the commits - right? 😉

workbaskets/admin.py
```diff
@@ -0,0 +1,32 @@
...
+
+class WorkBasketAdmin(admin.ModelAdmin):
+    change_list_template = "workbasket_change_list.html"
+
+    def get_urls(self):
+        urls = super().get_urls()
+        upload_url = path("upload/", self.upload)
+
+        return [upload_url] + urls
+
```
stuaxo, line 19: You might be able to use the @staff_member_required decorator here instead of checking the permission in the function.

workbaskets/tests/test_admin.py
```diff
@@ -0,0 +1,40 @@
...
+    def test_upload_calls_task(self, valid_user, client):
+        with mock.patch(
+            "exporter.tasks.upload_workbaskets.delay",
+        ) as mock_task:
+            valid_user.is_superuser = True
+            valid_user.save()
+            client.force_login(valid_user)
+            client.post(self.url)
+
+            mock_task.assert_called_once_with()
```
stuaxo, line 40: This might want to be assert_called_once()

workbaskets/tests/test_admin.py
```diff
@@ -0,0 +1,40 @@
...
+
+pytestmark = pytest.mark.django_db
+
+
+class TestWorkBasketUpload:
+    @classmethod
+    def setup_class(cls):
+        cls.url = "/admin/workbaskets/workbasket/upload/"
+
+    def test_upload_returns_302_for_valid_superuser(self, valid_user, client):
```
stuaxo, line 15: Maybe we should have fixture for something like valid_superuser ?

workbaskets/tests/test_admin.py
```diff
@@ -0,0 +1,40 @@
...
+        valid_user.is_superuser = True
+        valid_user.save()
```
andyhd, line 17: I am probably wrong, but I think this could affect other tests that use the `valid_user` fixture. Perhaps it would be safer and clearer to create a new fixture for the superuser, eg:
```python
@pytest.fixture
def superuser():
    return UserFactory.create(is_superuser=True)
```
And then this test could be parametrized, eg:
```python
@pytest.mark.parametrize("user, expected_response_status", [
    (valid_user, 403),
    (superuser, 302),
])
def test_upload(self, client, user, expected_response_status, request):  # this is a bit ugly
    user = request.getfixturevalue(user). # more ugly
    client.force_login(user)
    response = client.post(self.url)
    assert response.status_code == expected_response_status
    if response.status_code == 302:
        assert response.url == "../"
```
And then you could probably incorporate the mock call assertion and this class could be a single test function.

But this is just a suggestion and your code is fine as is, if there is no effect on other tests that use `valid_user`.

andyhd (commented):
Not sure if there is a potential cross-test dependency issue

workbaskets/admin.py
```diff
@@ -0,0 +1,32 @@
...
+
+class WorkBasketAdmin(admin.ModelAdmin):
+    change_list_template = "workbasket_change_list.html"
+
+    def get_urls(self):
+        urls = super().get_urls()
+        upload_url = path("upload/", self.upload)
+
+        return [upload_url] + urls
```
simonwo, line 18: Is this something that should be handled by `urls.py`? (Genuine question)

workbaskets/admin.py
```diff
@@ -0,0 +1,32 @@
...
+    def upload(self, request):
+        if not request.user.is_superuser:
+            return HttpResponseForbidden("Only superusers may upload workbaskets")
+
+        upload_workbaskets.delay()
+        self.message_user(
+            request,
+            f"Uploading workbaskets with status of '{WorkflowStatus.READY_FOR_EXPORT.label}'",
+        )
+        return HttpResponseRedirect("../")
```
simonwo, line 29: Would it be better to [explicitly request a URL](https://docs.djangoproject.com/en/3.2/ref/urlresolvers/#django.urls.reverse) using the name e.g. `workbasket-ui-list`?

workbaskets/templates/workbasket_change_list.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'admin/change_list.html' %}
+
+{% block object-tools %}
+    <div>
+        <form action="upload/" method="POST">
+            {% csrf_token %}
+                <button type="submit" style="cursor:pointer">Export workbaskets</button>
```
simonwo, line 7: It looks like if you use an `<a>` tag or add a `class="button"` then the button will get a consistent look and feel with the rest of the admin site.

workbaskets/templates/workbasket_change_list.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'admin/change_list.html' %}
+
+{% block object-tools %}
+    <div>
+        <form action="upload/" method="POST">
```
simonwo, line 5: Could you use the `{% url '…' %}` tag here instead of writing it explicitly?

simonwo (commented):
Putting the button in Django admin is a wise move I think! I wonder if that has any implications for where you put your the state movement stuff in your ticket @andyhd? 

The only thing I can see is that we probably want to use more of the [Django URL support](https://docs.djangoproject.com/en/3.2/topics/http/urls/#reverse-resolution-of-urls) and not hard-code URLs everywhere.

andyhd:
@simonwo I was already adding a WorkBasket admin view, so now I'm waiting for this PR to merge, otherwise we'll have conflicts

stuaxo:
Probably out of scope for this ticket: We should probably be tracking the upload tasks somewhere in the admin.

It's fine that we schedule new upload_workbaskets tasks - if there are none to upload it will do nothing.

It would be nice to see if there already was one and show that.

But again:  not really needed for this.

workbaskets/admin.py
stuaxo: Not a blocker:

A more accurate message might be along the lines of "Scheduled workbasket upload"   as opposed to "Uploading..." although celery will usually do the upload right away, there are some edge cases where it might not, and "Scheduled" provides a clue here.

workbaskets/tests/test_admin.py
```diff
@@ -1,40 +1,44 @@
...
 
 import pytest
-
-from common.tests.factories import UserFactory
+from django.urls import reverse
 
 pytestmark = pytest.mark.django_db
 
 
 class TestWorkBasketUpload:
```
stuaxo, line 9: Somehow I only noticed that this is a test using a class... is there a reason we are implementing this differently to the rest of the tests ?

workbaskets/tests/test_admin.py
```diff
@@ -1,40 +1,44 @@
...
-        valid_user.is_superuser = True
-        valid_user.save()
-        client.force_login(valid_user)
+    def test_upload_returns_302_for_valid_staff_user(self, staff_user, client):
+        client.force_login(staff_user)
         response = client.post(self.url)
 
         assert response.status_code == 302
-        assert response.url == "../"
+        assert response.url == reverse("admin:workbaskets_workbasket_changelist")
```
stuaxo, line 19: Not a blocker: usually I try avoid calls or logic the assert itself to try and stick more to the AAA (arrange-act-assert) test pattern (on the other hand, it's pretty clear what this does).

workbaskets/admin.py
andyhd: No, I don’t think so

andyhd (approved):

PR #377, opened 2021-09-08
Tp 794 install google tag manager in to service

gabelton:
## Why

This change is required, in order to support go-live for the service

## What

- In _generic.jinja conditionally add Google Tag Manager script into <head> and noscript into <body> on the basis of whether GOOGLE_ANALYTICS_ID environment variable has been set.
- In settings/common.py add GOOGLE_ANALYTICS_ID environment variable. This should be set in AWS EBS for production only.

## Checklist
- Needs GOOGLE_ANALYTICS_ID to be set as environment variable in AWS EBS prod before deployment.

(Original ticket https://uktrade.atlassian.net/browse/TP-794)

### DIFF
--- a/common/jinja2/layouts/_generic.jinja
+++ b/common/jinja2/layouts/_generic.jinja
@@ -31,7 +31,7 @@
   {% if settings.GOOGLE_ANALYTICS_ID %}
     <!-- Google Tag Manager (noscript) -->
       <noscript>
-        <iframe src="https://www.googletagmanager.com/ns.html?id="{{ settings.GOOGLE_ANALYTICS_ID }}""height="0" width="0" style="display:none;visibility:hidden"></iframe>
+        <iframe src="https://www.googletagmanager.com/ns.html?id={{ settings.GOOGLE_ANALYTICS_ID }}"height="0" width="0" style="display:none;visibility:hidden"></iframe>
       </noscript>
     <!-- End Google Tag Manager (noscript) -->
   {% endif %}
--- a/common/jinja2/layouts/_generic.jinja
+++ b/common/jinja2/layouts/_generic.jinja
@@ -8,7 +8,7 @@
         new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
         j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
         'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-        })(window,document,'script','dataLayer', {{settings.GOOGLE_ANALYTICS_ID}});
+        })(window,document,'script','dataLayer', "{{ settings.GOOGLE_ANALYTICS_ID }}");
       </script>
     <!-- End Google Tag Manager -->
   {% endif %}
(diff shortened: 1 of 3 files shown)


common/jinja2/layouts/_generic.jinja
```diff
@@ -1,6 +1,18 @@
...
 
 {% block head %}
+  {% if settings.GOOGLE_ANALYTICS_ID %}
+    <!-- Google Tag Manager -->
+      <script>
+        (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+        new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+        j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+        'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+        })(window,document,'script','dataLayer', {{settings.GOOGLE_ANALYTICS_ID}});
```
simonwo, line 11: Does the `{{ … }}` need to have string quotes around it? Or does Jinja know to do that automatically?

gabelton: I think Jinja does it automatically. After adding this I had a play around in my local and I was able to pass through a value.

stuaxo:
Make sure to add any new environment variables note in the environment variables section in the README.md.

settings/common.py
```diff
@@ -346,6 +346,9 @@
     },
 }
 
+# -- Google Tag Manager
```
stuaxo, line 349: Make sure to document this in the README as well.

If we have a page on our RTFM docs, it's worth making a link to it, from here or from the README.

stuaxo (approved):

common/jinja2/layouts/_generic.jinja
stuaxo: To make the indentation consistent, I think it should be `{{ settings.GOOGLE_ANALYTICS_ID }}`

andyhd: I think it should have quotes. I'm pretty sure Jinja does not do it automatically.
```python
>>> jinja2.Template("""<script>var some_var = {{value}};</script>""").render({"value": "a_string"})
'<script>var some_var = a_string;</script>'
```

settings/common.py
```diff
@@ -346,6 +346,9 @@
...
+GOOGLE_ANALYTICS_ID = os.environ.get("GOOGLE_ANALYTICS_ID", None)
```
andyhd, line 350: `None` is the default default value, so you can just do `.get("GOOGLE_ANALYTICS_ID")`

Obviously doesn't need fixing, I just wanted to write "default default" 😄

andyhd (changes requested):
I think @simonwo was right

common/jinja2/layouts/_generic.jinja
andyhd: I just thought - maybe check whether you have quotes in the `settings.GOOGLE_ANALYTICS_ID` string value (eg: your env var is `GOOGLE_ANALYTICS_ID="'some-id-string'"`)

common/jinja2/layouts/_generic.jinja
```diff
@@ -31,7 +31,7 @@
   {% if settings.GOOGLE_ANALYTICS_ID %}
     <!-- Google Tag Manager (noscript) -->
       <noscript>
-        <iframe src="https://www.googletagmanager.com/ns.html?id={{settings.GOOGLE_ANALYTICS_ID}}"height="0" width="0" style="display:none;visibility:hidden"></iframe>
+        <iframe src="https://www.googletagmanager.com/ns.html?id="{{ settings.GOOGLE_ANALYTICS_ID }}""height="0" width="0" style="display:none;visibility:hidden"></iframe>
```
simonwo, line 34: Nah, not this one… the quotes will end up ending the `href` element. It should have just been as it was in the previous version.

common/jinja2/layouts/_generic.jinja
gabelton: The env var in Vault doesn't have quotes around it

andyhd (approved):

PR #380, opened 2021-09-10
Tp 949 get returned more than one workbasket

gabelton:
## Why
Same error as https://github.com/uktrade/tamato/pull/378

## What
Replaces get() with last() when getting unapproved Workbaskets in require_current_workbasket decorator

### DIFF
--- a/common/tests/test_views.py
+++ b/common/tests/test_views.py
@@ -3,8 +3,10 @@ import xml.etree.ElementTree as etree
 import pytest
 from django.urls import reverse
 
+from common.tests import factories
 from common.views import HealthCheckResponse
 from workbaskets.models import WorkBasket
+from workbaskets.validators import WorkflowStatus
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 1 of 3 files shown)


common/tests/test_views.py
```diff
@@ -40,3 +42,18 @@ def test_healthcheck_response(response, status_code, status):
...
+
+
+def test_handles_multiple_unapproved_workbaskets(valid_user_client, new_workbasket):
+    workbasket = factories.WorkBasketFactory.create(
+        status=WorkflowStatus.NEW_IN_PROGRESS,
+    )
+    transaction = factories.TransactionFactory.create(workbasket=workbasket)
+    with transaction:
+        for _ in range(2):
+            factories.FootnoteTypeFactory.create()
```
stuaxo, line 54: You could probably use `factories.FootnoteTypeFactory.create_batch(2)` instead of an explicit loop here.

stuaxo (approved):

PR #381, opened 2021-09-15
TP-904: Import measures from a spreadsheet

simonwo:
## Why

In order to handle a large number of late changes to the tariff that can only be made when the EU have made and published all of their changes to commodity codes, our users need the ability to prepare a bulk set of changes in advance and then upload them all at once. The tactical way to achieve this is to allow them to prepare a spreadsheet offline and then upload it once the commodity codes are present.

## What

This PR introduces a prototype importer that can import measures from a spreadsheet. The structure of the spreadsheet is rigid and has to conform exactly to the spec. For each row in the spreadsheet, one measure is created.

So far this is just the function to run an import, and doesn't include a command line or UI interface because in the basic case we expect this to be run from a notebook by an engineer. A future ticket will add a simple user interface for this.

The implementation is simple but not particularly sophisticated. No attempt has been made to unify this with the XML importer because too much would have to change to make that work, in addition to including nearly everything that is here, but there are clearly some elements which look the same and it woulbe good to reuse that logic. That might be a good place to examine for future work. What we do have is lots of tests, so this should be useful for future iterations.

## Checklist
- Requires migrations?: No
- Requires dependency updates?: **Yes**

### DIFF
--- a/common/tests/factories.py
+++ b/common/tests/factories.py
@@ -845,6 +845,28 @@ class MeasureFactory(TrackedModelMixin, ValidityFactoryMixin):
     stopped = False
     export_refund_nomenclature_sid = None
 
+    class Params:
+        with_footnote = factory.Trait(
+            association=factory.RelatedFactory(
+                "common.tests.factories.FootnoteAssociationMeasureFactory",
+                factory_related_name="footnoted_measure",
+            ),
+        )
+
+        with_exclusion = factory.Trait(
+            exclusion=factory.RelatedFactory(
+                "common.tests.factories.MeasureExcludedGeographicalAreaFactory",
+                factory_related_name="modified_measure",
+            ),
+        )
+
+        with_condition = factory.Trait(
+            condition=factory.RelatedFactory(
+                "common.tests.factories.MeasureConditionFactory",
+                factory_related_name="dependent_measure",
+            ),
+        )
+
     @factory.lazy_attribute
     def terminating_regulation(self):
         if self.valid_between.upper is None:
--- a/common/tests/util.py
+++ b/common/tests/util.py
@@ -7,6 +7,7 @@ from io import BytesIO
 from itertools import count
 from typing import Any
 from typing import Dict
+from typing import Sequence
 
 import pytest
 from dateutil.parser import parse as parse_date
@@ -18,6 +19,7 @@ from django.urls import reverse
 from freezegun import freeze_time
 from lxml import etree
 
+from common.models.records import TrackedModel
 from common.renderers import counter_generator
 from common.serializers import validate_taric_xml_record_order
 from common.util import TaricDateRange
@@ -116,6 +118,51 @@ def make_non_duplicate_record(factory, identifying_fields=None):
     )
 
     return not_duplicate
+
+
+def get_checkable_data(model: TrackedModel, ignore=frozenset()):
+    """Returns a dict representing the model's data ignoring any automatically
+    set fields and fields with names passed to `ignore`."""
+    checked_field_names = {f.name for f in model.copyable_fields} - ignore
+    return {
+        name: model._meta.get_field(name).value_from_object(model)
+        for name in checked_field_names
+    }
+
+
+def assert_records_match(
+    expected: TrackedModel,
+    imported: TrackedModel,
+    ignore=frozenset(),
+):
+    """
+    Asserts that every value for every field in the imported model is the same
+    as the data in the expected model.
+
+    System fields that will change from model to model are not checked. Any
+    field names given to `ignore` will also not be checked.
+    """
+    assert get_checkable_data(expected, ignore=ignore) == get_checkable_data(
+        imported,
+        ignore=ignore,
+    )
+
+
+def assert_many_records_match(
+    expected: Sequence[TrackedModel],
+    imported: Sequence[TrackedModel],
+    ignore=frozenset(),
+):
+    """
+    Asserts that every value for every field in the imported models is the same
+    as the data in the expected models, and that the count of both is equal.
+
+    System fields that will change from model to model are not checked. Any
+    field names given to `ignore` will also not be checked.
+    """
+    assert [get_checkable_data(e, ignore=ignore) for e in expected] == [
+        get_checkable_data(i, ignore=ignore) for i in imported
+    ]
 
 
 _transaction_counter = count(start=1)
--- a/importer/utils.py
+++ b/importer/utils.py
@@ -14,6 +14,19 @@ from django.db.models.query_utils import DeferredAttribute
 from common.models import TrackedModel
 
 
+def col(label: str) -> int:
+    """Return the correct index given an Excel column letter."""
+    if not label.isalpha():
+        raise ValueError(f"'{label}' is not a valid column label.")
+    return (
+        sum(
+            (ord(char) - ord("A") + 1) * (26 ** position)
+            for position, char in enumerate(reversed(label.upper()))
+        )
+        - 1
+    )
+
+
 class LinksType(TypedDict):
     """
     A type defining the interface between a handler and its links.
--- a/measures/patterns.py
+++ b/measures/patterns.py
@@ -9,7 +9,6 @@ from typing import Sequence
 
 from django.db import transaction
 
-from additional_codes.models import AdditionalCode
 from certificates.models import Certificate
 from certificates.models import CertificateType
 from commodities.models import GoodsNomenclature
@@ -85,6 +84,10 @@ class MeasureCreationPattern:
         last_sid = MeasureCondition.objects.values("sid").order_by("sid").last()
         next_sid = 1 if last_sid is None else last_sid["sid"] + 1
         return counter_generator(next_sid)
+
+    @cached_property
+    def authorised_use_measure_types(self):
+        return set(MeasureType.objects.filter(description__contains="authorised use"))
 
     @cached_property
     def presentation_of_certificate(self) -> MeasureConditionCode:
@@ -267,17 +270,14 @@ class MeasureCreationPattern:
     def create_measure_tracked_models(
         self,
         duty_sentence: str,
-        geographical_area: GeographicalArea,
         goods_nomenclature: GoodsNomenclature,
-        measure_type: MeasureType,
         validity_start: date,
         validity_end: date,
         exclusions: Sequence[GeographicalArea] = [],
         order_number: Optional[QuotaOrderNumber] = None,
-        authorised_use: bool = False,
-        additional_code: AdditionalCode = None,
         footnotes: Sequence[Footnote] = [],
         condition_sentence: Optional[str] = None,
+        **data,
     ) -> Iterator[TrackedModel]:
         """
         Create a new measure linking the passed data and any defaults. The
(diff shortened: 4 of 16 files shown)


common/tests/util.py
```diff
@@ -118,6 +120,50 @@ def make_non_duplicate_record(factory, identifying_fields=None):
...
+        source_value = getattr(expected, field)
+        assert (
+            imported_value == source_value
+        ), f"imported '{field}' ({imported_value} - {type(imported_value)}) does not match source '{field}' ({source_value} - {type(source_value)})"
+
+
+def assert_many_records_match(
+    expected: Sequence[TrackedModel],
+    imported: Sequence[TrackedModel],
+    ignore=set(),
```
stuaxo, line 149: Annoyingly, python will persist classes (like set() here) called in the constructor, which isn't usually what you want.

common/tests/util.py
```diff
@@ -118,6 +120,50 @@ def make_non_duplicate_record(factory, identifying_fields=None):
...
+    System fields that will change from model to model are not checked. Any
+    field names given to `ignore` will also not be checked.
+    """
+    assert len(imported) == len(expected)
+    for model, expected in zip(imported, expected):
+        assert_records_match(
+            expected,
+            model,
+            ignore=ignore,
+        )
```
stuaxo, line 164: It would be good if we could use assertQuerysetEqual, or assertCountEqual.

These have good error reporting showing where things in differ over the whole queryset / list.

Here's a toy implementation (much can be improved):

```python
from pytest_django.asserts import assertQuerysetEqual
 
def build_get_checked_fields(ignore):
    def get_checked_fields(instance):
        field_names = (set(field.name for field in instance._meta.fields) - set(field.name for field in TrackedModel._meta.fields) - {"trackedmodel_ptr"}  - ignore)
        return {name: getattr(instance, name) for name in field_names}
    return get_checked_fields
    
get_checked_fields = build_get_checked_fields(ignore)

assertQuerysetEqual(imported, [*map(get_checked_fields, expected)], transform=get_checked_fields)
```

example output:

```
ipdb> assertQuerysetEqual(imported, [*map(get_checked_fields, _expected)], transform=get_checked_fields)
*** AssertionError: Lists differ: [{'stopped': False, 'dead_additional_code': None, 'mea[383 chars]one}] != [{'stopped': True, 'dead_additional_code': None, 'meas[382 chars]one}]

First differing element 0:
{'stopped': False, 'dead_additional_code': None, 'mea[382 chars]None}
{'stopped': True, 'dead_additional_code': None, 'meas[381 chars]None}

  [{'additional_code': None,
    'dead_additional_code': None,
    'dead_order_number': None,
    'generating_regulation': <Regulation: R2100000>,
    'geographical_area': <GeographicalArea: Region AAAA>,
    'goods_nomenclature': <GoodsNomenclature: 0000000000>,
    'measure_type': <MeasureType: 000>,
    'order_number': None,
    'reduction': None,
-   'stopped': False,
?              ^^^^

+   'stopped': True,
?              ^^^

    'terminating_regulation': None,
    'valid_between': TaricDateRange(datetime.date(2021, 9, 16), None, '[)')}]
Traceback (most recent call last):
  File "/home/stu/.virtualenvs/dit-pyston/lib/pyston3.8/site-packages/pytest_django/asserts.py", line 18, in assertion_func
    return func(*args, **kwargs)
  File "/home/stu/.virtualenvs/dit-pyston/lib/pyston3.8/site-packages/django/test/testcases.py", line 1052, in assertQuerysetEqual
    return self.assertEqual(list(items), values, msg=msg)
  File "/usr/lib/pyston3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/pyston3.8/unittest/case.py", line 1118, in assertListEqual
    self.assertSequenceEqual(list1, list2, msg, seq_type=list)
  File "/usr/lib/pyston3.8/unittest/case.py", line 1100, in assertSequenceEqual
    self.fail(msg)
  File "/usr/lib/pyston3.8/unittest/case.py", line 753, in fail
    raise self.failureException(msg)
```

Notes:  I probably would have used assertCountEqual here since these are not querysets, but it's not as easily available in pytest... assertQuerysetEqual calls down into it anyway.

common/tests/util.py
simonwo: I thought sets were immutable, but they aren't – I should be using `frozenset`.
```suggestion
    ignore=frozenset(),
```

stuaxo: Nice... they should have frozen versions of everything... since python is gradually turning into Perl5, maybe they will come with some weird new syntax for it...  `ignore=⛄set()`

simonwo: 🤢

common/tests/util.py
simonwo: Turns out that it is indeed quite hard to import assertCountEqual but apparently you don't really need it as the normal pytest `assert […] == […]` will also ensure that. So I've updated it.

importer/utils.py
```diff
@@ -14,6 +14,15 @@
 from common.models import TrackedModel
 
 
+def col(label: str) -> int:
+    """Return the correct index given an Excel column letter."""
```
stuaxo, line 18: It might be worth checking `str.alpha()`  here instead of the assertion later, as it will catch the user passing in an empty string, which currently returns -1.

common/tests/factories.py
```diff
@@ -845,6 +845,28 @@ class Meta:
...
+                factory_related_name="modified_measure",
+            ),
+        )
+
+        with_condition = factory.Trait(
+            condition=factory.RelatedFactory(
+                "common.tests.factories.MeasureConditionFactory",
+                factory_related_name="dependent_measure",
+            ),
+        )
```
stuaxo, line 868: This was good, got to go and learn about Trait in factoryboy :)

measures/patterns.py
```diff
@@ -86,6 +85,10 @@ def measure_condition_sid_counter(self) -> Counter:
         next_sid = 1 if last_sid is None else last_sid["sid"] + 1
         return counter_generator(next_sid)
 
+    @cached_property
+    def authorised_use_measure_types(self):
+        return set(MeasureType.objects.filter(description__contains="authorised use"))
```
kintisheff, line 90: is this robust enough or should we just define the AU-measure types explicitly (e.g. by `sid`) as they don't change a lot, or even consider a data migration that applies an AU-flag?

kintisheff (commented):
There's a general question (which may be the concern of a separate ticket) regarding how to inform the providers of the excel sheet input when some of their data fails to write - is it worth producing a report that lists import status by row (or another identifier with more context) and 
in the case of failure, the offending column values?

measures/sheet_importers.py
```diff
@@ -0,0 +1,203 @@
...
+from quotas.models import QuotaOrderNumber
+from regulations.models import Regulation
+from workbaskets.models import WorkBasket
+
+
+class MeasureSheetRow(SheetRowMixin):
+    """An importer that will parse values from a human-readable spreadsheet of
+    measures and will create one measure for each row that it finds."""
+
+    @column("A")
```
kintisheff, line 32: Is there an opportunity to make this more configurable rather than hard-coding the column arrangements and the associated parsers/handlers? E.g. we could define both the cell value parsers and the cached properties outside of this class, and in this class we just associate column numbers with the parsers and handlers (still using the `column` decorator). So in the future, the column arrangement here becomes configurable rather than hard-coded, but perhaps even better, if we want to process a different template that shares a number of these handlers, we can define the respective configuration class very quickly?

requirements.txt
```diff
@@ -33,6 +33,7 @@ gunicorn==20.0.4
 Jinja2==2.11.3
 lxml==4.6.3
 moto==2.1.0
+openpyxl==3.0.7
```
kintisheff, line 36: perhaps we can set the 'Requires dependency updates' flag on the PR to 'Yes'

measures/sheet_importers.py
```diff
@@ -0,0 +1,203 @@
...
+            desc.described_footnote
+            for desc in FootnoteDescription.objects.latest_approved()
+            .with_end_date()
+            .as_at(self.validity_start_date)
+            .filter(q)
+        ]
+
+    @column("L", optional=True)
+    def conditions(self, value: CellValue):
+        return str(value)
```
kintisheff, line 181: Is there room for a sanity check on the condition sentence here? I keep thinking of this slightly grey area where we know, for example, which action codes should go together for example, among other sane input criteria. I don't think the `ConditionSentenceParser` in `measures.parsers` deals with that kind of problem. I'm not sure whether we should be dealing with it here either, or rather just let providers of the input apply corrections post-factum where any mistakes are made?

measures/patterns.py
simonwo: I'm not sure… ultimately both of these soultions are trying to make this code more robust in the face of a database that isn't the UK tariff. At the moment we only have one case where we ever deal with that, and it is in using test data. So really the question is, do you want to denote authorised use in test data by the presence of `MeasureTypeFactory(sid=105)` or `MeasureTypeFactory(description="Test authorised use measure type")`? I thought the latter was preferable as it was at least clear what was happening, as opposed to having a magic number 105.

I agree that a better option would be to have a flag. I didn't want to chip off that work in this ticket because I'm not convinced a flag on the measure type is correct… it certainly might be. But really this authorised use thing is what I call a "pattern" e.g. with these specific data points, apply this special logic when making changes. There are others e.g. origin quotas – with these specific quotas, apply special conditions logic. So there might also be something more complex like a "pattern configuration" table that is more general. Basically, I agree, but didn't want to make that refactor yet. I can revisit if you think the current solution is unacceptable.

measures/sheet_importers.py
simonwo: Yes! Turning the column labels into something which can be mapped to and then maybe even exposing that mapping capability in the UI is a natural next step. But I see it as out of scope for this ticket and the subject of future work.

measures/sheet_importers.py
simonwo: I could certianly run this input through the `ConditionSentenceParser` at this stage to weed out any errors… but I'm not sure if it's worth it because it will be parsed in the almost immediately next step anyway?

Yeah, the parser doesn't have any opinions on what codes go together. Again, that is an example of a "pattern" that might change over time rather than being a hard rule, so the hard question of "how do we let users configure that" or "how do we provide a warning rather than an error" will need to be solved.

simonwo:
> There's a general question (which may be the concern of a separate ticket) regarding how to inform the providers of the excel sheet input when some of their data fails to write - is it worth producing a report that lists import status by row (or another identifier with more context) and in the case of failure, the offending column values?

Agreed, we should definitely do something here, ideally outputting rows that fail validation and business rules so that the users can take them away and amend them. I didn't want to do anything on that yet until we have more of a UI, because for now I see this as being run in a notebook where it would not be hard to add this (e.g. `try: parse(row); except: print(row)`). I'll create a ticket for this though.

measures/patterns.py
kintisheff: no I think this makes sense; that's pethaps scope for a future piece of work really

measures/sheet_importers.py
kintisheff: sounds like a separate scope again then

kintisheff (approved):
Since reporting will be a separate ticket, this one looks good to me otherwise.

simonwo:
I have included the outstanding things in TP-878.

PR #382, opened 2021-09-21
TP-928: Add UI to transition WorkBasket status

andyhd:
# TP-928: Add UI to transition WorkBasket status

## Why

As part of getting the policy team to use the Editor, we need a way to both:
 1. Control when their workbaskets are serialised to envelopes
 2. Update the system when envelopes are accepted by CDS

To do this we need to add a button to the as-yet secret workbasket UI to manually transition workbaskets through the workflow. This does not need to bypass the normal validation and routines.

So e.g. add a button to the workbasket details view that allows moving the workbasket to the next state in the workflow. For states where the “next state” can be multiple, show multiple buttons, one for each state.

In ADR 0008 we decided to reduce the number of workflow states we used to keep the app simple.

## What

Screenshot 2021-09-21 at 15 47 53

* Add custom admin form with "transition" field to enable transitioning a WorkBasket
  from one status to another
* Refactor workbasket status FSM as specified in [ADR-008](docs/adr/0008-simplify-workflow-states.rst)
* Refactor workbasket tests
* Migrate database to new status values
* Refactored WorkBasket session storage, because it was unnecessarily complicated
* Small tweaks to some XML tests to use XPath, because they did not work on my machine

## Checklist

See: [TP-289](https://uktrade.atlassian.net/browse/TP-289)


workbaskets/tests/test_admin.py
```diff
@@ -1,45 +1,109 @@
...
 
 
-def test_upload_returns_302_for_valid_staff_user(staff_user, client):
-    client.force_login(staff_user)
-    with mock.patch(
-        "exporter.tasks.upload_workbaskets.delay",
-    ):
-        response = client.post(reverse("admin:upload"))
+@pytest.fixture
+def celery_task():
```
stuaxo, line 14: Not a blocker:  

Maybe this could be `mock_celery_task`:  I can imagine a situation where you want a real celery task, from a fixture - if `settings.task_always_eager = True` is set, I believe this won't go via the workers etc.

workbaskets/tests/test_admin.py
```diff
@@ -1,45 +1,109 @@
...
+        "staff_user",
+        "valid_user",
+    ],
+)
+def test_upload(client, request, celery_task, user):
+    upload_url = reverse("admin:upload")
 
-        assert response.status_code == 302
+    client.force_login(request.getfixturevalue(user))
+    response = client.post(upload_url)
```
stuaxo, line 30: Not a blocker:

It can be useful to check the `response.status_code` before going further: while it does break AAA, it's basically zero-cost which can save confusion if something changes (breaks) in the view in the future.

workbaskets/models.py
```diff
@@ -103,41 +102,33 @@ def __str__(self):
...
     @transition(
         field=status,
-        source=WorkflowStatus.AWAITING_APPROVAL,
-        target=WorkflowStatus.READY_FOR_EXPORT,
-        permission="workbaskets.can_approve",
+        source=WorkflowStatus.PROPOSED,
+        target=WorkflowStatus.APPROVED,
+        # conditions=[has_approver],
+        # permission="workbaskets.can_approve",
+        custom={"label": "Approve"},
```
stuaxo, line 131: ^ some comments which may or may not need to be here.

footnotes/tests/test_xml.py
```diff
@@ -9,21 +9,16 @@
...
 
 
 @validate_taric_xml(factories.FootnoteDescriptionFactory)
 def test_footnote_description_xml(xml):
-    element = xml.find(".//oub:footnote.description.period", nsmap)
-    assert element is not None
-    element = xml.find(".//oub:footnote.description", nsmap)
-    assert element is not None
+    assert xml.xpath(".//oub:footnote.description.period", namespaces=nsmap)
+    assert xml.xpath(".//oub:footnote.description", namespaces=nsmap)
```
stuaxo, line 24: ^ Nice :)

common/tests/util.py
```diff
@@ -186,13 +186,12 @@ def generate_test_import_xml(obj: dict) -> BytesIO:
...
-            record.findtext(".//subrecord.code", namespaces=xml.nsmap),
-        )
-        for record in xml.findall(".//record", namespaces=xml.nsmap)
-    ]
+    records = xml.xpath(".//*[local-name() = 'record']")
+    codes = etree.XPath(
+        ".//*[local-name()='record.code' or local-name()='subrecord.code']/text()",
+    )
+
+    return [tuple(codes(record)) for record in records]
```
stuaxo, line 194: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark:

stuaxo (approved):

simonwo:
LGTM!

workbaskets/models.py
andyhd: I think we probably do need to require permission to perform (some) transitions, but maybe for a later PR

PR #383, opened 2021-09-21
TP-907

paulpepper-trade:
# TP-907 Optional footnotes and conditions

## Why
It should be possible to create a measure without any conditions or footnotes. The UI does not permit that at the moment and so this PR enables that capability.

## What
Add support in forms to allow empty conditions and footnotes to validate correctly.
Add support in Jinja2 templates to correctly display initial, single form in formsets of conditions and footnotes.
Correctly display empty conditions and footnotes on create measures summary page.
As per new design instructions, add space between the last form in the formset and the “Add new” button to convey to the user that it is not directly associated with the final form.

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -337,7 +337,7 @@ class AddAnother(forms.BaseFormSet):
     preserved.
     """
 
-    extra = 1
+    extra = 0
     can_order = False
     can_delete = True
     max_num = 1000
@@ -414,13 +414,18 @@ class AddAnother(forms.BaseFormSet):
         to redisplay the formset with an extra empty form or the selected form
         removed."""
 
+        # An empty set of forms is valid.
+        if not self.total_form_count():
+            return True
+
         # reshow the form with an extra empty form if "Add another" was submitted
         if f"{self.prefix}-ADD" in self.data:
             return False
 
         # reshow the form with the deleted form(s) removed if "Delete" was submitted
-        if any(field for field in self.data if field.endswith("-DELETE")):
-            return False
+        for field in self.data:
+            if field.endswith("-DELETE"):
+                return False
 
         return super().is_valid()
 
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -337,6 +337,10 @@ class AddAnother(forms.BaseFormSet):
     preserved.
     """
 
+    ACTION_ADD = "ADD"
+    ACTION_DELETE = "DELETE"
+    FORMSET_ACTIONS = (ACTION_ADD, ACTION_DELETE)
+
     extra = 0
     can_order = False
     can_delete = True
@@ -348,6 +352,17 @@ class AddAnother(forms.BaseFormSet):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+        # If we have form data, then capture the any user "add form" or
+        # "delete form" actions.
+        self.formset_action = None
+        if f"{self.prefix}-ADD" in self.data:
+            self.formset_action = AddAnother.ACTION_ADD
+        else:
+            for field in self.data:
+                if field.endswith("-DELETE"):
+                    self.formset_action = AddAnother.ACTION_DELETE
+                    break
 
         data = self.data.copy()
 
@@ -414,18 +429,14 @@ class AddAnother(forms.BaseFormSet):
         to redisplay the formset with an extra empty form or the selected form
         removed."""
 
+        # Re-present the form to show the result of adding another form or
+        # deleting an existing one.
+        if self.formset_action in (AddAnother.FORMSET_ACTIONS):
+            return False
+
         # An empty set of forms is valid.
         if not self.total_form_count():
             return True
-
-        # reshow the form with an extra empty form if "Add another" was submitted
-        if f"{self.prefix}-ADD" in self.data:
-            return False
-
-        # reshow the form with the deleted form(s) removed if "Delete" was submitted
-        for field in self.data:
-            if field.endswith("-DELETE"):
-                return False
 
         return super().is_valid()
 
--- a/measures/jinja2/measures/create-review.jinja
+++ b/measures/jinja2/measures/create-review.jinja
@@ -5,29 +5,49 @@
 {% set page_title = step_metadata[wizard.steps.current].title %}
 
 {% macro review_step(step) %}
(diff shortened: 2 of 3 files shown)


simonwo:
Looks good! Could you add a screenshot of what it looks like as an empty form?

paulpepper-trade:
> Looks good! Could you add a screenshot of what it looks like as an empty form pls?

I've just added screenshots to the "What" section of the body of the PR text, Simon.

gabelton:
Looks good to me - I don't know what the remainder of the sprint is looking like for you, but I've added some test fixtures and view tests on [my branch](https://github.com/uktrade/tamato/pull/384/files) now, if you wanted to adapt them and add some to this PR.

simonwo:
> 
> 
> > Looks good! Could you add a screenshot of what it looks like as an empty form pls?
> 
> I've just added screenshots to the "What" section of the body of the PR text, Simon.

Thanks! I wonder if @kathrynlangley thinks we need some content explaining that footnotes/conditions are optional? E.g. atm it says "add one or more" but really it should be "add zero or more" or "optionally add"?

measures/forms.py
```diff
@@ -337,7 +337,11 @@ class AddAnother(forms.BaseFormSet):
     preserved.
     """
 
-    extra = 1
+    ACTION_ADD = "ADD"
+    ACTION_DELETE = "DELETE"
```
andyhd, line 341: Minor issue: I'm not sure about `ACTION_ADD` and `ACTION_DELETE` - this makes the code more verbose without much benefit. For example, `self.formset_action = "ADD"` is straightforward, and `self.formset_action = AddAnother.ACTION_ADD` is perhaps a bit clunky.

measures/forms.py
```diff
@@ -349,6 +353,17 @@ class AddAnother(forms.BaseFormSet):
...
         super().__init__(*args, **kwargs)
 
+        # If we have form data, then capture the any user "add form" or
+        # "delete form" actions.
+        self.formset_action = None
+        if f"{self.prefix}-ADD" in self.data:
+            self.formset_action = AddAnother.ACTION_ADD
+        else:
+            for field in self.data:
+                if field.endswith("-DELETE"):
```
andyhd, line 363: Should use the constant here too

measures/forms.py
```diff
@@ -414,13 +429,14 @@ def is_valid(self):
         to redisplay the formset with an extra empty form or the selected form
         removed."""
 
-        # reshow the form with an extra empty form if "Add another" was submitted
-        if f"{self.prefix}-ADD" in self.data:
+        # Re-present the form to show the result of adding another form or
+        # deleting an existing one.
+        if self.formset_action in (AddAnother.FORMSET_ACTIONS):
```
andyhd, line 434: Unnecessary parentheses here

measures/forms.py
```diff
@@ -414,13 +429,14 @@ def is_valid(self):
...
             return False
 
-        # reshow the form with the deleted form(s) removed if "Delete" was submitted
-        if any(field for field in self.data if field.endswith("-DELETE")):
-            return False
+        # An empty set of forms is valid.
+        if not self.total_form_count():
```
andyhd, line 438: `total_form_count` always returns an integer, so it's probably clearer to say `if self.total_form_count() == 0:`

measures/jinja2/measures/create-review.jinja
```diff
@@ -51,23 +71,29 @@
...
 
   {% call(rows, data) review_step("conditions") %}
-    {% for item in data  %}
-      {% if not item.DELETE %}
-        {{ rows.append(("Condition code", item.condition_code)) }}
-      {% endif %}
-    {% endfor %}
+    {# Conditions are optional, so cater for empty data. #}
+    {% if data %}
+      {% for item in data  %}
```
andyhd, line 76: If there are no conditions, is data `None`? I would expect it to be an empty iterable, so `for item in data` will do nothing anyway.

andyhd (approved):
LGTM

measures/forms.py
paulpepper-trade: Similar to Django's Choices; provides up-front self-documenting vars and a single reference for checks.

measures/forms.py
```diff
@@ -349,6 +353,17 @@ class AddAnother(forms.BaseFormSet):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+        # If we have form data, then capture the any user "add form" or
+        # "delete form" actions.
+        self.formset_action = None
+        if f"{self.prefix}-ADD" in self.data:
```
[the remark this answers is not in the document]
paulpepper-trade: Done

measures/forms.py
paulpepper-trade: Done

measures/forms.py
paulpepper-trade: Done

measures/forms.py
paulpepper-trade: A truthy test is more robust and no ambiguity is introduced, but I don't think there's much in it. Done :-)

measures/jinja2/measures/create-review.jinja
paulpepper-trade: If data is `None` then the `for` statement blows up, hence the `if` check!

andyhd: Checking the django-formtools source code, `data` can be `None` if the form doesn't validate.

measures/forms.py
andyhd: I just find the `AddAnother.ACTION_ADD` variable name a bit inelegant - the `AddAnother` class could have a better name (perhaps `NestedForm`), and perhaps these string constants could actually be an Enum. If the enum class was an inner class (eg: `NestedForm.Action`) you could then do `self.formset_action = self.Action.ADD` and `if isinstance(self.formset_action, self.Action):` which IMO clearly reveal the intent of the code.

But of course this is just a style quibble, so feel free to ignore.

paulpepper-trade: `AddAnother` renamed to `FormSet`, which is what it is :)

I liked the idea of Enums in place of the class consts, but I've reverted back to the string literals `ADD` and `DELETE`. The intention had been to better document form set state and not provide part-values for use in button name.

PR #384, opened 2021-09-22
Tp 888 extend edit measures to allow editing footnotes

gabelton:
[TP-888](https://uktrade.atlassian.net/browse/TP-888)

## Why
Currently the "edit measures" page does not allow for adding or removing footnotes on an existing measure. We have upcoming TOPS tickets that require this capability.

## What
- Installs 3rd party library, [ajax-selects](https://github.com/crucialfelix/django-ajax-selects), and adds its AutoCompleteSelectMultipleField to MeasureForm, to enable searching and adding multiple footnotes at a time.
- Overrides _save_m2m method on MeasureForm, to allow for bulk creation of footnotes via the FootnoteAssociationMeasure.
- Updates edit.jinja to support footnotes and allow ajax-select to work [outside of the admin](https://django-ajax-selects.readthedocs.io/en/latest/Outside-of-Admin.html)
- Adds lookups.py file in measures and FootnotesLookup to support AutoCompleteSelectMultipleField and allow searching of footnotes on the basis of footnote_type_id and footnote_id
- Add two view tests for adding and removing footnotes
- Add existing_measure_data fixture to conftest

[ajax-selects docs](https://django-ajax-selects.readthedocs.io/en/latest/index.html)

Screenshot 2021-09-22 at 09 33 04

Screenshot 2021-09-22 at 09 33 12

### DIFF
--- a/common/tests/util.py
+++ b/common/tests/util.py
@@ -376,19 +376,23 @@ def get_form_data(form: forms.ModelForm) -> Dict[str, Any]:
     hence result in multiple HTML <input> objects."""
 
     data = {**form.initial}
-    for field in form.rendered_fields:
-        value = data[field] if field in data else form.fields[field].initial
-        if hasattr(form.fields[field].widget, "decompress"):
-            # If the widget can be decompressed, then it is not just a simple
-            # value and has some internal structure. So we need to generate one
-            # form item per decompressed value and append the name with _0, _1,
-            # etc. This mirrors the MultiValueWidget in django/forms/widgets.py.
-            if field in data:
-                del data[field]
-            value = form.fields[field].widget.decompress(value)
-            data.update(
-                **{f"{field}_{i}": v for i, v in enumerate(value) if v is not None}
-            )
-        elif value is not None:
-            data.setdefault(field, value)
+    if hasattr(form, "rendered_fields"):
+        for field in form.rendered_fields:
+            value = data[field] if field in data else form.fields[field].initial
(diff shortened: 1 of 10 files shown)


simonwo:
Is there a good reason to use a new design and new component over the one that's already present for create measures? The original design had us using a similar layout, where the existing autocomplete is paired with a "Add new" and "Delete" button: https://tqv1q6.axshare.com/#id=k1yxmj&p=change_footnotes&g=1

gabelton:
> Is there a good reason to use a new design and new component over the one that's already present for create measures? The original design had us using a similar layout, where the existing autocomplete is paired with a "Add new" and "Delete" button: https://tqv1q6.axshare.com/#id=k1yxmj&p=change_footnotes&g=1

I hadn't seen that design before, but I don't know about reusing the component for create measures - I just wasn't sure how to incorporate the footnotes formset in the govuk accordion while using the Update view instead of create. I was looking for some way to bulk create / remove all on that view

andyhd:
I might have misunderstood, but the TOPS ticket referenced is for footnotes associated with an additional code?

I feel slightly uneasy about the new dependency - it doesn't look like ajax-selects has an accessible fallback. This may not be a priority.

The design seems to only allow for removing all footnotes from a measure and adding footnotes (presumably with the "Add another" functionality from the "Create Measure > Add footnotes" form). I'm not sure if that addresses a case like the referenced TOPS ticket.

PR #386, opened 2021-09-23
TP-846 Geographical area detail view throws error for Taiwan

gabelton:
## Why
GeographicalMembership.other() raises a 500 ValueError complaining that 'Region TW is not part of membership geo_group__sid=168, member__sid=102'. This is because in memberships.jinja GeographicalArea.get_current_memberships() filters by 'member__sid', which returns both Taiwan the region AND country. This means that later in the template .other() is being called on the wrong membership.

## What
- Adds filtering on member__area_code to get_current_memberships method
- Adds test for this scenario in geo_areas.models


geo_areas/tests/test_models.py
```diff
@@ -51,6 +51,19 @@ def test_get_current_memberships_on_areas(membership_data, expected):
...
+def test_get_current_memberships_when_region_and_country_share_sid():
+    country = factories.CountryFactory()
+    region = factories.RegionFactory(sid=country.sid)
+    factories.GeographicalMembershipFactory(member=country)
+    factories.GeographicalMembershipFactory(member=region)
+    country_memberships = country.get_current_memberships()
+    region_memberships = region.get_current_memberships()
+
+    assert country_memberships.count() == 1
+    assert region_memberships.count() == 1
```
stuaxo, line 63: Would it be possible to check the content of `country_memberships` and `region_memberships` here ?   

Checking only the number of items returned can sometimes lead to tests that pass even though something unexpected is being returned.

geo_areas/tests/test_models.py
```diff
@@ -51,6 +51,19 @@ def test_get_current_memberships_on_areas(membership_data, expected):
     assert len(area.get_current_memberships()) == expected
 
 
+def test_get_current_memberships_when_region_and_country_share_sid():
+    country = factories.CountryFactory()
+    region = factories.RegionFactory(sid=country.sid)
+    factories.GeographicalMembershipFactory(member=country)
+    factories.GeographicalMembershipFactory(member=region)
```
andyhd, line 58: Tiny point - I think we explicitly use `Factory.create()` instead of just `Factory()` as a standard.

andyhd (approved):

geo_areas/models.py
```diff
@@ -87,7 +87,8 @@ class GeographicalArea(TrackedModel, ValidityMixin):
     def get_current_memberships(self):
         return (
             GeographicalMembership.objects.filter(
-                Q(geo_group__sid=self.sid) | Q(member__sid=self.sid),
+                Q(geo_group__sid=self.sid)
+                | Q(member__sid=self.sid, member__area_code=self.area_code),
```
simonwo, line 91: Do we need this bit now? In the Taiwan case, doesn't it just filter out memberships when it shouldn't?

geo_areas/models.py
```diff
@@ -179,9 +180,9 @@ def other(self, area: GeographicalArea) -> GeographicalArea:
         is passed this returns the geo group. Raises a ValueError if an object
         is passed that is neither of these.
         """
-        if area == self.geo_group:
+        if area.sid == self.geo_group.sid:
             return self.member
-        elif area == self.member:
+        elif area.sid == self.member.sid and area.area_code == self.member.area_code:
```
simonwo, line 185: Do you definitely need the area_code bit? Because again, if I pass in Taiwan(sid=1, code=REGION) but the membership is linked to Taiwan(sid=1, code=COUNTRY) won't that continue to raise a ValueError?

geo_areas/tests/test_models.py
```diff
@@ -61,6 +76,23 @@ def test_other_on_membership():
...
+    country = factories.CountryFactory.create()
+    geo_group = factories.GeoGroupFactory.create()
+    membership = factories.GeographicalMembershipFactory.create(
+        member=country,
+        geo_group=geo_group,
+    )
+    v2_country = factories.CountryFactory.create(
+        sid=country.sid,
+        area_code=country.area_code,
+    )
```
simonwo, line 89: Can you add a case that covers the area code being different? Because I think the current code will fail that, but it's valid. (Also, it may be worth using `new_version` rather than calling the factory twice because it preserves the version metadata – won't make a difference in the current implementation but it's more descriptive).

geo_areas/models.py
simonwo: E.g if a membership is created against Taiwan(sid=1, code=COUNTRY) and we then create a v2 that is Taiwan(sid=1, code=REGION), calling this on the v2 should still return the memberships against the v1.

simonwo (approved):
Noice!

PR #387, opened 2021-09-24
TP-857 Newly created descriptions don't appear on the Descriptions tab

gabelton:
## Why
The descriptions tab isn't displaying descriptions for footnotes after creating a footnote or editing one.

## What
- Updates the descriptions.jinja template to use object.descriptions.all() instead of object.get_descriptions(), as this was returning an empty queryset for footnotes.

Screenshot 2021-09-24 at 11 28 53


common/jinja2/includes/common/tabs/descriptions.jinja
```diff
@@ -2,7 +2,7 @@
 {% from 'macros/create_link.jinja' import create_link %}
 
 {% set description_rows = [] %}
-{% for description in object.get_descriptions() %}
+{% for description in object.descriptions.all() %}
```
andyhd, line 5: This will fetch all versions of descriptions. You want at least `object.descriptions.latest_approved()` or object.descriptions.approved_up_to_transaction(transaction)`.

Is it possible that no descriptions were returned because there are no currently approved descriptions?

andyhd (changes requested):

common/jinja2/includes/common/tabs/descriptions.jinja
simonwo: The latter one – so that the user can see the descriptions or changes to them that are contained in the draft workbasket they have open, which may not be approved yet.

It's possible there are no approved descriptions because you can create an object and an it's description, and neither will be approved until the workbasket is submitted.

gabelton: I think the problem is with get_descriptions. In this case it's returning query.latest_approved() which is empty because the footnote and description have only just been created and haven't been approved yet.

gabelton: (I wrote the above before I saw that you'd already summed it much more eloquently, Simon!)

stuaxo: The `Workbasket.from_session(…).last_transaction` to get the last_transaction of the current session seems like something we will want in other places, it's may be worth thinking about what an API for that should look like.

gabelton: I've just pushed some changes and a test now. Is there a scenario where we would want to return query.latest_approved() instead of query.approved_up_to_transaction(transaction=WorkBasket.get_current_transaction(request),)? I've been poking around in my local and it seems that either the two querysets are the same or query.approved_up_to_transaction covers the user's most recent change, which we want to display to them, even if it's not been approved yet.

andyhd (approved):
This is part of what I'm working on right now. It would be nice to not have to pass the request to `get_descriptions`, and have it be able to get the current transaction from the session itself.

common/jinja2/includes/common/tabs/descriptions.jinja
andyhd: `latest_approved` only really makes sense if there is no current workbasket/transaction, but we've changed the design so that there is always a workbasket. Perhaps if we allow non-authenticated users to browse (but not edit), or switch back to multiple concurrent workbaskets in future.

PR #388, opened 2021-09-29
DRAFT: Tp 888 extend edit measures to allow editing footnotes 

gabelton:
## Why
- Upcoming TOPS tickets require the ability to add or remove footnotes on an existing measure.
- Creating a draft because I'm currently blocked and would appreciate any advice on how to get the "Add another footnote" logic working on the edit form. Currently pressing "Add another footnote" button posts the edit form data when we need it to work like the footnotes section when creating a new measure.

## What
- A draft of current approach: passing existing MeasureFootnotesFormSet into MeasureUpdate view and borrowing from create-formset.jinja to display this as part of the edit accordion
- Also includes some changes merged in from Paul's branch here https://github.com/uktrade/tamato/pull/383

Screenshot 2021-09-29 at 14 05 08

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -78,12 +78,6 @@ class MeasureForm(ValidityPeriodForm):
         required=False,
         empty_label=None,
     )
-    footnote = AutoCompleteField(
-        label="Add a footnote",
-        help_text="Start typing the ID of the footnote or terms used in the description",
-        queryset=Footnote.objects.all(),
-        required=False,
-    )
 
     def __init__(self, *args, **kwargs):
         self.request = kwargs.pop("request", None)
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -78,6 +78,12 @@ class MeasureForm(ValidityPeriodForm):
         required=False,
         empty_label=None,
     )
+    footnote = AutoCompleteField(
+        label="Add a footnote",
+        help_text="Start typing the ID of the footnote or terms used in the description",
+        queryset=Footnote.objects.all(),
+        required=False,
+    )
 
     def __init__(self, *args, **kwargs):
         self.request = kwargs.pop("request", None)
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -332,7 +332,7 @@ class MeasureCommodityForm(forms.Form):
         )
 
 
-class AddAnother(forms.BaseFormSet):
+class FormSet(forms.BaseFormSet):
     """
     Adds the ability to add another form to the formset on submit.
 
@@ -343,7 +343,7 @@ class AddAnother(forms.BaseFormSet):
     preserved.
     """
 
-    extra = 1
+    extra = 0
     can_order = False
     can_delete = True
     max_num = 1000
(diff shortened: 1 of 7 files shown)


common/jinja2/includes/common/formset-delete-button.jinja
```diff
@@ -2,4 +2,5 @@
         class="govuk-button govuk-button--secondary"
         id="id_{{ field.html_name }}"
         value="1"
+        {{ "formaction={}".format(form.path) if form.path else "" }}
```
andyhd, line 5: I think this might be better written as
```jinja
{% if form.path %}formaction="{{ form.path }}"{% endif %}
```

measures/forms.py
```diff
@@ -130,6 +138,41 @@ def clean(self):
 
         return cleaned_data
 
+    def save(self, commit=True):
+        """Get the measure instance after form submission, get from session
+        storage any footnote pks created via the Footnote formset and any pks
+        not removed from the measure after editing and create footnotes via
+        FootnoteAssociationMeasure."""
+        instance = super(MeasureForm, self).save(commit=False)
```
andyhd, line 146: You can do `super().save(commit=False)` here

measures/forms.py
```diff
@@ -505,7 +548,12 @@ class MeasureFootnotesForm(forms.Form):
         queryset=Footnote.objects.all(),
     )
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, request, **kwargs):
+        # Check whether form is being used as part of measure edit view
+        # and pass edit-footnotes url if this is the case
+        self.request = request
+        path = self.request.path
+        self.path = path[:-1] + "-footnotes/" if "edit" in path else None
```
andyhd, line 556: I think it's better to avoid ternary conditional expressions in general, unless they are really simple. This one would be clearer (IMO) as:
```python
self.path = None
if "edit" in path:
    self.path = f"{path[:-1]}-footnotes/"
```

measures/jinja2/measures/create-footnotes-formset.jinja
```diff
@@ -0,0 +1,49 @@
+{% if form.request.session["instance_footnotes_"~form.instance.sid] %}
+        {% set table_rows = [] %}
+        {% for footnote in form.instance.footnotes.all() %}
+            {% if footnote.pk in form.request.session["instance_footnotes_"~form.instance.sid]%}
```
andyhd, line 4: `form.request.session["instance_footnotes_"~form.instance.sid]` might be worth putting into a variable, eg:
```jinja
{% with form_footnotes = form.request.session["instance_footnotes_" ~ form.instance.sid] %}
  {% if form_footnotes %}
    {% set table_rows = [] %}
    ...
{% endwith %}
```

measures/views.py
```diff
@@ -290,3 +310,48 @@ def form_valid(self, form):
...
+
+    def get_delete_key(self, footnote_key: str) -> str:
+        """
+        Expects a string of format 'form-0-footnote' or 'form-1-footnote' etc.
+
+        Outputs a string of format 'form-0-DELETE' or 'form-1-DELETE' etc.
+        """
+        split = footnote_key.split("-")
+
+        return split[0] + "-" + split[1] + "-DELETE"
```
andyhd, line 327: I think f-strings are the preferred way to concatenate strings. Also, `split` could have a better name.

You could do something like
```python
form_prefix, _ = footnote_key.rsplit("-", 1) # assuming there are never more than 2 `-`s
return f"{form_prefix}-DELETE"
# or
_, form_index, _ = footnote_key.split("-", 2)
return f"form-{form_index}-DELETE"
```

andyhd (approved):
👍  Only minor style comments

paulpepper-trade (approved):

PR #389, opened 2021-09-30
TP-859: Switch to using an in-memory SQLite export

simonwo:
## Why
We've found it too difficult to provision enough temporary disk space on PaaS to create ~3Gb database. 

## What
This commit changes the export to use an in-memory database instad. We've changed wrappers to use apsw instead of the native library so that we can subsequently deserialize the database into bytes and stream it directly to S3.

## Checklist
- Requires migrations? No.
- Requires dependency updates? Yes.

### DIFF
--- a/exporter/sqlite/__init__.py
+++ b/exporter/sqlite/__init__.py
@@ -25,7 +25,6 @@ date fields, and any other Postgres-specific features are ignored.
 """
 
 from itertools import chain
-from pathlib import Path
 
 from django.apps import apps
 from django.conf import settings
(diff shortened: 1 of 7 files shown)


rachaelcodes (approved):

stuaxo:
Ignore this if it's already in the PR:   

There should definitely be a note / info in the README about the memory requirements, otherwise it may be surprising to devs in the future.

Worth checking roughly what this uses in top/htop ... and note the size and the ratio compared to the database size e.g. roughly 1.2x the database size 2x, 3x / whatever.

Could be useful info later, if e.g. if we end up using up all the memory.

Michal actually worked on something to stream sqlite, which could be an option in the future, if the ram requirements get big.

simonwo:
> Ignore this if it's already in the PR:
> 
> There should definitely be a note / info in the README about the memory requirements, otherwise it may be surprising to devs in the future.
> 
> Worth checking roughly what this uses in top/htop ... and note the size and the ratio compared to the database size e.g. roughly 1.2x the database size 2x, 3x / whatever.
> 
> Could be useful info later, if e.g. if we end up using up all the memory.
> 
> Michal actually worked on something to stream sqlite, which could be an option in the future, if the ram requirements get big.

Agreed, I’ll add that.

Yes, in fact this change makes it easy to change over to that S3-based SQLite if we want to - it would just be a couple of lines to configure and swap out the virtual file system.

PR #392, opened 2021-10-01
Add ADR-0014 for simplifying current-transaction queries

andyhd:
# TP-826 ADR-0014 for simplifying current-transaction queries

## Why

Because our database is (mostly) bi-temporal, many queries need to filter data so that only data that is valid as of the latest approved transaction (including draft changes in the workbasket) is returned. This requires fetching the current transaction from the session workbasket and then passing down function call chains to be used in queries. This adds boilerplate to our code and is sometimes difficult to achieve, or easily forgotten, leading to bugs.

## What

ADR-0014 proposes using [Django-CRUM](https://django-crum.readthedocs.io/en/latest/) to store a reference to the current request (and hence, access to the current session/workbasket/transaction) in thread-local storage. This will allow us to fetch the current transaction at the point in the code where it is needed and not have to pass the transaction down long call chains.


docs/adr/0014-simplify-current-transaction-queries.rst
```diff
@@ -0,0 +1,124 @@
...
+            return descriptions_model.objects.approved().filter(**filter_kwargs).order_by(*ordering)
+
+
+and then
+
+.. code:: jinja
+
+    {% for description in footnote.get_descriptions() %}
+        <p>{{ description }}</p>
+    {% endfor %}
```
simonwo, line 90: Could we also examine how to achieve this in [this sort of case](https://github.com/uktrade/tamato/blob/master/measures/forms.py#L33)?

```python
    measure_type = AutoCompleteField(
        label="Measure type",
        help_text="Select the regulation which provides the legal basis for the measure.",
        queryset=models.MeasureType.objects.all(),
    )
```

E.g. it would be nice to be able to provide a QuerySet object (because they are a common interface) and have the queryset automatically pick up the current transaction at the moment they are called rather than when they are constructed?

andyhd: I have a prototype solution that I've tested and seems to work for this. I'll add what I have so far in, but I think it needs a bit more work than is in scope for this task.

andyhd: Ok, it took me longer than I thought, but I think I have a solution that works.

docs/adr/0014-simplify-current-transaction-queries.rst
```diff
@@ -0,0 +1,180 @@
...
+
+.. code:: python
+
+    from crum import get_current_request
+    from django.db.models import IntegerField, Value
+
+    from workbaskets.models import WorkBasket
+
+
+    class CurrentTransactionId(Value):
```
simonwo, line 64: 😍💯💯💯

docs/adr/0014-simplify-current-transaction-queries.rst
```diff
@@ -0,0 +1,180 @@
...
+
+
+Alternatives considered
+-----------------------
+
+Thread-local storage
+~~~~~~~~~~~~~~~~~~~~
+
+We store the current transaction in thread-local storage as a sort of global variable
+that provides the context to all queries in the current thread - this avoids
```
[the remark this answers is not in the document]
andyhd: That's a good point. I have assumed that there is only one thread per request. Django CRUM deletes the current request from thread-local storage when the response is sent and the request is effectively complete.

gabelton (approved):

docs/adr/0014-simplify-current-transaction-queries.rst
```diff
@@ -0,0 +1,180 @@
...
+
+    from crum import get_current_request
+    from django.db.models import IntegerField, Value
+
+    from workbaskets.models import WorkBasket
+
+
+    class CurrentTransactionId(Value):
+        def __init__(self, **kwargs):
+            kwargs["output_field"] = IntegerField()
```
paulpepper-trade, line 66: Nice.
Do we reference Django docs from our ADRs? If so, then perhaps a ref to say https://docs.djangoproject.com/en/dev/ref/models/expressions/#value-expressions in place of explanatory note?

paulpepper-trade (approved):

stuaxo:
Haven't commented so far, as it all looks good..

Workbasket.get_current_transaction looks like a good candidate for functools.cache

PR #400, opened 2021-10-21
TP-901 Create Measure wizard doesnt re-present filled form data when navigating backwards

gabelton:
## Why
Currently, if a user wants to revisit a step when creating a measure, the form will display as blank, even though data is stored. This means the user has to reenter data for all fields, even if they only wish to change one thing.

## What
- Using cleaned data in MeasureCreateWizard.get_form() to populate form fields
- Overriding ValidityPeriodForm.clean() to populate start_date and end_date fields
- Overriding GeographicalAreaFormMixin.clean() to accomplish the same for geo_area fields on MeasureDetails form
- Updating common/jinja2/components/geographical_area_select to check radio input if initial data is passed in
- Editing common/jinja2/layouts/form.jinja to avoid separate bug where an empty error box displays when navigating back to a step that uses formsets


paulpepper-trade (changes requested):

geo_areas/forms.py
```diff
@@ -132,6 +132,10 @@ def clean(self):
                 raise ValidationError({"geo_area": "A country or region is required."})
             cleaned_data["geographical_area"] = geo_area
```
paulpepper-trade, line 134: A slight niggle. If the user selects the "A single country or region" radio as Geographical area, but does not choose a country item from the exposed select, then the validation works correctly, but the radio selection is lost (because `self.fields["geo_area_type"].initial = geo_area_type` has not been set before `ValidationError` is raised).

paulpepper-trade (changes requested):

measures/views.py
```diff
@@ -226,6 +226,12 @@ def get_form(self, step=None, data=None, files=None):
                 for field in f.fields.values():
                     if hasattr(field, "queryset"):
                         field.queryset = field.queryset.approved_up_to_transaction(tx)
+
+        form.is_valid()
+        breakpoint()
+        if hasattr(form, "cleaned_data"):
+            form.initial = form.cleaned_data
```
stuaxo, line 233: Instead of checking if cleaned_data is present maybe it's worth checking the return value of `form.is_valid` ?

gabelton: I had that check originally, but found there were situations where the form was invalid and still had cleaned_data. E.g. the scenario Paul described above where the user has filled out all fields on MeasureDetailsForm, checked a radio option for geo_area, but forgotten to select a country from the dropdown. In that case, the form is invalid but we still want to use the cleaned_data to populate fields such as measure_type, which doesn't happen if the check is on form.is_valid()

paulpepper-trade (approved):

PR #410, opened 2021-11-02
TP-1022-Make pycharm a little happier: mostly backporting some of the typing …

stuaxo:
This is mostly fixing some of the typing annotations (backport stuff to 3.8), it certainly makes pycharm a little less underline-y.


kintisheff (commented):
so you and Andy need to duke it out on whether we use the future typings as of 3.10 (therefore the import of annotations from __future__) or make PyCharm happy :) otherwise I'm happy with that

kintisheff (approved):

PR #411, opened 2021-11-02
TP-1036: Prevent duplicate log outputs

kintisheff:
# TP-1036: Prevent duplicate log outputs

## Why
Due to a misconfiguration in our django common settings file, our system currently prints duplicate logs for most modules (because they allow propagation to root logger), and for celery (because celery unnecessarily prints to console as well as celery handler).

## What
- set the propagate attribute to false for each module-specific logger
- make the celery handler the only handler of the celery logger

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -1,6 +1,5 @@
 """Django settings for tamato project."""
 import json
-import logging.config
 import os
 import sys
 import uuid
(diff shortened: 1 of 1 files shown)


settings/common.py
```diff
@@ -353,55 +354,64 @@
 GOOGLE_ANALYTICS_ID = os.environ.get("GOOGLE_ANALYTICS_ID")
 
 # -- Logging
+LOGGING_CONFIG = None
```
andyhd, line 357: Isn't this redundant?

settings/common.py
```diff
@@ -353,55 +354,64 @@
...
-            "level": os.environ.get("LOG_LEVEL", "DEBUG"),
-        },
-        "exporter": {
-            "handlers": ["console"],
-            "level": os.environ.get("LOG_LEVEL", "DEBUG"),
-        },
-        "commodities": {
-            "handlers": ["console"],
-            "level": os.environ.get("LOG_LEVEL", "DEBUG"),
+logging.config.dictConfig(
```
andyhd, line 359: I assume by calling into the `logging` module directly, you are bypassing/not-clobbering some Django logging setup - if that is deliberate, it might be worth a comment.

andyhd (approved):
Looks good to me

settings/common.py
kintisheff: i think so yes

PR #412, opened 2021-11-02
TP-1037: Fix indent node end dating

kintisheff:
# TP-1037: Fix indent node end dating

## Why
GoodsNomenclatureIndents, part of the TARIC3 specification, do not have explicit end dates.
GoodsNomenclatureIndentNodes, a model specific to Tamato, use explicit end dates.
When new indents are imported into the system
(e.g. from when we synchronise our commodity codes with those of the EU),
we currently do not ensure that any affected existing indent nodes are updated
so that their explicit end date is in line with the new implicit end date
of existing indents related to the new one (preceding or succeeding).
The impact is a broken commodity tree and occasionally errors in imports as well.
This ticket is to fix that.

## What
- Add new properties and methods to GoodsNomenclature, GoodsNomenclatureIndent, and GoodsNomenclatureIndentNode for determining indent depth, preceding and succeeding nodes and indents, etc.
- Relieve the import handler of the responsibility to deal with all these, and just use them as needed
- Add to the import handler two methods - one for end-dating preceding indent nodes and one for applying an end-date to a new node if it has a succeeding node

_## NB:
This PR includes a necessary fix to import handler to apply a copy of the GoodsNomenclatureIndentNode tree each time an indent is replaced. This fix is necessary in order to avoid broken commodity trees. However, with it the space and time complexity of the implementation are now significantly worse. A decision has been made to dispense with MPTree altogether in separate PR-s. It is not in the scope of this PR to resolve the issue._


commodities/import_handlers.py
```diff
@@ -193,9 +263,11 @@ def save(self, data: dict):
                 pk=data.pop("indented_goods_nomenclature_id"),
             )
         item_id = data["indented_goods_nomenclature"].item_id
-        suffix = data["indented_goods_nomenclature"].suffix
+        data["indented_goods_nomenclature"].suffix
```
stuaxo, line 266: ^ is this meant to be assigned to anything?

If not it may be worth a short comment as it looks odd.

commodities/import_handlers.py
```diff
@@ -205,37 +277,12 @@ def save(self, data: dict):
             "creating_transaction_id": data["transaction_id"],
         }
 
-        if depth == 0 and item_id[2:] == "00000000":
-            # This is a root indent (i.e. a chapter heading)
-            # Race conditions are common, so reduce the chance of it.
-            time.sleep(random.choice([x * 0.05 for x in range(0, 200)]))
+        if indent.is_root:
```
stuaxo, line 280: It's good you remove the sleep, they generally hide bad things, in this case it may be have been to mask the fact that tree beard really only wants one thing to create nodes in it at once (and can be slow enough that it's possible to hit this when you have something with multiple processes doing an updates, like a couple of celery jobs).

commodities/import_handlers.py
```diff
@@ -396,7 +432,11 @@ def find_and_replace_inappropriate_parent_nodes(
         This method finds all possible children of a new indent and queries to
         see if any should be moved from their current parent to the new indent.
         """
-        nodes = obj.nodes.all()
+        try:
+            nodes = obj.nodes.all()
+        except (AttributeError, TypeError):
```
stuaxo, line 437: I guess there is a good reason for this, could be worth a comment.

commodities/import_handlers.py
simonwo: Ah yes, interesting. The importer supports a degree of paralellism across multiple chapters (which are independent) so if treebeard is expecting things to not be multi-threaded that would explain this.

commodities/models/orm.py
```diff
@@ -195,6 +199,44 @@ def dependent_measures(self):
...
+        """
+        chapter = self.code.chapter
+        indent_shift = 0
+
+        # In some cases, where there are phantom headers at the 4 digit level
+        # in a chapter, the depth is shifted by + 1.
+        # A phantom header is any good with a suffix != "80". In the real world
+        # this represents a good that does not appear in any legislature and is
+        # non-declarable. i.e. it does not exist outside of the database and is
+        # purely for "convenience". This algorithm doesn't apply to chapter 99.
```
simonwo, line 225: Duplicated in the doc string

commodities/models/orm.py
```diff
@@ -231,6 +283,88 @@ def get_parent_indents(self):
 
         return GoodsNomenclatureIndent.objects.filter(parent_path_query)
 
+    @property
+    def good_indents(self) -> QuerySet:
+        """Return the related goods indents based on approval status."""
+        good = self.indented_goods_nomenclature
+        return good.indents.approved_up_to_transaction(
+            self.transaction,
+        )
```
simonwo, line 292: I'm a little worried about the use of `self.transaction` here… 

Let's say that in transaction X I create this indent, and there is a preceeding indent for the same commodity code.
Then in transaction X+1 I create a succeding indent for the same commodity code.

If I call `succeeding_indents` on this indent, I will get `None`, because the succeeding indent didn't exist when this one was created. Is that intuitive behaviour?

Would it be better to force the user to pass a transaction explicitly?

commodities/models/orm.py
```diff
@@ -231,6 +283,88 @@ def get_parent_indents(self):
...
+        # The end dates on some historically created nodes
+        # may have not been synced with the implied end date of their indent
+        # when a succeeding indent has been introduced at a later point in time.
+        # This can cause the above query to yield the wrong parent.
+        # The extra logic below catches and remedies such potential cases.
+        # TODO: Handle situations with multiple wrong patterns
+        effective_end_date = parent.effective_end_date
+
+        if effective_end_date and effective_end_date < self.validity_start:
+            parent = parent.succeeding_node
```
simonwo, line 366: Where does this method return the parent?

commodities/models/orm.py
```diff
@@ -358,6 +492,85 @@ def _get_restricted_valid_between(
...
+        (e.g. due to the introduction of a succeeding indent),
+        we need to be able to tell the effective end dates of such nodes,
+        which are constrained by the implicit end date of the related indent.
+        """
+        indent = self.indent.succeeding_indent
+
+        if not indent:
+            return self.valid_between.upper
+
+        return indent.validity_start + timedelta(days=-1)
```
simonwo, line 554: Could we just fix the data?

commodities/models/orm.py
```diff
@@ -358,6 +492,85 @@ def _get_restricted_valid_between(
...
+        """
+        return TaricDateRange(
+            self.valid_between.lower,
+            self.effective_end_date,
+        )
+
+    @property
+    def parent(self) -> GoodsNomenclatureIndentNode:
+        """Returns the parent of the node."""
+        return self.indent.get_parent_node(self.parent_depth)
```
simonwo, line 572: I'm worried this is a little confusing if a `Node` has both `get_parent` and `parent` … which should I be using?

commodities/models/orm.py
kintisheff: yes but that is out of scope for this ticket and the above helps meantime

commodities/models/orm.py
kintisheff: that was just fixed! not sure how it fell out, that was breaking the tests too

commodities/models/orm.py
kintisheff: agree

commodities/import_handlers.py
kintisheff: well spotted, this is a relic, removing

commodities/models/orm.py
kintisheff: yes, this is quite simply redundant (and not used even). removing

commodities/models/orm.py
kintisheff: i suppose it got promoted to a docstring :)

simonwo:
Would it be useful to have a test for this new functionality, that the preceding indent node is end-dated? I.e. something that would fail without this change.

commodities/models/orm.py
```diff
@@ -362,6 +362,9 @@ def get_parent_node(
             .first()
         )
 
+        if not parent:
+            return
+
```
simonwo, line 367: Should this ever be true if we've gone past the "root node" check?? It might hide errors?

commodities/tests/test_importer.py
```diff
@@ -647,3 +647,86 @@ def test_footnote_association_goods_nomenclature_importer(imported_fields_match)
...
+            validity_start=imported_indent_start_date,
+            indented_goods_nomenclature=good,
+            indent=0,
+        ),
+        serializers.GoodsNomenclatureIndentSerializer,
+    )
+
+    if current_indent_start_date:
+        current_node = current_indent.nodes.first()
+        assert current_node.valid_between.upper == curent_node_end_date
```
simonwo, line 729: Does this update the existing node? Or does it need to create a new node with the same data but with an updated `creating_transaction` field?

kintisheff: i think it should be the existing node.

commodities/import_handlers.py
stuaxo: Treebeard creates predictable keys for it's tree, so you can imagine how they clash.

The keys are strings, something like this:

```
1
10
110
```

But you can imagine different structures using different keys, we don't care about implicit ordering of the tree as much.
There's some good work on threadsafe/multiprocess safe data structures, most of it probably lives in the world of functional programming languages.

commodities/models/orm.py
kintisheff: parasite from a test i was doing! removing

commodities/tests/test_importer.py
simonwo: If so, how do we go back to an existing transaction before the indent was imported and run something like ME32? Because the node will claim that the commodity code ends early, but there is no indent as of that transaction that actually makes that true

kintisheff: wait you might be right, it should create a new one because otherwise this compromises prior view of the tree potentially! fixing

commodities/import_handlers.py
```diff
@@ -184,7 +184,102 @@ def clean(self, data: dict) -> dict:
...
+        we are trying to attach to the tree.
+        """
+        preceding_indent = indent.get_preceding_indent()
+
+        if not preceding_indent:
+            return
+
+        preceding_node = models.GoodsNomenclatureIndentNode.objects.get(
+            indent=preceding_indent,
+        )
```
simonwo, line 224: Is it possible for `get` to return more than one object? If I have an indent, with a node, and then it gets updated by this method, there are now two nodes for the same indent. And then if it gets updated by this method again, this call will fail because multiple objects?

Should we really `.order_by(creating_transaction_id).last()`?

commodities/import_handlers.py
```diff
@@ -184,7 +184,102 @@ def clean(self, data: dict) -> dict:
...
+        then we need to use the implied end date for the new indent
+        as the explicit end date for the new indent's related
+        `GoodsNomenclatureIndentNode` object we are about to create.
+        """
+        succeeding_indent = indent.get_succeeding_indent()
+
+        if not succeeding_indent:
+            return
+
+        return succeeding_indent.validity_start - timedelta(days=-1)
```
simonwo, line 280: Is there a good reason for this to not just be `GoodsNomenclatureIndent.objects.with_end_date().get(pk=indent.pk).validity_end`?

commodities/import_handlers.py
```diff
@@ -254,43 +333,43 @@ def save(self, data: dict):
...
             node_data["valid_between"] = TaricDateRange(indent_start, indent_end)
-            next_parent.add_child(**node_data)
+            indent_node = next_parent.add_child(**node_data)
+
+            if preceding_node:
+                preceding_node.copy_tree(
+                    parent=indent_node,
+                    valid_between=node_data["valid_between"],
+                    transaction=indent.transaction,
+                )
```
simonwo, line 367: I could do with a comment to explain this… don't we do this above, and also in the `set_preceding_node_end_date` method?

commodities/import_handlers.py
simonwo: This could still do with a comment…

commodities/models/orm.py
```diff
@@ -354,6 +496,89 @@ def _get_restricted_valid_between(
...
         return new_valid_between
 
+    @property
+    def parent_depth(self) -> int:
+        """
+        Returns parent depth factoring in the good's indent shift.
+
+        See the docs to `GoodsNomenclature.indent_shift` for context.
+        """
+        return self.depth + 1 + self.good.indent_shift
```
simonwo, line 506: Definitely +1? Not -1?

commodities/models/orm.py
```diff
@@ -354,6 +496,89 @@ def _get_restricted_valid_between(
...
+        (e.g. due to the introduction of a succeeding indent),
+        we need to be able to tell the effective end dates of such nodes,
+        which are constrained by the implicit end date of the related indent.
+        """
+        indent = self.indent.get_succeeding_indent(as_of_transaction)
+
+        if not indent:
+            return self.valid_between.upper
+
+        return indent.validity_start + timedelta(days=-1)
```
simonwo, line 565: IT looks like this code is duplicated somewhere earlier in this review?

commodities/import_handlers.py
kintisheff: unfortunately, copy tree requires a target indent node, i.e. it needs to be copied over for each new node we create; will add a comment

commodities/import_handlers.py
kintisheff: they seem equivalent, will use our established way of getting it

commodities/import_handlers.py
kintisheff: agree

commodities/models/orm.py
kintisheff: yes. here is the original code on this + 1 from `master` (when it was still in `import_handler`), which comment is also now promoted to the docs in this method.

# In some cases, where there are phantom headers at the 4 digit level
# in a chapter, the depth is shifted by + 1.

commodities/models/orm.py
kintisheff: the earlier code is to get an indent's implicit end date;
this line is to get what the node's end date should be when it's still none so we can change it (as opposed to the indent)

commodities/models/orm.py
kintisheff: got it!

simonwo (approved):
Let's do it 🚀

PR #414, opened 2021-11-02
TP-1040: Tweaks to the excel importer

kintisheff:
TP-1040: Tweaks to the excel importer

## Why
Upon loading a create measures file provided by MTPO, we noticed we needed to make a few small tweaks to the excel importer.

## What
A few small tweaks to the excel importer

### DIFF
--- a/importer/sheet.py
+++ b/importer/sheet.py
@@ -1,4 +1,3 @@
-import logging
 from abc import abstractmethod
 from dataclasses import dataclass
 from datetime import date
@@ -13,14 +12,11 @@ from typing import Union
 from openpyxl.cell import Cell
 from openpyxl.worksheet.worksheet import Worksheet
 
-from commodities.models.orm import GoodsNomenclature
 from common.util import classproperty
 from importer.utils import col
 from workbaskets.models import WorkBasket
 
 CellValue = Union[date, str, int, None]
-
-logger = logging.getLogger(__name__)
 
 
 def column(label, optional=False, many=False):
@@ -95,13 +91,4 @@ class SheetRowMixin:
         (header) row."""
         for row in islice(sheet.iter_rows(), 1, None):
             row_model = cls(row, *args, **kwargs)
-
-            try:
-                yield row_model.import_row(workbasket)
-            except GoodsNomenclature.DoesNotExist:
-                logger.warning(
-                    f"Commodity {row_model.item_id}: not imported from EU Taric files yet.",
-                )
-            except Exception as e:
-                logger.error(e)
-                raise
+            yield row_model.import_row(workbasket)
--- a/importer/sheet.py
+++ b/importer/sheet.py
@@ -1,3 +1,4 @@
+import logging
 from abc import abstractmethod
 from dataclasses import dataclass
 from datetime import date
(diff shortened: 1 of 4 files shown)


measures/sheet_importers.py
```diff
@@ -25,6 +25,18 @@
...
 
+def process_date_value(value: CellValue) -> date:
+    """
+    Returns a date instance corresponding to the cell value.
+
+    The date part of the value must be in "%Y-%m-%d" format. If a value
+    representing date and time is provided, this method will take the date part
+    only.
+    """
+    value = str(value).split(" ")[0]
```
stuaxo, line 36: Just because I like to optimise ~even~ especially in slow python land, let's add the ,1 -

`value = str(value).split(" ", 1)[0]`

importer/sheet.py
```diff
@@ -91,4 +95,13 @@ def import_sheet(cls, sheet: Worksheet, workbasket: WorkBasket, *args, **kwargs)
...
+
+            try:
+                yield row_model.import_row(workbasket)
+            except GoodsNomenclature.DoesNotExist:
+                logger.warning(
+                    f"Commodity {row_model.item_id}: not imported from EU Taric files yet.",
+                )
+            except Exception as e:
+                logger.error(e)
+                raise
```
simonwo, line 107: Could we move this to the `MeasureSheetImporter` instead? Because that keeps the domain specific-ness in one place. We could reuse this sheet importer concept for other things unrelated to the EU or Goods Nomenclatures.

measures/sheet_importers.py
```diff
@@ -77,6 +89,9 @@ def excluded_origin_descriptions(self, value: CellValue) -> str:
 
     @cached_property
     def excluded_origins(self) -> Sequence[GeographicalArea]:
+        if not self.excluded_origin_descriptions:
+            return []
+
```
simonwo, line 94: I don't think this should be necessary – because `many=True` the property should always return a sequence. If it doesn't, we should fix that instead.

importer/sheet.py
kintisheff: indeed!

measures/sheet_importers.py
kintisheff: there is no `many=True` on this one as it's not decorated with `colulmn`, it's just called as an arg to `creator.create` in `import_row`

simonwo: The `many=True` is on the `excluded_origin_descriptions` and so it should never be `None`.
https://github.com/uktrade/tamato/blob/master/measures/sheet_importers.py#L75

measures/sheet_importers.py
```diff
@@ -24,6 +25,20 @@
 from regulations.models import Regulation
 from workbaskets.models import WorkBasket
 
+logger = logging.getLogger(__name__)
+
+
+def process_date_value(value: CellValue) -> date:
```
gabelton, line 31: Is it worth adding a unit test where date and time value are passed in to check that only date is returned ?

kintisheff: Yes! I'll add it

measures/sheet_importers.py
kintisheff: Yes I was just able to confirm also. Removing

kintisheff: Yes I was just able to confirm also. Removing

simonwo (approved):

PR #415, opened 2021-11-03
TP-1041: Enable selective workbasket xml export

kintisheff:
# TP-1041: Selective workbasket xml export

## Why
At the moment, the management command in `dump_transactions.py` automatically picks up workbaskets with an APPROVED status.
In the context of HS22, where we may want to lump several workbaskets with EU commodity change imports and MTPO create measures imports, we need to also allow users to select the specific workbaskets they want to package for publishing

## What
- Tweak `Command()` in  the `dump_transactions` management script to accept a list of comma-separated workbasket id-s as an additional optional argument
- Tweak the django db query to fetch the custom workbaskets if the optional argument is provided, and APPROVED workbaskets (existing default behaviour) if not

### DIFF
--- a/exporter/management/commands/dump_transactions.py
+++ b/exporter/management/commands/dump_transactions.py
@@ -50,7 +50,8 @@ class Command(BaseCommand):
                 "Override the default selection of APPROVED workbaskets "
                 "with a comma-separated list of workbasket ids."
             ),
-            type="str",
+            nargs="*",
+            type=int,
             default=None,
             action="store",
         )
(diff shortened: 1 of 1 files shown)


exporter/management/commands/dump_transactions.py
```diff
@@ -44,9 +44,27 @@ def add_arguments(self, parser):
...
+        parser.add_argument(
+            "workbasket_ids",
+            help=(
+                "Override the default selection of APPROVED workbaskets "
+                "with a comma-separated list of workbasket ids."
+            ),
+            type="str",
+            default=None,
+            action="store",
+        )
```
simonwo, line 56: Would it be better to use [`nargs *`](https://docs.python.org/3/library/argparse.html#nargs) for multiple arguments rather than comma-separating?

kintisheff: makes sense, will tweak

simonwo (approved):

PR #417, opened 2021-11-03
Duplicate many-to-many fields for new_versions

simonwo:
## Why
Previously any many-to-many fields that have auto-generated through models would not be retained by `new_version`. So if you create a `new_version` of a quota, it will forget about any `required_certificates`.

## What
This commit allows `new_version` to keep the models if they haven't been overriden by the user, or set them correctly to a new iterable.

This won't work for `save=False`, so in the future we will need to find some way to handle this. It won't be a problem until we start editing quotas, which is currently the only place we use many-to-many fields with an auto-generated through model.

## Checklist
- Requires migrations? No.


common/tests/test_models.py
```diff
@@ -312,6 +312,23 @@ def test_new_draft_uses_passed_transaction(sample_model):
...
+    )
+    assert sample_model.descriptions.get() == description
+
+    new_model = sample_model.new_version(
+        sample_model.transaction.workbasket,
+    )
+    assert new_model.descriptions.get() == description
+
+
+def test_new_version_works_for_all_models(trackedmodel_factory):
```
gabelton, line 327: Outside of the scope of this ticket, but is it worth putting a TODO next to this / creating a ticket to add a couple more unit tests with assertions for new_version? e.g. checking for ValueError, save=False, and that new_object_kwargs contain what they should contain

simonwo: Done as TP-1043.

stuaxo (approved):

PR #418, opened 2021-11-04
TP-1045 Allow passing partition schema by name to avoid serialisation over ce…

stuaxo:
Pass schemas by name so that celery is happy.

Note: this changes the command line use slightly (see the help), to the IMO slightly less friendly ALL_CAPS names, but it seemed better than maintaining two sets of names and making the name resolution even more complex.

### DIFF
--- a/importer/tests/test_tasks.py
+++ b/importer/tests/test_tasks.py
@@ -2,7 +2,6 @@ from unittest import mock
 
 import pytest
 
-from common.models.transactions import TransactionPartition
 from common.tests import factories
 from importer import tasks
 from importer.models import ImporterChunkStatus
@@ -21,7 +20,7 @@ def test_import_chunk(
     tasks.import_chunk(
         chunk.pk,
         "PUBLISHED",
-        REVISION_ONLY,
+        "REVISION_ONLY",
         valid_user.username,
     )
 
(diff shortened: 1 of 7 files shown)


importer/tests/test_tasks.py
```diff
@@ -61,7 +60,7 @@ def test_setup_chunk_task_already_running(mock_import_chunk, batch, valid_user):
     tasks.setup_chunk_task(
         batch,
         "PUBLISHED",
-        REVISION_ONLY,
+        "REVISION_ONLY",
```
simonwo, line 63: Is it worth using something like `REVISION_ONLY.__name__` or something?

simonwo (approved):

importer/tests/test_tasks.py
stuaxo: ~-  :+1: -~    (bad attempt at making a strikethrough on my thumbs up)

Wait ... I caught myself out..   REVISION_ONLY is an instance of UserPartitionScheme not a class, so to have a name itself I'd need to implement a global registry of schemes.

I didn't do this feature (even though it would be fun to write) as it seemed overkill.

stuaxo: The third time I scrolled past this and still makes me a little sad I didn't write this, worked with someone who was very strictly YAGNI a couple of projects ago and seems to have limited my free-spirit in writing code that's just fun :)

PR #419, opened 2021-11-04
WIP: Add pytest-split-tests to speed up tests on CI.

stuaxo:
See if pytest-split-tests can help CI times.


stuaxo:
Finding here is that github actions give you CPU seconds.

If we want the tests to run faster on CI we need to ask for a bigger plan.

PR #420, opened 2021-11-05
TP-1050: ME12 using reverse relations

gabelton:
## Why
- Currently measure business rule ME12 (If the additional code is specified then the additional code type must have a relationship with the measure type.) uses reverses relationships in its query. This has meant that since measure type 105 has been updated, two measures (20151527, 20151528) are now in violation of the rule.

## What
- Updates the rule to use the through model and adds a test for the scenario described above

### DIFF
--- a/measures/business_rules.py
+++ b/measures/business_rules.py
@@ -542,10 +542,14 @@ class ME12(BusinessRule):
     have a relationship with the measure type."""
 
     def validate(self, measure):
+        AdditionalCodeTypeMeasureType = (
+            measure.measure_type.additional_code_types.through
+        )
         if (
             measure.additional_code
-            and not measure.measure_type.additional_code_types.filter(
-                sid=measure.additional_code.type.sid,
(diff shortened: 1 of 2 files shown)


measures/business_rules.py
```diff
@@ -542,10 +542,14 @@ class ME12(BusinessRule):
...
+        AdditionalCodeTypeMeasureType = (
+            measure.measure_type.additional_code_types.through
+        )
         if (
             measure.additional_code
-            and not measure.measure_type.additional_code_types.filter(
-                sid=measure.additional_code.type.sid,
+            and not AdditionalCodeTypeMeasureType.objects.filter(
+                additional_code_type__sid=measure.additional_code.type.sid,
+                measure_type__sid=measure.measure_type.sid,
```
simonwo, line 552: You will also need an `approved_up_to_transaction(self.transaction)` otherwise it will catch deleted ACTMTs as well.

simonwo (approved):

PR #421, opened 2021-11-05
Fix update validity business rule to consider transactions in order.

simonwo:
## Why
We have a business rule that is applied to all models that ensures, per version group, we only have:
- The first version being a CREATE
- Subsequent versions being an UPDATE or DELETE
- No further versions after a DELETE

At the moment this business rule only expects to be run against the very latest version, and does not work when we run it against earlier versions. This happens routinely when we CREATE and then UPDATE a model in the same workbasket, because when we run the rule against the CREATE it picks up the UPDATE too.

Instead, the business rule needs to be transaction-aware – i.e. only consider data that exists as of a certain transaction.

## What

This PR creates a new filter called `up_to_transaction` – which is simplified using @stuaxo's work on partitions. This is broadly the same filter as is required for `approved_up_to_transaction` but split out and able to take a join prefix, so that method is also refactored to use it.

**Please read the commit messages for more context!**

## Checklist
- Requires migrations? No.

### DIFF
--- a/commodities/models/orm.py
+++ b/commodities/models/orm.py
@@ -185,18 +185,14 @@ class GoodsNomenclature(TrackedModel, ValidityMixin):
     def autocomplete_label(self):
         return f"{self} - {self.get_description().description}"
 
-    @property
-    def dependent_measures(self):
+    def get_dependent_measures(self, transaction=None):
         return self.measures.model.objects.filter(
             goods_nomenclature__sid=self.sid,
-        ).approved_up_to_transaction(self.transaction)
+        ).approved_up_to_transaction(transaction)
 
     @property
     def is_taric_code(self) -> bool:
         return self.code.is_taric_code
-
-    def in_use(self):
-        return self.dependent_measures.exists()
 
 
 class GoodsNomenclatureIndent(TrackedModel, ValidityStartMixin):
--- a/common/business_rules.py
+++ b/common/business_rules.py
@@ -491,9 +491,13 @@ class UpdateValidity(BusinessRule):
     """
 
     def validate(self, model):
-        existing_objects = model.__class__.objects.filter(
-            version_group=model.version_group,
-        ).exclude(id=model.id)
+        existing_objects = (
+            model.__class__.objects.filter(
+                version_group=model.version_group,
+            )
+            .exclude(id=model.id)
+            .versions_up_to(self.transaction)
+        )
 
         if existing_objects.exists():
             if model.update_type == UpdateType.CREATE:
--- a/common/business_rules.py
+++ b/common/business_rules.py
@@ -395,7 +395,7 @@ class MustExist(BusinessRule):
             raise self.violation(model)
 
 
-class DescriptionsRules(BusinessRule):
+class ValidityStartDateRules(BusinessRule):
     """Repeated rule pattern for descriptions."""
 
     messages: Mapping[str, str] = {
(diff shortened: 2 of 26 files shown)


commodities/models/orm.py
```diff
@@ -185,19 +185,15 @@ def __str__(self):
     def autocomplete_label(self):
         return f"{self} - {self.get_description().description}"
 
-    @property
-    def dependent_measures(self):
+    def get_dependent_measures(self, transaction=None):
         return self.measures.model.objects.filter(
             goods_nomenclature__sid=self.sid,
-        ).approved_up_to_transaction(self.transaction)
+        ).approved_up_to_transaction(transaction)
```
stuaxo, line 191: :+1:

common/business_rules.py
```diff
@@ -438,6 +440,15 @@ def validate(self, model):
             raise self.generate_violation(model, "start after end")
 
 
+class DescriptionsRules(ValidityStartDateRules):
+    """Repeated rule pattern for descriptions."""
+
+    item_name = "description"
+
+    def get_objects(self, model):
+        return model.get_descriptions(self.transaction)
```
rachaelcodes, line 449: 👍

common/business_rules.py
```diff
@@ -480,9 +491,13 @@ class UpdateValidity(BusinessRule):
     """
 
     def validate(self, model):
-        existing_objects = model.__class__.objects.filter(
-            version_group=model.version_group,
-        ).exclude(id=model.id)
+        existing_objects = (
+            model.__class__.objects.filter(
```
stuaxo, line 495: Not a blocker:

I normally use `__class__` but in this project we seem to have chosen `type(model)` so could be worth using it here.

```python
    type(model).objects.filter(version_group=model.version_group)
```

common/models/records.py
```diff
@@ -68,57 +70,20 @@ def latest_approved(self) -> TrackedModelQuerySet:
...
-                    filter=(
-                        Q(version_group__versions__transaction=transaction)
-                        | Q(
-                            version_group__versions__transaction__workbasket=transaction.workbasket,
-                            version_group__versions__transaction__order__lt=transaction.order,
-                        )
-                        | self.approved_query_filter("version_group__versions__")
+                    filter=self.as_at_transaction_filter(
+                        transaction,
+                        "version_group__versions__",
```
stuaxo, line 86: :100: cheers, I got stuck on some of the complexity round here and what was needed, good to see this shake out into something simpler.

rachaelcodes (approved):

common/models/records.py
```diff
@@ -294,6 +260,38 @@ def approved_query_filter(self, prefix=""):
...
+        if transaction.partition not in TransactionPartition.approved_partitions():
+            this_partition[
+                f"{prefix}transaction__workbasket__id"
+            ] = transaction.workbasket_id
+
+        earlier_parition = {
+            f"{prefix}transaction__partition__lt": transaction.partition,
+        }
+
+        return Q(**this_partition) | Q(**earlier_parition)
```
stuaxo, line 293: Not a blocker:

Might be worth flipping the if statement so it's a positive check `in ... approved_partitions():` and the later Q statement so it's in natural order earlier_partition, this_partition ?

stuaxo (approved):
Pretty mega-change, but sensible ... everything seems a little more convenient than before.

PR #422, opened 2021-11-05
Bugfix for TP-1046

stuaxo:
Part the API I missed for the change from partition scheme to str-referencing-partition-scheme.


simonwo (approved):

stuaxo:
Totally tab-completed into the wrong name with this branch and one test fail, hrmf.

PR #423, opened 2021-11-05
Point to trade org's crispy-forms-gds package.

paulpepper-trade:
# TP-1052 Use the trade org version crispy-forms-gds

## Why
We now have our own, forked version of the crispy-forms-gds Python package. This package fixes issues in the up-stream version of the package and gives us control over future fixes and changes. We should therefore use our own package.

## What
This PR makes no functional change to the application, but substitutes a like-for-like package (with fixes).

## Checklist
- Requires migrations? No
- Requires dependency updates? This is a Tamato dependency update, but doesn't draw in any related dependencies.

Links to relevant material
See: [uktrade / crispy-forms-gds](https://github.com/uktrade/crispy-forms-gds)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ black>=20.8b1
 boto3==1.18.51
 celery[redis]==5.0.5
 coverage[toml]==5.4
-crispy-forms-gds==0.2.2
+crispy-forms-gds @ git+https://github.com/uktrade/crispy-forms-gds.git@b50168d0e23ffacbdd30e7819ec8f9a08e055c8e
 dj-database-url==0.5.0
 django==3.1.13
 django-dotenv==1.4.2


gabelton (approved):

stuaxo (approved):

requirements.txt
```diff
@@ -5,7 +5,7 @@ black>=20.8b1
 boto3==1.18.51
 celery[redis]==5.0.5
 coverage[toml]==5.4
-crispy-forms-gds==0.2.2
+crispy-forms-gds @ git+https://github.com/uktrade/crispy-forms-gds.git@b50168d0e23ffacbdd30e7819ec8f9a08e055c8e
```
stuaxo, line 8: From teams chat:  Not a blocker, but I normally point at the zipball on github so users don't need git installed:

```
https://github.com/uktrade/crispy-forms-gds/zipball/b50168d0e23ffacbdd30e7819ec8f9a08e055c8e#egg=crispy-forms-gds
```

No idea if one is more "correct" than the other.

paulpepper-trade: Sorry @stuaxo  I thought I'd changed this to your suggested non-git option before I'd merged. I'll update with your suggested approach in another PR.

PR #425, opened 2021-11-05
Add pattern that helps manage suspensions implemented with additional codes

simonwo:
This commit adds a new pattern that can `suspend` and `unsuspend` rates on a code using the common pattern of two additional codes. This is used for policies such as INN, TAPI and UK suspensions for which there is no longer an appropriate EU code.

Modifying these regimes is difficult to get right because there are lots of moving parts – one either needs to delete or restore the original MFN or not depending on the presence of other suspensions. There is also a lot of boilerplate
associated with correctly setting up the measure data.

This code is useful immediately for notebook users making data changes and may also be useful later in the UI.

### DIFF
--- a/measures/patterns.py
+++ b/measures/patterns.py
@@ -526,11 +526,14 @@ class SuspensionViaAdditionalCodePattern:
         given duty rate between the two validity dates."""
         if not copy_from:
             # If there is no MFN measure we have a problem because we don't know what
-            # rate to use on the subsequent measure, so just error out.
+            # rate to use on the subsequent measure, so just skip that.
             existing_measures = self.get_mfn_measures(code, validity_start)
-            assert (
-                existing_measures.exists()
-            ), f"No MFN found on code {code} @ {validity_start}"
+            if not existing_measures.exists():
+                self.logger.warning(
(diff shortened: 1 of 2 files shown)


measures/patterns.py
```diff
@@ -391,3 +395,215 @@ def create(self, *args, **kwargs) -> Measure:
...
+    ):
+        """Implement a new suspension on tariffs on the passed code down to the
+        given duty rate between the two validity dates."""
+        if not copy_from:
+            # If there is no MFN measure we have a problem because we don't know what
+            # rate to use on the subsequent measure, so just error out.
+            existing_measures = self.get_mfn_measures(code, validity_start)
+            assert (
+                existing_measures.exists()
+            ), f"No MFN found on code {code} @ {validity_start}"
```
stuaxo, line 533: (probably not specific to this) - we are OK that asserts can be optimised out, e.g. when running with `-O`  ?

stuaxo (approved):

measures/patterns.py
simonwo: Agreed, this should not be an assert. Really it is only a warning – it's possible and maybe sometimes fine that a code does not have an MFN rate, and the pattern already handles that by doing nothing.

PR #427, opened 2021-11-08
Fix required request keywork  using subclasses

paulpepper-trade:
# TP-1057 Fix required keyword-only argument error

## Why
This PR is required to fix a production issue on the footnotes step of Create Measure. Attempting to navigate to that step causes a 500 error response from the server.
This is caused by a absent but required `request` param on the `MeasureFootnotesForm.__init__()`, which is an outcome of using the form for in both create and edit measure cases.

## What
This PR fixes the issue by specialising `MeasureFootnotesForm` as `MeasureUpdateFootnotesForm` and supplying an extra `kwargs` argument into only its `__init__()` function.

## Checklist
- Requires migrations? No


measures/forms.py
```diff
@@ -573,5 +566,26 @@ class MeasureFootnotesFormSet(FormSet):
...
 
 
+class MeasureUpdateFootnotesForm(MeasureFootnotesForm):
+    # Used with edit measure, this form has two buttons each submitting
+    # to different routes: one for submitting to the edit measure view
+    # (MeasureUpdate) and the other for submitting to the edit measure
+    # footnote view (MeasureFootnotesUpdate). This is done by setting the
+    # submit button's "formaction" attribute.
+    # This requires that the path is passed here on kwargs, allowing it to
+    # be accessed and used when rendering the edit forms' submit buttons.
```
andyhd, line 576: Maybe this should be a docstring

andyhd (approved):
Looks good to me, just a suggestion about a comment

PR #433, opened 2021-11-15
Add infrastructure to test class based views render without raising a…

stuaxo:
Fix a bug in measures detail ... but also -

Adds infrastructure to test our class based views and uses it to verify all our detail views return a 200.

While this could be just one big parametrised test, I separated things out: there should be a bit of intentionality to what's in a test, and while this automates things quite a lot - automating things too much can lead to tests where it's hard to tell if it is passing because it should, or not.

This only tests detail views right now, to limit the size of the PR, testing other views should be straightforward in a follow-up PR, it'll be good to see where feedback takes this first.

### DIFF
--- a/additional_codes/tests/test_views.py
+++ b/additional_codes/tests/test_views.py
@@ -1,0 +1,23 @@
+import pytest
+
+from additional_codes.models import AdditionalCode
+from common.tests.util import assert_model_view
+from common.tests.util import get_detail_class_based_view_urls_matching_url
+from common.tests.util import view_urlpattern_ids
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.mark.parametrize(
+    ("view", "url_pattern"),
+    get_detail_class_based_view_urls_matching_url("additional_codes/"),
+    ids=view_urlpattern_ids,
+)
+def test_additional_codes_detail_views(view, url_pattern, valid_user_client):
+    """Verify that additional code detail views are under the url
+    additional_codes/ and don't return an error."""
+    model_overrides = {
+        "additional_codes.views.AdditionalCodeCreateDescription": AdditionalCode,
+    }
+
+    assert_model_view(view, url_pattern, valid_user_client, model_overrides)
--- a/measures/jinja2/includes/measures/tabs/conditions.jinja
+++ b/measures/jinja2/includes/measures/tabs/conditions.jinja
@@ -1,16 +1,16 @@
-{% from 'macros/create_link.jinja' import create_link %} 
+{% from 'macros/create_link.jinja' import create_link %}
 
 <h2 class="govuk-heading-l">Conditions</h2>
-{% if object.get_conditions() %}
+{% if object.get_conditions(get_current_workbasket(request).last_transaction) %}
     {% set table_rows = [] %}
-    {% for condition in object.get_conditions().with_duty_sentence() -%}
+    {% for condition in object.get_conditions(get_current_workbasket(request).last_transaction).with_duty_sentence() -%}
         {{ table_rows.append([
             {"text": condition.condition_code.code ~ condition.component_sequence_number},
             {"text": condition.duty_sentence if condition.duty_sentence else '-'},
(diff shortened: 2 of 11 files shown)


additional_codes/tests/test_views.py
```diff
@@ -0,0 +1,23 @@
...
+    ("view", "url_pattern"),
+    get_detail_class_based_view_urls_matching_url("additional_codes/"),
+    ids=view_urlpattern_ids,
+)
+def test_additional_codes_detail_views(view, url_pattern, valid_user_client):
+    """Verify that additional code detail views are under the url
+    additional_codes/ and don't return an error."""
+    model_overrides = {
+        "additional_codes.views.AdditionalCodeCreateDescription": AdditionalCode,
+    }
```
stuaxo, line 21: Question -

Do we need model_overrides, or can/should we add

`model = AdditionalCode` in the AdditionalCodeCreateDescription and friends ?

measures/jinja2/includes/measures/tabs/conditions.jinja
```diff
@@ -1,16 +1,16 @@
-{% from 'macros/create_link.jinja' import create_link %} 
+{% from 'macros/create_link.jinja' import create_link %}
 
 <h2 class="govuk-heading-l">Conditions</h2>
-{% if object.get_conditions() %}
+{% if object.get_conditions(object.transaction) %}
     {% set table_rows = [] %}
-    {% for condition in object.get_conditions().with_duty_sentence() -%}
+    {% for condition in object.get_conditions(object.transaction).with_duty_sentence() -%}
```
simonwo, line 6: I think this needs to be the last transaction in the workbasket so as to take into account any updates in the workbasket that may have been made. Do we have access to that in this template?

stuaxo: We have the `get_current_workbasket()` (which points at `Workbasket.current`) - I can get the last transaction contained in there - is there any filtering I should do, or is the one with the highest order OK ?

simonwo: You can just do `workbasket.transactions.last()` and the implict `(partition, order)` ordering will automatically be applied, which is what we need.

stuaxo: There are 14 results for `object.transaction` in the templates, and may for other object-dot-something.

Maybe we should be setting `object` to be the latest version of itself ?
(Though right now, I'm not sure where it comes from)

simonwo: Even the latest version of itself is not really enough, because in this case object will be a `Measure`, but the `MeasureConditions` can be updated independently of the parent `Measure`…

All of those `object.transaction` calls will most likely be causing similar bugs

simonwo: Perhaps we could just inject a generic `current_transaction` into every view, which is set to `get_current_workbasket().last_transaction`?

simonwo: And also it's not a big worry, because @rachaelcodes work will make it all unnecessary when implemented ✨

stuaxo: I've set these calls to get_conditions to `get_current_workbasket(request).last_transaction`

The whole "current-ness", bi-temporal stuff, gets incredibly mind-boggling, incredibly quickly - I started thinking about these views being able to display older versions of objects and ended up writing down four questions before deleting them as they would just confuse things.

I'm glad @rachaelcodes working on that as it gets confusing so quickly - no idea why it's easy to imagine many more than 3 spacial dimensions, yet N-dimensional changes in graph-structure seem very tricky - maybe I just need a nice vis of the commodity tree over time, drawn on many many slices.. 

I guess in the temporal PR, or afterwards we can check what data is set in the contexts rendered by the view, to ensure everything matches whatever version of "current" we need.

simonwo (approved):

PR #434, opened 2021-11-16
TP-1081: Models inheriting from TrackedModel should also inherit from DescribedMixin

gabelton:
## Why
- get_descriptions, get_description, and structure_description used to live on TrackedModel. Now that they belong to DescribedMixin, classes that inherit from TrackedModel will also need to inherit from the mixin

## What
- Adds DescribedMixin to lots of model classes and adds a test for TrackedModel.subclasses
- Updates docs, where they point to common.models.records, which no longer exists
- Removes a line from MeasureCondition.description which currently throws an AttributeError

### DIFF
--- a/common/tests/test_models.py
+++ b/common/tests/test_models.py
@@ -562,3 +562,10 @@ def test_structure_description(trackedmodel_factory):
         assert description == model.description
     else:
         assert description == None
+
+
+def test_described(description_factory):
+    description = description_factory.create()
+    described = description.get_described_object()
+
+    assert described.get_description() == description
--- a/common/tests/test_models.py
+++ b/common/tests/test_models.py
@@ -551,7 +551,10 @@ def test_structure_description(trackedmodel_factory):
     model = trackedmodel_factory.create()
     description = model.structure_description
 
-    if "structure_description" not in model.__class__.__dict__:
+    if description:
+        assert type(description) == str
+
+    if "structure_description" in type(model).__dict__:
         pass
     elif hasattr(type(model), "descriptions") and model.get_descriptions().last():
         assert description == model.get_descriptions().last().description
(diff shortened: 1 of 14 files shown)


common/tests/test_models.py
```diff
@@ -545,3 +545,17 @@ def test_save_drafts_transaction_updates(unordered_transactions):
         == unordered_transactions.new_transaction.order
     )
     assert_transaction_order(Transaction.objects.all())
+
+
+def test_structure_description(trackedmodel_factory):
+    model = trackedmodel_factory.create()
+    description = model.structure_description
+
+    if "structure_description" not in model.__class__.__dict__:
```
stuaxo, line 554: I'm not quite sure what branch tests, is it another way of saying hasattr("structure_description"), though that must be the case because of the previous line ?

re: `model.__class__`, I normally do this in other projects, but this one seems to have settled on `type(model)`

I wonder if it's worth checking if the model_description is a string while we're checking it ?

stuaxo (approved):

common/tests/test_models.py
gabelton: It's because a model like Regulation overrides structure_description property, so follows different logic, which I didn't want to get into in this test. I've added another small test for it. Possibly trivial...

simonwo (changes requested):
Why are we adding `DescribedMixin` to things that don't have associated descripiton models?

measures/models.py
```diff
@@ -810,9 +810,6 @@ def description(self) -> str:
         elif self.is_certificate_required():
             out.append("On presentation of no certificate,")
 
-        if self.reference_price_string:
-            out.append(f"If reference price > {self.reference_price_string},")
-
```
simonwo, line 815: This looks unrelated? `reference_price_string` can sometimes be added by the queryset.

simonwo (approved):

stuaxo:
I tried rebasing this on top of TP-1069 and can confirm the detail view tests pass, with:

`pytest -v --no-cov -n0 -s -k detail_view`

measures/models.py
gabelton: Just added it back in now but with a hasattr check

PR #435, opened 2021-11-16
Fix measure effective validity calculation

simonwo:
## Why
A measure can have either an explicit or implicit end date, where the implicit version is specified by the regulation. Our previous algorithm for finding the implicit end date was not correct, because it reported lots of EU measures as still being in force when clearly (as enforced by ME32) they could not be.

## What
This commit fixes the algorithm, and also ensures that when the value is requested from a measure for which it was not pre-loaded, it is calculated in the same way.

The new algorithm has been found through examination of the existing data and comparison with the Online Tariff service as a known good source. E.g. I found EU measures that the system considers still active using:

```python
Measure.objects
.with_effective_valid_between()
.filter(
    sid__lt=20000000, # exclude UK measures
    db_effective_valid_between__contains=date.today(),
)
```

With this new algorithm, the above query returns a small number of measures which it would make sense if still active e.g. measures on Meursing codes. There are still some anomalies, but the situation is much improved.

### DIFF
--- a/common/tests/factories.py
+++ b/common/tests/factories.py
@@ -47,6 +47,10 @@ def date_ranges(name):
     return factory.LazyFunction(lambda: getattr(Dates(), name))
 
 
+def end_date(name):
+    return factory.LazyFunction(lambda: getattr(Dates(), name).upper)
+
+
 class ValidityFactoryMixin(factory.django.DjangoModelFactory):
     valid_between = date_ranges("no_end")
 
@@ -210,9 +214,6 @@ class RegulationFactory(TrackedModelMixin, ValidityFactoryMixin):
     regulation_id = factory.Sequence(lambda n: f"R{Dates().now:%y}{n:04d}0")
     approved = True
     role_type = 1
-    valid_between = factory.LazyAttribute(
-        lambda o: Dates().no_end if o.role_type == 1 else None,
-    )
     community_code = 1
     regulation_group = factory.SubFactory(
         RegulationGroupFactory,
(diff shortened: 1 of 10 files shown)


common/tests/factories.py
```diff
@@ -247,11 +262,12 @@ class Meta:
...
 
     target_regulation = factory.SubFactory(
-        RegulationFactory,
+        BaseRegulationFactory,
         transaction=factory.SelfAttribute("..transaction"),
     )
     enacting_regulation = factory.SubFactory(
-        RegulationFactory,
+        ModificationRegulationFactory,
+        amendment=None,
```
rachaelcodes, line 270: Is it worth adding a test with a different kind of amendment, if nothing else so that we can see how it works?

measures/querysets.py
```diff
@@ -151,76 +160,80 @@ def with_duty_sentence(self) -> QuerySet:
...
+        There are four ways in which measures can get end dated:
 
-        Annotate the queryset with the db_effective_valid_between based on the regulations and measure.
+        1. Where the measure is given an explicit end date on the measure record
+           itself
+        2. Where the measure's generating regulation is a base regulation, and
+           the base regulation is end-dated
+        3. Where the measure's generating regulation is a modification
+           regulation, and the modification regulation is end-dated
+        4. Where the measure's generating regulation is a modification
```
gabelton, line 171: Sorry, just for my own understanding: what's the difference between 3 and 4 ?

measures/querysets.py
```diff
@@ -151,76 +160,80 @@ def with_duty_sentence(self) -> QuerySet:
 class MeasuresQuerySet(TrackedModelQuerySet, DutySentenceMixin):
     def with_effective_valid_between(self):
         """
-        In many cases the measures regulation effective_end_date overrides the
-        measures validity range.
+        There are four ways in which measures can get end dated:
```
gabelton, line 163: should this be five ?

measures/querysets.py
```diff
@@ -151,76 +160,80 @@ def with_duty_sentence(self) -> QuerySet:
...
-                   END AS "db_effective_valid_between"
-              FROM "measures_measure"
-             INNER JOIN "regulations_regulation"
-                ON "measures_measure"."generating_regulation_id" = "regulations_regulation"."trackedmodel_ptr_id"
-             INNER JOIN "common_trackedmodel"
-                ON "measures_measure"."trackedmodel_ptr_id" = "common_trackedmodel"."id"
+        Numbers 2–5 also have to take account of the "effective end date" which
+        if set should be used over any explicit date. The effective end date is
+        set when other types of regulations are used (abrogation, prorogation,
+        etc).
```
rachaelcodes, line 180: I really like the comments in this file - they make the logic much easier to follow.

rachaelcodes (approved):
👍 The comments make the logic very clear.

measures/querysets.py
simonwo: Oops, that's a typo!

measures/querysets.py
simonwo: Yes!

common/tests/factories.py
simonwo: Ah, so, an `Amendment` represents the link between a `BaseRegulation` and `ModificationRegulation` (both of them are just `Regulation` objects with an appropriately-set `role_type`). So the `enacting_regulation` field for a `Amendment` can only ever be a `ModificationRegulation`.

Does that remove the need in your head for another test? Or did I not understand? :)

PR #436, opened 2021-11-17
TP-935 Standardise transaction time

rachaelcodes:
# TP-935 Transaction-time context for queries

## Why

We have a lot of database queries that specify the transaction-time (one of the time
axes of our bitemporal data) to filter results. The most common case is to filter by the
latest approved transaction including draft changes in the current workbasket. This
often results in having to pass the transaction down long function call chains.

We also need to perform queries against different transactions in the same code, and so
we need a way to temporarily override the current transaction.

## What
* Add thread-local storage using [Django-CRUM](https://github.com/ninemoreminutes/django-crum) to keep a reference to the current transaction, so that database querying code can access it directly without having to pass it around.
* Add a ``contextmanager``, which sets the thread-local stored transaction to a new value within the scope of a ``with`` block, then resets it afterwards.
* Add a lazily evaluated query value, this allows defining a query once and then reusing it with different transactions without having to build a separate query for each transaction.

See: [ADR 0014 - Simplify Current Transaction Queries](https://github.com/uktrade/tamato/blob/master/docs/source/adr/0014-simplify-current-transaction-queries.rst)


common/models/tracked_qs.py
```diff
@@ -26,21 +31,62 @@ class TrackedModelQuerySet(
...
-        """
-        return self.filter(is_current__isnull=False).exclude(
-            update_type=UpdateType.DELETE,
+    def latest_approved(
+        self,
+        transaction: Optional[Transaction] = None,
+    ) -> TrackedModelQuerySet:
+        """Get the approved versions of the model, or the latest draft versions
+        if they exist within a transaction preceding (and including) the given
+        transaction in the workbasket of the given transaction."""
```
stuaxo, line 40: It may be possible to improve clarity on here by splitting this into parts about approved vs draft.
The `given transaction` part may need tweaking as this is an optional parameter.

Almost tempting to resort to some ASCII art in these.

common/tests/test_models_utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+    assert (
+        Transaction.objects.get(
+            id=CurrentTransactionId().value,
+        ).workbasket
+        == session_workbasket
+    )
+
+
+def test_current_workbasket_id(client, session_workbasket):
```
stuaxo, line 29: This test should create at least one other workbasket.

If the only workbasket around is the session_workbasket, then this test may pass even if something wonky happens in the implementation.

stuaxo:
crum needs adding to requirements.txt

common/models/utils.py
```diff
@@ -0,0 +1,72 @@
...
+    def get_value(self):
+        request = get_current_request()
+        if request is None:
+            return 1
+
+        from workbaskets.models import WorkBasket
+
+        transaction = WorkBasket.get_current_transaction(request)
+        if transaction:
+            return transaction.order
```
simonwo, line 71: This will most likely need to become `(transaction.partition, transaction.order)` in order to be used in `approved_up_to_transaction` and friends.

common/tests/test_models_utils.py
```diff
@@ -0,0 +1,51 @@
...
+pytestmark = pytest.mark.django_db
+
+
+# TODO: this checks that the transaction is in the workbasket, not necessarily that it's the right one
+def test_current_transaction_id(
+    client,
+    session_workbasket,
+):
+
+    set_current_request(client)
```
simonwo, line 19: We will need something like this method but a bit lower level that we are able to invoke from non-test code, i.e. we need to be able to set `CurrentTransactionId` without having a request object.

Why? Because when we run business rule cleans, we run them in order against each transaction in turn. So we need to be able to say "set the `current transaction` to be transaction X". When we do that, we most likely won't have a request that is asking for that either because it will be checking rules in a background task or because we are using it from a Jupyter notebook. Also, the workbasket is what we will get from the request, but we will need to have more precise scoping over which transaction in that request is in use.

Thoughts?

andyhd: We can set the current transaction using `queryset.latest_approved(transaction)` - if there is no request, what does "current transaction" mean? If it isn't the last approved transaction, how can we determine it?

More precise targeting of transactions in the session workbasket is currently impossible, because we don't store a "current transaction" in the session. What are the circumstances in which the current transaction is not the last transaction in the workbasket?

simonwo: I suppose I was imagining this mechanism as more general than just storing the workbasket in the session, and would allow us to change code that looks like:

```python
Measure.objects.approved_up_to_tranasction(transaction).filter(…)
```

into

```python
with current_transaction_as(transaction):
    Measure.current_objects.filter(…)
```

I.e. so we don't have to pass the transaction into every function that ever makes a database call, we can just define it once at quite a high level, and then have it automatically picked up by any lower level function.

It would have made fixes to things like `in_use` much easier because we wouldn't have needed to refactor it to take a `transaction` parameter.

andyhd: Ok, I've just pushed a change with a context manager in `common/models/utils.py` which allows you to do the following:
```python
from common.models.utils import override_current_transaction
from footnotes.models import Footnote


qs = Footnote.objects.filter(footnote_type__footnote_type_id="CA")

# this current() name is WIP, I'm not sure if it can replace approved_up_to_transaction()
qs = qs.current()

print(qs.values("pk").first())
# {"pk": 10000}

tx = WorkBasket.objects.get(pk=63).transactions.last()

with override_current_transaction(tx):
    print(qs.values("pk"))
# {"pk": 9951551}
```

stuaxo:
@andyhd nice last day database poking :)

common/models/tracked_qs.py
```diff
@@ -26,6 +33,88 @@ class TrackedModelQuerySet(
...
+            ),
+        )
+
+    def annotate_overridden_in_workbasket(
+        self,
+        transaction: Transaction | None = None,
+    ) -> TrackedModelQuerySet:
+        """Annotate results with whether they are overridden in the current
+        workbasket, or in the workbasket containing the specified Transaction if
+        given."""
```
stuaxo, line 78: ^ This could do with a little elaboration.

common/models/utils.py
```diff
@@ -0,0 +1,54 @@
...
+from django.db.models import Value
+
+
+class LazyValue(Value):
+    """
+    Lazily evaluated Value.
+
+    Calls a function passed to the constructor when ``value`` property is
+    accessed. This allows the value to change with each access.
+    """
```
stuaxo, line 14: A short sentence on when this is useful could be handy.

common/tests/test_models_utils.py
```diff
@@ -0,0 +1,58 @@
...
+def test_lazy_value():
+    counter = iter(range(10))
+    val = LazyValue(get_value=lambda: next(counter))
+
+    assert val.value == 0
+    assert val.value == 1
+    assert val.value == 2
+
+
+def test_get_current_transaction_from_thread_locals():
```
stuaxo, line 24: Adding a check for the value of `thread_locals.transaction` will make this more robust.

It could be worth using a fixture so it can be checked with a non-None and None value.

common/models/utils.py
```diff
@@ -0,0 +1,54 @@
...
+def get_current_transaction():
+    tx = getattr(_thread_locals, "transaction", None)
+
+    if tx is None:
+        request = get_current_request()
+
+        if request:
+            from workbaskets.models import WorkBasket
+
+            return WorkBasket.get_current_transaction(request)
```
stuaxo, line 35: Off topic and (not a comment on this PR):   ^ `WorkBasket.get_current_transaction(...)`  seems really easy to make unhappy during unit tests, probably needs some thought.

common/tests/test_models_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+def test_get_current_transaction_from_request(client):
+    set_current_request(client)
+
+    with mock.patch("workbaskets.models.WorkBasket") as WorkBasket:
+        tx = get_current_transaction()
+
+        WorkBasket.get_current_transaction.assert_called_once()
+
+        assert tx == WorkBasket.get_current_transaction.return_value
```
stuaxo, line 37: Looks good, probably wants an assert on the value of `tx` as well - if this always returned None, the test would still pass.

stuaxo:
Not a blocker since all the tests are small .. but it still may be worth adding docstrings to them - once people tweak them, it can be hard to know what the original intention was.

stuaxo (approved):

simonwo:
Superseded by #486.

PR #438, opened 2021-11-18
Fix: Unordered Track Models are constrained to those linked to the TransactionQueryset

kintisheff:
# TP-1082 Unordered Track Models are constrained to those linked to the TransactionQueryset

## Why
At the moment `common.models.transactions.TransactionQueryset.unordered_tracked_model` returns any and all models in the database. 
This is unintended and the filter condition `transaction__in=self` applied later in `ordered_tracked_model` should be moved to this method instead.

## What
Move the  filter condition `transaction__in=self` to the unordered method

### DIFF
--- a/common/models/transactions.py
+++ b/common/models/transactions.py
@@ -70,20 +70,16 @@ class TransactionsAlreadyApproved(Exception):
 class TransactionQueryset(models.QuerySet):
     def unordered_tracked_models(self):
         """Usually 'ordered_tracked_models' is required."""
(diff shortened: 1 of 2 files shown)


stuaxo (approved):
:+1: 

Ideally should have a test, OTOH I should have written it in the first place :grimacing:

kintisheff:
> 👍
> 
> Ideally should have a test, OTOH I should have written it in the first place 😬

oh yes i'll add it now actually before merge!

PR #439, opened 2021-11-18
TP-1083 Add measures-as-defined style reports for changed measures in a workbasket

kintisheff:
# TP-1083 Add measures-as-defined style reports for changed measures in a workbasket

## Why
In some cases, e.g. HS22 commodity code changes, it is useful to produce reports on affected and related measures for any current or past workbasket in the Measures-as-Defined format familiar to many end users.
The two reports enabled with this ticket are:
1. Affected Measures: Any changed measures in the workbasket, in Measures-as-Defined format
2. Related Measures: A full list of measures linked to any changing commodities in the workbasket, in MaD format

## What
- Add a new `reports.py` module in the `importer` app to provide this functionality
- Add a mock_responses fixture to `commodities/conftest.py` for use in future commodity import handler tests that involve reports.

### DIFF
--- a/importer/reports.py
+++ b/importer/reports.py
@@ -5,6 +5,7 @@ import logging
 from collections import OrderedDict
 from contextlib import closing
 from typing import Any
+from typing import Dict
 from typing import Generator
 from typing import Iterable
 from typing import Optional
@@ -96,8 +97,8 @@ MAPPING_MAD_CMT = (
 )
 
 
-TMadItemGenerator = Generator[None, OrderedDict[str, Any], None]
-TMadItemCollection = Tuple[OrderedDict[str, Any]]
+TMadItemGenerator = Generator[None, Dict[str, Any], None]
+TMadItemCollection = Tuple[Dict[str, Any]]
 TGoods = Tuple[GoodsNomenclature]
 
 # Acronyms:
--- a/importer/reports.py
+++ b/importer/reports.py
@@ -1,0 +1,276 @@
+from __future__ import annotations
+
+import csv
+import logging
+from collections import OrderedDict
+from contextlib import closing
+from typing import Any
+from typing import Generator
+from typing import Iterable
+from typing import Optional
+from typing import Tuple
+from typing import Union
+
+import requests
+
+from commodities.models.orm import CommodityCode
+from commodities.models.orm import GoodsNomenclature
+from common.validators import UpdateType
+from measures.models import Measure
+from workbaskets.models import WorkBasket
+
+logger = logging.getLogger(__name__)
+
+API_PROTOCOL = "https"
+API_HOST = "data.api.trade.gov.uk"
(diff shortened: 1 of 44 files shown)


importer/reports.py
```diff
@@ -0,0 +1,277 @@
...
+    if not isinstance(value, str):
+        return value
+    return value.replace("#NA", "")
+
+
+def get_filtered_mad_items(
+    key: str,
+    values: Iterable[str],
+    as_defined: Optional[bool] = True,
+) -> TMadItemGenerator:
```
stuaxo, line 120: Probably at worth documenting as_defined, at least.

importer/reports.py
```diff
@@ -0,0 +1,277 @@
...
+
+        for item in filtered_reader:
+            regulation = (
+                Measure.objects.filter(
+                    sid=item[MAD_MEASURE_SID],
+                )
+                .order_by("id")
+                .last()
+                .generating_regulation
+            )
```
stuaxo, line 137: This may become a bit query heavy.

I can't tell if that's an issue as I'm not fully sure what this does from the docstring.

importer/reports.py
```diff
@@ -0,0 +1,277 @@
...
+
+            yield mapped_item
+
+
+def get_goods_mad_items(
+    goods: Iterable[GoodsNomenclature],
+    as_defined: Optional[bool] = True,
+) -> TMadItemGenerator:
+    """Returns CM rows that match a set of commodity codes."""
```
stuaxo, line 153: CM = Commodity ?

stuaxo (approved):
Approved

Note - it does look like it may be easy to generate a lot of queries with some of these interfaces.
Some of this could be ameliorated by prefetching or using select_related, for various things - e.g. tracked_models, and it may be worth noting that in docstrings.

I'm not sure how many records these interfaces will be used against, if it's not many - then OK, if not it may be worth casting an eye over these with an aim of reducing the query count.

simonwo:
My main concern is that we are reaching out to our open data in order to generate these results. Not bad in principle (in fact, great to reuse it like this!). But I am concerned about how the open data lags our own database sometimes by several days, and only includes data that has been validated by CDS. Will that make it possible for the change report to have rows missing because they aren't in the open data yet, especially if we are doing imports much more regularly (e.g. daily)?

I wonder if a more strategic solution would be to generate the report from our own db. It's not trivial in SQL, but doing it with Django's help should make it a bit easier. But clearly as a tactical solution for now, this works and we probably shouldn't do anything different this side of 2022. 

So would you agree that we should look to replace this at some point to avoid the above problems?

kintisheff:
> 

I agree with this concern. I propose a 2-step approach:
1. Update MoD data now (as soon as Paul finishes updating imports on prod I guess?)
2. Create a ticket in Jira to sort this more strategically

Vote

PR #440, opened 2021-11-18
TP-1084 Verify list views return 200 and are in available on the expected URLs.

stuaxo:
While there is a bit of duplication here, it's probably better than having one big test in common.
It's also useful to separate these from the detail view tests, even if the code is pretty similar.

In future these tests, and the view tests should probably be refactored and expanded to verify
the expected data is included in the output.

# TP-1084

## Add tests to ensure list views don't fail.
This was split out from TP-1067.
The tests here ensures list views are at the expected locations, and render without raising any exceptions (the simplest / dumbest but still useful test).

## What
This adjusts the infrastructure built for TP-1067, (which is does things like grab every class based view and verify the URL matches some starting string) and generalizes a little to also work with list views.

In theory this could all be one big test, but splitting over our apps, and different kinds of views is probably easier on the developer when / if something goes wrong.

It is expected that in future we may wish to do further verification - that expected data is output, for instance.

### DIFF
--- a/common/tests/util.py
+++ b/common/tests/util.py
@@ -20,6 +20,7 @@ from django.db.models.base import ModelBase
 from django.template.loader import render_to_string
 from django.urls import get_resolver
 from django.urls import reverse
+from django_filters.views import FilterView
 from freezegun import freeze_time
 from lxml import etree
 
@@ -31,7 +32,6 @@ from common.tests import factories
 from common.util import TaricDateRange
 from common.util import get_accessor
 from common.util import get_field_tuple
-from common.views import TrackedModelDetailMixin
 
 INTERDEPENDENT_IMPORT_IMPLEMENTED = True
 UPDATE_IMPORTER_IMPLEMENTED = True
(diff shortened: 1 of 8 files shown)


common/tests/util.py
```diff
@@ -236,32 +236,43 @@ def prefix(url_pattern=None, **kwargs):
...
-def get_detail_class_based_view_urls_matching_url(url_start, must_contain_views=None):
+def get_class_based_view_urls_matching_url(
+    url_start,
+    prefix=lambda **kwargs: True,
+    assert_contains_view_classes=None,
+):
     """
-    Return class based views, and url patterns, where the url_pattern starts
-    with a supplied string.
+    Return a list of typles of (view, url pattern), where url_pattern starts
```
rachaelcodes, line 245: tuples?

common/tests/util.py
```diff
@@ -236,32 +236,43 @@ def prefix(url_pattern=None, **kwargs):
...
+    with the supplied string.
+
+    Users may pass optionally must_contain_views to verify if certain views were found.
```
rachaelcodes, line 248: Has this changed to `assert_contains_view_classes`?

rachaelcodes (approved):
LGTM, a couple of the comments might need a little tweak (mentioned in my other comments).

common/tests/util.py
stuaxo: Oh, heh - good spot.

paulpepper-trade (approved):

PR #441, opened 2021-11-18
TP-1087 Find a version group created in the same draft workbasket

kintisheff:
# TP-1087 Find a version group created in the same draft workbasket

## Why
Historically, `TrackedModel._get_version_group()` has only looked at `latest_approved()` versions of a model when looking for a version group.
However, when importing a large taric envelope, it is possible to have the same model created and then updated in the same workbasket.
This ticket is to enable that.

## What
Enhance `TrackedModel._get_version_group()` to support the extra lookup functionality

### DIFF
--- a/common/models/trackedmodel.py
+++ b/common/models/trackedmodel.py
@@ -210,7 +210,30 @@ class TrackedModel(PolymorphicModel):
     def _get_version_group(self) -> VersionGroup:
         if self.update_type == validators.UpdateType.CREATE:
             return VersionGroup.objects.create()
-        return self.get_versions().latest_approved().last().version_group
+
+        latest_version = self.get_versions().latest_approved().last()
+
+        if not latest_version:
+            # An object may be created and deleted/updated in the same workbasket.
+            # If the workbasket status is not WorkflowStatus.PUBLISHED,
+            # then latest_approved() in the above line of code will return None.
+            # Trying to get the version group off that will throw an exception.
+            # The extra bit of logic below deals with such cases
+            # It will attempt to find the corresponding CREATE record
+            # in the current workbasket and return that as the latest_version.
+            try:
+                latest_version = [
+                    record
+                    for transaction in self.transaction.workbasket.transactions.all()
+                    for record in transaction.tracked_models.all()
+                    if type(record) == type(self)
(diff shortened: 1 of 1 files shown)


simonwo (commented):
I literally don't understand why this method exists… 

Version groups don't change on a transaction by transaction basis _by definition_ – they are one thing that identifies a model from one transaction to the next. So the version group of version 1 is the same version group of version 1,000,000.

So can't this just be `return self.version_group`.

stuaxo (changes requested):
Might have some feedback ... will fill this out more usefully shortly, or approve.

common/models/trackedmodel.py
```diff
@@ -210,7 +210,30 @@ def get_versions(self):
...
+            # in the current workbasket and return that as the latest_version.
+            try:
+                latest_version = [
+                    record
+                    for transaction in self.transaction.workbasket.transactions.all()
+                    for record in transaction.tracked_models.all()
+                    if type(record) == type(self)
+                    if record.update_type == UpdateType.CREATE
+                    if record.get_identifying_fields() == self.get_identifying_fields()
+                ][0]
```
stuaxo, line 232: ^ this looks like it can be a filter on workbasket.transactions.all() - will update with something closer to an implementation.

kintisheff:
> I literally don't understand why this method exists…
> 
> Version groups don't change on a transaction by transaction basis _by definition_ – they are one thing that identifies a model from one transaction to the next. So the version group of version 1 is the same version group of version 1,000,000.
> 
> So can't this just be `return self.version_group`.

Per our conversation, the method is useful in dealing with `UpdateType.CREATE` vs other `UpdateType`-s; however, `self.version_group` indeed should be all we need in the cases of `UPDATE` and `DELETE` and that's a nice resolution

kintisheff:
> Might have some feedback ... will fill this out more usefully shortly, or approve.

sounds good, keep in mind we've simplified this also

stuaxo (approved):

stuaxo (approved):
Not having that block is even better.

simonwo (approved):

stuaxo:
If this really should only be called on draft workbaskets it may be worth gating on that.

PR #442, opened 2021-11-18
TP-1088 Support reordering when workbasket transaction order is not contiguous

kintisheff:
# TP-1088 Support reordering when workbasket transaction order is not contiguous

## Why
In some cases, e.g. we import commodity code changes, we may add preemptive transactions to the draft import workbasket, e.g. to deal with side effects on other models and avoid business rule violations.
This ticket enhances workbasket.approve(), which is also responsible for reordering workbasket transactions, to support non-contiguous transaction orders in a workbasket.

## What
Add a new method to TransactionQueryset that assumes all responsibility for reordering workbasket transactions

### DIFF
--- a/common/models/transactions.py
+++ b/common/models/transactions.py
@@ -8,7 +8,6 @@ from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
-from django.db.models import F
 from django.db.transaction import atomic
 from django_fsm import FSMIntegerField
 from django_fsm import transition
@@ -16,6 +15,7 @@ from django_fsm import transition
 from common.business_rules import BusinessRuleChecker
 from common.business_rules import BusinessRuleViolation
 from common.models.mixins import TimestampedMixin
+from common.renderers import counter_generator
 
 logger = getLogger(__name__)
 PREEMPTIVE_TRANSACTION_SEED = -100000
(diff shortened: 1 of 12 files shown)


common/tests/test_transactions.py
```diff
@@ -86,3 +86,37 @@ def test_ordered_tracked_models_are_sorted_on_partition_and_order():
...
+    transaction_second = factories.TransactionFactory(order=1)
+
+    ids = (transaction_first.id, transaction_second.id)
+    qs = Transaction.objects.filter(id__in=ids)
+
+    qs.preorder_negative_transactions()
+    assert all(tx.order < 0 for tx in qs.all())
+
+
+def test_ordering_of_querysets_with_negative_transaction_orders():
```
gabelton, line 105: Should this test also show that order 0 belongs to transaction_first and 1 to transaction_second ? Or is that implied somewhere that I'm missing ?

kintisheff: good point, i'll enhance the test to assert that explicitly!

common/models/transactions.py
```diff
@@ -93,6 +94,21 @@ def approved(self):
...
+        Negative order numbers happen in preemptive transactions, e.g. when we
+        import commodity code changes
+        """
+        if self.count() and self.order_by("order").first().order < 0:
+            order = PREEMPTIVE_TRANSACTION_SEED
+            for tx in self.order_by("order").all():
+                order += 1
+                tx.order = order
+                tx.save()
+
```
simonwo, line 111: It looks like this method mostly just duplicates [this logic](https://github.com/uktrade/tamato/pull/442/files#diff-5ba262a8622d0afa3970dd730fa419287c5798279b57190fb4e0210239162cc7R166)… is it possible to just modify that other method to handle negative orders instead of adding a whole new thing?

kintisheff: it seems like the [this logic] link in your comment is pointing to the same method as the one you are commenting on. did you mean to link elsewhere?

simonwo: Yes, this: https://github.com/uktrade/tamato/blob/e4687930156d4e70c8524fdc82eb5ce8b2fedbcd/common/models/transactions.py#L150

kintisheff: i see what you mean; no i would rather keep the preorder method separate - makes it more testable and all

simonwo: I think that more testability is a good goal. Hmm, I'm trying to balance that between the fact we now have the same logic in two places and the reduction in maintainability because of that. E.g. we just modified the other method to be more performant with the use of a queryset instead, now we'd have to do that in two places.

Could we just extract the original logic into it's own method that could be unit-tested? And then modify it to handle the negative orders? That way we get the best of both worlds?

kintisheff: ^ i think i agree with that suggestion yes - put all ordering in one place! will do

kintisheff: save_draft implicitly embedding that responsibility is not right anyway!

kintisheff (commented):
based on our thread on teams, `django_cte` seems to have a reported bug when compiling the SQL for update queries (it appears not to be formatting one of the parameters). so won't use that for dynamic updates after all (unfortunately!)

simonwo (approved):

PR #443, opened 2021-11-19
Add ADR 15 around formalising immutability in our data.

stuaxo:
This ADR is about making our Transactions read-only, by default, and adding MutableTransaction as a proxy for the times we need to `.save` and `objects.update`.

## Why
Conventions are great, but what happens after we have all gone ?
This helps formalise the convention behind an API, while we can't control what future developers will do, we can make people stop for thought.

Separating reading from writing makes it very explicit that we are trying to be immutable.

https://github.com/uktrade/tamato/blob/ADR-15-make-transaction-immutable-by-default/docs/source/adr/0015-make-transaction-immutable-by-default.rst

### DIFF
--- a/docs/source/adr/0015-make-transaction-immutable-by-default.rst
+++ b/docs/source/adr/0015-make-transaction-immutable-by-default.rst
@@ -1,0 +1,49 @@
+.. _15-record-architecture-decisions:
+
+15. Make Transactions Immutable by Default
+==========================================
+
+Date: 2021-11-19
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+Data in Tamato is written as an set of "changes to things", stored in the "Tracked Model" model grouped by the Transaction model.
+The exception to this rule are Tracked Models with Transactions in the DRAFT partition, which may be modified.
+
+Currently immutability is a convention, but unenforced.
+This ADR attempts to make the places where writing happens more visible and default to being read-only elsewhere.
+
+
+Decision
+--------
+
+1. Use typing to make read-only transactions the default:
+
+i.   MutableTransaction will be added as a proxy to the Transaction class.
+
+MutableTransaction will be used in places where we need to modify transactions, i.e. save_drafts.
+
+This will be a proxy to the existing Transaction class, so will work in the same way as Transaction does today.
+
+
+ii.  Transaction to become read-only by default.
+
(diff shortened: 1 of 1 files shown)


docs/source/adr/0015-make-transaction-immutable-by-default.rst
```diff
@@ -0,0 +1,49 @@
...
+Decision
+--------
+
+1. Use typing to make read-only transactions the default:
+
+i.   MutableTransaction will be added as a proxy to the Transaction class.
+
+MutableTransaction will be used in places where we need to modify transactions, i.e. save_drafts.
+
+This will be a proxy to the existing Transaction class, so will work in the same way as Transaction does today.
```
paulpepper-trade, line 32: Is it a little more than a proxy, since 'MutableTransaction` would override `Transaction.save()` and `Transaction.update()` in order to get the mutability behaviour?

simonwo (commented):
Encoding conventions is good!

Immutability is somewhat enforced: `can_write` on `TrackedModel` prevents saving changes to a model that already exists. It can be overridden by specifying `force_write=True`. For me this has worked quite well because because with the primitives we have such as copy sometimes the easiest thing to do is copy a model and then force a change to it afterwards.

Is subtyping definitely the right answer? Could you add some discussion to the ADR about what the trade-off might be compared with just using a `can_write` method?

Are we sure that all of the data associated with the transaction will always be immutable? What about things like the error state – there is discussion elsewhere about how transactions can become errored even after they are approved, eg if CDS disagrees with us about their validity. Our system will never be bug free so we should always assume this is possible! 

So maybe it’s just the contents of the transaction or only some of its fields that are immutable? In which case does calling something a mutable or an immutable transaction not really tell the whole story?

stuaxo:
can_write is interesting, I didn't know about it - I'll will need to have a play, and think about the implications.

This is more about the Transaction side than the TrackedModel side, but obviously they are related.

I guess something like new_transaction() will give you a MutableTransaction, vs the default queryset which may give you Transactions.

docs/source/adr/0015-make-transaction-immutable-by-default.rst
stuaxo: This is a good point, the writing here could definitely be clearer.

stuaxo:
Closing for now, this may be back with a more hardcore solution such as readonly cursors by default :)

PR #445, opened 2021-11-22
TP-990 Common docstrings

andyhd:
# TP-990 Common docstrings

## Why

Docstrings are valuable both for generating documentation automatically with Sphinx, and for use in some IDEs contextual help. Code can be self-documenting, but it is difficult to make the intention of the code author or the reasons for the decisions they take immediately comprehensible. In these cases, docstrings and inline comments help express that context.

## What

* Add missing docstrings to the common app code
* Tidy up envvars documentation
* Simplify testing documentation to reduce the detail in the introduction, hopefully making it easier to read

### DIFF
--- a/common/app_config.py
+++ b/common/app_config.py
@@ -3,7 +3,6 @@ from importlib import import_module
 
 from django.apps import AppConfig
 
-
 logger = logging.getLogger(__file__)
 
 
@@ -12,7 +11,19 @@ IMPORT_HANDLER_NAME = "import_handlers"
 
 
 class CommonConfig(AppConfig):
+    """
+    Extends the default Django AppConfig to load importer parser and handler
+    modules for the app.
+
+    Each app that defines tariff component models may also provide parser and
+    handler modules to be used by the importer when reading the TARIC XML that
+    describes those models. This class encapsulates the loading of those
+    modules, so that we do not have to write the same code in each app's
+    AppConfig.ready method.
+    """
+
     def ready(self):
+        """Load importer parser and handler modules, if they exist."""
         modules_to_import = [
             f"{self.name}.{IMPORT_PARSER_NAME}",
             f"{self.name}.{IMPORT_HANDLER_NAME}",
--- a/common/business_rules.py
+++ b/common/business_rules.py
@@ -1,4 +1,6 @@
 """Base classes for business rules and violations."""
+from __future__ import annotations
+
 import logging
 from datetime import date
 from datetime import datetime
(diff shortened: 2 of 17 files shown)


stuaxo:
@andyhd thanks for doing this tireless docstring-ifying.

common/app_config.py
```diff
@@ -1,9 +1,10 @@
+"""common app config."""
```
stuaxo, line 1: Feedback .. possibly for whoever picks this up after Andy, or for andy :)

This one may be a docstring too far, since it doesn't add more than the filename already tells us.

common/business_rules.py
```diff
@@ -26,6 +28,13 @@ class BusinessRuleViolation(Exception):
     """Base class for business rule violations."""
 
     def default_message(self) -> Optional[str]:
+        """
+        Get the first paragraph of the class docstring (if it exists) to use as
+        the error message.
+
+        :rtype Optional[str]: The error message
```
stuaxo, line 35: I think the `The error message` is something for an `:return` ?

common/business_rules.py
```diff
@@ -200,8 +250,13 @@ def validate(self, model):
     return decorator
 
 
-def skip_when_update_type(cls, update_types):
-    """Skip business rule validation for given update types."""
+def skip_when_update_type(cls: Type[BusinessRule], update_types: Iterable[int]):
```
simonwo, line 253: Could this be `Iterable[UpdateType]`?

common/business_rules.py
```diff
@@ -270,10 +349,23 @@ class PreventDeleteIfInUse(BusinessRule):
...
+        """
         names = [self.via_relation] if self.via_relation else []
         return getattr(model, self.in_use_check)(self.transaction, *names)
 
     def validate(self, model):
+        """
+        Check whether the specified model violates this business rule.
+
+        :param model TrackedModel: The model to check
+        :rtype bool: True if the specified model violates this business rule.
```
simonwo, line 367: It doesn't return anything, but does raise a violation?

common/exceptions.py
```diff
@@ -1,14 +1,18 @@
 class NoIdentifyingValuesGivenError(Exception):
-    pass
+    """Raised when TrackedModelQuerySet.get_versions is called without model
+    identifying fields as keyword arguments."""
 
 
 class AlreadyHasSuccessorError(Exception):
+    # XXX unused
     pass
```
simonwo, line 8: Can we remove it then?

common/querysets.py
```diff
@@ -37,6 +64,13 @@ def not_current(
         self,
         asof_transaction=None,
     ) -> QuerySet:
+        """
+        Filter the queryset to only those TrackedModels that are the latest approved
+        versions as of the given Transaction.
```
simonwo, line 69: Isn't it the opposite, i.e. only those TrackedModels that are *not* the latest approved versions?

simonwo (approved):

simonwo:
`black` has been updated which necessitates updating our requirements.txt and pre commit config.

Please run `python3.9 -m pip install -U -r requirements.txt` and `pre-commit autoupdate` when you get this.

PR #446, opened 2021-11-23
TP-931 Import Commodity Changes from EU Taric files

kintisheff:
# TP-931 Import Commodity Changes from EU Taric files

## Why
When we import commodity changes from EU Taric files, we need to ignore unrelated records (such as EU measures), pre-process the commodity code changes, detect side effects, add preemptive transactions for the side effects, and then proceed importing the relevant and preemptive records as normal.
This rerouting for the EU taric files case needs to be implemented with zero impact on the normal importer workflow

## What

- Introduce a `CommodityChangeRecordLoader`, which converts relevant records from EU Taric files into `CommodityChange` instances
- Extend the `SideEffect` data model so it can be exported to a preemptive transaction and produce a detailed _explain_ log for the reason why a side effect was detected.
- Add a hook to the `TransactionParser` to trigger the commodity change handler 
- Propagate `record_group` all the way down to ensure commodity change handler is only triggered on *commodity-only* imports
- Exclude `FootnoteAssociationGoodNomenclature` records from the record group as we don't use EU's associations (ours are handled in side effects where needed)
- Add minor fixes, e.g. to detecting dependent measures and dependent footnote associations in the change handler.
- Add extra sugar methods and properties to some models to enhance code readability elsewhere

## Note:
There is a long commit history due to the iterative nature of the implementation, however ultimately a lot of pre-requisite work was spun out in separate PR-s and the ultimate changes here are the ones described in the *What* section above.


commodities/models/dc.py
```diff
@@ -551,15 +608,20 @@ def snapshot_date(self) -> Optional[date]:
...
         """Retruns the snapshot transaction if uses the transaction clock."""
         if self.clock_type.is_transaction_clock:
             return self.moments[1]
 
     @property
     def snapshot_transaction(self) -> Optional[Transaction]:
-        """Retruns the snapshot transaction if uses the transaction clock."""
         if self.clock_type.is_transaction_clock:
-            return Transaction.objects.get(id=self.moments[1])
+            return self.max_transaction
```
simonwo, line 624: Hm, are we sure about this? If we create a snapshot with a transaction X, and then `snapshot_transaction` returns e.g. transaction X-1, wouldn't that be a bit suprising? Wouldn't we expect it to return the actual transaction that was specified?

commodities/models/dc.py
```diff
@@ -601,6 +663,9 @@ def _traverse(self):
         for commodity in self.commodities:
             indent = commodity.get_indent()
 
+            if indent is None:
+                continue
+
```
simonwo, line 668: Do we know for what use cases we'd expect this to be triggered? I'm nervous that it might cover up bugs where we should have an indent but do not.

commodities/models/dc.py
```diff
@@ -773,53 +843,42 @@ def current_snapshot(self) -> CommodityTreeSnapshot:
...
 
-        commodities = self._get_snapshot_commodities(snapshot_date, transaction_id)
+        commodities = self._get_snapshot_commodities(snapshot_date, transaction)
+        order = transaction.order if transaction else None
 
         return CommodityTreeSnapshot(
             commodities=commodities,
             clock_type=clock_type,
-            moments=(snapshot_date, transaction_id),
+            moments=(snapshot_date, order),
```
simonwo, line 869: Hm, as above, doesn't seem like calling `moments[1].id` can be correct?

commodities/models/dc.py
```diff
@@ -845,7 +904,48 @@ class SideEffect(BaseModel):
...
+
+    def _get_preemptive_transaction_order(self, workbasket: WorkBasket) -> int:
+        preemptive_transactions = [
+            transaction
+            for transaction in workbasket.transactions.all()
+            if transaction.order < 0
+        ]
+
+        if not preemptive_transactions:
+            return PREEMPTIVE_TRANSACTION_SEED
```
simonwo, line 929: Would it be more efficient to use a queryset here? E.g. `if not workbasket.transactions.filter(order__lt=0).exists()`?

commodities/models/dc.py
```diff
@@ -845,7 +904,48 @@ class SideEffect(BaseModel):
...
+
+        return max([transaction.order for transaction in preemptive_transactions]) + 1
```
simonwo, line 931: Would definitely be better to use a queryset or generator here to prevent Python from materializing all of the transaction objects…

commodities/models/dc.py
```diff
@@ -845,7 +904,48 @@ class SideEffect(BaseModel):
...
+
+    def explain(self) -> dict[str, Any]:
+        """Returns a dict explaining the context of the side effect."""
+        identifier = "|".join(map(str, self.obj.get_identifying_fields().values()))
```
simonwo, line 935: Would it be good to reuse `get_identifying_fields_to_string`?

commodities/models/dc.py
```diff
@@ -875,7 +975,6 @@ class CommodityChange(BaseModel):
     current: Optional[Commodity] = None
     candidate: Optional[Commodity] = None
     ignore_validation_rules: bool = False
-    # side_effects: Optional[dict[TTrackedModelIdentifier, SideEffect]] = None
```
simonwo, line 878: comment

commodities/models/dc.py
```diff
@@ -845,7 +904,48 @@ class SideEffect(BaseModel):
 
     obj: TrackedModel
     update_type: UpdateType
-    attrs: Dict[str, Any] = None
+    code: CommodityCode
+    rule: BusinessRule
+    variant: Optional[str] = None
+    attrs: Optional[Dict[str, Any]] = None
```
simonwo, line 910: Could we add some documentation about what these attributes represent? I can guess most of them, but `variant` is a bit of a mystery.

commodities/models/dc.py
```diff
@@ -1236,10 +1376,25 @@ def _get_dependent_measures(
...
 
+    @property
+    def as_at_date(self):
+        if self.candidate:
+            if self.update_type == UpdateType.UPDATE:
+                return self.candidate.get_valid_between().upper
+            else:
+                return self.candidate.get_valid_between().lower
+        return self.current.get_valid_between().upper
+
```
simonwo, line 1397: Could we get some explanation of what this logic represents? It sets of spidey sense that we're sometimes using the upper or lower date…

commodities/tests/test_dataclasses.py
```diff
@@ -281,7 +282,7 @@ def test_collection_current_snapshot(collection_full):
 
     commodities = set(x.identifier for x in snapshot.commodities)
 
-    assert snapshot.moments == (snapshot_date, transaction_id)
+    assert snapshot.moments == (snapshot_date, transaction.order)
```
simonwo, line 285: Thinking more on the above, do we not need the `partition` in this data structure here as well? It might be useful to have a `@dataclass Moment`?

importer/chunker.py
```diff
@@ -272,24 +272,33 @@ def filter_transaction_records(
...
+    pending_removals = []
+
     for message in Tags.ENV_APP_MESSAGE.iter(elem):
         for transmission in Tags.OUB_TRANSMISSION.iter(message):
             for record in Tags.OUB_RECORD.iter(transmission):
                 record_code = Tags.OUB_RECORD_CODE.first(record).text
                 subrecord_code = Tags.OUB_SUBRECORD_CODE.first(record).text
+                identifier = f"{record_code}{subrecord_code}"
+
+                Tags.OUB_UPDATE_TYPE.first(record).text
```
simonwo, line 287: What does this do?

importer/taric.py
```diff
@@ -132,6 +137,44 @@ def save(
...
+
+                        EnvelopeTransaction.objects.create(
+                            envelope=envelope,
+                            transaction=transaction,
+                            order=order,
+                        )
+
+            if reasons:
+                with open(f"env/{envelope.envelope_id}.log", "a") as f:
+                    f.write("\n".join(map(str, reasons)) + "\n")
```
simonwo, line 176: Do we want to be doing this in production?

importer/taric.py
```diff
@@ -132,6 +137,44 @@ def save(
...
+
+            for chapter_changes in loader.chapter_changes.values():
+                for change in chapter_changes.changes:
+                    for side_effect in change.side_effects.values():
+                        obj = side_effect.to_transaction(self.parent.workbasket)
+
+                        transaction = obj.transaction
+                        order = transaction.order
+                        commodity = change.candidate or change.current
+                        commodity.code.dot_code
```
simonwo, line 159: This doesn't appear to be used.

importer/taric.py
```diff
@@ -159,6 +202,23 @@ def end(self, element: etree.Element):
...
+            get_record_code(record)
+            for transmission in data["message"]
+            for record in transmission["record"]
+        ]
+
+        matching_codes = [
+            code for code in codes if code in TARIC_RECORD_GROUPS["commodities"][:2]
+        ]
+
+        return len(matching_codes) != 0
```
simonwo, line 220: Can we make this `any` and then use generators and bug out early if we find just one?

simonwo (commented):
Nothing major…

commodities/models/dc.py
```diff
@@ -428,9 +479,9 @@ def __post_init__(self) -> None:
             raise CommodityTreeSnapshotException(msg)
 
         try:
-            assert len(chapters) == 1
+            assert len(chapters) <= 1
         except AssertionError:
-            print("All commodities in a group must be from the same HS chapter.")
+            logger.error("All commodities in a group must be from the same HS chapter.")
             raise
```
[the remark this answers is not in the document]
simonwo: Actually, I also changed this to just `if …: raise ValueError(…)` rather than asserting and then catching the assertion…

commodities/models/dc.py
```diff
@@ -233,6 +260,13 @@ def is_current_as_of_transaction_order(
         )
         return self.version == current_version
 
+    def _get_transactions(self) -> tuple[Transaction]:
+        """Returns the id-s of all transactions in the version group of the
+        record."""
+        return tuple(
+            [version.transaction for version in self.obj.version_group.versions.all()],
+        )
```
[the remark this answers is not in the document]
kintisheff: yes!

commodities/models/dc.py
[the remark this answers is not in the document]
kintisheff: ah nice, i don't think we want assert statements outside of tests! will change

commodities/models/dc.py
```diff
@@ -551,15 +608,20 @@ def snapshot_date(self) -> Optional[date]:
...
     def snapshot_transaction_id(self) -> Optional[int]:
+        """Retruns the snapshot transaction if uses the transaction clock."""
+        if self.clock_type.is_transaction_clock:
+            return self.moments[1].id
+
+    @property
+    def snapshot_transaction_order(self) -> Optional[int]:
         """Retruns the snapshot transaction if uses the transaction clock."""
         if self.clock_type.is_transaction_clock:
             return self.moments[1]
```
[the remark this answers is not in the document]
kintisheff: i think the type hint is outdated, will update

commodities/models/dc.py
kintisheff: yes, and i'll add a comment!

commodities/models/dc.py
kintisheff: agreed!

commodities/models/dc.py
kintisheff: also agreed!

commodities/models/dc.py
kintisheff: now i'm aware of it, yes :)

commodities/models/dc.py
kintisheff: will remove!

commodities/models/dc.py
kintisheff: will  add to the docstring!

commodities/models/dc.py
kintisheff: yes => a) will check, b) will doc

importer/chunker.py
kintisheff: artefact of a prior iteration, thanks for spotting!

importer/taric.py
kintisheff: NO!!

importer/taric.py
kintisheff: will check

importer/taric.py
kintisheff: yes!

commodities/import_handlers.py
```diff
@@ -342,6 +343,8 @@ def save(self, data: dict):
                     as_of_transaction=indent.transaction,
                     start_date=start_date,
                 )
+
+                # we may be accidentall
```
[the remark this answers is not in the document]
stuaxo: We may.   This reads better than a lot of my comments !

importer/chunker.py
```diff
@@ -329,13 +338,14 @@ def chunk_taric(
 
         transaction = filter_transaction_records(elem, record_group)
 
-        if transaction is not None:
-            write_transaction_to_chunk(elem, chunks_in_progress, batch, envelope_id)
+        if transaction is None:
+            continue
```
stuaxo, line 342: :+1:  :+1:

importer/chunker.py
```diff
@@ -329,13 +338,14 @@ def chunk_taric(
...
 
-        elem.clear()
+        write_transaction_to_chunk(transaction, chunks_in_progress, batch, envelope_id)
+        transaction.clear()
 
         element_counter += 1
-        if element_counter % 100000 == 0:
+        if element_counter % 100 == 0:
```
stuaxo, line 348: My as well put the 100 into some named constant while editing it.

importer/namespaces.py
```diff
@@ -45,7 +45,7 @@
         "40005",
         "40010",
         "40015",
-        "40020",
+        # "40020",
```
stuaxo, line 48: Probably wants a note as to why it's commented it out, if it's staying here.

importer/taric.py
```diff
@@ -144,7 +187,7 @@ def end(self, element: etree.Element):
         super().end(element)
         if element.tag == self.tag:
             logging.debug(f"Saving import {self.data['id']}")
-            if int(self.data["id"]) % 1000 == 0:
+            if int(self.data["id"]) % 100 == 0:
```
stuaxo, line 190: ^ move this 100 to a constant, so user doesn't have to think as much when they see it.

importer/namespaces.py
kintisheff: yes, will add!

importer/chunker.py
kintisheff: i think i should just revert it to what it was tbh

importer/chunker.py
```diff
@@ -272,24 +272,33 @@ def filter_transaction_records(
...
         for transmission in Tags.OUB_TRANSMISSION.iter(message):
             for record in Tags.OUB_RECORD.iter(transmission):
                 record_code = Tags.OUB_RECORD_CODE.first(record).text
                 subrecord_code = Tags.OUB_SUBRECORD_CODE.first(record).text
+                identifier = f"{record_code}{subrecord_code}"
+
+                Tags.OUB_UPDATE_TYPE.first(record).text
 
-                record_identifier = f"{record_code}{subrecord_code}"
+                if identifier not in record_group:
```
stuaxo, line 289: Me:  the micro-optimiser:

I guess the order in record_group probably is important ?
If not it, would be nice to make it a set (at least before these loops), and the `in` will be cheaper than the list.

(OTOH, I doubt record_group is big enough to make a difference)

simonwo (approved):

commodities/tests/test_importer.py
```diff
@@ -741,3 +743,58 @@ def test_sync_indent_node_end_dates_on_indent_import(
...
+    ),
+    ids=("future", "earlier", "current", "shortlived"),
+)
+def test_correct_affected_measures_are_selected(
+    run_xml_import,
+    date_ranges,
+    measure_validity,
+    is_affected,
+    update_type,
+):
```
stuaxo, line 768: This is a complicated test + 6000%* needs a docstring explaining what it's validating, I have no way of telling if it is doing what it's meant to without this.

*Try and follow "don't make the reader think" (or not too hard anyway!)

kintisheff: added a doc and will pop into the next commit

commodities/models/orm.py
```diff
@@ -625,7 +610,10 @@ class GoodsNomenclatureDescription(DescriptionMixin, TrackedModel):
     description = LongDescription()
 
     indirect_business_rules = (business_rules.NIG12,)
-    business_rules = (UpdateValidity,)
+    business_rules = (
+        NoBlankDescription,
+        UpdateValidity,
+    )
```
simonwo: We need to leave this out because the EU will send objects with no description. So instead, can we add a `skip` for the `GoodsNomenclatureDescription` case to that test?

kintisheff: ah of course that makes sense!

PR #447, opened 2021-11-24
TP-856: Structure code links in the tariff changes view don't work for descriptions

gabelton:
## Why
On the tariff changes table the link for a newly created description is empty ("#") and doesn't link to the described model descriptions tab as it should. This is because get_url in DescriptionMixin is looking for a url with a name of the format "'certificate_description-ui-detail", which doesn't exist.

## What
It seemed simpler to add two lines to the DescriptionMixin.get_url method than to create a redirect view and register another url for different description models (e.g. https://stackoverflow.com/questions/63467337/django-hashtags-in-url)

### DIFF
--- a/common/models/mixins/description.py
+++ b/common/models/mixins/description.py
@@ -44,6 +44,9 @@ class DescriptionMixin(ValidityStartMixin):
         if action != "list":
             kwargs = self.get_identifying_fields()
             described_object = self.get_described_object()
+            if action == "detail":
+                return described_object.get_url() + "#descriptions"
+
             for field, value in described_object.get_identifying_fields().items():
                 kwargs[f"{self.described_object_field.name}__{field}"] = value
         try:


common/models/mixins/description.py
```diff
@@ -44,6 +44,9 @@ def get_url(self, action="detail"):
         if action != "list":
             kwargs = self.get_identifying_fields()
             described_object = self.get_described_object()
+            if action == "detail":
+                return described_object.get_url() + "#descriptions"
```
stuaxo, line 48: As long as described_object can never be self this is OK, otherwise it will get stuck looping here.
(But maybe that would just be bad semantics?)

simonwo: I think that situation would break a whole bunch of other things, so probably safe to assume that it won't happen.

simonwo (approved):

PR #448, opened 2021-11-25
use existing util.contained_date_range

kintisheff:
# TP-1094 Fix _get_restricted_valid_between

## Why
At the moment `GoodsNomenclatureIndentNode._get_restricted_valid_betwen` does not handle well cases with non-overlapping dates, which can legitimately occur. This has emerged as a secondary issue blocking 887.

## What
- `GoodsNomenclatureIndentNode._get_restricted_valid_betwen` now calls `commodities.util.contained_date_range` instead of trying to replicate its logic. The latter method is already well covered with tests, including cases with non-overlapping dates.
- Allow `commodities.util.contained_date_range` to return a customisable fallback value in case of non-overlapping end dates

### DIFF
--- a/commodities/models/orm.py
+++ b/commodities/models/orm.py
@@ -16,6 +16,7 @@ from treebeard.mp_tree import MP_Node
 from commodities import business_rules
 from commodities import validators
 from commodities.querysets import GoodsNomenclatureIndentQuerySet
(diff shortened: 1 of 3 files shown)


paulpepper-trade (approved):
LGTM

PR #449, opened 2021-11-25
TP-1020: ME58 should cover conditions without certificates as well

gabelton:
## Why
"Our current implementation doesn't run the rule if there is no certificate, but actually, this rule should still apply. I.e. there should only be one measure condition with "no required certificate" per measure condition code."

## What
- Update ME58 to filter by kwargs set conditionally on whether a measure_condition has a required_certificate
- Add case to business rule test

### DIFF
--- a/measures/business_rules.py
+++ b/measures/business_rules.py
@@ -920,19 +920,25 @@ class ME58(BusinessRule):
     the same condition type."""
 
     def validate(self, measure_condition):
+        kwargs = {
+            "condition_code__code": measure_condition.condition_code.code,
+            "dependent_measure__sid": measure_condition.dependent_measure.sid,
+        }
         if measure_condition.required_certificate is None:
-            return
+            kwargs.update({"required_certificate": None})
+        else:
+            kwargs.update(
+                {
+                    "required_certificate__sid": measure_condition.required_certificate.sid,
+                    "required_certificate__certificate_type__sid": measure_condition.required_certificate.certificate_type.sid,
+                },
+            )
 
         if (
             type(measure_condition)
             .objects.exclude(pk=measure_condition.pk or None)
             .excluding_versions_of(version_group=measure_condition.version_group)
-            .filter(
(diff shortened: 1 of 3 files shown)


measures/business_rules.py
```diff
@@ -920,19 +920,25 @@ class ME58(BusinessRule):
...
         if measure_condition.required_certificate is None:
-            return
+            kwargs.update({"required_certificate": None})
+        else:
+            kwargs.update(
+                {
+                    "required_certificate__sid": measure_condition.required_certificate.sid,
+                    "required_certificate__certificate_type__sid": measure_condition.required_certificate.certificate_type.sid,
+                },
+            )
```
simonwo, line 935: You could also use `required_certificate__version_group` to abstract over the identifying fields.

measures/tests/test_business_rules.py
```diff
@@ -1369,6 +1369,16 @@ def test_ME58():
...
+    not_required = factories.MeasureConditionFactory.create(required_certificate=None)
+    duplicate = factories.MeasureConditionFactory.create(
+        condition_code=not_required.condition_code,
+        dependent_measure=not_required.dependent_measure,
+        required_certificate=None,
+    )
+
+    with pytest.raises(BusinessRuleViolation):
+        business_rules.ME58(duplicate.transaction).validate(not_required)
+
```
simonwo, line 1382: Could this be factored using a parameterized test instead? E.g. make the `None` or `factories.CertificateFactory` into a test parameter?

Also whilst not caused by your change, a positive test (that checks the rule doesn't fire when the conditions are good) would be nice if you have time.

simonwo (approved):

measures/tests/test_business_rules.py
simonwo: Also, you could/should test that the business rule also fires for `duplicate` as well as `not_required`.

simonwo (approved):

PR #450, opened 2021-11-25
TP-1096 Fix exploding db round trips in commodity tree snapshots

kintisheff:
# TP-1096 Fix exploding db round trips in commodity tree snapshots

## Why
At the moment, when `CommodityCollection._get_snapshot_commodities` selects the matching commodities for a given transaction or calendar moment, it makes a large number of individual select queries involving `GoodsNomenclature`, `TrackedModel`, and `VersionGroup` objects.
As a result, if we need to take a snapshot of a larger branch (e.g. when a 2-level code and all its descendants), this would become a very expensive process

## What
- Implement a more efficient algorithm in `CommodityCollection._get_snapshot_commodities` that has exactly ONE db round trip regardless of the size of the tree.
- Implement an enabling util method `get_latest_versions` that returns only the most recent version of each model within a queryset result 
- As a related necessary change, switch transaction clock in `CommodityTreeSnapshot` to be based on transaction order, not transaction ids.

### DIFF
--- a/commodities/models/dc.py
+++ b/commodities/models/dc.py
@@ -776,14 +776,15 @@ class CommodityCollection(CommodityTreeBase):
     def _get_snapshot_commodities(
         self,
         snapshot_date: date,
-        transaction_order: int = None,
+        transaction_order: int,
     ) -> List[Commodity]:
         """
         Returns the list of commodities than belong to a snapshot.
 
         This method needs to be very efficient -
-        In particular, it should make only ONE database round trip
-        regardless of the level of the root commodity in the tree.
+        In particular, it should require the same number
+        of database round trips regardless of the level
+        of the root commodity in the tree.
 
         The solution is fetch all goods matching the snapshot moment
         (incl. potentially multiple versions of each)
@@ -794,22 +795,21 @@ class CommodityCollection(CommodityTreeBase):
         We then efficiently find the commodities in our collection
         that match the latest_version goods.
         """
-        if snapshot_date is None:
-            snapshot_date = date.today()
-
-        if transaction_order is None:
-            transaction_order = self.max_transaction_order
-
         item_ids = {c.get_item_id() for c in self.commodities if c.obj}
+        transaction = Transaction.objects.get(
+            order=transaction_order,
+            partition__in=(
+                TransactionPartition.DRAFT,
+                TransactionPartition.REVISION,
+            ),
+        )
 
         goods = (
-            GoodsNomenclature.objects.filter(
+            GoodsNomenclature.objects.approved_up_to_transaction(transaction)
+            .filter(
                 item_id__in=item_ids,
                 valid_between__contains=snapshot_date,
                 transaction__order__lte=transaction_order,
-            )
-            .exclude(
-                transaction__partition=TransactionPartition.SEED_FILE,
             )
             .order_by(
                 "item_id",
--- a/commodities/models/dc.py
+++ b/commodities/models/dc.py
@@ -24,13 +24,13 @@ from commodities.models.orm import GoodsNomenclatureIndent
 from commodities.util import clean_item_id
 from commodities.util import contained_date_range
 from commodities.util import date_ranges_overlap
-from commodities.util import get_latest_versions
 from common.business_rules import BusinessRuleViolation
 from common.models.constants import ClockType
 from common.models.dc.base import BaseModel
 from common.models.trackedmodel import TrackedModel
 from common.models.transactions import Transaction
 from common.util import TaricDateRange
+from common.util import get_latest_versions
 from common.validators import UpdateType
 from measures import business_rules as mbr
 from measures.models import FootnoteAssociationMeasure
--- a/commodities/models/dc.py
+++ b/commodities/models/dc.py
@@ -29,6 +29,7 @@ from common.models.constants import ClockType
 from common.models.dc.base import BaseModel
 from common.models.trackedmodel import TrackedModel
 from common.models.transactions import Transaction
+from common.models.transactions import TransactionPartition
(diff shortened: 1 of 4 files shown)


common/util.py
```diff
@@ -340,3 +340,23 @@ def get_model_indefinite_article(model_instance: Model) -> Optional[str]:
...
+    only the one with the one with the latest transaction order
+    (which should be latest version) is yielded.
+    """
+    keys = set()
+
+    for model in qs.order_by("-transaction__order"):
+        key = tuple(model.get_identifying_fields().values())
+        if key not in keys:
+            keys.add(key)
+            yield model
```
simonwo, line 362: Isn't this just (the perhaps badly named) `approved_up_to_transaction`?

commodities/models/dc.py
```diff
@@ -774,60 +740,78 @@ def current_snapshot(self) -> CommodityTreeSnapshot:
...
+            item_id__in=item_ids,
+            valid_between__contains=snapshot_date,
+            transaction__order__lte=transaction_order,
+        ).order_by("item_id", "-transaction__order")
+
+        latest_versions = get_latest_versions(goods)
+        pks = {good.pk for good in latest_versions}
+
+        keyed_collection = {c.obj.pk: c for c in self.commodities if c.obj}
+        return [commodity for pk, commodity in keyed_collection.items() if pk in pks]
```
simonwo, line 814: Ah, I lied. You will also need to handle the partition, e.g. the same order can appear multiple times in different partitions (but once per partition). The partition is basically just like the "most significant bit" of the order. So I'd recommend using `approved_up_to_transaction` or at least the filter to handle that. This might mean you need to refactor again to not pass around transaction orders but complete transaction objects (which is maybe also a good refactor?).

kintisheff: oh the partition, yes! forgot to add it in from the notebook where i was drafting this :)

kintisheff: although in the context of a draft workbasket, when we want to take an after snapshot so we can preview side effects, we may need to work with 2 partitions - approved and draft

kintisheff: and also approved_up_to_transaction means extra db queries. so will address the issue differently

common/util.py
kintisheff: no. 2 reasons:
- this one doesn't involve a round-trip to the db.
- latest version may not necessarily mean latest approved version (in the context of previewing side effects for a commodity change import, the after snapshot will contain goods from two partitions - approved and draft)

simonwo: On the final point, `approved_up_to_transaction` has that behaviour i.e. it will consider draft data from the same workbasket as the transaction if the transaction is in draft. That's why it's badly-named… >_>

commodities/models/dc.py
```diff
@@ -794,23 +795,22 @@ def _get_snapshot_commodities(
...
+            ),
+        )
 
         goods = (
-            GoodsNomenclature.objects.filter(
+            GoodsNomenclature.objects.approved_up_to_transaction(transaction)
+            .filter(
                 item_id__in=item_ids,
                 valid_between__contains=snapshot_date,
                 transaction__order__lte=transaction_order,
```
simonwo, line 812: You won't need this line anymore … in fact, it might remove some objects

commodities/models/dc.py
```diff
@@ -774,60 +740,81 @@ def current_snapshot(self) -> CommodityTreeSnapshot:
...
         )
 
     def _get_snapshot(
         self,
         snapshot_date: Optional[date] = None,
-        transaction_id: Optional[int] = None,
+        transaction_order: Optional[int] = None,
     ) -> CommodityTreeSnapshot:
-        if transaction_id is None == snapshot_date is None:
+        if transaction_order is None == snapshot_date is None:
```
stuaxo, line 755: Is it meant to be legal these can both be None ?

simonwo (approved):

commodities/models/dc.py
```diff
@@ -774,60 +740,81 @@ def current_snapshot(self) -> CommodityTreeSnapshot:
...
+
+        goods = GoodsNomenclature.objects.approved_up_to_transaction(
+            transaction,
+        ).filter(
+            item_id__in=item_ids,
+            valid_between__contains=snapshot_date,
+        )
+
+        latest_versions = get_latest_versions(goods)
+        pks = {good.pk for good in latest_versions}
```
stuaxo, line 814: I got a bit confused that `self.commodities` isn't already a queryset ?

If you made self.commodities a queryset then things can get a bit simpler, along these lines:

```python
pks = latest_versions.values_list("pk", flat=True)
return self.commodities.filter(pk__in=pks)
```

common/util.py
```diff
@@ -340,3 +340,26 @@ def get_model_indefinite_article(model_instance: Model) -> Optional[str]:
...
+
+    These may not be the current versions of each model,
+    e.g. because the queryset may be filtered as of a given transaction.
+
+    But if there are two versions of the same tracked model in the queryset,
+    only the one with the one with the latest transaction order
+    (which should be latest version) is yielded.
+    """
+    keys = set()
+
```
stuaxo, line 357: I guess `qs` here could be a mix of types, so we can't just query the identifying fields all in one go ?

commodities/models/dc.py
kintisheff: self.commodities needs to be a list i'm afraid as we may need to add candidates to it

commodities/models/dc.py
kintisheff: we will be making it illegal to not have transaction_order as of 931 yes, not a concern of this one though

PR #453, opened 2021-11-30
TP-851: When creating a footnote with the prefix TN - TaMaTo assigned a footnote number already in use

gabelton:
## Why
get_next_id is currently only used once in footnotes/forms.py. The queryset used by this method is filtered on valid_between__overlap, which means that currently Footnote TN008 may not be part of this qs and, as a result, the method may return "008" as next id. This causes an error on the confirm-create view which expects a unique combination of footnote_id and footnote_type_id, but instead finds more than one footnote with the same combination

## What
- Removes the filter on valid_between overlap and adds test for this scenario
- Also updates get_next_id to handle empty queryset - currently the user will get an error if they try to create a footnote of a type where no footnote currently exists.


footnotes/tests/test_forms.py
```diff
@@ -0,0 +1,46 @@
...
+
+from common.tests import factories
+from common.util import TaricDateRange
+from footnotes import forms
+from workbaskets.validators import WorkflowStatus
+
+pytestmark = pytest.mark.django_db
+
+
+def test_form_save_creates_new_footnote_id_and_footnote_type_id_combo(client):
```
stuaxo, line 14: Could definitely do with a docstring so developers can evaluate it does what it's meant to.

footnotes/tests/test_forms.py
```diff
@@ -0,0 +1,46 @@
...
+
+    data = {
+        "footnote_type": footnote_type.pk,
+        "start_date_0": 2,
+        "start_date_1": 2,
+        "start_date_2": 2022,
+        "description": "A note on feet",
+    }
+    form = forms.FootnoteCreateForm(data=data)
+    form.is_valid()
```
stuaxo, line 41: I'm not quite sure what the section above is doing ?

Is the is_valid check, a sanity check at this point ?
Probably could do with a comment.

footnotes/tests/test_forms.py
```diff
@@ -0,0 +1,46 @@
...
+
+    with mock.patch("workbaskets.models.WorkBasket.current", return_value=workbasket):
+        new_footnote = form.save(commit=False)
```
stuaxo, line 44: Hard to evaluate what this mock is for ?
3rd party APIs are good places to inject mocks, if we need to mock our APIs then comments are good.

stuaxo: [10:49] stuart.axon (Guest)
It looks like Workbasket.current returns None if the workbasket is not approved.

Would it be OK to use ApprovedWorkbasketFactory so you have an approved one ?

Then the you might not need the mock, since Workbasket.current will return your workbasket in the form.

gabelton: I've removed the mock, but still leaves the issue of mocking the request somehow. I've just used the django request factory. The setup is a bit convoluted, but could make a fixture, if we want to reuse it elsewhere.

stuaxo: Yeah, request factory is fine, a fixture also makes sense - I don't believe any of the ones we have fit this exact use case yet.

stuaxo: As we test more of the forms, some patterns might fall out.

simonwo: Would something similar to `use_update_form` help here?

stuaxo (approved):

stuaxo:
Approved - the only marginally useful thing I discovered is the "rf" fixture for RequestFactory which gives a really marginal space saving:

```python
def test_form_save_creates_new_footnote_id_and_footnote_type_id_combo(client, rf):
    """Tests that when two non-overlapping footnotes of the same type are
    created that these are created with a different footnote_id, to avoid
    duplication of footnote_id and footnote_type_id combination e.g. TN001."""
    workbasket = factories.ApprovedWorkBasketFactory.create()
    session = client.session
    session["workbasket"] = {"id": workbasket.pk}
    session.save()

    request = rf.post('/footnotes/create')
    request.session = session

    footnote_type = factories.FootnoteTypeFactory.create()
    valid_between = TaricDateRange(date(2021, 1, 1), date(2021, 12, 1))
    earlier = factories.FootnoteFactory.create(
        footnote_type=footnote_type,
        valid_between=valid_between,
        footnote_id="001",
    )

    data = {
        "footnote_type": footnote_type.pk,
        "start_date_0": 2,
        "start_date_1": 2,
        "start_date_2": 2022,
        "description": "A note on feet",
    }
    form = forms.FootnoteCreateForm(data=data, request=request)
    new_footnote = form.save(commit=False)

    assert earlier.footnote_id != new_footnote.footnote_id
```

footnotes/tests/test_forms.py
stuaxo: @simonwo I'd totally forgotten about this - I think so... also, to cover a ticket I opened about testing all of our forms.

footnotes/tests/test_forms.py
```diff
@@ -0,0 +1,42 @@
...
+from django.test.client import RequestFactory
+
+from common.tests import factories
+from common.util import TaricDateRange
+from footnotes import forms
+
+pytestmark = pytest.mark.django_db
+
+# https://uktrade.atlassian.net/browse/TP-851
+def test_form_save_creates_new_footnote_id_and_footnote_type_id_combo(client):
```
stuaxo, line 13: Not a blocker - but there's a margin saving to be had using the `rf` fixture for RequestFactory - which I only just heard about -

```python
def test_form_save_creates_new_footnote_id_and_footnote_type_id_combo(client, rf):
    """Tests that when two non-overlapping footnotes of the same type are
    created that these are created with a different footnote_id, to avoid
    duplication of footnote_id and footnote_type_id combination e.g. TN001."""
    workbasket = factories.ApprovedWorkBasketFactory.create()
    session = client.session
    session["workbasket"] = {"id": workbasket.pk}
    session.save()

    request = rf.post('/footnotes/create')
    request.session = session
```

footnotes/tests/test_forms.py
stuaxo: Caught up with @gabelton yesterday, I'm having a go at implementing use_create_form since I need to it to test the additional_code fix.

stuaxo: Hi @gabelton - use_create_form is now merged.

It should be possible to use it to implement a test, similar to [test_additional_code_create_form](https://github.com/uktrade/tamato/blob/master/additional_codes/tests/test_views.py#L48) for additional codes.

More detail fields tend to be needed than use_update_form tests, since use_create_form can't infer data from existing models (in form_valid, and in form base classes).

To work out which extra data is needed, it's worth looking in form_valid for your model as well as your models form class and base classes.

footnotes/tests/test_forms.py
stuaxo: @gabelton might be worth considering use of `rf` at least - but not a blocker.

Probably out of scope, I wonder if we need an easier way of specifying a session with a workbasket ?

stuaxo (approved):

PR #454, opened 2021-11-30
TP-1107  Additional code form was broken trying to get cleaned_data field not …

stuaxo:
AdditionCode form was broken.

This PR implements Simons suggestion on Gabes ticket TP-851 and implements use_create_form and uses it to verify the fix for additional_codes.

use_create_form is ported from use_update_form, and is just different enough I didn't combine them as use_create_form already does a lot.

Screenshot of additional_codes form before this fix:

Before:
image

### DIFF
--- a/common/fields.py
+++ b/common/fields.py
@@ -18,7 +18,7 @@ from common.util import TaricDateTimeRange
 
 def get_next_by_max(field):
     return lambda: RawSQL(
-        sql=f'SELECT MAX("{field.column}") + 1 FROM "{field.model._meta.db_table}"',
+        sql=f'SELECT COALESCE(MAX("{field.column}"), 0) + 1 FROM "{field.model._meta.db_table}"',
         params=[],
     )
 
@@ -28,7 +28,7 @@ class NumericSID(models.PositiveIntegerField):
         kwargs["editable"] = False
         kwargs["validators"] = [validators.NumericSIDValidator()]
         kwargs["db_index"] = True
-        kwargs["default"] = get_next_by_max(self)
+        kwargs.setdefault("default", get_next_by_max(self))
         super().__init__(*args, **kwargs)
 
     def deconstruct(self):
@@ -44,7 +44,7 @@ class SignedIntSID(models.IntegerField):
     def __init__(self, *args, **kwargs):
         kwargs["editable"] = False
         kwargs["db_index"] = True
-        kwargs["default"] = get_next_by_max(self)
+        kwargs.setdefault("default", get_next_by_max(self))
         super().__init__(*args, **kwargs)
 
     def deconstruct(self):
--- a/conftest.py
+++ b/conftest.py
@@ -18,6 +18,7 @@ from django.contrib.auth.models import Group
 from django.contrib.auth.models import Permission
 from django.core.exceptions import ValidationError
 from django.test.html import parse_html
+from django.urls import reverse
 from factory.django import DjangoModelFactory
 from lxml import etree
 from moto import mock_s3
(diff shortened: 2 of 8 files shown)


gabelton (approved):

stuaxo:
Cheers @gabelton I really should borrow the code from your test to make sure this has coverage :)

conftest.py
```diff
@@ -275,6 +276,83 @@ def description_factory(request):
...
+        # Submit the edited data and if we expect success ensure we are redirected
+        realised_data = new_data(data)
+        data.update(realised_data)
+
+        # There is no model instance to fetch identifying_fields from, construct
+        # them from the provided form data.
+        identifying_values = {k: data.get(k) for k in Model.identifying_fields}
+        versions = set(
+            Model.objects.filter(**identifying_values).values_list("pk", flat=True),
+        )
```
simonwo, line 321: This should always be an empty set, because you shouldn't be able to create a new thing with the same identifying fields as an old one.

conftest.py
```diff
@@ -275,6 +276,83 @@ def description_factory(request):
...
+        if response.status_code not in (301, 302):
+            assert (
+                set(
+                    Model.objects.filter(**identifying_values).values_list(
+                        "pk",
+                        flat=True,
+                    ),
+                )
+                == versions
+            )
```
simonwo, line 335: This could just be `not Model.objects.filter(**identifying_values).exists()`?

conftest.py
```diff
@@ -275,6 +276,83 @@ def description_factory(request):
...
+
+        new_version = Model.objects.filter(**identifying_values).get()
+        assert new_version not in versions
+
+        # Check that the new version is an update and is not approved yet
+        assert (
+            new_version.update_type == UpdateType.UPDATE
+            if versions
+            else UpdateType.CREATE
+        )
```
simonwo, line 348: This should always be a CREATE if we are using a CREATE form?

conftest.py
```diff
@@ -417,7 +495,8 @@ def check(
             else:
                 raise
 
-        assert_records_match(model, imported)
+        # Ignore export_refund_nomenclature_sid:- it's a link to a key that isn't implemented in Tamato
+        assert_records_match(model, imported, ignore={"export_refund_nomenclature_sid"})
```
simonwo, line 499: I don't think this should be ignored – it should be NULL in both created and imported data?

conftest.py
stuaxo: Ah, very good point.

stuaxo: Maybe I didn't understand the original test I was porting this from in use_update_form (just above) ?

This is meant to be equivalent to the check at the top of there where it gets `versions` in a similar way by using get_identifying_fields - is there something I'm missing about the other use_update_form that means this check is not equivalent ?

Can we definitely not have two different versions of the same model that have the same identifying fields ?

simonwo: It's because `use_update_form` is meant to be updating an existing instance, and else will already have at least 1 version. But the `use_create_form` will be creating a new instance which must by definition have a unique set of identifying fields.

conftest.py
stuaxo: I couldn't work out why it was always 1 in `imported` but got sorted it now, I haven't looked at that bit of code for > 1hr, will push up shortly.

common/fields.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, *args, **kwargs):
         kwargs["editable"] = False
         kwargs["validators"] = [validators.NumericSIDValidator()]
         kwargs["db_index"] = True
-        kwargs["default"] = get_next_by_max(self)
+        kwargs.setdefault("default", get_next_by_max(self))
```
simonwo, line 31: Excellent!

conftest.py
```diff
@@ -417,6 +479,7 @@ def check(
             else:
                 raise
 
+        # Ignore export_refund_nomenclature_sid:- it's a link to a key that isn't implemented in Tamato
```
simonwo, line 482: Maybe remove this?

simonwo (approved):

conftest.py
stuaxo: :boom:

common/fields.py
stuaxo: Phew!...     Had a few false starts earlier before getting to this simple line.

PR #455, opened 2021-12-01
TP-758: ME32 without the materialized tree

simonwo:
## Why
Our previous implementation of ME32 used the Nodes contained in the Goods Nomenclature tree to check for overlaps between commodities. We have found these Nodes difficult to maintain (it's a 3D-cube of commodity trees!) and subsequently the data structure is wrong in many places.

This leads to false positives and (presumably) false negatives from ME32.

## What
This commit introduces a new implementation of ME32 that does not use the tree and insteadmakes use of the in-memory tree used for the commodity code importer.

## Checklist
- Requires migrations? No.

### DIFF
--- a/commodities/models/dc.py
+++ b/commodities/models/dc.py
@@ -1,6 +1,7 @@
 """Includes dataclasses used in goods nomenclature hierarchy tree management."""
 from __future__ import annotations
 
+import datetime
 import logging
 from copy import copy
 from dataclasses import dataclass
@@ -16,7 +17,9 @@ from typing import Set
 from typing import Tuple
 from typing import Union
 
-from django.db.models import Q
+from django.db.models.expressions import F
+from django.db.models.expressions import Subquery
+from django.db.models.query_utils import Q
 
 from commodities import business_rules as cbr
 from commodities.models.constants import SUFFIX_DECLARABLE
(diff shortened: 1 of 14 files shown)


commodities/models/dc.py
```diff
@@ -97,79 +104,63 @@
...
+    indent: Optional[GoodsNomenclatureIndent] = None
 
+    def get_item_id(self) -> str:
+        """Returns the item_id attribute if set, or the object item_id field
+        otherwise."""
         return self.obj.item_id
 
-    def get_suffix(self) -> Optional[str]:
+    def get_suffix(self) -> str:
         """Returns the suffix attribute if set, or the object suffix field
```
gabelton, line 129: Don't this and the above method just return the object fields now ?

simonwo: Yes… I thought that [maybe it was best to keep them](https://hackernoon.com/object-oriented-tricks-2-law-of-demeter-4ecc9becad85) but perhaps they at least be properties instead?

gabelton: Oh, I just meant that the doc strings seems to describe the old functionality

commodities/models/dc.py
```diff
@@ -97,79 +104,63 @@
...
-            .order_by("transaction_id")
-            .last()
-        )
-
-        if obj is None:
-            return
-
-        return int(obj.indent)
-
     def get_valid_between(self) -> TaricDateRange:
```
kintisheff, line 133: is this method still necessary then?

kintisheff (approved):
lgtm

PR #456, opened 2021-12-01
Fix view tests that fail around month end.

simonwo:
## Why
We have some view tests that try to test update forms. These often based on the date because they try to naively add 1 to a date or month counter, and this often results in invalid dates (e.g. the 32nd March).

## What
This commit changes the system to add amounts to the date using a relative delta. To do this, the model needed to change from being based on a lambda for each field, to a single lambda that returns all of the new form data.

## Checklist
- Requires migrations? No

### DIFF
--- a/footnotes/tests/test_views.py
+++ b/footnotes/tests/test_views.py
@@ -5,6 +5,8 @@ from common.tests import factories
 from common.tests.util import assert_model_view_renders
 from common.tests.util import get_class_based_view_urls_matching_url
 from common.tests.util import raises_if
+from common.tests.util import valid_between_start_delta
+from common.tests.util import validity_start_delta
 from common.tests.util import view_is_subclass
 from common.tests.util import view_urlpattern_ids
 from common.views import TamatoListView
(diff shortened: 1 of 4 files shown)


footnotes/tests/test_views.py
```diff
@@ -18,11 +20,11 @@
...
-        ({}, True),
-        ({"start_date_0": lambda d: d + 1}, True),
-        ({"start_date_0": lambda d: d - 1}, False),
-        ({"start_date_1": lambda m: m + 1}, True),
-        ({"start_date_2": lambda y: y + 1}, True),
+        (lambda f: {}, True),
+        (valid_between_start_delta(days=+1), True),
+        (valid_between_start_delta(days=-1), False),
+        (valid_between_start_delta(months=1), True),
+        (valid_between_start_delta(years=1), True),
```
stuaxo, line 27: This feels like an improvement in readability :)

stuaxo (approved):

gabelton (approved):

stuaxo:
Cheers, should get my PR passing again.

PR #457, opened 2021-12-02
TP-1109 Prevent Duplicate Preemptive Transactions in a Workbasket

kintisheff:
# TP-1109 Prevent Duplicate Preemptive Transactions in a Workbasket

## Why
At the moment, the commodity code change handler does not control for situations where a business rule (e.g ME32) affects the same model instance twice. This ticket aims to fix that .


commodities/models/dc.py
```diff
@@ -986,6 +986,11 @@ def preempt_side_effects(self) -> None:
         if self.update_type != UpdateType.CREATE:
             self._handle_hierarchy_side_effects(before, after)
 
+    @property
+    def workbasket(self) -> WorkBasket:
+        """Returns the workbasket for the commodity change."""
+        return (self.candidate or self.current).obj.transaction.workbasket
+
```
simonwo, line 993: Hm, if `candidate` is `None` because this is a deleted code(?), it is appropriate to the use the `current` workbasket? Couldn't that workbasket be a very old/published if the `current` is some commodity version from months/years ago?

simonwo (commented):
I'm just not sure if this *should* be necessary…

1. If in a commodity update (let's say, as of transaction 1) we need to delete a measure, the system will do so and insert a pre-emptive transaction (say -1). So from the view of transaction 1, the measure has been deleted and doesn't cause a business rule failure, great.
2. Then the system considers a commodity update in transaction 2. How can the system conclude that the measure that was deleted in transaction -1 needs updating, because it has already been deleted?

It sounds like the queries we are doing are not sufficiently scoped down, i.e. it is not considering what is live as of transaction 2, because if it was, this shouldn't be necessary. So is the proper fix to ensure those queries are scoped correctly?

commodities/models/dc.py
kintisheff: yes, the hack worked for this round because we don't delete codes, just end date them, but the point is valid.
i also agree with the broader review comment that a circuit breaker ultimately should not be needed if the right checks are done upstream, and will redesign the ticket implementation accordingly

commodities/models/dc.py
```diff
@@ -1229,13 +1234,38 @@ def _get_measure_key(m: Measure) -> Tuple[str, str, str, str, int]:
...
+        obj: TrackedModel,
+        verbose: Optional[bool] = True,
+    ) -> bool:
+        """Returns True if a side effect has already been preempted."""
+        key = get_model_identifier(obj)
+
+        preempted = key in (
+            get_model_identifier(model)
+            for model in self.workbasket.tracked_models.all()
+            if type(model) == type(obj)
```
stuaxo, line 1248: Depending on if you expect `workbasket.tracked_models` to be prefetched, you might either want to make the `==` an `is` OR:

If you want to filter by just the type every django polymorphic model has a "polymorphic_ctype" value you can use.
I need to check, but it's something along these lines:

`workbasket.tracked_models.filter(polymorphic_ctype=ContentType.objects.get_for_model(obj))`

commodities/models/dc.py
stuaxo: On the theme of not knowing if this is necessary, it sounds like this may be a symptom of something else.

simonwo:
Can this be closed @kintisheff?

PR #461, opened 2021-12-09
TP-1070: Find common ancestors of multiple Commodities

simonwo:
## Why
For 999L we need to be able to work out what the common ancestor of a group of commodities is. This PR adds that function.

## What
Finding a common ancestor is easy when we already have `get_ancestor` and just use sets. But…

Previously it wasn't possible to put `Commodity` objects into containers because they were mutable and therefore unhashable.

This PR makes more use of the dataclass features to automatically generate hashes and string representations, and therefore makes these objects immutable by default.

It also changes to using SIDs as identifiers instead of combinations of other fields which is a more accurate and straightforward way to identify each object.

### DIFF
--- a/commodities/models/dc.py
+++ b/commodities/models/dc.py
@@ -418,6 +418,18 @@ class CommodityTreeSnapshot(CommodityTreeBase):
         """Returns all descendants of a commodity in the snapshot tree."""
         return self.descendants.get(commodity, [])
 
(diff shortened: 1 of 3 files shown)


kintisheff (approved):
LGTM - if any existing use case in comm change handler is affected by this (should be unlikely), we will adapt that use case.

PR #462, opened 2021-12-10
TP-1123: MeasureSnapshot.overlaps() should check commodity exists before calling get_branch_measures(commodity)

gabelton:
## Why
This is to avoid an AttributeError when self.ancestors[commodity.identifier] is called in commodities/models/dc.

## What
MeasureSnapshot.overlaps returns an empty queryset, if no commodity is found


kintisheff (approved):
lgtm

PR #465, opened 2021-12-15
TP-1125: Make ME58 consider volumes and units as well

simonwo:
## Why
In TP-1020 we made it so that users could not add more than one condition with the same certificate. This works for all certificate-based conditions, but it is not correct for conditions with volumes (under e.g. condition code E).

ME58 will now stop users from adding conditions that use volume amounts, because these conditions don’t have certificates and the rule thinks this is a duplicate of the base case where nothing is supplied. 

## What
So this commit fixes the rule to only enforce that the combination of certificate + volume + unit is unique.

### DIFF
--- a/measures/business_rules.py
+++ b/measures/business_rules.py
@@ -882,10 +882,22 @@ class ME58(BusinessRule):
                 },
             )
 
+        if measure_condition.condition_measurement is None:
+            kwargs.update({"condition_measurement": None})
+        else:
+            kwargs.update(
+                {
+                    "condition_measurement__version_group": measure_condition.condition_measurement.version_group,
+                },
+            )
+
+        kwargs.update({"duty_amount": measure_condition.duty_amount})
+
         if (
             type(measure_condition)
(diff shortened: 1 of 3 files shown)


kintisheff (approved):
lgtm, i'll regen the hs sync import measures workbasket using this;
one q - can monetary units potentially loop into volumes also?

measures/business_rules.py
```diff
@@ -882,10 +882,22 @@ def validate(self, measure_condition):
...
+                },
+            )
+
+        kwargs.update({"duty_amount": measure_condition.duty_amount})
+
         if (
             type(measure_condition)
-            .objects.exclude(pk=measure_condition.pk or None)
-            .excluding_versions_of(version_group=measure_condition.version_group)
+            .objects.excluding_versions_of(
```
kintisheff, line 898: is this `.objects.excluding_versions_of` or just `excluding_versions_of`?

kintisheff (approved):

PR #468, opened 2021-12-20
Ensure quotas have a unit or currency

simonwo:
## Why
In the TARIC XSD it's possible for neither the monetary unit or measurement unit to be set on a quota definition. But this doesn't represent a real situation – the volume on a quota definition must refer to one of the two, so one (and only one) of them should always be present. 

This is confirmed by the fact that no previous data has ever had neither or both present.

## What
This introduces a check constraint that ensures that only one of the two columns is always non-null.

## Checklist
- Requires migrations? ✨**Yes.**


gabelton (approved):

stuaxo (approved):
@simonwo Might be worth adding a docstring matching your comment on teams about "technically allowed but never and should never happen" in case it comes up one day.

PR #469, opened 2021-12-20
TP-1128 tweaks to the commodity importer

kintisheff:
# TP-1128 tweaks to the commodity importer

## Why
Adding some code optimizations resulting from experience with EU Taric files

## What
- handle ME32 conflicts more robustly
- filter taric envelope transaction records and footnote associations more accurately
- add a check to avoid duplicate preemptive deletes or an update to a deleted record
- add optional as_of_transaction arg to get_dependent_measures
- fix ME7 to work with changes in both the measure and the good
- improve scenario fixtures


commodities/models/dc.py
```diff
@@ -447,7 +448,8 @@ def get_dependent_measures(
         if self.moment.clock_type.is_transaction_clock:
             qs = qs.approved_up_to_transaction(self.moment.transaction)
         else:
-            qs = qs.latest_approved()
+            logger.debug("Filtering by moment transaction: %s", self.moment.transaction)
+            qs = qs.approved_up_to_transaction(self.moment.transaction)
```
simonwo, line 452: Aren't these two branches the same now?

commodities/models/dc.py
```diff
@@ -966,6 +968,11 @@ def preempt_side_effects(self) -> None:
         if self.update_type != UpdateType.CREATE:
             self._handle_hierarchy_side_effects(before, after)
 
+    @property
+    def workbasket(self) -> WorkBasket:
+        """Returns the workbasket for the commodity change."""
+        return (self.candidate or self.current).obj.transaction.workbasket
```
simonwo, line 974: When will it not be the candidate?

commodities/util.py
```diff
@@ -93,3 +93,15 @@ def get_snapshot_from_good_chapter(good):
...
+    containing_date_range: TaricDateRange,
+) -> bool:
+    """Returns true iof the date range is fully contained by the containing
+    range."""
+    contained_range = contained_date_range(
+        date_range,
+        containing_date_range,
+    )
+
+    return contained_range == date_range
```
simonwo, line 107: I think we already have this in `validity_range_contains_range`? :)
https://github.com/uktrade/tamato/blob/085bdcdb8a7638db827c5f81d1c60b9517a9bdeb/common/util.py#L170

simonwo (approved):

PR #470, opened 2021-12-21
Ensure ME32 always considers draft measure versions

simonwo:
Previously, the measure set used by ME32 would sometimes only return measures in `latest_approved`, whereas it should always be scoped to the passed transaction.

We had no test that draft data was considered, but now do.

### DIFF
--- a/measures/business_rules.py
+++ b/measures/business_rules.py
@@ -20,6 +20,7 @@ from common.util import TaricDateRange
 from common.util import validity_range_contains_range
 from common.validators import ApplicabilityCode
 from geo_areas.validators import AreaCode
+from measures.querysets import MeasuresQuerySet
 from quotas.validators import AdministrationMechanism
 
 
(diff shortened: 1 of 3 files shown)


gabelton (approved):

kintisheff (approved):

measures/tests/test_business_rules.py
```diff
@@ -465,26 +465,50 @@ def related_goods_nomenclature(request, existing_goods_nomenclature):
...
 )
-def existing_measure(request, date_ranges, existing_goods_nomenclature):
-    return factories.MeasureWithQuotaFactory.create(
-        **{
-            "goods_nomenclature": existing_goods_nomenclature,
-            "additional_code": factories.AdditionalCodeFactory.create(),
-            **request.param(date_ranges),
-        }
-    )
+def existing_measure(request, existing_goods_nomenclature):
```
stuaxo, line 492: This definitely wants a docstring, I can kind of see what it does now, but if someone changes it later it will be hard to know if it's correct.

measures/business_rules.py
```diff
@@ -336,11 +337,9 @@ def compile_query(self, measure):
...
         return query
 
-    def validate(self, measure):
-        if measure.goods_nomenclature is None:
-            return
-
-        # build the query for measures matching the given measure
+    def clashing_measures(self, measure) -> MeasuresQuerySet:
+        """Returns all of the measures that clash with the passed measure over
+        its lifetime."""
```
stuaxo, line 342: Not a blocker:

An example of how they can clash would be good context.

PR #474, opened 2022-01-20
Pin to major versions of Node and npm.

paulpepper-trade:
TP-1160 Jenkins build failing (part 2)

## Why
* Tamato is currently pinning to a specific patch-specific version of Node.
* Cloud Foundry no longer makes the pinned version available and so builds are failing (because of the missing version of Node).
* Minor versions of Node become unavailable.

## What
* Don't pin to a minor version of Node and npm.
* Pin to a major version of Node (16.x.x) and npm (7.x.x), which (hopefully) guarantees Node and npm availability within buildpacks.

## Checklist
* Requires migrations? No
* Requires dependency updates? No

## Links to relevant material
See: [Report to SRE with suggested action](https://ditdigitalteam.slack.com/archives/C1Q2EKZK3/p1642616596104100)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "repository": "https://github.com/uktrade/tamato.git",
   "licence": "MIT",
   "engines": {
-    "node": "~16.8.0",
-    "npm": "~7.19.1"
+    "node": "^16.8.0",
+    "npm": "^7.19.1"
   },
   "dependencies": {
     "accessible-autocomplete": "^2.0.3",


gabelton (approved):

package.json
```diff
@@ -7,8 +7,8 @@
   "repository": "https://github.com/uktrade/tamato.git",
   "licence": "MIT",
   "engines": {
-    "node": "~16.8.0",
-    "npm": "~7.19.1"
+    "node": "^16.8.0",
```
rachaelcodes, line 10: looks like 16.13.0 is the only minor supported version of 16: https://github.com/cloudfoundry/nodejs-buildpack/releases/tag/v1.7.66

rachaelcodes (changes requested):
Unfortunately, it looks like 16.8.0 wouldn't be supported

rachaelcodes (approved):

package.json
paulpepper-trade: I think the `^` pins to the major version only, allowing the minor version to be incremented: https://stackoverflow.com/a/22345808/471735

rachaelcodes: Yeah, just spotted that - node brain not fully working in the morning. Sorry about that!

simonwo:
The version used in automated testing is the one that is named… so now it sounds like the Github tests will be using a different version to the deployment? IS there any way to avoid that?

package.json
paulpepper-trade: It isn't the most intuitive notation!

PR #476, opened 2022-01-25
Tp 1015 mvp tariff managers are presented with the ability to download an envelope

gabelton:
## Why
To enable Tariff Managers to self service and make changes to the tariff, they need to be able to download the envelope to send to HMRC

## What
- Updates upload_workbasket_envelopes task to accept workbasket_ids param
- Updates WorkBasketSubmitView so that that task is called for the workbasket being submitted
- Updates DashboardView so that most recently approved workbasket dates are passed in to jinja download template
- Creates download_workbasket function view which gets the most recently uploaded envelope from s3 and serves it as attachment on the response.
- Adds view tests

Requires migration

Original design ticket https://uktrade.atlassian.net/browse/TP-1014

Screenshot 2022-01-25 at 17 58 49


exporter/tasks.py
```diff
@@ -104,14 +104,16 @@ def upload_and_create_envelopes(
...
+def upload_workbasket_envelopes(self, upload_status_data, workbasket_ids=None) -> Dict:
     """
     Upload workbaskets.
 
     :return :class:`~exporter.util.UploadTaskResultData`: object with user readable feedback on task status.
     """
     upload_status = UploadTaskResultData(**upload_status_data)
     workbaskets = WorkBasket.objects.filter(status=WorkflowStatus.APPROVED)
+    if workbasket_ids:
+        workbaskets = WorkBasket.objects.is_approved().filter(id__in=workbasket_ids)
```
stuaxo, line 116: It looks like like the line 114 does the same thing as the first part of this, could it be re-used and filtered by id ?

```python
workbaskets = workbaskets.filter(id__in=workbasket_ids)
```

workbaskets/tests/test_views.py
```diff
@@ -1,16 +1,29 @@
...
+from exporter.tasks import upload_workbasket_envelopes
 from workbaskets.models import WorkBasket
 from workbaskets.validators import WorkflowStatus
 
 pytestmark = pytest.mark.django_db
 
 
-def test_submit_workbasket(unapproved_transaction, valid_user, client):
+@patch("exporter.tasks.upload_workbasket_envelopes.s")
+@patch("exporter.tasks.send_upload_notifications.s")
```
stuaxo, line 19: Patching the celery task signature will work, but feels a brittle, I think we set celery to "always eager" during tests, so we can call the tasks, we need to make sure they don't *actually* upload anything to S3 or call APIs.

workbaskets/tests/test_views.py
```diff
@@ -30,6 +43,9 @@ def test_submit_workbasket(unapproved_transaction, valid_user, client):
 
     assert client.session["workbasket"]["status"] == WorkflowStatus.SENT
 
+    assert notifications.called
+    assert envelopes.called
```
stuaxo, line 47: It's best to check things were called with expected arguments, where possible.

workbaskets/views/__init__.py
```diff
@@ -142,3 +162,38 @@ def post(self, request, *args, **kwargs):
 
 class WorkBasketDeleteChangesDone(TemplateView):
     template_name = "workbaskets/delete_changes_confirm.jinja"
+
+
+def download_envelope(request):
+    """
+    Creates lambda function for sorting s3 objects by `last_modified` attribute.
```
stuaxo, line 169: Instead of grabbing the whole list of s3 objects and finding the oldest one, we could use information in our database.

We have an Upload object with a FileField, for every Envelope.
If there isn't a date for last modified, one could be added.

If you need further tracking of s3 objects, then the etag is a unique id for every s3 object, you can use it to request a particular object as needed.

workbaskets/views/__init__.py
```diff
@@ -142,3 +162,38 @@ def post(self, request, *args, **kwargs):
...
+        obj for obj in sorted(bucket.objects.all(), key=get_last_modified, reverse=True)
+    ][0]
+
+    with tempfile.NamedTemporaryFile() as f:
+        bucket.download_file(last_added.key, f.name)
+        with open(f.name, "rb") as fh:
+            response = HttpResponse(fh.read(), content_type="text/xml")
+            response[
+                "Content-Disposition"
+            ] = "attachment; filename=" + os.path.basename(last_added.key)
```
stuaxo, line 197: I'm assuming the bucket isn't accessible on the client side, for security (otherwise you could just send a redirect pointing at it).

Streaming the file is probably more optimal than downloading the whole thing.

boto has StreamingBody and iter_lines which allow you to stream, from s3.
On the django side [StreamingHttpResponse](https://docs.djangoproject.com/en/3.2/ref/request-response/#streaminghttpresponse-objects) sends data out as it's received.

common/views.py
```diff
@@ -93,6 +93,18 @@ def workbasket(self):
     def paginator(self):
         return Paginator(self.workbasket.tracked_models, per_page=10)
 
+    @property
+    def approved_workbasket(self):
```
stuaxo, line 97: There is only one approved workbasket now, but in the future there may be many, maybe the name could reflect that this is the last approved_workbasket.

Definitely add a docstring too :)

exporter/tasks.py
```diff
@@ -104,14 +104,16 @@ def upload_and_create_envelopes(
     retry_jitter=True,
 )
 @TableLock.acquire_lock(Envelope, lock="SHARE")
-def upload_workbasket_envelopes(self, upload_status_data) -> Dict:
+def upload_workbasket_envelopes(self, upload_status_data, workbasket_ids=None) -> Dict:
```
stuaxo, line 107: There's something to think about, re: failure modes and ordering.

When things are working OK:
- the user activates the job (e.g. by clicking a button),
- celery queues the job
- the job runs straight away
- everything is OK.

If there are issues with celery tasks may queue up and order may not be guaranteed.
It makes me wonder if we need to separately store the ordered list of manually queued workbaskets[1] ?

As a side effect it could make the task idempotent again.

[1] - Although this gets closer to Bos "task" design.

workbaskets/views/__init__.py
```diff
@@ -67,6 +76,17 @@ def get_redirect_url(self, *args, **kwargs):
...
+        workbasket_ids = [workbasket.id]
+
+        upload = (
+            upload_workbasket_envelopes.s(
+                upload_status_data={},
+                workbasket_ids=workbasket_ids,
+            )
+            | send_upload_notifications.s()
+        )
+        upload.delay()
```
stuaxo, line 88: It looks like `upload` wants to be a task itself.

stuaxo (changes requested):
Looking good so far, with a few tweaks todo.

exporter/tasks.py
gabelton: I was unsure on this one because `is_approved` allows for workbaskets with SENT, PUBLISHED, and ERRORED status. I needed to access PUBLISHED. Maybe it would be worth adding another method to WorkBasketQueryset that doesn't include the ERRORED status?

simonwo: The answer might actually be that `ERRORED` should not be in `is_approved` or that we have a different method like `is_current` or something like that? Basically `ERRORED` data should be treated as not visible to the rest of the system?

workbaskets/jinja2/includes/workbaskets/workbasket-selectable-items.jinja
```diff
@@ -39,9 +39,36 @@
...
-
+    </div>
+  </div>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-warning-text">
+        <span class="govuk-warning-text__icon">!</span>
+        <strong class="govuk-warning-text__text">
+          <span class="govuk-warning-text__assistive">Warning</span>
+          "We've created an envelope which contains your published changes between {{ "{:%d %b %Y}".format(approved_dates.start) }} and {{ "{:%d %b %Y}".format(approved_dates.end) }}. It's ready for your action"
```
simonwo, line 52: Can we not have the quotes around the test?

workbaskets/views/__init__.py
simonwo: I agree that this idea sounds like it will be easier to scale into letting TMs download other (not the most recent) envelopes later.

workbaskets/views/__init__.py
gabelton: Actually I've just realised that I can redirect to the bucket. Do you know how to trigger download from that redirect? I thought I could just add the download attribute in the <a> tag. At the moment I'm seeing this view :

Screenshot 2022-01-26 at 16 01 29

workbaskets/views/__init__.py
gabelton: I was looking for a way to update upload_workbaskets, so as to pass in the workbasket_ids kwarg to the first task, but couldn't find a way to do this.

exporter/tasks.py
stuaxo: In that case it sounds like we really just want `(status=WorkflowStatus.APPROVED)`.

SENT is later in `upload_and_create_envelopes` so we shouldn't' be resending these, we shouldn't be resending PUBLISHED - as in theory that would get sent at the end of this process, though I can't seem to find it - maybe it went when we removed the code that automatically triggers the upload task ?

workbaskets/views/__init__.py
stuaxo: I realised after that, you could probably just have an ordinary link to the bucket.

If the browser complains then we may need to add cross-domain rule to get it working.

workbaskets/views/__init__.py
stuaxo: We can pair on this, sometimes it's not 100% intuitive, which things should be tasks vs which should just be functions.

workbaskets/views/__init__.py
simonwo: http://iwantmyreal.name/s3-download-only-presigned-upload

You might be able to add the content disposition header as metadata?

exporter/tasks.py
gabelton: I don't think we'd be resending PUBLISHED in this current workflow. I think currently once the user clicks "submit all changes" the workbasket becomes PUBLISHED and I wanted to send the envelope at that same stage, which I needed is_approved for

common/views.py
```diff
@@ -93,6 +94,29 @@ def workbasket(self):
     def paginator(self):
         return Paginator(self.workbasket.tracked_models, per_page=10)
 
+    @property
+    def latest_approved_workbasket(self):
+        """Returns the most recently created WorkBasket with a status of
+        "APPROVED", "SENT", "PUBLISHED", or "ERRORED"."""
+        return WorkBasket.objects.is_approved().last()
```
simonwo, line 101: I'm guessing these are sorted by primary key – is there any possibility we might need to order by e.g. created time instead?

common/views.py
```diff
@@ -93,6 +94,29 @@ def workbasket(self):
...
+
+    @property
+    def download_ready(self):
+        return Upload.objects.count() > 0
```
simonwo, line 105: Does the message need to disappear whilst a workbasket is being generated? E.g. should we be filtering the Upload to just the last_approved_workbasket?

exporter/storages.py
```diff
@@ -17,6 +17,12 @@ def get_default_settings(self):
             default_acl="private",
         )
 
+    def get_object_parameters(self, name):
+        self.object_parameters.update(
+            {"ContentDisposition": f"attachment; filename={path.basename(name)}"},
+        )
+        return super().get_object_parameters(name)
```
simonwo, line 24: 👌

exporter/tasks.py
```diff
@@ -104,14 +104,18 @@ def upload_and_create_envelopes(
...
 
     :return :class:`~exporter.util.UploadTaskResultData`: object with user readable feedback on task status.
     """
     upload_status = UploadTaskResultData(**upload_status_data)
-    workbaskets = WorkBasket.objects.filter(status=WorkflowStatus.APPROVED)
+
+    workbaskets = WorkBasket.objects.is_approved()
+    if workbasket_ids:
+        workbaskets = workbaskets.filter(id__in=workbasket_ids)
+
```
simonwo, line 118: Does it definitely need to take specific IDs? Orginally when this was built the idea was that it would just upload "any workbaskets that are ready to send". Our assumption is that it'd be only one at a time until we build more task features. So is this necessary? Or could we stick with just "upload any that are ready"?

workbaskets/jinja2/includes/workbaskets/workbasket-selectable-items.jinja
```diff
@@ -39,9 +39,40 @@
...
+          <span class="govuk-warning-text__assistive">Warning</span>
+          We've created an envelope which contains your published changes between {{ "{:%d %b %Y}".format(approved_dates.start) }} and {{ "{:%d %b %Y}".format(approved_dates.end) }}. It's ready for your action.
+        </strong>
+      </div>
+      <details class="govuk-details" data-module="govuk-details">
+        <summary class="govuk-details__summary">
+          <span class="govuk-details__summary-text">Help with your actions</span>
+        </summary>
+        <div class="govuk-details__text">
+          This is a place holder for help text about tariff managers' actions and/or next steps
```
simonwo, line 62: Looks like we need to fill this out… something to bring up with PM?

common/views.py
gabelton: I was forgetting that WorkBasket has a created_at field. Updated to order_by this

exporter/tasks.py
gabelton: I can't remember why I did it this way now... I think I was imagining that loads could end up being sent, forgetting that the status gets set to SENT, so there would only ever be one sent with the way we have it setup currently. I don't know why I needed is_approved even because I should only have been sending an approved basket not PUBLISHED ?

Long way of saying: I think you're right and we shouldn't need this.

common/views.py
gabelton: Could we possibly have a quick call at some point next week to discuss the workflow logic? I think I need to write down exactly when the WorkBasket is meant to change states in this workflow and what the user should be seeing in terms of download options at each point

workbaskets/jinja2/includes/workbaskets/workbasket-selectable-items.jinja
SteveCN7: Definitely worth a quick passover with a content person.

simonwo: Or if we don't have anything to put here, we can just remove this and not block the PR.

SteveCN7: I can accept that - would be good to be clear this - I want to do a full content review of the pages at sometime

exporter/management/commands/quota_export.py
```diff
@@ -0,0 +1,143 @@
+from collections import defaultdict
```
simonwo, line 1: Did you mean to commit this?

workbaskets/views/__init__.py
```diff
@@ -142,3 +147,41 @@ def post(self, request, *args, **kwargs):
...
+
+    Returns `HttpResponseRedirect` with presigned url passed as only argument.
+    """
+    s3 = boto3.resource(
+        "s3",
+        endpoint_url=settings.AWS_S3_ENDPOINT_URL,
+        aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
+        aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
+        config=Config(signature_version="s3v4"),
+    )
```
simonwo, line 168: Do we gain anything by referring to the `HMRCStorage` instead of setting this up manually?
https://github.com/uktrade/tamato/blob/74381ac3dfa2542c4d539939d01a342011b3453a/exporter/storages.py#L9

simonwo (approved):

PR #477, opened 2022-01-27
ADR-15: Make reverse relations aware of versions.

simonwo:
ADR for fixing reverse relations – please read the text and discuss :)

### DIFF
--- a/docs/source/adr/0015-make-reverse-relations-aware-of-versions.rst
+++ b/docs/source/adr/0015-make-reverse-relations-aware-of-versions.rst
@@ -1,0 +1,83 @@
+15. Make reverse relations aware of versions
+============================================
+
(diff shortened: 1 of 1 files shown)


stuaxo:
It sounds like a sane default (even an exception being thrown would be better than what happens now).

Are there any circumstances where you wouldn't want this, and would want to opt for an older version for instance ?

simonwo:
> Are there any circumstances where you wouldn't want this, and would want to opt for an older version for instance ?

I can't think of any – this should always just work. As it notes there is a TODO around whether those reverse relations should include magic versioning, but I'm not including that for now.

gabelton (approved):

paulpepper-trade (approved):

PR #478, opened 2022-01-31
Fix reverse relations on all TrackedModels

simonwo:
Implements ADR-15.

When tracked models exist in a one-to-many
relationship (such as how one parent object can
have many descriptions) each description object
will hold a foreign key back to the parent object
that specifically links to the version it was
created against.

    Parent X (primary key = 1)
    Description X (primary key = 2, parent = 1)

Calling the reverse relation on the parent will
correctly discover the descriptions:

    parent.descriptions
    # => [<Description: pk=2>]

If that parent object is updated, it receives a
new row in the database, but the descriptions are
not updated. This means that the reverse
relationship now no longer returns any objects,
because the reverse relationship by default tries
to find descriptions that have the primary key of
the *new* parent object:

    parent_v2 = parent.new_version(...)
    parent_v2.descriptions # => []

What needs to be done instead is not use the
database primary key of the parent object but
instead refer to that object's version group,
which will stay constant between versions.

    Parent.objects = TrackedModelManager.from_queryset(...)
    parent_v2.descriptions # => [<Description: pk=2>]

This commit therefore creates a new Manager class
(that will be inherited by RelatedManager) and
that replaces PKs with the version group of the
instance. For more information on why this works
see django/db/models/fields/related_descriptors.py.

### DIFF
--- a/common/models/managers.py
+++ b/common/models/managers.py
@@ -1,0 +1,85 @@
+from typing import Any
+from typing import Dict
+from typing import Tuple
+
+from django.db.models import Model
+from django.db.models.fields import Field
+from polymorphic.managers import PolymorphicManager
+
+
+class TrackedModelManager(PolymorphicManager):
+    """
+    Fixes reverse relations so that always return correct results for versioned
+    models.
+
+    For why, see ADR `15. Make reverse relations aware of versions`_.
+
+    This mixin swaps out the foreign keys used by the RelatedManager for
+    querying and replaces them with the version group of the instance. The
+    apporach is to take over control of the `core_filters` attribute used in
+    django/db/models/fields/related_descriptors.py.
+
+    The attribute is only set in two places – either once for a RelatedManager
+    or (in the case of a many-to-many relation) updated in a loop, but the loop
+    is only dependent on context provided by the relation itself. Hence it is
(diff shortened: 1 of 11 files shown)


common/models/managers.py
```diff
@@ -0,0 +1,85 @@
...
+            return {f"{self.field.name}{suffix}": object}
+        else:
+            # ManyRelatedManager: this is a many-to-many relationship with an
+            # auto-generated through model. The auto-generated model does not
+            # take part in the TrackedModel system – hence each version of the
+            # many-to-many models needs an independent set of links, otherwise
+            # it wouldn't be possible for the links to be updated. The
+            # implementation is the same as the default and it is up to the code
+            # that creates new versions to duplicate the many-to-many links.
+            return {
```
gabelton, line 73: Is it worth adding a test this 'else' scenario ?

simonwo: This is a test that covers it: https://github.com/uktrade/tamato/blob/master/quotas/tests/test_models.py#L51. It was originally designed to test `new_version` but it actually tests that this works/continues to work too.

But if you think we need another test, let me know.

gabelton: nope, lgtm in that case!

gabelton (approved):

PR #479, opened 2022-01-31
Exception in measures wizard on invalid data when no duty sentence entered yet.

stuaxo:
When the duties field isn't present (it turns out this shouldn't happen - all fields should be present - but that may be a fix for another time).

The wizard dies with an exception - this makes the logic use .get() so the wizard doesn't die in this way.


simonwo:
Hm, is a blank string the correct thing to return here? If we do end up not having the duties field in the wizard (which would make sense because sometimes we create measures without duties, e.g. if they are import/export controls) then would the system try to parse a blank string and error? And if so would it override any existing duties that are present, if we are editing??

Would `None` be better? And then downstream we don't try and modify duties if the value is None?

stuaxo:
I should have put a comment - the parser complaining that this was empty, is probably the behaviour I was aiming for.

This certainly could be changed, what I wasn't sure of with the `None` approach was exactly what should happen at the other end - though your comment clarifies it a bit + can definitely try the approach where None == don't modify.

simonwo:
Basically it being empty will not be an error in some situations – there is a business rule that takes care of that.

simonwo (approved):

paulpepper-trade (approved):

PR #480, opened 2022-01-31
Increase timeout to 120 seconds to help large upload.

stuaxo:
Increase gunicorn timeout to 120 seconds to stop large uploads timing out.


simonwo (approved):
It's worth trying!

In the future it might be wise as a security measure to work out if this can be scoped down on a per endpoint basis or not.

PR #483, opened 2022-02-02
Bugfix - described objects with no url should not break the homepage.

stuaxo:
If a described_object get_url() returned None an error occured on the index page.

### DIFF
--- a/common/models/mixins/description.py
+++ b/common/models/mixins/description.py
@@ -45,7 +45,10 @@ class DescriptionMixin(ValidityStartMixin):
             kwargs = self.get_identifying_fields()
             described_object = self.get_described_object()
             if action == "detail":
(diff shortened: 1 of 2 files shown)


simonwo (approved):
From looking at https://github.com/uktrade/tamato/blob/b4b930e4aacd9399c3a4cec6e9208e94e3f7ba34/common/models/trackedmodel.py#L605, it looks like get_url will return None if there's not a path configured for that model. So could be a description we don't currently allow editing of, like a GoodsNomenclatureDescription? Anyway, it seems like handling that here is reasonable.

stuaxo:
OK, YOLOing this, and will do a part two afterwards once I work out how to get our factories to recreate data like this for the frontend.

PR #484, opened 2022-02-03
Create 0016-split-transactions-into-partitions-to-simplify-global-tra…

stuaxo:
When I went to explain transactions partitioning earlier I had to refer back to the PR, which was a good sign this needed documenting.

I'm adding this a retrospective ADR, aware this isn't in the spirit of ADRs, but equally this seems like the best place to document this.


kintisheff (approved):

kintisheff:
> This may be in a bit of a raw state - I need to go back and re-read this and fix for clarity.

i think it looks good even as is

simonwo:
I think your ADR looks likely to land first @kintisheff, so if you take 16 we can always easily update this?

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,75 @@
...
+TAMATO follows TARICs concept of a "transaction" as a record of changes to the Tariff.
+
+In TAMATO Transaction is a concrete database model with records reflecting TRANSACTION elements imported from TARIC XML.
+
+Order of transactions is important, beforet the PR for partitioning was written global ordering of transactions used the following logic:
+
+1.  Transactions with DELETED status were excluded.
+
+2. Transactions were ordered into groups:
+ - The first workbasket (considered the "seed file")
```
paulpepper-trade, line 24: Could say a little about what the 'first' workbasket is, what a seed file is and why they're important.

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,75 @@
...
+Retrospective ADR for TP-841
+
+Context
+-------
+
+TAMATO follows TARICs concept of a "transaction" as a record of changes to the Tariff.
+
+In TAMATO Transaction is a concrete database model with records reflecting TRANSACTION elements imported from TARIC XML.
+
+Order of transactions is important, beforet the PR for partitioning was written global ordering of transactions used the following logic:
```
paulpepper-trade, line 19: A few words about why ordering is important? "Ordering of transactions is important because... ". Or reference our other documentation that remarks on ordering - if it exists.

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,75 @@
...
+
+1.  Transactions with DELETED status were excluded.
+
+2. Transactions were ordered into groups:
```
paulpepper-trade, line 23: I know it's history now, but were these groups somehow inferred previously If so, perhaps say how.

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,75 @@
...
+    SEED_FILE = 1
+
+    REVISION = 2
+
+    DRAFT = 3
+
+Maintaining compatibility with the existing system
+--------------------------------------------------
+
+To ensure the system works as before, the concept of a Transaction Partition Schema is introduced.
```
paulpepper-trade, line 47: What does 'works as before' encompass?

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,75 @@
...
+`TransactionPartitionScheme`s implement the logic that is called when a workbasket is saved and it's Transactions need to move from `DRAFT` to `SEED_FIRST` or `REVISION`.
```
paulpepper-trade, line 48: > `TransactionPartitionScheme`s implement the logic that is called

Is it possible to say what that logic is or characterise it briefly?

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,75 @@
...
+
+
+SEED_FIRST
+..........
+The SEED_FIRST Transaction Schema implements the same logic the app had before partitions were implemented.
+Transactions in the first workbasket are set to SEED_FIRST and any others go in the REVISION partition.
+
+SEED_ONLY, REVISION_ONLY
+........................
+These schemas can be used when there is a need to manually specify a partition from the commandline.
```
paulpepper-trade, line 61: > These schemas can be used when there is a need to manually specify a partition from the commandline.

In what circumstances is this necessary? An example or two may help illustrate how and why partitions are used more generally?

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,75 @@
...
+
+
+Consequences
+------------
+
+- Global ordering is simplified to sorting by partition, order.
+- Ordering in views is much simpler
+- More complex logic is moved to `TransactionPartitionScheme` and only called during workbasket approval.
+- In the future multiple seed file transactions may be used if required.
```
paulpepper-trade, line 70: What does it mean to have multiple seed file transactions? (Probably related to comment above about  what a seed file is.)

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,75 @@
...
+
+SEED_ONLY, REVISION_ONLY
+........................
+These schemas can be used when there is a need to manually specify a partition from the commandline.
+
+
+Consequences
+------------
+
+- Global ordering is simplified to sorting by partition, order.
```
paulpepper-trade, line 67: Is the simplification one of less complex queries, or is there more to it than that?

paulpepper-trade (approved):

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
stuaxo: Quick dump of some info from @simonwo 

image

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
stuaxo: Note to self:  There should be info about TARIC transaction ids somewhere.

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
stuaxo: Yep - note to self, see the chat with Simon, pasted above.

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,164 @@
...
+Retrospective ADR for TP-841
+
+Context
+-------
+
+TAMATO follows TARICs concept of a "transaction", that groups individual changes to the Tariff.
+
+In TAMATO a Transaction is a concrete database model that reflects TRANSACTION elements imported from TARIC XML files known as envelopes.
+
+The Tariff is made up of an ordered sequence of transactions and transactions in the TaMaTo database must follow the same order.
```
LaurenQurashi, line 19: Is this supposed to be the Taric is made up of an ordered sequence ... and transactions in the Tamato db must follow the same order?

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,172 @@
...
+2. Transactions were grouped by workbasket, within the workbasket the Transaction order field dictated ordering.
+
+
+Workbaskets themselves were ordered in the following order:
+
+ - The first workbasket (workbasket with pk=1)
+   This workbasket only contains data imported from the initial "seed file".
+
+ - Published workbaskets (workbaskets with any of the approved statuses).
+   These workbaskets were imported from "delta files" or created in TaMaTo.
```
paulpepper-trade, line 34: "delta files" created in TaMaTo

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Revision Transactions
+.....................
+
+Subsequent updates are imported from "delta XML" files and called revisions (in the earlier version of the system
+this naming was not formalised).
+
+Keeping revision transactions separate from those from the seed file has advantages:
+
+- For parallelization (the seed file can be imported in parallel with the delta files).
```
paulpepper-trade, line 67: I must be misunderstanding this. Wouldn't the seed file need to be in place (fully imported)  before deltas are applied?

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,172 @@
...
+- For debugging versioning (you don't need to know how many transactions are in the seed file),
```
paulpepper-trade, line 68: "you don't need to know how many transactions are in the seed file" why is that?

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
```diff
@@ -0,0 +1,172 @@
...
+The existing system didnt centralise ordering splitting it between the view and the importer.
+Ordering was complex - with information having to be fetched from the workbasket to order transactions, this can add up when there is a lot of data (for example there are 9 million items in the seed file workbasket).
+
+
+Decision
+--------
+
+Where possible replace grouping logic with concrete data in the database.
+
+This is achieved by adding a `partition` enum field, to Transaction with values corresponding to the above groups and with values that reflect the groups ordering:
```
paulpepper-trade, line 114: Is it worth saying why the partitioning is applied at the transaction level, rather than the workbasket that groups them?

paulpepper-trade (approved):

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
stuaxo: How about -

The Tariff represented by the TAMATO database must respect the ordering of the data imported from TARIC XML envelopes.

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
stuaxo: Good point - and I think we in fact don't do that... I think it probably is possible.

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
stuaxo: 1 workbasket != 1 envelope.

On export transactions from the passed in workbaskets are all passed to the exporter and exported as an envelope [1].
The exported data may be split before reaching a size, this means some data in a workbasket could be published and some other workbasket not.


[1] - Note, we mostly DO do 1 workbasket == 1 envelope, but the system isn't constrained in this way.


Some of these questions may have to open ones, since the ADR is retrospective.. if I can't think of a good answer I may put the questions at the end of the doc.

stuaxo: Funnily enough, the other day I was thinking a simpler scheme may have been to use a field on workbasket as the ordering - this would make more sense if all the workbasket models were brought into the common app.

docs/source/adr/0016-split-transactions-into-partitions-to-simplify-global-transaction-ordering.rst
stuaxo: Re-written I think this should make sense now.

PR #485, opened 2022-02-04
Tp 1162 edit duties

gabelton:
## Why
The ability to edit duties is in the original designs, but currently missing from editor functionality

## What
- Adds a diff_components method to `Measure` to determine whether to create, update, or delete components 
- Adds logic to `MeasureForm` to ensure that this method is only called when user has changed or added a duty value
- Adds unit tests for three `diff_components` scenarios and form logic
- Adds conftest fixtures to make it easier to create an instance of `MeasureForm`
- Deletes unused measure util, `clean_duty_sentence`

This [bug](https://uktrade.atlassian.net/browse/TP2000-100) was discovered while doing this ticket and so we should let the tariff managers know that until it's fixed the displayed value will look like it's duplicated (see screenshot below)

Screenshot 2022-02-04 at 16 05 56

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1,6 +1,5 @@
 import logging
 from collections import defaultdict
-from datetime import date
 
 from crispy_forms_gds.helper import FormHelper
 from crispy_forms_gds.layout import HTML
@@ -27,7 +26,6 @@ from geo_areas.forms import GeographicalAreaFormMixin
 from geo_areas.forms import GeographicalAreaSelect
 from geo_areas.models import GeographicalArea
 from measures import models
-from measures.parsers import DutySentenceParser
 from measures.validators import validate_duties
 from quotas.models import QuotaOrderNumber
 from regulations.models import Regulation
@@ -98,11 +96,15 @@ class MeasureForm(ValidityPeriodForm):
 
         WorkBasket.get_current_transaction(self.request)
 
-        if hasattr(self.instance, "duty_sentence"):
-            self.initial["duty_sentence"] = self.instance.duty_sentence
-            self.request.session[
-                f"instance_duty_sentence_{self.instance.sid}"
-            ] = self.instance.duty_sentence
+        if not hasattr(self.instance, "duty_sentence"):
+            raise AttributeError(
+                "Measure instance is missing `duty_sentence` attribute. Try calling `with_duty_sentence` queryset method",
+            )
+
+        self.initial["duty_sentence"] = self.instance.duty_sentence
+        self.request.session[
+            f"instance_duty_sentence_{self.instance.sid}"
+        ] = self.instance.duty_sentence
 
         self.initial_geographical_area = self.instance.geographical_area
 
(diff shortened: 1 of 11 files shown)


measures/forms.py
```diff
@@ -91,6 +98,12 @@ def __init__(self, *args, **kwargs):
 
         WorkBasket.get_current_transaction(self.request)
 
+        if hasattr(self.instance, "duty_sentence"):
+            self.initial["duty_sentence"] = self.instance.duty_sentence
+            self.request.session[
+                f"instance_duty_sentence_{self.instance.sid}"
+            ] = self.instance.duty_sentence
+
```
simonwo, line 106: What happens if `self.instance` does not have a `duty_sentence`? It looks like from the code below that it will try to access `self.request.session[f"instance_duty_sentence_{self.instance.sid}"]` anyway and get a `KeyError`?

So should it really be an error to pass in an instance without a duty sentence?

measures/forms.py
```diff
@@ -115,9 +128,58 @@ def __init__(self, *args, **kwargs):
...
+                )
+
     def clean(self):
         cleaned_data = super().clean()
 
+        duty_sentence = cleaned_data["duty_sentence"]
+        valid_between = self.initial.get("valid_between")
+        if valid_between is not None:
+            validate_duties(duty_sentence, valid_between.lower)
+
```
simonwo, line 182: Would this be better factored as the content of a `clean_duty_sentence` method?

measures/forms.py
```diff
@@ -153,6 +215,19 @@ def save(self, commit=True):
 
         sid = instance.sid
 
+        if (
+            self.request.session[f"instance_duty_sentence_{self.instance.sid}"].replace(
+                " ",
+                "",
+            )
```
simonwo, line 222: This replacing spaces with blanks triggers my spidey sense… what problem is this solving? It looks like we might be trying to work around some issue with the duty sentence formatting?

measures/tests/test_forms.py
```diff
@@ -0,0 +1,87 @@
...
+    original_component = factories.MeasureComponentFactory.create(
+        component_measure=measure_form.instance,
+        duty_amount=9.000,
+        duty_expression=percent_or_amount,
+    )
+    measure_form.diff_components(
+        "8.000%",
+        measure_form.instance,
+        measure_form.instance.valid_between.lower,
+    )
```
simonwo, line 25: Does this suggest that actually `diff_components` could/should be a method independent to a form? I can see it coming in handy within notebooks, for example.

measures/forms.py
gabelton: When `self.instance` doesn't have a `duty_sentence` it doesn't reach line 103

measures/tests/test_forms.py
gabelton: I think it could live in measures/util.py maybe?

gabelton: Also just noticed that there's a clean_duty_sentence in measures/util.py that doesn't appear to be used anywhere in the codebase. Do you know if we use it for TOPS work ?

measures/forms.py
simonwo: Yes – won't that cause a KeyError on 219?

measures/tests/test_forms.py
simonwo: I think I authored it and probably factored it out of some early TOPS work, yes. If it looks useful, we should probably use it!

measures/forms.py
gabelton: doh, sorry yeah of course. I don't know why that didn't come out while testing - I did run it against some measures without duties

gabelton: I think I might have added this while I was still testing against v old measures. I don't think it's necessary and hasn't been causing a key error, as measures without a duty will just have an empty string value for `duty_sentence`

simonwo: Yes – as long as the measure being passed in has come back from a queryset with `with_duty_sentence()` called… maybe it's worth a defensive error of some sort if the measure has been passed without `with_duty_sentence()` being called?

measures/tests/test_forms.py
simonwo: Actually that method is not all that helpful – just some code to handle when Excel represented things as numbers. So maybe not that relevant if we always know we'll bet getting form content as a string.

simonwo (approved):

PR #486, opened 2022-02-07
TP-935 Standardise current transaction time - #2

paulpepper-trade:
# TP-935 Standardise current transaction time - #2

## Why

We have a lot of database queries that specify the transaction-time (one of the time
axes of our bitemporal data) to filter results. The most common case is to filter by the
latest approved transaction including draft changes in the current workbasket. This
often results in having to pass the transaction down long function call chains.

We also need to perform queries against different transactions in the same code, and so
we need a way to temporarily override the current transaction.

## What

- Add thread-local storage to keep a reference to the current transaction, so that database querying code can access it directly without having to pass it around.
- Add a contextmanager, which sets the thread-local stored transaction to a new value within the scope of a with block, then resets it afterwards.
- Add a lazily evaluated query value, this allows defining a query once and then reusing it with different transactions without having to build a separate query for each transaction.

## Checklist
- Requires migrations? No

This PR updates and replaces [TP-935 Standardise transaction time](https://github.com/uktrade/tamato/pull/436).


common/models/tracked_qs.py
```diff
@@ -43,6 +45,15 @@ def latest_approved(self) -> TrackedModelQuerySet:
             update_type=UpdateType.DELETE,
         )
 
+    def current(self) -> TrackedModelQuerySet:
+        """Get the approved versions of the model, or the latest draft versions
+        if they exist within a transaction preceding (and including) the given
+        transaction in the workbasket of the given transaction."""
```
simonwo, line 51: We should update this docstring to talk about it's automatic behaviour.

common/models/trackedmodel.py
```diff
@@ -97,6 +98,10 @@ class TrackedModel(PolymorphicModel):
     objects: TrackedModelQuerySet = TrackedModelManager.from_queryset(
         TrackedModelQuerySet,
     )()
+    # current_objects filters instances to the 'current' transaction.
+    current_objects: TrackedModelQuerySet = CurrentTrackedModelManager.from_queryset(
+        TrackedModelQuerySet,
+    )()
```
simonwo, line 104: If you add a docstring _after_ the field, it will get attached and be visible to e.g. architecture docs…

common/models/utils.py
```diff
@@ -0,0 +1,74 @@
...
+
+class LazyValue(Value):
+    """
+    Lazily evaluated Value.
+
+    Calls a function passed to the constructor when ``value`` property is
+    accessed. This allows the value to change with each access.
+    """
+
+    allow_list: FrozenSet[str] = frozenset()
```
simonwo, line 18: We could do with some docs on what this does.

common/models/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        from workbaskets.models import WorkBasket
+
+        set_current_transaction(
+            WorkBasket.get_current_transaction(request),
+        )
+        response = self.get_response(request)
```
simonwo, line 60: Should this use `override_current_transaction` instead? So that whatever default value is restored once the response is finished?

common/models/tracked_qs.py
paulpepper-trade: Updated.

common/models/trackedmodel.py
paulpepper-trade: Done.

common/models/utils.py
paulpepper-trade: Done.

common/models/utils.py
paulpepper-trade: Done.

common/tests/test_models_utils.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@pytest.fixture
+def model1_with_history(date_ranges):
+    return model_with_history(
+        factories.TestModel1Factory,
+        date_ranges,
+        version_group=factories.VersionGroupFactory.create(),
+        sid=1,
+    )
```
simonwo, line 23: How about putting this (and maybe `models_with_history` etc) in a `common/tests/conftest.py`? For sharing between this file and the other file that has this fixture.

common/tests/test_models_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    with override_current_transaction(tx2):
+        assert get_current_transaction() is tx2
+
+    assert get_current_transaction() is tx
+
+
+def test_current_objects_model_manager(model1_with_history):
+    for model_version in model1_with_history.all_models:
+        set_current_transaction(model_version.transaction)
+        assert TestModel1.current_objects.count() == 1
```
simonwo, line 64: Would it also be worth adding `assert TestModel1.current_objects.get() == model_version`? Because right now that could return any arbitrary version.

simonwo (approved):

common/tests/test_models_utils.py
paulpepper-trade: Done.

common/tests/test_models_utils.py
paulpepper-trade: Done.

PR #490, opened 2022-02-09
TP2000-89 "described objects with no url break the front page" The tests for this.

stuaxo:
Tests for TP2000-89.

E2E test that data in the current workbasket is actually displayed on the frontpage (currently only GoodsNomenclature and related, as in the bug)

Call getURL on every tracked model and make sure it outputs either None or something nominally like a URL.


common/tests/test_views.py
```diff
@@ -29,6 +31,21 @@ def test_index_doesnt_creates_workbasket_if_not_needed(
     assert WorkBasket.objects.is_not_approved().count() == 1
 
 
+def test_index_displays_links_to_described_objects(
+    valid_user_client,
+    workbasket,
+):
+
+    with workbasket.new_transaction():
+        GoodsNomenclatureFactory.create()
```
simonwo, line 42: Would it be good to use the `trackedmodel_factory` and check this for every model type?

common/tests/test_models.py
```diff
@@ -458,6 +458,19 @@ def test_get_description_dates(description_factory, date_ranges):
     assert future == future_description
 
 
+def test_trackedmodel_get_url(trackedmodel_factory):
+    """Verify no getUrl() returns something sensible and doesn't crash."""
```
simonwo, line 462:
```suggestion
    """Verify get_url() returns something sensible and doesn't crash."""
```

common/tests/test_models.py
```diff
@@ -458,6 +458,19 @@ def test_get_description_dates(description_factory, date_ranges):
...
+    instance = trackedmodel_factory.create()
+    url = instance.get_url()
+
+    if url is None:
+        # None is returned for models that have no URL
+        return
+
+    # Very basic check that this looks like a url.
+    assert "/" in url
```
simonwo, line 471: Could we do something with urllib to check that it's a valid relative url?

common/tests/util.py
```diff
@@ -770,3 +770,45 @@ def assert_transaction_order(transactions):
...
+    table_data = list(get_table_content(table, col_start=1))
+
+    tracked_models = list(tracked_models)
+    for i, obj in enumerate(tracked_models):
+        expected = {
+            "ID": f'{obj.get_described_object() if hasattr(obj, "get_described_object") else obj.structure_code}',
+            "Type": f"{obj._meta.verbose_name.title()}",
+            "Description": obj.structure_description or "-",
+            "Activity date": f"{obj.transaction.updated_at:%d %b %Y}",
+        }
```
simonwo, line 810: I'm worried this is a little brittle because it's just replicating the logic that's in the jinja?

simonwo (approved):

common/tests/test_models.py
stuaxo: Relative, or just local ?

I think we want local so have put that in.

common/tests/util.py
stuaxo: This is true, and I felt the same doing this.

Some of the logic in Jinja should probably be in the models themselves, so it's easier to test, 

```python
get_id_display()
get_type_display()
get_description_display()
get_activity_date_display()
```
This would mean less bits in Jinja that try and call some method that isn't on a model before defaulting to the next thing.

One thing I considered was setting actual values in the factories and checking them later, there are qutie a lot though, maybe there is some middle ground that doesn't make the test data really big.

simonwo: Are these sorts of display properties meant to be set on the views? I.e. treating the views as "view models"? Or something.

common/tests/util.py
```diff
@@ -770,3 +770,48 @@ def assert_transaction_order(transactions):
...
+        cols = list(row.findAll("td"))[col_start:]
+        if not cols:
+            continue
+
+        data = {headers[i]: col.getText().strip() for i, col in enumerate(cols)}
+        yield data
+
+
+def assert_table_displays_models(table, tracked_models):
+    """Given a beautiful soup object containtain a table, check that it contains
```
gabelton, line 799:
```suggestion
    """Given a beautiful soup object containing a table, check that it contains
```

common/tests/util.py
stuaxo: I guess the view could expose a method, add a function to the view make the object suitable for consumption by the list.

Where data is just transformed for display I think that's suited to for this.
When it comes to anything where methods are available on some objects but not others, I think we should have the one method name available on all the classes.

common/tests/util.py
stuaxo: Cheers :)

common/tests/test_views.py
paulpepper-trade: The dashboard's "Your tariff changes" list of TrackedModels is split across pages of 10 items (models) per page. So we'd need to constrain a test using the current approach to a maximum of 10 models - `common.tests.factories.MeasureFactory` breaks this for example, because it generates a Measure instance and 11 associated models.

The current approach to this test ensures that a basket with 5 models (the total number generated by our `GoodsNomenclatureFactory`) is displayed correctly on the dashboard page. Perhaps our WorkBasket tests would be a better place to test for mixtures of model types, if they don't already do that?

common/tests/util.py
paulpepper-trade: This has been reworked now, so the test checks that the workbasket representation on the page (the dashboard's paged list of "Your tariff changes") contains all model instances by referencing their IDs. I don't think there's a need to get the extra information shown here. Or is there?

workbaskets/forms.py
```diff
@@ -16,23 +16,38 @@ class SelectableObjectsForm(forms.Form):
...
+            self.fields[
+                SelectableObjectsForm.field_name_for_object(obj)
+            ] = SelectableObjectField(
                 required=False,
                 obj=obj,
                 initial=str(obj.id) in [str(k) for k in self.initial.keys()],
             )
 
+    @classmethod
+    def field_name_for_object(cls, obj):
```
stuaxo, line 36: Will this never be overridden ?  - Maybe it could be a standard method ?

paulpepper-trade: Although the form field's name-prefix could be overridden, I think we're unlikely to need to do so, and this approach simplifies accessing that prefix.
A class method also helps with testability, since there's no access to the Form instance within the test - for instance there's no context available on the `response` object.

common/tests/test_views.py
paulpepper-trade: `test_index_with_each_type_of_object_in_current_workbasket` now added in order to test displaying the workbasket using every type of object on the dashboard.

PR #492, opened 2022-02-10
TP2000-50 Enable creation of measures for multiple commodities

eadpearce:
Updated the create measure wizard so users can select multiple commodity codes to create measures for. There are no changes to the models.


measures/tests/test_forms.py
```diff
@@ -35,3 +38,96 @@ def test_error_raised_if_no_duty_sentence(session_request):
...
+
+
+@pytest.fixture()
+def measure_type():
+    return factories.MeasureTypeFactory.create(
+        valid_between=TaricDateRange(datetime.date(2020, 1, 1), None, "[)")
+    )
+
+
+@pytest.fixture()
```
gabelton, line 55: Perhaps you could move these fixtures to the conftest file, so that all the fixtures are in the same place ?

measures/tests/test_forms.py
```diff
@@ -35,3 +38,96 @@ def test_error_raised_if_no_duty_sentence(session_request):
...
+def regulation():
+    return factories.RegulationFactory.create()
+
+
+@pytest.mark.django_db
```
gabelton, line 60: We normally set `pytestmark = pytest.mark.django_db` below the imports at the top of a tests file, so that you don't have to repeat it on each test

measures/tests/test_forms.py
eadpearce: yep I only just realised I duplicated these when I was looking at the diff! can do

measures/tests/test_forms.py
eadpearce: ooh that's what that does. will update

measures/tests/test_forms.py
```diff
@@ -35,3 +38,96 @@ def test_error_raised_if_no_duty_sentence(session_request):
         match="Measure instance is missing `duty_sentence` attribute. Try calling `with_duty_sentence` queryset method",
     ):
         MeasureForm(data={}, instance=measure, request=session_request)
+
+
+@pytest.fixture()
+def additional_code():
```
gabelton: I guess there's a more general question about whether we want to be creating fixtures that just make a vanilla create() call? We do have things like approved_workbasket in conftest, so might be a good shout

measures/tests/test_forms.py
gabelton: Also means you won't have to create the same fixture again in you test_views file

stuaxo:
Nice work on adding tests to the measures wizard.

It looks like you basically added some of the tests I needed to for
https://github.com/uktrade/tamato/pull/479

measures/tests/test_forms.py
```diff
@@ -35,3 +36,74 @@ def test_error_raised_if_no_duty_sentence(session_request):
...
+
+def test_measure_forms_additional_code_valid_data(additional_code):
+    data = {
+        "additional_code": additional_code.pk,
+    }
+    form = forms.MeasureAdditionalCodeForm(data, prefix="")
+    assert form.is_valid()
+
+
+def test_measure_forms_additional_code_invalid_data():
```
gabelton, line 92: Perhaps you could parametrize this and the above?

gabelton: And maybe do the same for the MeasureDetailsForm ones

measures/views.py
```diff
@@ -158,63 +167,71 @@ def done(self, *args, **kwargs):
...
+                            duty_amount=condition.duty_amount,
+                            monetary_unit=measure_creation_pattern.condition_sentence_parser.monetary_units[
+                                "GBP"
+                            ],
+                            component_measurement=None,
+                        )
+                        component.save()
+
+            except AssertionError as e:
+                raise ValidationError(e) from e
```
simonwo, line 234: If we are creating multiple measures, and the first one succeeds but the second one fails for some reason, we will present a validation error to the user. If they fix the error and then resubmit, the form will create the first measure again despite the fact it was successfully created previously.

Do we need to try and create all of the measures in a single database transaction and roll it back if any of them fail?

simonwo (commented):
Looks great!!

measures/forms.py
```diff
@@ -2,15 +2,17 @@
...
+    HTML,
+    Button,
+    Div,
+    Field,
+    Fieldset,
+    Fluid,
+    Layout,
+    Size,
+    Submit,
+)
```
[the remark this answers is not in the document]
eadpearce: I'm not sure I do. Is the red the correct formatting?

measures/forms.py
```diff
@@ -348,26 +349,20 @@ def clean(self):
                 cleaned_data["valid_between"],
             ):
                 raise ValidationError(
-                    f"The date range of the measure can't be outside that of the measure type: {measure_type.valid_between} does not contain {cleaned_data['valid_between']}",
+                    f"The date range of the measure can't be outside that of the measure type: "
+                    "{measure_type.valid_between} does not contain {cleaned_data['valid_between']}",
```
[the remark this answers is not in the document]
eadpearce: I'll need to double check but I think this still works

measures/views.py
```diff
@@ -158,63 +167,71 @@ def done(self, *args, **kwargs):
...
-
-                # XXX the design doesn't show whether the condition duty_amount field
-                # should handle duty_expression, monetary_unit or measurements, so this
-                # code assumes some sensible(?) defaults
-                if condition.duty_amount:
-                    component = MeasureConditionComponent(
-                        condition=condition,
+        measures_data = []
+
+        for commodity_data in cleaned_data.get("formset-commodities", []):
```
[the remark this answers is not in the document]
eadpearce: It should fail further up when we try to get the cleaned data for the form. Still trying to wrangle how to test this though https://github.com/uktrade/tamato/pull/492/files#diff-3806f913d8afec873da85c1bf042083bd9a459178e7c195dbde32a8751ff1d1fR158

measures/forms.py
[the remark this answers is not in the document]
eadpearce: It doesn't work! Fixed and updated test

measures/tests/test_views.py
```diff
@@ -260,3 +262,162 @@ def test_measure_form_save_called_on_measure_update(
...
+        ],
+    }
+
+    wizard = MeasureCreateWizard(request=mock_request)
+
+    measures = wizard.create_measures(form_data)
+
+    assert len(measures) == 2
+    assert len(measures[0].footnotes.all()) == 1
+    assert len(measures[1].footnotes.all()) == 1
```
simonwo, line 421: Is there any value in checking that `measures[0].footnotes.get() == footnote1`? So that it's saving the right footnote?

measures/views.py
```diff
@@ -97,156 +96,187 @@ class MeasureCreateWizard(
...
+        },
+        SUMMARY: {
+            "title": "Review your measure",
+            "link_text": "Summary",
+        },
+        COMPLETE: {
+            "title": "Finished",
+            "link_text": "Success",
+        },
+    }
```
simonwo, line 177: Nice 👌

measures/views.py
```diff
@@ -97,156 +96,187 @@ class MeasureCreateWizard(
...
+                    "duty_sentence": data["duties"],
+                }
+
+                measures_data.append(measure_data)
+
+        created_measures = []
+
+        for measure_data in measures_data:
             measure = measure_creation_pattern.create(**measure_data)
             for i, condition_data in enumerate(
```
simonwo, line 221: I think the enumerate needs `start=1` to avoid the bug with component sequence number being zero. But maybe @rachaelcodes has fixed that elsewhere?

measures/views.py
```diff
@@ -97,156 +96,187 @@ class MeasureCreateWizard(
...
-                        ],
-                        component_measurement=None,
+                        transaction=measure.transaction,
+                        duty_amount=condition_data["duty_amount"],
+                        condition_code=condition_data["condition_code"],
+                        action=condition_data.get("action"),
+                        required_certificate=condition_data.get("required_certificate"),
                     )
-                    component.save()
+                    condition.save()
```
simonwo, line 237: Should we be calling `condition.clean()` to make sure it's valid? I don't think save calls it by default. Or perhaps @rachaelcodes has already done this in another PR?

simonwo (approved):

measures/forms.py
```diff
@@ -574,15 +593,16 @@ def __init__(self, *args, **kwargs):
                 "duties",
                 HTML.details(
                     "Help with duties",
-                    "Enter the duty that applies to the measure. This is expressed as a percentage (e.g. 4%), a specific duty (e.g. 33 GBP/100kg) or a compound duty (e.g. 3.5% + 11 GBP/LTR).",
+                    "Enter the duty that applies to the measure. This is expressed as a percentage (e.g. 4%), a"
+                    "specific duty (e.g. 33 GBP/100kg) or a compound duty (e.g. 3.5% + 11 GBP/LTR).",
```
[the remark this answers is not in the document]
eadpearce: oh good spot. I forgot I edited this.

measures/tests/test_views.py
eadpearce: yeah good idea. I can add this

measures/views.py
rachaelcodes: @stuaxo and I were working on that last week (as part of https://uktrade.atlassian.net/browse/TP2000-156 ) . As the testing for this fix relied a lot on @eadpearce 's work in this ticket, the plan is to get that fix in after this one has been merged.

measures/views.py
rachaelcodes: It's not in another PR, so would be good to get that fix in here.

eadpearce: does the same thing need to be done with component further down?

PR #493, opened 2022-02-11
TP-2000-130 Measure edit form is broken on POST

gabelton:
## Why
The measures edit form is throwing an AttributeError on POST because MeasureUpdateView.get_form() instantiates the form before the new object has been saved and so it doesn't appear in the with_duty_sentence queryset and therefore doesn't have a duty_sentence attribute

## What
- Removes `raise AttributeError` from MeasureForm and instead only sets the initial duty data, if duty_sentence attribute is present. Adds check that key exists on session in MeasureForm.save(), to avoid KeyError
- Adds a basic MeasureUpdate view test that would've caught this bug
- Also fixes a bug in Measure.diff_components where the original measure transaction was being set on a component instead of creating a new transaction

Original sentry error: https://sentry.ci.uktrade.digital/organizations/dit/issues/56414/?environment=production&project=200&query=is%3Aunresolved

### DIFF
--- a/common/views.py
+++ b/common/views.py
@@ -373,7 +373,8 @@ class TrackedModelChangeView(
 
     def get_result_object(self, form):
         # compares changed data against model fields to prevent unexpected kwarg TypeError
-        # e.g. `geographical_area_group` field on MeasureUpdateForm
+        # e.g. `geographical_area_group` is a field on `MeasureUpdateForm` and included in cleaned data,
+        # but isn't a field on `Measure` and would cause a TypeError on model save()
         model_fields = [f.name for f in self.model._meta.get_fields()]
         form_changed_data = [f for f in form.changed_data if f in model_fields]
         changed_data = {name: form.cleaned_data[name] for name in form_changed_data}
--- a/common/views.py
+++ b/common/views.py
@@ -1,6 +1,7 @@
 """Common views."""
 import time
 from typing import Optional
+from typing import Tuple
 from typing import Type
 
 import django.contrib.auth.views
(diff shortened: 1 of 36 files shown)


measures/forms.py
```diff
@@ -177,9 +176,11 @@ def save(self, commit=True):
 
         sid = instance.sid
 
+        duty_key = f"instance_duty_sentence_{self.instance.sid}"
+        session = self.request.session
         if (
-            self.request.session[f"instance_duty_sentence_{self.instance.sid}"]
-            != self.cleaned_data["duty_sentence"]
+            duty_key in session.keys()
+            and session[duty_key] != self.cleaned_data["duty_sentence"]
```
simonwo, line 183: If the measure form isn't loaded `with_duty_sentence` then when the user submits a modified duty sentence, will this just silently ignore it? 

I think the previous behaviour of throwing an error when the form was loaded incorrectly was right… can we fix the fact that objects being passed to the form don't have a duty sentence attached???

common/views.py
```diff
@@ -372,7 +372,11 @@ def success_url(self):
         return self.object.get_url(self.success_path)
 
     def get_result_object(self, form):
-        changed_data = {name: form.cleaned_data[name] for name in form.changed_data}
+        # compares changed data against model fields to prevent unexpected kwarg TypeError
+        # e.g. `geographical_area_group` field on MeasureUpdateForm
```
paulpepper-trade, line 376: Could the example expand a little on how a TypeError is prevented?

paulpepper-trade (approved):

PR #494, opened 2022-02-15
TP2000-77: Run workbasket transitions in background job

simonwo:
## Why
We have previously had issues transitioning workbaskets through states because the transition would run in a web thread and would time out after 30ish seconds.

## What
This commit moves those transitions to a background task so that they can take as long as necessary and will continue after the lifetime of the request.

We also introduce the archive state defined in ADR 008 that was not implemented at the time.


workbaskets/models.py
```diff
@@ -294,13 +295,24 @@ class WorkBasket(TimestampedMixin):
...
     def __str__(self):
         return f"({self.pk}) [{self.status}]"
 
+    @transition(
+        field=status,
+        source=WorkflowStatus.EDITING,
+        target=WorkflowStatus.ARCHIVED,
+        custom={"label": "Archive"},
+    )
+    def archive(self):
```
gabelton, line 313: All looks really good. I'm just wondering what the scenario(s) for calling archive() would look like? On the ADR it looks like we would be moving workbaskets from editing to archived and archived to editing.

conftest.py
```diff
@@ -829,8 +829,8 @@ def check(
             "exporter.tasks.upload_workbaskets.delay",
         ):
             workbasket.approve(
-                valid_user,
-                get_partition_scheme(),
+                valid_user.pk,
+                settings.TRANSACTION_SCHEMA,
```
gabelton, line 833: Does this work without adding TRANSACTION_SCHEMA to settings/test.py ?

workbaskets/models.py
simonwo: So I think the purpose was that tariff managers could decide that changes they were going to make are no longer relevant and could "hide" the workbasket out of their list of things they're working on. 

It's a bit moot atm whilst we only have 1 editing workbasket, but it does mean that if we end up in the situation again where we have a large workbasket in EDITING (by accident, say) then we can do something with it to get rid of it with a simple UI button press rather than having to hack around with the database.

conftest.py
simonwo: I think it inherits it from settings/common.py?

workbaskets/models.py
simonwo: I figured that to fulfill that last task at the moment only the archive direction would be necessary, and if we add a UI for this we can implement the reverse. But maybe that's shortsighted and we should just have both?

gabelton (approved):

PR #497, opened 2022-02-17
TP2000-143: 500 on Measure Detail view

gabelton:
## Why
It shouldn't be possible to create a measure with the same `sid` as another measure from a different `version_group`. There's still a separate issue being explored in this ticket ( https://uktrade.atlassian.net/browse/TP2000-42 ) around synchronising measure sids so that data engineers can make changes alongside those of Tariff Managers on the frontend, but this should add a bit of extra security in the meantime.

## What
- Adds custom validation in Measure.clean()
- Adds unit test 
- Specifies marksupsafe version in requirements.txt (https://github.com/aws/aws-sam-cli/issues/3661)

### DIFF
--- a/commodities/models/orm.py
+++ b/commodities/models/orm.py
@@ -177,7 +177,6 @@ class GoodsNomenclature(TrackedModel, ValidityMixin, DescribedMixin):
         business_rules.NIG31,
         business_rules.NIG34,
         business_rules.NIG35,
-        UniqueIdentifyingFields,
         UpdateValidity,
     )
 
--- a/commodities/models/orm.py
+++ b/commodities/models/orm.py
@@ -11,6 +11,7 @@ from polymorphic.managers import PolymorphicManager
 from commodities import business_rules
 from commodities import validators
 from commodities.querysets import GoodsNomenclatureIndentQuerySet
+from common.business_rules import UniqueIdentifyingFields
 from common.business_rules import UpdateValidity
 from common.fields import LongDescription
 from common.models import NumericSID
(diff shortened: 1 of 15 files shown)


simonwo (commented):
Is it possible to do this in a general way and apply it anywhere else we have a SID? Also, are there any implications for this being a validation and not a business rule e.g. will it be possible to import duplicate SIDs and fix them later, or will this prevent that?

gabelton:
> 

I think a 500 error would block import, so maybe a business rule is better. Might make generalising it for other components a bit easier too

gabelton:
I added a test for measures just because that was what originally motivated the PR, but I think tests in common/tests/test_business_rules should already have us covered.

measures/models.py
```diff
@@ -279,6 +281,7 @@ class MeasureType(TrackedModel, ValidityMixin):
         business_rules.MT4,
         business_rules.MT7,
         business_rules.MT10,
+        UniqueIdentifyingFields,
```
simonwo, line 284: I think MT1 already does this one.

measures/models.py
```diff
@@ -53,6 +54,7 @@ class MeasureTypeSeries(TrackedModel, ValidityMixin):
     business_rules = (
         business_rules.MTS1,
         business_rules.MTS2,
+        UniqueIdentifyingFields,
```
simonwo, line 57: MTS1 already does this I think.

commodities/models/orm.py
```diff
@@ -176,6 +177,7 @@ def footnote_application_codes(self) -> Set[ApplicationCode]:
         business_rules.NIG31,
         business_rules.NIG34,
         business_rules.NIG35,
+        UniqueIdentifyingFields,
```
simonwo, line 180: Ah I think in this case NIG1 actually has a slightly different requirement.

simonwo (commented):
There are a few other models that already have rules that cover uniqueness… can you double check we're not doubling up?

simonwo (approved):

PR #498, opened 2022-02-18
Only `EDITING` workbaskets should be used for editing

simonwo:
## Why

At the moment, any workbasket in a non-approved state can be selected as the currently active workbasket. This isn't right because users shouldn't be editing baskets once they are `PROPOSED` and shouldn't be stuck editing a basket that is `ARCHIVED`. Essentially, they should only be editing baskets in the `EDITING` state.

## What

This commit makes sure that the current workbasket is in `EDITING` and if not, picks the next available `EDITING` workbasket. If there is none, it creates one as before.

### DIFF
--- a/workbaskets/models.py
+++ b/workbaskets/models.py
@@ -250,6 +250,11 @@ class WorkBasketQueryset(QuerySet):
             approver__isnull=False,
         )
 
+    def editable(self):
+        return self.filter(
+            status=WorkflowStatus.EDITING,
+        )
+
 
 class WorkBasket(TimestampedMixin):
     """
@@ -259,7 +264,7 @@ class WorkBasket(TimestampedMixin):
     See https://uktrade.atlassian.net/wiki/spaces/TARIFFSALPHA/pages/953581609/a.+Workbasket+workflow
     """
 
-    objects = WorkBasketQueryset.as_manager()
+    objects: WorkBasketQueryset = WorkBasketQueryset.as_manager()
 
     title = models.CharField(
         max_length=255,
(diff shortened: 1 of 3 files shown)


workbaskets/views/decorators.py
```diff
@@ -10,7 +10,7 @@ def require_current_workbasket(view_func):
     @wraps(view_func)
     def check_for_current_workbasket(request, *args, **kwargs):
         if WorkBasket.current(request) is None:
-            workbasket = WorkBasket.objects.is_not_approved().last()
+            workbasket = WorkBasket.objects.editable().last()
```
paulpepper-trade, line 13: I'm unsure of the rationale for choosing the last editable workbasket. Might this result in a user editing an unexpected workbasket (perhaps one that has been created outside the scope of the UI)? Or is this preparatory work for multiple workbaskets, or something else?

simonwo: So the assumption right now is that the system only hosts one workbasket for editing, and users put all thier changes into that one basket. So under normal operation there should only ever be one. But using `last()` instead of e.g. `get()` protects us against the situation where for some reason (db restore, bugs) there ends up being more than one (we have encountered this already), and makes the assumption that any present editing workbaskets is a valid place to put changes.

In the future we will have parallel editing of workbaskets, and once that lands I'd expect this code to be replaced with some sort of UI-driven picker.

workbaskets/models.py
```diff
@@ -426,7 +431,7 @@ def current(cls, request):
         if "workbasket" in request.session:
             workbasket = cls.load_from_session(request.session)
 
-            if workbasket.status in WorkflowStatus.approved_statuses():
+            if workbasket.status != WorkflowStatus.EDITING:
```
gabelton, line 434: Should `TrackedModel._can_write` also be updated in this way?

gabelton: I think that scenario you've just described about modifying an archived TrackedModel is a good argument for not updating _can_write

gabelton (approved):

PR #501, opened 2022-02-23
TP2000-166: Ensure ME73 et. al. run against latest model versions. 

simonwo:
This overall purpose of this PR is to ensure that ME73 consdiers the latest versions of each model up to the given transaction, which is particularly important when the object being validated is nethier the container or the contained object. However, along the way we have taken the option of doing some refactoring to replace all of the custom validity checking code with a single business rule that can handle all of the cases. This has fixed at least 3 other bugs in validity checking rules that were unseen.

I would encourage looking at the commits individually to understand each change in isolation.

The summary of the changes are:

- Sort out ordering of factories, so that subfactories are correctly created in transactions before the current one
- Refactor the validity method for computing the valid between, so that it's done on a queryset rather than model
- Add a utility for expanding from one set of models to another by following a relation path
- Refactor all validity business rules

This also adds a new testing suite to this style of test that checks for this case and also ensures we handle deletes properly.


common/business_rules.py
```diff
@@ -405,6 +405,20 @@ def query_contains_validity(self, container, contained, model):
         ):
             raise self.violation(model)
 
+    def get_latest_object(self, related_object):
+        if related_object is None:
+            return None
+
+        try:
+            related_object = related_object.version_at(self.transaction)
```
gabelton, line 413: Could you call this variable 'latest_object' so that it fits the function name more ?

gabelton (approved):

common/business_rules.py
simonwo: This method ended up getting removed :)

common/util.py
```diff
@@ -215,6 +217,39 @@ def validity_range_contains_range(
     return True
 
 
+def resolve_path(model: Type[Model], path: str):
```
kintisheff, line 220: Is there an opportunity to just load all resolved paths upfront when the application starts up? that involves only reading meta data from the models and should be quick and cheap

common/models/tracked_qs.py
```diff
@@ -294,3 +296,45 @@ def with_workbasket(self, workbasket):
         in_workbasket = self.model.objects.filter(transaction__workbasket=workbasket)
         # add latest version of models from the current workbasket
         return self.filter(query) | in_workbasket
+
+    def follow_path(self, path: str) -> TrackedModelQuerySet:
```
kintisheff, line 300: I like that - that's similar to the graph traverser we had in labs, makes a lot of sense!

kintisheff (approved):
The approach to fundamentally fix all related cases makes a lot of sense, minor comments only

common/util.py
simonwo: Yeah… I imagine we will have to contend with cycles appearing, unless we pre-specified all of the paths? But even so we could just lob a `@lru_cache` on this method and it would at least remember all of the results?

common/models/tracked_qs.py
simonwo: I realised as I was writing this that you had probably already done this elsewhere :). And also that it will be useful for TP2000-146!

common/util.py
simonwo: ^ I've done that.

PR #502, opened 2022-02-24
TP2000-145 - Users cannot see measure conditions that use amounts as a requirement

gabelton:
## Why
The measure conditions tab doesn't support the display of conditions that use amounts as a requirement, leading to a confusing display where it looks as if a certificate can both allow and disallow an import at the same time.

## What
- Adds conditions as an itertools.groupby object to the MeasureDetail view context data
- Updates jinja to allow for either amount or certificate as condition
- Also adds indexes to Transaction.Meta and updates WorkBasket.get_current_transaction to return latest approved transaction instead of None.

Original ticket https://uktrade.atlassian.net/browse/TP2000-145


simonwo (commented):
LGTM, but someone else should probably provide the approval!

simonwo:
Also can you rename the PR title to contain TP2000-145 exactly, that way it gets autolinked and also appears in JIRA?

common/models/transactions.py
```diff
@@ -204,6 +204,7 @@ class Transaction(TimestampedMixin):
     class Meta:
         # See TransactionPartition for information on workbasket_status order.
         ordering = ("partition", "order")
+        indexes = (models.Index(fields=("partition", "order")),)
```
kintisheff, line 207: amen!

PR #506, opened 2022-03-04
Run imports inside a single database transaction

simonwo:
## Why
We have recently had issues that have prevented our imports from completing successfully. In these cases we've been left with the database in a inconsistent state and in order to reattempt the import we have to restore the database from a backup.

## What
As discussed, this commit changes the behaviour so that now an entire import is run inside a single database transaction. If there is an error that causes the import to bomb out, the changes are removed, and the issue can be fixed and the import reattempted more easily.

This applies to imports at the "chunk" level. Most import jobs will only generate one chunk, so that is equivalent to a single file. For imports that span across multiple independent chunks, each chunk will have to be manually restarted.

### DIFF
--- a/importer/taric.py
+++ b/importer/taric.py
@@ -267,6 +267,7 @@ class EnvelopeParser(ElementParser):
             nursery.clear_cache()
 
 
+@atomic
 def process_taric_xml_stream(
     taric_stream,
     workbasket_status,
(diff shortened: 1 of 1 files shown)


paulpepper-trade (approved):
Amen.

PR #508, opened 2022-03-08
Fix Exploding Paginator Count Queries using Limited Paginators

kintisheff:
[TP2000-204](https://uktrade.atlassian.net/browse/TP2000-204) Fix Exploding Paginator Count Queries in ListViews

## Why
The Find and Edit Measures page in TaMaTo is notoriously slow, not only affecting UX, but in cases contributing to `504` (gateway timeout) errors when the page takes more than 1 minute to load.
Upon investigation, we identified a single query is responsible for the bulk of this delay (see the attached screenshot).
Upon further investigation, we determined that this was a `SELECT COUNT(*)` query that performs a large number of outer joins, in some cases on all rows in the measures table.
The sole purpose of this query is to help the django `Paginator` to determine the total number of pages for a result set.

## What
The solution we propose is to introduce a configurable `LimitedPaginator`, which returns the true count for small record sets with sizes lower than a configurable `max_count` threshold, but caps the count for larger result sets at this threshold value to avoid exploding count queries.

If the paginator's limit is breached then this is indicated in the UI against the total objects count and the page navigation numbers.


simonwo:
What is the impact of this on the UI? Does the UI select the correct cound and number of pages? Sometimes it is appropriate to want to go to the "last page"… is that still possible?

kintisheff:
> What is the impact of this on the UI? Does the UI select the correct cound and number of pages? Sometimes it is appropriate to want to go to the "last page"… is that still possible?

the impact is simple (Paul has kindly agreed to sort that out, but we can merge if it's a blocker and resolve later):
- just say instead of "200 results" when the cap is 200 and the result set is large, ay something like "at least 200 results, refine your search")

simonwo:
> yes, correct count and number of pages are selected; yes should be still possible to use "last page", we are not changing the behaviour of the paginator in that respect with our solition. the UI impact is small really (Paul has kindly agreed to sort that out, but we can merge if it's a blocker and resolve later):


Makes sense… but how does the paginator know how many pages there are if it doesn't get a full count?

settings/common.py
```diff
@@ -494,3 +494,5 @@
 }
 
 TRANSACTION_SCHEMA = os.getenv("TRANSACTION_SCHEMA", "workbaskets.models.SEED_FIRST")
```
stuaxo, line 496: Chuck in a comment here to explain the purpose of this - I've had very bad luck at adding things in that fix issues like this and future developers stripping them out.

stuaxo:
I would add a test that checks the performance hasn't regressed on that view.

I had a run of 2/2 projects where we did careful work to optimise something for performance, for future devs to strip that work out + decided that in future I would make a test that would complain that they would have to remove as well.

Though I guess more broadly having a set of performance tests for the UI could be a whole thing in itself.

simonwo:
> I would add a test that checks the performance hasn't regressed on that view.

@stuaxo Did you have any success with making performance tests that are "relative" rather than "absolute"? Because it would be nice if Github Actions and/or my computer is just slow, that they won't fail a test that is like "make sure this takes less than a second"?

stuaxo:
I'm not sure if this is a thing that's possible  -

To calculate pagination, we need to know "how many things are there", but we don't need all the same fields at all.

So we have some complicated query that grabs all the data needed for display, to show the view - but if you are working out pages you don't need all that, just the ids.

Maybe the paginator already does something like this though ?

stuaxo:
> > I would add a test that checks the performance hasn't regressed on that view.
> 
> @stuaxo Did you have any success with making performance tests that are "relative" rather than "absolute"? Because it would be nice if Github Actions and/or my computer is just slow, that they won't fail a test that is like "make sure this takes less than a second"?

Sometimes testing query count can be a good proxy, (we may have already done this somewhere) - it's relatively straightforward to make a decorator for this, the only tricky bit is to ensure DEBUG is enabled in it otherwise the counter doesn't work.

Benchmarking performance and having a dashboard is probably the best way to avoid regressions (even a text with some blerb text is invisible enough it may disappear) 

Firefox's "arewefastyet" dashboard
https://arewefastyet.com/win10/benchmarks/overview?numDays=60

Pypy - Speed.pypy.org
https://speed.pypy.org/

I don't know the state of general purpose dashboards for measuring performance (these are both custom).

paulpepper-trade:
> I'm not sure if this is a thing that's possible -
> 
> To calculate pagination, we need to know "how many things are there", but we don't need all the same fields at all.
> 
> So we have some complicated query that grabs all the data needed for display, to show the view - but if you are working out pages you don't need all that, just the ids.
> 
> Maybe the paginator already does something like this though ?

I'm unsure whether or not the main bottlenecks are data transfer related. There are ten joins (a combination of left joins and inner joins) on the select that is used to calculate the measures count. Perhaps the query structure could be optimised, but the filtering likely needs to reference related entities (because that's what the filtering is about). It's probably worth us scrutinising the query plan to identify where time is actually spent.

settings/common.py
paulpepper-trade: Done.

paulpepper-trade:
> I'm not sure if this is a thing that's possible -
> 
> To calculate pagination, we need to know "how many things are there", but we don't need all the same fields at all.
> 
> So we have some complicated query that grabs all the data needed for display, to show the view - but if you are working out pages you don't need all that, just the ids.
> 
> Maybe the paginator already does something like this though ?

I don't think we need performance tests to go with this PR's changes. Performance testing our queries would seem useful, as pointed out, but right now these changes avoid the poorly performing count query altogether (using a threshold value set by `MEASURES_PAGINATOR_MAX_COUNT`) to void application stalls and redeployments.

paulpepper-trade:
> > What is the impact of this on the UI? Does the UI select the correct cound and number of pages? Sometimes it is appropriate to want to go to the "last page"… is that still possible?
> 
> yes, correct count and number of pages are selected; yes should be still possible to use "last page", we are not changing the behaviour of the paginator in that respect with our solition. the UI impact is small really (Paul has kindly agreed to sort that out, but we can merge if it's a blocker and resolve later):
> 
>     * just say instead of "200 results" when the cap is 200 and the result set is large, ay something like "at least 200 results, refine your search")

Latest changes indicate to the user when there are more than the max number of objects. Similarly with page numbering. For screenshots of UI changes used see this comment on the associated Jira task: https://uktrade.atlassian.net/browse/TP2000-204?focusedCommentId=91091.

stuaxo (approved):

PR #509, opened 2022-03-09
TP2000-51: Allow users to add a duty per commodity code

eadpearce:


simonwo (approved):
Nice!

Is there any mileage in a test that the right duties end up on the right commodities?

eadpearce:
@simonwo Yeah I can add that. I think Stuart suggested a similar thing about testing the correct commodities are created too

stuaxo:
> @simonwo Yeah I can add that. I think Stuart suggested a similar thing about testing the correct commodities are created too

I often got told off for by the person that taught me about testing for not testing the right thing, in particular just testing len(thing) or count(thing) instead of more detailed checks.

stuaxo (approved):

PR #510, opened 2022-03-09
TP2000-205 Add new fields to MeasureConditionCode and MeasureAction

gabelton:
## Why
In order to support the ui changes in [this ticket](https://uktrade.atlassian.net/browse/TP2000-199) we need to add some flags to the MeasureConditionCode and MeasureAction models and business rules to prevent the creation of conditions with incompatible combinations of code with certificate or reference price.

## What
- Adds `accepts_certificate` and `accepts_price` BooleanField to MeasureConditionCode
- Adds `requires_duty` field to MeasureAction
- Adds two business rules on MeasureCondition for action and condition code
- Adds tests
- Adds custom migration to set values for new fields on the basis of a list of different codes
- Adds a tiny ordering tweak to readme - I got an error running `docker-compose build` before copying sample env to my .env when I was setting up the project again recently.


measures/business_rules.py
```diff
@@ -970,6 +977,36 @@ def validate(self, component):
...
+        if self.duty_amount and not code.accepts_price:
+            raise self.violation(message=message + "a price")
+
+
+class ActionCodeRequired(BusinessRule):
+    """ 
+    If a condition's action code requires a duty, then an associated condition component must be created with a duty amount.
+    """
+    def validate(self):
+        components = self.components.current()
```
simonwo, line 1005: This is correct, but at the moment business rules don't have `set_current_transaction(self.transaction)` called for them, so this will most likely not return the things we expect.

The future-proof thing to do is probably modify `BusinessRuleChecker` to use `with override_current_transaction(…)`?

measures/business_rules.py
```diff
@@ -89,6 +89,13 @@ class MC4(PreventDeleteIfInUse):
     condition component."""
 
 
+class AcceptedConditionTriggers(BusinessRule):
+    """A measure condition code must be created with one or both of accepts_certificate and accepts_price set to True"""
+    def validate(self):
+        if not (self.accepts_certificate or self.accepts_price):
+            raise self.violation(message="Condition codes must be created with at least one of accepts_certificate or accepts_price set to True")
```
simonwo, line 96: You could potentially also do this as a constraint on the model, that would be enforced by the DB… a la https://github.com/uktrade/tamato/blob/master/quotas/models.py#L234. I don't think causes a problem for migrations as the check is only enforced on write and does not need to be generally true when it is created.

simonwo (commented):
Looks tremendous so far!

measures/business_rules.py
rachaelcodes: :+1: @simonwo  What would you say is the guiding principle between what is enforced as a model constraint and what gets checked as a business rule?

measures/migrations/0010_auto_20220309_1725.py
```diff
@@ -0,0 +1,63 @@
+# Generated by Django 3.1.14 on 2022-03-09 17:25
+
+from django.db import migrations, models
+
+CERTIFICATE_ONLY_CODES = ["A", "B", "C", "H", "Q", "Y", "Z"]
+PRICE_ONLY_CODES = ["F", "L", "M", "R", "U", "V"]
+CERTIFICATE_AND_PRICE_CODES = ["E", "I"]
```
simonwo, line 7: I wrote this code to look at existing conditions that use the codes:

```python
for x in MeasureConditionCode.objects.latest_approved():
    print(
        x.code,
        x.description,
        MeasureCondition.objects.filter(condition_code=x, required_certificate__isnull=False).exists(),
        MeasureCondition.objects.filter(condition_code=x, duty_amount__isnull=False).exists(),
     )
```

And got these results:

| Code | Description | Accepts certificate | Accepts amount |
|---|---|---|---|
| A | Presentation of an anti-dumping/countervailing document | True | False |
| B | Presentation of a certificate/licence/document | True | False |
| C | Presentation of a certificate/licence/document | True | False |
| D | Intended for processing | False | False |
| E | The quantity or the price per unit declared, as appropriate, is equal or less than the specified maximum, or presentation of the required document | True | True |
| F | The net free at frontier price before duty must be equal to or greater than the minimum price (see components) | False | True |
| G | The CIF price plus the duty to be paid/ton must be equal to or greater than the minimum price (see components) | False | True |
| H | Presentation of a certificate/licence/document | True | False |
| I | The quantity or the price per unit declared, as appropriate, is equal or less than the specified maximum, or presentation of the required document | True | True |
| K | Also applicable simultaneously with tariff quota shown in the field "certificates" | True | False |
| L | CIF price must be higher than the minimum price (see components) | False | True |
| M | Declared price must be equal to or greater than the minimum price/reference price (see components) | False | True |
| N | The CIF price before duty must be equal to or greater than the minimum price (see components) | False | True |
| P | Only particular ingredients are eligible for export refund | True | False |
| Q | Presentation of an endorsed certificate/licence | True | False |
| R | Ratio "net weight/supplementary unit" is equal to or higher than the condition amount | False | True |
| S | Lodgement of a security | False | True |
| U | Ratio "declared value/supplementary unit" should be higher than the condition amount | False | True |
| V | Import price must be equal to or greater than the entry price (see components) | False | True |
| W | Washington Convention | False | False |
| Y | Other conditions | True | False |
| Z | Presentation of more than one certificate | True | False |

So it looks like we could add K and P to `CERTIFICATE_ONLY_CODES` and G and S to `PRICE_ONLY_CODES`. There also appear to be a few False, False codes which I don't really understand but I guess we can ignore them.

measures/migrations/0010_auto_20220309_1725.py
```diff
@@ -0,0 +1,63 @@
...
+
+REQUIRES_DUTY_CODES = ["01", "02", "03", "11", "12", "13", "15"]
```
simonwo, line 9: I ran this code:

```python
for x in MeasureAction.objects.latest_approved():
    print(
        x.code,
        x.description,
        MeasureCondition.objects.filter(action=x, components__isnull=False).exists(),
        sep="\t",
    )
```

And got these results:

| Code | Description | Requires duty |
|---|---|---|
| 01 | Apply the amount of the action (see components) True | True |
| 02 | Apply the difference between the amount of the action (see components) and the price at import | True |
| 03 | Apply the difference between the amount of the action (see components) and CIF price | False |
| 04 | The entry into free circulation is not allowed | False |
| 05 | Export is not allowed | False |
| 06 | Import is not allowed | False |
| 07 | Measure not applicable | False |
| 08 | Declared subheading not allowed False | False |
| 09 | Import/export not allowed after control False | False |
| 10 | Declaration to be corrected  | False |
| 11 | Apply the difference between the amount of the action (see components) and the free at frontier price before duty | True |
| 12 | Apply the difference between the amount of the action (see components) and the CIF price before duty | True |
| 13 | Apply the difference between the amount of the action (see components) and the CIF price augmented with the duty to be paid per tonne | True |
| 14 | The exemption/reduction of the anti-dumping duty is not applicable | False |
| 15 | Apply the difference between the amount of the action (see components) and the price augmented with the countervailing duty (3,8%) | True |
| 16 | Export refund not applicable | False |
| 24 | Entry into free circulation allowed | False |
| 25 | Export allowed | False |
| 26 | Import allowed | False |
| 27 | Apply the mentioned duty | False |
| 28 | Declared subheading allowed | False |
| 29 | Import/export allowed after control | False |
| 30 | Suspicious case | False |
| 34 | Apply exemption/reduction of the anti-dumping duty | False |
| 36 | Apply export refund | True |

So it looks like we should add 36 to this list. I know that 13 is showing as `False` but I think that's because it doesn't have any conditions using it. So I think we should leave it as requiring a duty.

measures/models.py
```diff
@@ -350,8 +350,8 @@ class MeasureConditionCode(TrackedModel, ValidityMixin):
         db_index=True,
     )
     description = ShortDescription()
-    accepts_certificate = models.BooleanField()
-    accepts_price = models.BooleanField()
+    accepts_certificate = models.BooleanField(null=True)
+    accepts_price = models.BooleanField(null=True)
```
simonwo, line 354: Should this and the below have `default=False` instead? Which stops us having a tri-state boolean?

measures/migrations/0010_auto_20220309_1725.py
gabelton: 03 is also showing as False. Is that the same deal as 13 do you know ?

measures/business_rules.py
stuaxo: @simonwo is it worth considering a custom linter to check stuff like this ?

gabelton: I think this is just my new machine not being setup with pre-commit hooks

measures/business_rules.py
simonwo: Aha, good question indeed… I think that anything that is really a policy decision and can possibly change in the future shouldn't be implemented as a DB constraint. I said a bit more about it here: https://www.register-dynamics.co.uk/blog/james-bond-data-modelling

simonwo: …But given what we found below, it looks like condition code W doesn't necessarily fit this pattern. It seems that measures with condition code W seem to only have 1 condition, which is the negative case? So I'm not sure what they are really meant to be doing. 

So if we did want to enforce this as a constraint, perhaps we'd have to set W to have True, True? As we have nothing more specific to go on.

measures/migrations/0010_auto_20220309_1725.py
simonwo: Oops, I meant 03 in the above comment. Code 03 has zero conditions, yes, so that's why.

measures/business_rules.py
rachaelcodes: Do we want to enforce this as a constraint then? And if we do, it feels more like a policy constraint rather than a permanent one.

measures/business_rules.py
```diff
@@ -970,6 +971,51 @@ def validate(self, component):
...
+
+
+class ActionRequiresDuty(BusinessRule):
+    """If a condition's action code requires a duty, then an associated
+    condition component must be created with a duty amount."""
+
+    def validate(self, condition):
+        # components = self.components.current()
+        current = get_current_transaction()
+        components = condition.components.approved_up_to_transaction(current)
```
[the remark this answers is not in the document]
simonwo: Or if you don't want to think about that now, just replace `current` with `self.transaction`.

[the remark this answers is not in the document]
gabelton: I was getting confused because I thought the condition would be created before the component was and that I'd need to use a later transaction to get the correct components, but then I guess the components will actually be created with the same transaction as the condition, so it should be alright to use self.transaction

simonwo (approved):

measures/migrations/0010_add_requires_to_action_and_accepts_to_condition_code.py
```diff
@@ -0,0 +1,97 @@
...
+    for code in codes:
+        code.accepts_certificate = False
+        code.accepts_price = False
+        code.save()
+
+    actions = MeasureAction.objects.all()
+    for action in actions:
+        action.requires_duty = False
+        action.save()
+
```
stuaxo, line 57: I guess there isn't anywhere to get `accepts_certifcate`, `accepts_price` and `requires_duty` from when doing the reverse migration ?

stuaxo (approved):

PR #511, opened 2022-03-14
Implement ME27 for regulation replacements

simonwo:
We never implemented this rule because we thought we weren't using replacements in the UK. We're still not, but if a user  accidentally?) selects an EU regulation it's possible for the rule to be broken. Selecting EU regulations is still necessary because the UK has "retained" some EU legislation and it is still valid to choose it.

### DIFF
--- a/measures/business_rules.py
+++ b/measures/business_rules.py
@@ -16,6 +16,7 @@ from common.business_rules import UniqueIdentifyingFields
 from common.business_rules import ValidityPeriodContained
 from common.business_rules import only_applicable_after
 from common.business_rules import skip_when_deleted
+from common.models.utils import override_current_transaction
 from common.util import TaricDateRange
 from common.util import validity_range_contains_range
 from common.validators import ApplicabilityCode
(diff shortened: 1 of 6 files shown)


measures/business_rules.py
```diff
@@ -541,6 +542,30 @@ class ME24(MustExist):
...
+    # This method only checks that a single Replacement does this whereas it
+    # might be possible for multiple Replacements to cover the full validity
+    # period. However, the very few Regulations that have >1 Replacement have
+    # been manually checked and don't require this extra complexity, and we
+    # don't use Replacements in the UK so it won't be possible to create them.
+
+    def validate(self, measure):
+        with override_current_transaction(self.transaction):
+            measures = type(measure).objects.filter(pk=measure.pk)
+            regulation_validity = measures.follow_path("generating_regulation").get()
```
gabelton, line 560: This is tangential and not a blocker, but I was working backwards and reminding myself how follow_path works. Do you think you could add the expected return type on the resolve_path function in common/util.py?

gabelton (approved):

PR #513, opened 2022-03-16
TP2000-199 create measures journey doesn't support the correct fields for measure conditions

gabelton:
TP2000-199 create measures journey doesn't support the correct fields for measure conditions

## Why
The current logic for creating measure condition components doesn't work as it should. At the mo we're using the condition duty_amount field to create the component. We should be creating these components with their own separate duty amount (applicable duty) and only creating these components when that value is provided.

## What
- Updates DutySentenceParser to handle creation of MeasureConditionComponent as well as MeasureComponent
- Updates MeasureConditionsForm and ui to add applicable_duty field beneath the action code 
- Also fixes a bug with TrackedModel.get_url when called with "create" where kwargs are passed when they shouldn't be.
- Adds various parser, view, form, and model tests

[This ticket ](https://uktrade.atlassian.net/browse/TP2000-214)to add extra conditional field display logic to the UI depends on this PR


measures/parsers.py
```diff
@@ -131,6 +132,7 @@ def __init__(
         duty_expressions: Iterable[DutyExpression],
         monetary_units: Iterable[MonetaryUnit],
         permitted_measurements: Iterable[Measurement],
+        component_output: Optional[TrackedModel] = MeasureComponent,
     ):
```
rachaelcodes, line 136: Should we add a comment to the docstring to state that this is relevant to MeasureComponentConditions as well?

rachaelcodes (approved):

measures/views.py
```diff
@@ -233,22 +234,17 @@ def create_measures(self, data):
...
-                            duty_amount=condition.duty_amount,
-                            monetary_unit=measure_creation_pattern.condition_sentence_parser.monetary_units[
-                                "GBP"
-                            ],
-                            component_measurement=None,
+                    if condition_data.get("applicable_duty"):
+                        parser = DutySentenceParser.get(
+                            measure.valid_between.lower,
+                            component_output=MeasureConditionComponent,
                         )
```
simonwo, line 241: This will get done once per condition – may be worth pulling it out of the loop?

measures/views.py
```diff
@@ -233,22 +234,17 @@ def create_measures(self, data):
...
-                        component.clean()
-                        component.save()
+                        components = parser.parse(condition_data["applicable_duty"])
+                        for c in components:
+                            c.condition = condition
+                            c.transaction = condition.transaction
+                            c.update_type = UpdateType.CREATE
+                            c.save()
```
simonwo, line 247: Should we have a `c.clean()` here?

common/jinja2/components/measure_condition_component_duty/template.jinja
```diff
@@ -0,0 +1,22 @@
...
+            </summary>
+            <div class="govuk-details__text">
+                You can use one of the following:
+                <ul>
+                    <li>a percentage, such as 4%</li>
+                    <li>a specific duty, such as 1.23 GBP/kg</li>
+                    <li>a compound duty, such as 3.5% + 11 GBP/LTR</li>
+                </ul>
+            </div>
+        </details>
```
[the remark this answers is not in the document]
simonwo: I think given it contains html, you may as well use a jinja include.

measures/views.py
gabelton: I guess I should add this to the diff_components method too ?

gabelton: I'm not sure I can call `full_clean()` anyhow because until save is called a missing field validation error is called because `version_group` hasn't been saved yet

simonwo (approved):

PR #514, opened 2022-03-18
TP2000-217 duplicate footnotes displaying in measure edit form

gabelton:
## Why
Currently the MeasureUpdate view and MeasureForm both use the reverse relationship when getting footnotes, but really they should be using the FootnoteAssociationMeasure model. Otherwise we get duplicated footnotes and footnotes associated with a deleted FootnoteAssociationMeasure displaying on the edit form.

## What
- Updates view and form to filter using FootnoteAssociationMeasure
- Adds test to check this

[Original teams chat](https://teams.microsoft.com/l/message/19:MU-0VlKdDiGxdC3Iv8oKQbAlshz1N3ZhH3f6qrOoW7g1@thread.tacv2/1647511203316?tenantId=8fa217ec-33aa-46fb-ad96-dfe68006bb86&groupId=16ae1baa-31cc-4fda-8ea7-cfcd00d254ed&parentMessageId=1647511203316&teamName=Tariff%20Team&channelName=General&createdTime=1647511203316)

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -128,8 +128,16 @@ class MeasureForm(ValidityPeriodForm):
         # If no footnote keys are stored in the session for a measure,
         # store all the pks of a measure's footnotes on the session, using the measure sid as key
         if f"instance_footnotes_{self.instance.sid}" not in self.request.session.keys():
+            tx = WorkBasket.get_current_transaction(self.request)
+            associations = (
+                models.FootnoteAssociationMeasure.objects.approved_up_to_transaction(
+                    tx,
+                ).filter(
+                    footnoted_measure=self.instance,
+                )
+            )
             self.request.session[f"instance_footnotes_{self.instance.sid}"] = [
-                footnote.pk for footnote in self.instance.footnotes.all()
+                a.associated_footnote.pk for a in associations
             ]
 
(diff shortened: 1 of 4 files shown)


simonwo (approved):
[As discussed](https://teams.microsoft.com/l/message/19:80f6ef27ef454a87a8c4844d3aa5b3f7@thread.tacv2/1647594383345?tenantId=8fa217ec-33aa-46fb-ad96-dfe68006bb86&groupId=7dec05eb-83aa-4ca6-8bff-bf9b86f579e7&parentMessageId=1647594383345&teamName=TAP&channelName=Dev%20and%20Data&createdTime=1647594383345), I think this is worthy of more discussion about the right way to fix. I think what we want is where possible for Django features to work as intended with the versioning being a transparent thing that you don't have to think about most of the time… which means doing `measure.footnotes` really should return the right footnotes. But to fix that we'll have to start making use of `current_objects` in reverse relations, to take account of the fact that footnote associations might be deleted and their footnote therefore shouldn't be returned… which maybe sounds OK? But also makes it hard to find footnotes from previous versions… not sure if that is a concern.

@gabelton has created [a ticket for the long term fix](https://uktrade.atlassian.net/browse/TP2000-218).

PR #515, opened 2022-03-18
TP2000-156 - Component sequence number start at 1, fixing the importer.

stuaxo:
Component sequence numbers start at 1 not 0, fixing the importer.
Make `test_measure_form_wizard_create_measures` a little more batch oriented making it easier to test this, should save some queries too.

### DIFF
--- a/measures/tests/conftest.py
+++ b/measures/tests/conftest.py
@@ -362,6 +362,11 @@ def measure_type():
 
 
 @pytest.fixture()
+def erga_omnes():
+    return factories.GeoGroupFactory(area_id=1011)
+
+
+@pytest.fixture()
 def regulation():
     return factories.RegulationFactory.create()
 
--- a/measures/tests/conftest.py
+++ b/measures/tests/conftest.py
@@ -362,11 +362,6 @@ def measure_type():
 
 
 @pytest.fixture()
-def erga_omnes():
-    return factories.GeoGroupFactory(area_id=1011)
-
-
-@pytest.fixture()
 def regulation():
     return factories.RegulationFactory.create()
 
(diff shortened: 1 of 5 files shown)


measures/tests/test_views.py
```diff
@@ -413,64 +414,60 @@ def test_measure_form_wizard_create_measures(
...
+        measures.values_list(
+            "pk", "conditions__component_sequence_number", "conditions__condition_code"
         )
-    )
+    ) == [
+        (measure_data[0].pk, 1, condition_code1.pk),
+        (measure_data[0].pk, 2, condition_code2.pk),
+        (measure_data[1].pk, 1, condition_code1.pk),
+        (measure_data[1].pk, 2, condition_code2.pk),
+    ]
```
simonwo, line 457: Using lists here is asserting an ordering, whereas we are not asking for any specific ordering in the query, which looks brittle. Would a set be a better choice?

measures/tests/conftest.py
```diff
@@ -361,6 +361,11 @@ def measure_type():
     )
 
 
+@pytest.fixture()
+def erga_omnes():
+    return factories.GeoGroupFactory(area_id=1011)
+
+
```
simonwo, line 368: Where is this used?

stuaxo: Good catch I believe this came from a previous version of this where we were building a different test for this work, will remove, I'm can imagine this fixture coming back in the future though :)

stuaxo: Removed.

simonwo (approved):

gabelton (approved):

PR #518, opened 2022-03-22
TP2000-128: Speed up commodity code autocomplete

simonwo:
There were three main issues that made the commodity code autocomplete slow:

1. Searching on aggregated strings is slow, and we can speed this up by only searching on item ID if the search input looks like an item ID.
2. Getting the current transaction (which happens on every request) prefetches all of the tracked models, which is expensive. This basically isn't necessary at all.
3. Not having minimum lengths specified means that lots of requests get sent for the first few characters in a search, which causes a backlog as Django runs through them (aborting the req. doesn't abort the DB query, it seems). So now we have a minimum length of search which has been picked based on the individual model.

### DIFF
--- a/common/models/tracked_qs.py
+++ b/common/models/tracked_qs.py
@@ -161,6 +161,20 @@ class TrackedModelQuerySet(
             ),
         )
 
+    def record_ordering(self) -> TrackedModelQuerySet:
+        """
+        Returns objects in order of their TARIC record code and subrecord code.
+
+        This is primarily useful for querysets that contain multiple types of
+        tracked model, e.g. when exporting the tracked models to XML.
+        """
+        return self.annotate_record_codes().order_by(
+            "transaction__partition",
+            "transaction__order",
+            "record_code",
+            "subrecord_code",
+        )
+
     def version_ordering(self) -> TrackedModelQuerySet:
         """
         Returns objects in canonical order, i.e. by the order in which they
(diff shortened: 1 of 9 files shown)


common/tests/test_transactions.py
```diff
@@ -23,7 +23,7 @@ def test_unordered_tracked_models_are_linked_to_queryset():
     )
 
     qs = Transaction.objects.filter(id=transaction_this.id)
-    unordered_tracked_models = qs.unordered_tracked_models()
+    unordered_tracked_models = qs.tracked_models.record_ordering()
```
gabelton, line 26: Should this variable name be updated?

gabelton (approved):
Looks great. Just so that I'm clear, on point 2 is it the TransactionManager.get_queryset() and the prefetch in there that is causing slowness on autocomplete?

stuaxo:
Can we get some numbers on the improvement - even if it's slightly unscientific "on my computer" or even wallclock type measurements ?

common/models/tracked_qs.py
```diff
@@ -161,6 +161,20 @@ def annotate_record_codes(self) -> TrackedModelQuerySet:
...
+        This is primarily useful for querysets that contain multiple types of
+        tracked model, e.g. when exporting the tracked models to XML.
+        """
+        return self.annotate_record_codes().order_by(
+            "transaction__partition",
+            "transaction__order",
+            "record_code",
+            "subrecord_code",
+        )
+
```
stuaxo, line 177: :+1:

simonwo:
On my machine:

* The original query took several seconds. Only looking up by item ID chops this to 400ms.
* The prefetch of the tracked models invoked a number of queries which added up to a few seconds.
* The requests don’t abort the database query when the user keeps typing, so the original query has to complete before another request is processed. 

Overall on my machine it now takes about a second to return a result. More could probably be done by trying to combine the description query into the same thing with a CTE.

PR #520, opened 2022-03-23
TP2000-190: Commodities importer

eadpearce:
# TP2000-190: Commodities importer

## Why
- This enables non-admin users to upload commodity changes in bulk via XML file

## What
- This reuses (and slightly modifies) the preexisting importer code and hooks it up to a new simpler UI
- Introduces the python-magic dependency to enable checking of MIME types in the file upload


commodities/forms.py
```diff
@@ -0,0 +1,99 @@
...
+        except lxml.etree.XMLSyntaxError:
+            raise ValidationError(generic_error_message)
+
+        with open(settings.PATH_XSD_TARIC) as xsd_file:
+            xmlschema = lxml.etree.XMLSchema(file=xsd_file)
+
+        try:
+            xmlschema.assertValid(xml_file)
+        except lxml.etree.DocumentInvalid:
+            raise ValidationError(generic_error_message)
```
simonwo, line 68: If the file fails schema validation, do we expect uploading it again to make a difference? Are we assuming it was just corrupted in transit? Do we need to provide a longer error message that it might not be a valid file??

commodities/forms.py
```diff
@@ -0,0 +1,99 @@
...
+
+        chunk_taric(self.cleaned_data["taric_file"], batch, record_group=record_group)
+        run_batch(
+            batch=batch.name,
+            status=WorkflowStatus.EDITING,
+            partition_scheme_setting=settings.TRANSACTION_SCHEMA,
+            username=user.username,
+            workbasket_id=workbasket_id,
+            record_group=record_group,
+        )
```
simonwo, line 93: These are potentially long running operations and could take many minutes for a large file. Do these need to run in a background task?

commodities/tests/test_files/valid.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
simonwo, line 1: This file doesn't contain any commodities… does it make sense to test with one that does? To check that a commodity actually appears in the DB?

commodities/views.py
```diff
@@ -31,3 +37,22 @@ def get_queryset(self):
...
+
+@method_decorator(require_current_workbasket, name="dispatch")
+class CommodityImportView(FormView):
+    template_name = "commodities/import.jinja"
+    form_class = CommodityImportForm
+    success_url = reverse_lazy("commodities-import-success")
+
+    @property
+    def workbasket(self) -> WorkBasket:
+        return WorkBasket.current(self.request)
```
simonwo, line 50: You should be able to use [`WithCurrentWorkBasket`](https://github.com/uktrade/tamato/blob/master/workbaskets/views/mixins.py#L4) for this?

simonwo (commented):
Looks pretty great!!

commodities/forms.py
```diff
@@ -0,0 +1,99 @@
...
+
+        with open(settings.PATH_XSD_TARIC) as xsd_file:
+            xmlschema = lxml.etree.XMLSchema(file=xsd_file)
+
+        try:
+            xmlschema.assertValid(xml_file)
+        except lxml.etree.DocumentInvalid:
+            raise ValidationError(generic_error_message)
+
+        # https://code.djangoproject.com/ticket/7812
```
gabelton, line 70: Is it worth maybe quoting the first line of the issue description maybe?

requirements.txt
```diff
@@ -46,6 +46,7 @@ pytest-cov==2.12.1
 pytest-django==4.1.0
 pytest-responses==0.5.0
 pytest-xdist==2.3.0
+python-magic==0.4.25
```
gabelton, line 49: Worth just giving a heads up in PR summary, that new dependencies need to be installed

importer/tests/test_tasks.py
```diff
@@ -18,6 +18,7 @@ def test_import_chunk(
 ):
     tasks.import_chunk(
         chunk.pk,
+        None,
```
gabelton, line 21: Is None the default value ?

commodities/forms.py
eadpearce: `run_batch` here eventually calls `import_chunk.delay()` so yeah it's running in a background task

importer/tests/test_tasks.py
eadpearce: It should be. I had some trouble with the ordering of args/kwargs in all the importer functions which I might be able to go back and fix so it has a default of None in every function

commodities/views.py
eadpearce: if I use this do I still need to use the `require_current_workbasket` decorator?

commodities/forms.py
eadpearce: I've tried to stick to the error messages suggested by GDS here but this one in particular is not very helpful. Is there a way of showing this to the user but then logging something in the console so that when a dev goes to reproduce it they can easily see what happened? Or have it show up in sentry?

commodities/views.py
simonwo: Yes I think you do. It only really helps with this property.

commodities/forms.py
simonwo: Yes you can do: https://github.com/uktrade/tamato/blob/master/importer/taric.py#L177

eadpearce: Awesome! I will add that

simonwo (approved):

PR #521, opened 2022-03-24
TP2000-206: Automatically select terminating regulations

simonwo:
## Why
At the moment it's not possible for users to add or remove end dates to measures because they aren't able to make a corresponding edit to the terminating regulation. This behaviour is also handled in a number of bespoke places in the code base.

## What
This commit unifies that logic into a signal that is fired pre-save, so no matter where the edit to the measure is happening the logic will be applied. This also lets us remove some redundant code.

We also take the changed commodity code handling logic out into a signal.

The ``mute_signals`` decorator is necessary for ME33 and ME34 so that it _is_ possible to create test data that ignore this logic.

### DIFF
--- a/measures/signals.py
+++ b/measures/signals.py
@@ -1,0 +1,54 @@
+from typing import Type
+
+from django.db.models.signals import pre_save
+from django.dispatch import receiver
+
+from common.validators import UpdateType
+from measures import models
+
+
+@receiver(pre_save, sender=models.Measure, dispatch_uid="update_terminating_regulation")
+def update_terminating_regulation(sender: Type, **kwargs):
+    """
+    Update the terminating regulation on the measure to not be present if the
+    measure has no explicit end date or to be equal to the generating regulation
+    if the measure has an explicit end date and the termination regulation is
+    not already set.
+
+    This automatically avoids issues with ~`measures.business_rules.ME33` and
+    ~`measures.business_rules.ME34`.
+    """
+    instance: models.Measure = kwargs["instance"]
+    should_have_reg = not instance.valid_between.upper_inf
+    if should_have_reg and instance.terminating_regulation is None:
+        instance.terminating_regulation = instance.generating_regulation
+    elif not should_have_reg:
+        instance.terminating_regulation = None
+
+
+@receiver(pre_save, sender=models.Measure, dispatch_uid="handle_changed_commodity_code")
+def handle_changed_commodity_code(sender: Type, **kwargs):
+    """If the commodity code is being changed on an existing measure, the
+    measure is deleted instead of doing an `UPDATE` and a new measure created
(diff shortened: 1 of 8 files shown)


measures/signals.py
```diff
@@ -0,0 +1,43 @@
...
+
+from django.db.models.signals import pre_save
+from django.dispatch import receiver
+
+from common.validators import UpdateType
+from measures import models
+
+
+@receiver(pre_save, sender=models.Measure, dispatch_uid="update_terminating_regulation")
+def update_terminating_regulation(sender: Type, **kwargs):
```
stuaxo, line 11: Not a blocker - any chance of a docstring here ?

stuaxo:
:+1:

stuaxo (approved):

measures/signals.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@receiver(pre_save, sender=models.Measure, dispatch_uid="handle_changed_commodity_code")
+def handle_changed_commodity_code(sender: Type, **kwargs):
+    """If the commodity code is being changed on an existing measure, the
+    measure is deleted instead of doing an `UPDATE` and a new measure created
+    with the updated commodity code."""
+    instance: models.Measure = kwargs["instance"]
+    is_update = instance.update_type == UpdateType.UPDATE
+    previous: models.Measure = instance.get_versions().version_ordering().last()
```
gabelton, line 36: Should this query exclude the instance ?

measures/signals.py
```diff
@@ -0,0 +1,54 @@
...
+    instance: models.Measure = kwargs["instance"]
+    should_have_reg = not instance.valid_between.upper_inf
+    if should_have_reg and instance.terminating_regulation is None:
+        instance.terminating_regulation = instance.generating_regulation
+    elif not should_have_reg:
+        instance.terminating_regulation = None
+
+
+@receiver(pre_save, sender=models.Measure, dispatch_uid="handle_changed_commodity_code")
+def handle_changed_commodity_code(sender: Type, **kwargs):
```
gabelton, line 30: Is there a test for this logic somewhere?

measures/signals.py
simonwo: This is the same code that used to exist before as part of `Measure.save`… and it seems to work? I think because we don't save instances twice, `instance` should be unsaved?

measures/signals.py
simonwo: Yes, the tests are the same ones that existed before, i.e.: https://github.com/uktrade/tamato/blob/master/measures/tests/test_models.py#L216

simonwo: I thought it best to not move them to ensure I hadn't broken anything in changing the code to be a signal?

gabelton (approved):

PR #523, opened 2022-03-24
Optionally omit the 'name' attr of dummy autocomplete fields.

paulpepper-trade:
# TP2000-229 Don't submit '_autocomplete' fields

## Why
Autocomplete fields are only used in the client and have no purpose on the server, but current behaviour submits their contents to the server (as URL encoded params on many of Tamato's forms). This can cause undesirable side-effects.

## What

 [the auto](https://github.com/alphagov/accessible-autocomplete) allows setting the name attribute of the autocomplete field. This PR allows setting it to an empty value and thereby avoids submitting its value to the server.

### DIFF
--- a/common/static/common/js/application.js
+++ b/common/static/common/js/application.js
@@ -8,6 +8,9 @@ import initStepNav from './step-by-step-nav';
 import { initAll } from 'govuk-frontend';
 import initCheckAllCheckboxes from './checkAllCheckboxes';
 showHideCheckboxes();
+// Initialise accessible-autocomplete components without a `name` attr in order
+// to avoid the "dummy" autocomplete field being submitted as part of the form
+// to the server.
 initAutocomplete(false);
 initStepNav();
 initAll();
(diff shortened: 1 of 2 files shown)


common/static/common/js/autocomplete.js
```diff
@@ -26,7 +26,7 @@ const initAutocomplete = () => {
       },
       minLength: element.dataset.minLength ? element.dataset.minLength : 0,
       defaultValue: element.dataset.originalValue,
-      name: `${hiddenInput.name}_autocomplete`,
+      name: "",
```
stuaxo, line 29: Not a blocker - but a ternery operator may be useful here, we use it with minLength above and it avoids having to scan further down the code to see what's going on.

```javascript
name: includeNameAttr ? `${hiddenInput.name}_autocomplete` : "",
```

common/static/common/js/application.js
```diff
@@ -8,7 +8,7 @@ import initStepNav from './step-by-step-nav';
 import { initAll } from 'govuk-frontend';
 import initCheckAllCheckboxes from './checkAllCheckboxes';
 showHideCheckboxes();
-initAutocomplete();
+initAutocomplete(false);
```
stuaxo, line 11: ^ I reckon we really want a comment to say why we are not setting the name attribute - it's the kind of thing someone could easily change in future not realising it's a breaking change.

stuaxo (approved):

common/static/common/js/autocomplete.js
paulpepper-trade: It's a Pythonic approach ;)

PR #524, opened 2022-03-25
TP2000-169 The edit measures journey doesn't support adding measure conditions

gabelton:
# TP2000-169 The edit measures journey doesn't support adding measure conditions

## Why
* To enable tariff managers to add multiple conditions to an existing measure

## What
* Update `MeasureForm.save()` to use `MeasureConditionsFormSet` to create new conditions. (Currently this process involves deleting all the footnotes currently attached to a measure and then creating them from the new data. This is a bit inefficient, but we have existing tickets to look at editing / removing conditions and I think we could refactor as part of these)
* Add `create_conditions_and_components` method to `MeasureCreationPattern` and call within `MeasureUpdate` view.
* Add view and form tests

## Checklist
- No migrations necessary
- No dependency updates necessary


measures/forms.py
```diff
@@ -212,11 +212,37 @@ def __init__(self, *args, **kwargs):
...
+            if isinstance(data["start_date_1"], list)
+            else int(data["start_date_1"])
+        )
+        day = (
+            int(data["start_date_0"][0])
+            if isinstance(data["start_date_0"], list)
+            else int(data["start_date_0"])
+        )
+
+        return datetime.date(year, month, day)
```
simonwo: Aren't the form fields meant to do this for us?

measures/forms.py
```diff
@@ -38,6 +39,268 @@
...
+        to redisplay the formset with an extra empty form or the selected form
+        removed."""
+        # Re-present the form to show the result of adding another form or
+        # deleting an existing one.
+        if self.formset_action == "ADD" or self.formset_action == "DELETE":
+            return False
+
+        # An empty set of forms is valid.
+        if self.total_form_count() == 0:
+            return True
```
simonwo, line 147: Is this true if `min_num` was not zero?

measures/forms.py
```diff
@@ -220,8 +483,45 @@ def save(self, commit=True):
...
+
         return instance
 
+    def is_valid(self) -> bool:
+        conditions_formset = MeasureConditionsFormSet(self.data)
+
+        if not conditions_formset.is_valid():
+            # for error in conditions_formset.errors[0]:
+            #     self.add_error(field=None, error=conditions_formset.errors[0][error])
+
```
simonwo, line 520: Commented out code

measures/patterns.py
```diff
@@ -270,6 +270,42 @@ def create_measure_footnotes(
...
+            condition_measurement=data.get(
+                "condition_measurement",
+            ),
+        )
+        condition.clean()
+        condition.save()
+
+        # XXX the design doesn't show whether the condition duty_amount field
+        # should handle duty_expression, monetary_unit or measurements, so this
+        # code assumes some sensible(?) defaults
```
simonwo, line 300: I don't think this is true anymore

simonwo (approved):

PR #526, opened 2022-03-29
TP2000-214 Add conditional rendering to MeasureCondition form

rachaelcodes:
# TP2000-214 Add conditional rendering to `MeasureCondition` form

## Why
As different measure conditions expect different supporting data, to make data entry easier, this hides any measure condition form fields that are not relevant for a particular `MeasureConditionCode` or `MeasureAction`. _For more information, see: https://uktrade.github.io/tariff-data-manual/documentation/data-structures/measure-conditions.html#condition-codes_

## What
This adds a new JavaScript file that allows for conditional rendering of individual fields in the measure conditions form. (The form will still work without JavaScript, but will not have the option of hiding fields for clarity, so there is some progressive enhancement).

New Jinja templates have been added so that the triggers for hiding particular fields are generated from the `MeasureConditionCode` and `MeasureAction` instances rather than a hardcoded list of ids. `data-*` attributes have been used to pass this information to the JS frontend.

### Before
All fields visible
Before

### After
Only relevant fields visible
After

## Checklist
Will require a rebuilding of static files to view locally

### DIFF
--- a/.github/actions/setup-python/action.yml
+++ b/.github/actions/setup-python/action.yml
@@ -25,7 +25,6 @@ runs:
           ${{ runner.os }}-pip-
 
     - name: Install dependencies
-      if: steps.cache.outputs.cache-hit != 'true'
       shell: bash
       run: |
         python -m pip install --upgrade pip
--- a/common/jinja2/components/measure_condition_action_code/template.jinja
+++ b/common/jinja2/components/measure_condition_action_code/template.jinja
@@ -1,0 +1,18 @@
+<div id="div_id_{{ field.html_name }}" class="govuk-form-group">
+    {% set selected_action = field.initial | int %}
+    <label for="id_{{ field.html_name }}" class="govuk-label">
+        Action code
+    </label>
+    <select name="{{ field.html_name }}" class="govuk-select" id="id_{{ field.html_name }}"> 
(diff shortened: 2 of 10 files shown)


gabelton (commented):
Looks really good ! We mentioned moving action code, so that it's not nested beneath condition code. Is that something we still want to do? I think it would be nice visually because action code isn't conditional on condition code and the nesting normally indicates that something is conditional on the field above.

simonwo (approved):
Wow, looks great!

We could probably do with rewording "Reference price (where applicable)" to something like "Reference price or quantity"

.github/actions/setup-python/action.yml
```diff
@@ -25,7 +25,6 @@ runs:
           ${{ runner.os }}-pip-
 
     - name: Install dependencies
-      if: steps.cache.outputs.cache-hit != 'true'
```
gabelton, line 28: Why does this have to be removed? (not implying that it doesn't have to be)

rachaelcodes: We were having some problem loading dependencies in the Github automated tests after the recent linter dependency change. @simonwo suggested this fix - which did the trick.

gabelton (approved):

PR #529, opened 2022-04-05
Added provision to clear workbaskets of tracked models.

paulpepper-trade:
# TP2000-237 Clear workbasket via Django admin site

## Why

In order for TMs and data engineers to cooperatively work together in the current environment, It’s sometimes necessary to clear down the current editing workbasket.

## What

Add a provision to allow a workbasket’s tracked models to be deleted, leaving the workbasket intact, but empty.


simonwo (approved):

workbaskets/util.py
```diff
@@ -32,5 +32,12 @@ def delete_workbasket(workbaset):
             ).first()
             version_group.save()
 
-    workbaset.transactions.all().delete()
-    workbaset.delete()
+    workbasket.transactions.all().delete()
```
stuaxo, line 35: Not a blocker - 

^ A shame this has to be this fiddly, it would be nice if `nursery.remove_objects_from_cache(...)` existed that did the cache and version group work.

PR #530, opened 2022-04-06
TP2000-238 Duties field is required on Measure create wizard commodities / duties step

gabelton:
## Why
The duties field on the commodities / duties step of the wizard should not be required because it should be possible to create a measure without a duty amount.

The wizard get_form_initial method also should be update because there is no longer a step called "duties", which means that duty validation is not being called and we are seeing 500 parse errors. This validation should also be called for the "conditions" step.

## What
- Make duties field required on form and update get_form_initial method

https://uktrade.atlassian.net/browse/TP2000-238

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -512,6 +512,7 @@ class MeasureCommodityAndDutiesForm(forms.Form):
 
     duties = forms.CharField(
         label="Duties",
+        required=False,
     )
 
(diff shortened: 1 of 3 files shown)


paulpepper-trade (approved):

eadpearce (approved):
LGTM

PR #532, opened 2022-04-07
TP2000-221: Users cannot select footnotes etc. that they've just created

gabelton:
## Why
Users can't select footnotes (or other described objects) that they've just created because `autocomplete_label` errors. This happens because get_description is returning none, due it not filtering by current transaction.

## What
- Updates `get_descriptions` to use `Transaction.objects.last()` where no request or transaction object is passed in.
- Updates factories for described objects to ensure that description and described object use the same transaction, reflecting the logic we have in create views.
- Also adds a fix for [this bug](https://uktrade.atlassian.net/browse/TP2000-196) where current footnotes aren't displaying after measure update by making `MeasureUpdate.get_footnotes` filter by measure sid, rather than pk


simonwo (changes requested):
Unfortunately I don't think that's the right fix – it will choose the latest transaction across all workbaskets. So draft data from another workbasket might appear, and changes in the correct workbasket still won't be visible.

Hm, I'm not I understand why we even reach that line. Is the method not being passed a `request` parameter? Is a fix to make sure that it gets one, or to refactor this to use `current_objects`?

gabelton:
> Unfortunately I don't think that's the right fix – it will choose the latest transaction across all workbaskets. So draft data from another workbasket might appear, and changes in the correct workbasket still won't be visible.
> 
> Hm, I'm not I understand why we even reach that line. Is the method not being passed a `request` parameter? Is a fix to make sure that it gets one, or to refactor this to use `current_objects`?

I was [chatting with Stuart](https://teams.microsoft.com/l/message/19:80f6ef27ef454a87a8c4844d3aa5b3f7@thread.tacv2/1649343005527?tenantId=8fa217ec-33aa-46fb-ad96-dfe68006bb86&groupId=7dec05eb-83aa-4ca6-8bff-bf9b86f579e7&parentMessageId=1649343005527&teamName=TAP&channelName=Dev%20and%20Data&createdTime=1649343005527) a bit about this before I went on sick leave. I wasn't able to get it working with current_objects. I think the solution is to find a way to pass in the `request` parameter, as you suggest because it doesn't currently have access to it, which is why it's reaching that line.

gabelton:
Actually I think it was just me using it wrong. Testing in my local while using current objects seems to be working. I've broken some sheet importer tests, though, and I'm trying to work out if it's an issue with test setup

common/models/mixins/description.py
```diff
@@ -124,7 +125,18 @@ def get_descriptions(self, transaction=None, request=None) -> TrackedModelQueryS
                 transaction=WorkBasket.get_current_transaction(request),
             )
```
simonwo, line 126: Can we also clean up the `if request` branch now, as `request` doesn't seem to be used anywhere and using `current` is the better way to handle that anyway?

common/models/mixins/description.py
```diff
@@ -117,14 +118,18 @@ def get_descriptions(self, transaction=None, request=None) -> TrackedModelQueryS
...
             return query.approved_up_to_transaction(transaction)
 
-        if request:
-            from workbaskets.models import WorkBasket
-
-            return query.approved_up_to_transaction(
-                transaction=WorkBasket.get_current_transaction(request),
-            )
+        if get_current_transaction():
+            return query.current()
```
paulpepper-trade, line 122: The relationship between `get_current_transaction()` and the `TrackedModelQuerySet` value returned from `query.current()` isn't clear in this context. Is it perhaps worth adding a brief explanation in a comment.

simonwo (approved):

simonwo (approved):

PR #533, opened 2022-04-08
Get ALLOWED_HOSTS list from environment commas/space seperated ALLOWE…

stuaxo:
Get ALLOWED_HOSTS list from environment commas/space separated ALLOWED_HOSTS environment variable.

### DIFF
--- a/docs/source/envvars.rst
+++ b/docs/source/envvars.rst
@@ -160,7 +160,7 @@ Environment variables
 
     (default ``False``)
 
-    Do call the HMRC API notification endpoint after each upload
+    Do not call the HMRC API notification endpoint after each upload
 
 .. envvar:: DJANGO_SETTINGS_MODULE
 
--- a/docs/source/envvars.rst
+++ b/docs/source/envvars.rst
@@ -156,12 +156,6 @@ Environment variables
 
     Password for local MinIO instance
 
-.. envvar:: ALLOW_PAAS_HOSTS
-
-    (default ``False``)
-
-    Add URIs from the PaaS environment format variable VCAP_APPLICATION to ALLOWED_HOSTS
-
 .. envvar:: EXPORTER_DISABLE_NOTIFICATIONS
 
     (default ``False``)
--- a/docs/source/envvars.rst
+++ b/docs/source/envvars.rst
@@ -154,7 +154,19 @@ Environment variables
 
 .. envvar:: MINIO_SECRET_KEY
 
-    Password for local MinIO instance 
+    Password for local MinIO instance
+
+.. envvar:: ALLOW_PAAS_HOSTS
+
+    (default ``False``)
+
+    Add URIs from the PaaS environment format variable VCAP_APPLICATION to ALLOWED_HOSTS
+
+.. envvar:: EXPORTER_DISABLE_NOTIFICATIONS
+
+    (default ``False``)
+
+    Do call the HMRC API notification endpoint after each upload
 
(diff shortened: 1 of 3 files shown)


stuaxo:
```
[15:57] Paul Pepper (Guest)
    I think the set of valid host name are available in the VCAP_APPLICATION[uris] env var. I guess it's possible to fix-up ALLOWED_HOSTS from there?
​[16:03] stuart.axon (Guest)
    Interesting, yeah I can definitely hook it up to that - maybe as a default if none are specified, or as-well-as ?
​[17:28] Paul Pepper (Guest)
    
Hmm, not sure. I think VCAP_APPLICATION contains the canonical set of hostnames for each service instance and takes values from the same source as is used to create DNS (A, CNAME, etc) records for those services. So taking values from VCAP_APPLICATION[uris]may be the safer option. Probably worth checking!
<https://teams.microsoft.com/l/message/19:80f6ef27ef454a87a8c4844d3aa5b3f7@thread.tacv2/1649429852224?tenantId=8fa217ec-33aa-46fb-ad96-dfe68006bb86&amp;groupId=7dec05eb-83aa-4ca6-8bff-bf9b86f579e7&amp;parentMessageId=1649416659001&amp;teamName=TAP&amp;channelName=Dev and Data&amp;createdTime=1649429852224>
```

settings/common.py
```diff
@@ -182,7 +183,11 @@
 
 # Whitelist values for the HTTP Host header, to prevent certain attacks
 # App runs inside GOV.UK PaaS, so we can allow all hosts
-ALLOWED_HOSTS = ["*"]
+ALLOWED_HOSTS = re.split(r"\s|,", os.environ.get("ALLOWED_HOSTS", ""))
+if is_truthy(os.environ.get("ALLOW_PAAS_HOSTS", False)):
+    # Under PaaS, if ALLOW_PAAS_URIS is set, fetch trusted domains from VCAP_APPLICATION env var
+    paas_hosts = json.loads(os.environ["VCAP_APPLICATION"])["uris"]
+    ALLOWED_HOSTS.extend(paas_hosts)
```
simonwo, line 190: Why the need for `ALLOW_PAAS_HOSTS`? Why not always do this if `VCAP_APPLICATION` is set, as we do for e.g. postgres and redis?

stuaxo: Good question, I prefer to make it explicit since I've had this set before locally (though I guess this isn't usually the case - and I was probably using conduit or similar).

I thought it would follow the principle of least surprise, that if you do have this set you won't try and connect.

Though now I've typed this twice it seems pretty unlikely - I guess it's an acknowledgement that you really do want this (and a way of turning this off on paas too by not setting it).

stuaxo: Oh wait, I must have been distracted and not read your full sentence that we do this already, will make that change.

stuaxo: Done

docs/source/envvars.rst
```diff
@@ -154,7 +154,13 @@ Environment variables
...
 .. envvar:: MINIO_SECRET_KEY
 
-    Password for local MinIO instance 
+    Password for local MinIO instance
+
+.. envvar:: EXPORTER_DISABLE_NOTIFICATIONS
+
+    (default ``False``)
+
+    Do call the HMRC API notification endpoint after each upload
```
gabelton, line 163: This line reads weirdly to me. Could you rephrase as "Whether to call..." ?

PR #534, opened 2022-04-12
Firebreak bag

simonwo:
A selection of tweaks encountered during migration work over firebreak:

- Typing fixes for maybe_*
- Declarable commodities
- Improve item ID cleaner to ignore all kinds of whitespace.
- Automatically create a description when creating a described record
- Tweak sheet importer to better handle splits and return itself by default.


commodities/models/dc.py
```diff
@@ -447,19 +448,16 @@ def get_dependent_measures(
...
             qs = qs.latest_approved()
 
         if as_at is not None and as_at is not NOT_PROVIDED:
-            logger.debug("Filtering by supplied date: %s", as_at)
             qs = qs.with_effective_valid_between().filter(
                 Q(db_effective_valid_between__contains=as_at)
                 | Q(valid_between__startswith__gte=as_at),
             )
         elif as_at is NOT_PROVIDED and self.moment.clock_type.is_calendar_clock:
-            logger.debug("Filtering by moment date: %s", self.moment.date)
```
gabelton, line 462: Are these logs not being used ?

gabelton (approved):

commodities/models/dc.py
simonwo: I don't think so, and they get executed a lot when doing anything with the commodity code tree.

PR #536, opened 2022-04-14
Tools to filter and renumber TARIC XML files.

simonwo:
## Why

As part of our operational process to keep Production in line with data changes made on engineer machines, it is sometimes necessary to manually renumber or filter TARIC XML files to avoid data created on Production clashing on auto-picked ID numbers.

## What

This PR adds three new tools (`renumber_records`, `renumber_transactions` and `filter_transactions`) that can be run against XML files, and tests for them.

### DIFF
--- a/common/tests/test_xml_util.py
+++ b/common/tests/test_xml_util.py
@@ -1,0 +1,89 @@
+import pytest
+from django.db import transaction
+
+from common.models.transactions import Transaction
+from common.tests import factories
+from common.tests.util import serialize_xml
+from common.xml.util import remove_transactions
+from common.xml.util import renumber_records
+from common.xml.util import renumber_transactions
+
+pytestmark = pytest.mark.django_db
+
+
+def test_record_renumbering_produces_valid_import_data(import_xml, export_xml):
+    """Tests that when the renumbering function is used, the XML file it
+    produces is valid and when imported produces records in the database that
+    are correctly renumbered."""
+    code_type = factories.AdditionalCodeTypeFactory.create()
+    workbasket = factories.ApprovedWorkBasketFactory()
+    first = factories.AdditionalCodeFactory.create(
+        type=code_type,
+        transaction__workbasket=workbasket,
+    )
+    second = factories.AdditionalCodeFactory.create(
+        type=code_type,
+        transaction__workbasket=workbasket,
+    )
+
+    for instance in (first, second):
+        assert not type(instance).objects.filter(sid=instance.sid + 2).exists()
+
+    xml = export_xml(workbasket=workbasket)
+    renumber_records(xml, second.sid + 1, "oub:additional.code.sid")
+    import_xml(serialize_xml(xml))
+
+    for instance in (first, second):
+        new_instance = type(instance).objects.get(sid=instance.sid + 2)
+        assert new_instance.type == instance.type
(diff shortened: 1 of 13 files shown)


common/tests/test_xml_util.py
```diff
@@ -0,0 +1,80 @@
...
+    for instance in (first, second):
+        assert not type(instance).objects.filter(sid=instance.sid + 2).exists()
+
+    xml = export_xml(workbasket=workbasket)
+    renumber_records(xml, second.sid + 1, "oub:additional.code.sid")
+    import_xml(serialize_xml(xml))
+
+    for instance in (first, second):
+        new_instance = type(instance).objects.get(sid=instance.sid + 2)
+        assert new_instance.type == instance.type
```
stuaxo, line 35: This looks like it might want to be `is` not `==`  (I may well be wrong too)

common/tests/test_xml_util.py
```diff
@@ -0,0 +1,80 @@
...
+from common.tests import factories
+from common.tests.util import serialize_xml
+from common.xml.util import filter_transactions
+from common.xml.util import renumber_records
+from common.xml.util import renumber_transactions
+
+pytestmark = pytest.mark.django_db
+
+
+def test_record_renumbering_produces_valid_import_data(import_xml, export_xml):
```
stuaxo, line 14: A docstring to explain a little about what this tests would be useful.

common/xml/util.py
```diff
@@ -0,0 +1,165 @@
...
+from xml.etree import ElementTree
+
+from common.validators import UpdateType
+from common.xml.namespaces import nsmap
+
+
+@contextlib.contextmanager
+def rewrite(filename: str):
+    """Open the passed XML file and return it, and then write back whatever
+    changes are made to it afterwards."""
```
stuaxo, line 14: Might be worth tweaking the line to be more specific than `return it` and mention it yields the Element Tree for users to affect their changes on.

Re-writes original file (which should be obvious by the docstring, but might warrant a warning on it's own line.

common/xml/util.py
```diff
@@ -0,0 +1,165 @@
...
+        root = tree.getroot()
+        yield root
+
+    tree.write(filename, "utf-8", True)
+
+
+ATTRIBUTE_XPATH = re.compile(r"\[@([a-z]+)\]")
+
+
+def renumber_paths(element: ElementTree.Element, start_from: int, *paths: str):
```
stuaxo, line 27: Nice :)

common/xml/util.py
```diff
@@ -0,0 +1,165 @@
...
+            return True
+
+    return False
+
+
+def filter_transactions(
+    envelope: ElementTree.Element,
+    element_name: str,
+    element_values: Collection[str],
+):
```
stuaxo, line 150: It may be too late, but I wonder if it this could be named differently to reflect this removes items ?
We extensively use the ORM `objects.filter` where `filter` choses not exclude objects, and the having the meaning flipped here may make it easy to make a mistake in the future.

stuaxo (approved):

common/tests/test_xml_util.py
simonwo: I think this is OK – it's comparing an attribute called type which is just another model rather than a Python type.

PR #537, opened 2022-04-21
New content to be added to the start page of the Create Measures journey

LaurenQurashi:
# TP2000-256 New content to be added to the start page of the Create Measures journey

## Why
This PR adds additional text to the create measures start page in order to inform a user of the information they need in order to create a measure. 

This will prevent users from partially completing the form, only to find that they have to go back and lose all their work as they've missed an important piece of data, or didn't create something they needed for this process first. 

Due to the nature of the tool not having drafts, or versioning for creating a measure, the next most appropriate measure is to ensure that the user is adequately prepared to complete this process in one sitting. 

## What
This PR adds text to the page to highlight what information is needed to create a measure. It goes from looking like this: 

Screenshot 2022-04-21 at 11 55 09

...To looking like this: 

Screenshot 2022-04-21 at 11 55 09

### DIFF
--- a/measures/jinja2/measures/create-start.jinja
+++ b/measures/jinja2/measures/create-start.jinja
@@ -4,7 +4,29 @@
 
 {% block form %}
   <p class="govuk-body">
-    It is essential to plan your activities in advance of using this measure creation functionality.
+    Before you start, you'll need:
+    <ul class="govuk-list govuk-list--bullet">
+      <li>the measure type</li>
+      <li>the measure start date</li>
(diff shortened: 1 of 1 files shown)


gabelton (approved):
Looks good. I think I've fixed the failing tests that you're seeing in the CI on my currently open branch. Slightly worrying that they passed before and not now though...

stuaxo:
Looks good to me, it may be an idea to hold off until a fix for the test fail (not from this branch is ready)  ?

gabelton (approved):

PR #539, opened 2022-04-22
TP2000-195 Users can get a 500 error if they fail to select a geographical area

gabelton:
# TP2000-195 Users can get a 500 error if they fail to select a geographical area

## Why
Currently users can go through the create measure journey without selecting a geographical area in the measure details form. This causes a 500 on the summary page when trying to render the geographical area's description.

## What
- Add a validation error on the measure details form when a user hasn't selected a geographical area
- Updates tests to include a default erga omnes value

## Checklist
- Requires migrations? No


paulpepper-trade (approved):
LGTM

stuaxo (approved):
LGTM

PR #540, opened 2022-04-25
Parallelise and persist checks of business rules.

simonwo:
## Why

Users are able to produce models using the UI that do not pass validation of TARIC business rules. When they do this, the need to be informed of the issues so that they can go and manually correct them. Business rule checking is slow, so running all of the rules to find and display errors cannot be run during a single web request.

So the requirements on the checks subsystem are:

1. Show the user all of the errors in their (1000+) model changes in the UI.
2. When the user has finished all of their editing, allow the final output to be downloaded quickly without them having to wait for the checks to run.

## What
This commit introduces a new `checks` subsystem that can run background checks against models.

Checks are designed to be run asyncronously using Celery. Syncronous checking is available but is only designed for use in testing and debugging.

Whilst the checks are designed to run business rules, this is not hard-coded and it is easy to implement new types of checks that aren't business rule focused if desired. Any new business rules that are implemented will automatically be picked up.

In order to make web requests fast, the checks system includes a cache of what checks have have been run, whether any have failed, and how. Like any cache, checks can become outdated if transactions in draft are modified. The system includes the ability to detect this and re-run new checks accordingly.

The system is designed to be tolerant towards background tasks being killed at any point during their execution. The system does not assume that a task will be completed if it is started. The database therefore is the only stateful part of the system and only what is present in the database controls the execution of new checks and whether or not a workbasket is valid and can be
sent.

The system is also designed to be tolerant of poor integration with the rest of the app. It does not require that all of the places that edit transactions also know how to start checks. Instead, a scheduled task is used to hunt for transactions that have not been checked yet and pre-emptively check them. This means that developers on the whole do not need to think
about how to start checks and can just leave the system to run checks at appropriate times.

The assumptions in the system are:

1. Transactions are immutable once they have been approved (i.e. they enter a partition in ``TransactionPartition.approved_partitions()``). Checks against approved transactions will not be checked again.
2. There is at least one transaction in an approved partition.
4. Changes made to a draft transaction take the form of a removal or an addition of a model – models are never edited directly. Unpicking this assumption is hard, so it's recommended that the system sticks to this if it can.
5. Checks can be run independently of each other and do not modify data. (Actually, checks can modify data as long as they do it in the manner described in point 3 – all of other checks will then immediately be invalidated, so in the worst case this is a waste of resources.)

## Checklist
- Requires migrations? **Yes.**


common/tests/factories.py
```diff
@@ -149,18 +149,31 @@ class Meta:
...
+
+        seed = factory.Trait(
+            partition=TransactionPartition.SEED_FILE,
+        )
+
+        draft = factory.Trait(
+            partition=TransactionPartition.DRAFT,
+            workbasket=factory.SubFactory(WorkBasketFactory),
+        )
+
```
stuaxo, line 165: ^ nice, TIL about factory.Trait.

checks/checks.py
```diff
@@ -0,0 +1,162 @@
...
+    A ``Checker`` is an object that knows how to perform a certain kind of check
+    against a model.
+
+    Checkers can by applied against a model. The logic of the checker will be
+    run and the result recorded as a ``TrackedModelCheck``.
+    """
+
+    @cached_property
+    def name(self) -> str:
+        """
```
stuaxo, line 35: (Just because I am a sucker for keywords) - it may be good to use the word "unique" in this docstring as this appears to be a unique name.

checks/tasks.py
```diff
@@ -0,0 +1,150 @@
...
+from common.models.transactions import TransactionPartition
+from common.models.utils import override_current_transaction
+
+logger = logging.getLogger(__name__)
+
+
+@app.task(time_limit=60)
+def check_model(trackedmodel_id: int, context_id: int):
+    """Runs all of the applicable checkers on the passed model ID, and records
+    the results."""
```
stuaxo, line 20: Since this has a time limit, it's worth mentioning this, and possibly a few words on why it was chosen.

checks/tasks.py
```diff
@@ -0,0 +1,150 @@
...
+
+    check, model_ids = setup_or_resume_transaction_check(transaction_id)
+    if check.completed and not any(model_ids):
+        logger.debug(
+            "Skipping check of transaction %s "
+            "because an up-to-date check already exists",
+            transaction_id,
+        )
+        return
+
```
stuaxo, line 107: Some words about the workflow structure being built here might be good since many devs experience of celery is just "thing that sends emails for me", and they don't delve into building workflows.

Similar for the .replace and why it's there, as it's not something often used.

checks/checks.py
simonwo: Yeah, it's kinda hard to describe what it is unique over. It should uniquely identify "a check that it is necessary to run to prove validity" but that doesn't necessarily equate to a `Type[Checker]` or a `Checker`.

stuaxo:
Its good there are docstrings about what each function does, it would be good to also mention the intended architecture of the task, at least that -

The task is intended to be idempotent, this is achieved by running integrity checks before and after the it.*

Since this seems really crucial to the design, it's probably good to be explicit about it, and will help if in future we break or even have to fix / extend the idempotency check.


* I'm assuming I've understood things correctly here.

paulpepper-trade:
> Changes made to a draft transaction take the form of a removal or an addition of a model – models are never edited directly. Unpicking this assumption is hard, so it's recommended that the system sticks to this if it can.
Could you remark a little more on why editing is omitted and why it would be hard to implement this?

This PR's description could make a good ADR, or the basis for one, if we get the chance.

checks/querysets.py
```diff
@@ -0,0 +1,152 @@
...
+    freshness_fields = {
+        "tracked_model_count": Count("transaction__tracked_models"),
+        "latest_tracked_model": Max("transaction__tracked_models__id"),
+    }
+
+    freshness_annotations = {
+        f"real_{field}": expr for field, expr in freshness_fields.items()
+    }
+
+    freshness_filter = models.Q(
```
gabelton, line 58: Would it be possible to add some more comments around this freshness filter?

checks/tasks.py
```diff
@@ -0,0 +1,150 @@
...
+        for check in applicable_to(model):
+            if not context.model_checks.filter(
+                model=model,
+                check_name=check.name,
+            ).exists():
+                check.apply(model, context)
+
+
+@app.task
+def decide_if_transaction_check_complete(check_id: int) -> bool:
```
gabelton, line 36: Could you rename to something like 'check_transaction_complete' to make it slightly shorter?

checks/tasks.py
```diff
@@ -0,0 +1,150 @@
...
+        context.save()
+
+    return (
+        context,
+        transaction.tracked_models.values_list("pk", flat=True),
+    )
+
+
+@app.task(bind=True)
+def check_transaction(self, transaction_id: int):
```
gabelton, line 96: could this be called check_transaction_async?

simonwo:
> > Changes made to a draft transaction take the form of a removal or an addition of a model – models are never edited directly. Unpicking this assumption is hard, so it's recommended that the system sticks to this if it can.
> 
> Could you remark a little more on why editing is omitted and why it would be hard to implement?

The metadata the system collects can only detect if models are added or removed at the moment. I couldn't find any other metadata that we already collect that would consistently show that a model has been edited. 

I guess it could be as simple as a `updated_at` timestamp on the tracked model table but at the moment we don't collect them… 

> This PR's description could make a good ADR, or the basis for one, if we get the chance.

It's actually cribbed from the module-level docstring! Which will appear in the architecture doc.

> The task is intended to be idempotent, this is achieved by gathering metadata unique to the data to be processed, then once business rules have been run verifying it hasn't been edited in the meantime by checking that the metadata has the same values as it had at the beginning.

Cool, I will add that to the module-level docstring.

checks/checks.py
stuaxo: > so that the system can record the name in the database

A guess: Is a unique id formed, `checker name` + `data to be checked` ?

If so, maybe that's the context I was missing from the docstring ?

checks/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+so running all of the rules to find and display errors cannot be run during a
+single web request.
+
+So the requirements on the checks subsystem are:
+
+1. Show the user all of the errors in their (1000+) model changes in the UI.
+2. When the user has finished all of their editing, allow the final output to be
+   downloaded quickly without them having to wait for the checks to run.
+
+Checks are designed to be run asyncronously using Celery. Syncronous checking is
```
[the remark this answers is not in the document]
simonwo: Ah, yes… I can't spell

checks/checks.py
```diff
@@ -0,0 +1,162 @@
...
+
+Self = TypeVar("Self")
+
+
+class Checker:
+    """
+    A ``Checker`` is an object that knows how to perform a certain kind of check
+    against a model.
+
+    Checkers can by applied against a model. The logic of the checker will be
```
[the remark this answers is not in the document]
simonwo: Ah yes… I can't grammar.

checks/checks.py
```diff
@@ -0,0 +1,162 @@
...
+        return type(self).__name__
+
+    @classmethod
+    def checkers_for(cls: Type[Self], model: TrackedModel) -> Collection[Self]:
+        """
+        Returns instances of this ``Checker`` that should apply to the model.
+
+        What checks apply to a model is sometimes data-dependent, so it is the
+        responsibility of the ``Checker`` class to tell the system what
+        instances of itself it would expect to run against the model.
```
stuaxo, line 54: An example, (real or made up) would be really good here.

checks/tasks.py
```diff
@@ -0,0 +1,150 @@
...
+    context: TransactionCheck = TransactionCheck.objects.get(pk=context_id)
+    transaction = context.transaction
+
+    with override_current_transaction(transaction):
+        for check in applicable_to(model):
+            if not context.model_checks.filter(
+                model=model,
+                check_name=check.name,
+            ).exists():
+                check.apply(model, context)
```
stuaxo, line 32: Oooh I was in celery mode, where `.apply` is one of their APIs (and wondering why it is passed a model).

Add a comment to mention this is not celerys .apply to disambiguate, otherwise it's easily confusing since we are already in a celery task.

checks/models.py
```diff
@@ -0,0 +1,129 @@
...
+    transaction_check = models.ForeignKey(
+        TransactionCheck,
+        on_delete=models.CASCADE,
+        related_name="model_checks",
+    )
+
+    check_name = fields.CharField(max_length=255)
+    """A string identifying the type of check carried out."""
+
+    successful = fields.BooleanField()
```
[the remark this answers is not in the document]
simonwo: So… no, the `TrackedModelCheck` only gets created once the check is complete, and hence success should always be known. This is unlike the `TransactionCheck` which gets created on the check starts and has a flag to track completeness. The reason is that a single model check is an atomic thing and so its either done or not, and it can't be "resumed". The docs should probably point this out?

checks/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+available but is only designed for use in testing and debugging.
```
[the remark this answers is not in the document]
simonwo: In Jupyter notebooks? Using this system with a concurrent Celery worker could speed it up a lot.

But the data would have to be commited so that it can be seen outside of the notebook process, which is not the workflow we use. So it's probably still a synchronous process…

checks/tasks.py
simonwo: It could… I left off the `async` to suggest that this is the one that you **should** be using, i.e. this is the **normal** one. But can change if you think?

checks/tasks.py
```diff
@@ -0,0 +1,150 @@
...
+        check.successful = not check.model_checks.filter(successful=False).exists()
+        logger.info("Completed checking transaction %s", check.transaction.id)
+
+    check.save()
+    return check.completed
+
+
+def setup_or_resume_transaction_check(transaction_id: int):
+    """Return a current, fresh transaction check for the passed transaction ID
+    and a list of model IDs that need to be checked."""
```
stuaxo, line 67: Since this is a standard function, not a Task it can probably take a plain transaction, which _may_ be kinder to the ORM.

checks/tasks.py
```diff
@@ -0,0 +1,150 @@
...
+    return (
+        context,
+        transaction.tracked_models.values_list("pk", flat=True),
+    )
+
+
+@app.task(bind=True)
+def check_transaction(self, transaction_id: int):
+    """Run and record checks for the passed transaction ID, asynchronously."""
+
```
stuaxo, line 98: Since setup_or_resume_transaction_check is a standard function, not a Task it can probably be passed a Transaction, which _may_ be kinder to the ORM down the line.

workbaskets/tasks.py
```diff
@@ -18,3 +22,28 @@ def transition(instance_id: int, state: str, *args):
...
+def check_workbasket_sync(workbasket_id: int):
+    """
+    Run and record transaction checks for the passed workbasket ID,
+    synchronously.
+
+    This method will run all of the checks one after the other and won't return
+    until they are complete. This is useful for testing and debugging.
+    """
+    workbasket: WorkBasket = WorkBasket.objects.get(pk=workbasket_id)
+    transactions = workbasket.transactions.values_list("pk", flat=True)
```
stuaxo, line 46: Since `check_transaction_sync` is a plain function, not a task it could take a plain Transaction which _may_ be a little kinder on the ORM.

checks/checks.py
simonwo: Yeah, I guess a unique ID is formed by checker name + model.

checks/tasks.py
```diff
@@ -0,0 +1,150 @@
...
+    ids_require_update = (
+        Transaction.objects.exclude(
+            pk__in=TransactionCheck.objects.requires_update(False).values(
+                "transaction__pk",
+            ),
+        )
+        .filter(partition=TransactionPartition.DRAFT)
+        .values_list("pk", flat=True)
+    )
```
stuaxo, line 148: This is either correct or I brainwashed you into writing code I would have written :thinking:

checks/models.py
```diff
@@ -0,0 +1,129 @@
+from django.db import models
+from django.db.models import fields
+
+from checks.querysets import TransactionCheckQueryset
+from common.models.trackedmodel import TrackedModel
+from common.models.transactions import Transaction
+
+
+class TransactionCheck(models.Model):
+    """Represents the result of checking a transaction for correctness."""
```
stuaxo, line 10: Mabe something for the future - but it feels like these Checks want to extend something, even if it's just an abstract base class.

checks/models.py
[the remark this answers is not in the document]
stuaxo: Definitely worth mentioning the lifecycle in both since they are different.

checks/models.py
```diff
@@ -0,0 +1,129 @@
...
+    key. This is used to detect if the check is now stale.
+    """
+
+    model_checks: models.QuerySet["TrackedModelCheck"]
+
+    objects: TransactionCheckQueryset = models.Manager.from_queryset(
+        TransactionCheckQueryset,
+    )()
+
+    def save(self, *args, **kwargs):
```
stuaxo, line 76: Probably worth a comment that this is saving data about ... "uniqueness"  (sorry, bit fuzzy, there is probably a better way of putting this).

It feels like the stuff that identifies the record is something we want to be explicit about, but I don't have time before you leave to come up with a good suggestion for an API.

checks/tasks.py
stuaxo: Feel free to disregard this bikeshedding comment:

`is_transaction_check_complete` is a little less ambiguous.   Otherwise it can read like the transaction itself is complete (which isn't a thing we have, but in English you might do).

checks/tasks.py
```diff
@@ -0,0 +1,150 @@
...
+    return self.replace(workflow)
+
+
+def check_transaction_sync(transaction_id: int):
+    """
+    Run and record checks for the passed transaction ID, syncronously.
+
+    This method will run all of the checks one after the other and won't return
+    until they are complete. This is useful for testing and debugging.
+    """
```
stuaxo, line 121: Oh, I wasn't expecting one of these in `checks`, and one in `workbaskets` !

The same comment as the other one:

Since check_transaction_sync is a plain function, not a task it could take a plain Transaction which may be a little kinder on the ORM.

checks/tasks.py
stuaxo: I think `_async` as a suffix leads to an expectation this uses the python async stuff, whereas this is a uses Celery to run asyncronously.

I have a slight preference for no suffix, but if there is one, then `_task` might communicate that it's a celery thing if the dev is looking at function names.

gabelton (approved):

stuaxo (approved):

PR #542, opened 2022-04-26
Save a workbasket's title and description in the Django admin site.

paulpepper-trade:
# TP2000-284 Save WorkBasket titles in Django admin

## Why
As per title. Currently only WorkBasket state transitions are saved. This is a pain, in particular because our new process for coordinating work between TMs and engineers requires transitioning WorkBaskets through various states. Having no way of recording the current intent of a WorkBasket (say when it has been temporarily archived) on the object itself makes this process error-prone.

## What

Save WorkBasket data when `WorkBasketAdmin.save_data()` is called.

- Requires migrations? No


simonwo (approved):
Wow. Does it need a test?

stuaxo (approved):
This seems like this must have been a bug.

paulpepper-trade:
> Wow. Does it need a test?

Unit tests: TP2000-287

PR #544, opened 2022-04-29
TP2000 290 measure edit conditions form doesn't display correct initial data

gabelton:
# TP2000-290 measure edit conditions form doesn't display correct initial data

## Why
- The measure edit form conditions section has required fields that should be pre-populated when conditions already exist.

- The with_reference_price_string qs method should return a string that passes validation in validate_duties because it is used to pre-populate the reference_price field on the conditions form and, therefore, the initial data should be valid.

## What
- Borrows logic from with_duty_sentence_string and updates with_reference_price_string, to ensure that a valid duty string is returned.
- Updates MeasureUpdate view to use foreign key pk when populating initial data


measures/querysets.py
```diff
@@ -255,8 +255,9 @@ def with_reference_price_string(self):
                     "duty_amount",
                     Case(
                         When(
-                            monetary_unit__isnull=True,
-                            then=Value(""),
+                            monetary_unit=None,
```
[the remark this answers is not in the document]
gabelton: We have some and I will do that, but it's a little bit terse. I'm going to do a reread of that enormous teams chat as part of this work because I think there's still info hanging around there that could be added here or in the data manual itself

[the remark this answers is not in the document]
SteveCN7: The rule of thumb with reference price is that it can be either a monetary value in £ and/or unit of measurement or a percentage so very similar to duties eg, £33/100 ltrs or something like that. Where it differs from duties is that it can't be a compound value, eg £33/100 ltrs + 3.5%
I don't know if that helps or confuses things here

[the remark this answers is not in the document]
gabelton: I think they also differ in one other respect, which is that it should be possible to enter "kg" as a value for duties, but not for a condition reference price. I'm looking in measures/tests/conftest.py where we have `reversible_duty_sentence_data`. I think for condition reference price we can have, to use the terms in reversible_duty_sentence_data, ad valorem, simple specific duty, unit with qualifier, and monetary unit without measurement. I don't think we want a non amount expression because it's compound and I don't think we want a supplementary unit (e.g. "kg") because that doesn't make sense as a point of reference for a condition.

measures/querysets.py
```diff
@@ -245,6 +245,21 @@ def with_effective_valid_between(self):
...
+        """
+        Returns a MeasureCondition queryset annotated with
+        ``reference_price_string`` query expression.
+
+        This expression should evaluate to a valid reference price string (https://uktrade.github.io/tariff-data-manual/documentation/data-structures/measure-conditions.html#condition-codes)
+        If a condition has no duty_amount value, then this expression evaluates to an empty string value ("").
+        Else it returns the result of three Case() expressions chained together.
+        The first Case() expression evaluates to "%" if the condition has a duty amount and no monetary unit,
+        else " ".
+        The second evaluates to "" when a condition has no condition_measurement or its measuement has no measurement unit or the measurement unit has no abbreviation,
```
paulpepper-trade, line 257: I guess our precommit linting checks aren't picking up these long lines?

paulpepper-trade (approved):
LGTM

measures/querysets.py
gabelton: That is a bit weird. Is it just the url making it longer than usual or are the other lines also too long?

paulpepper-trade: That line and another one, plus the one containing the URL (perhaps put the URL on a separate line?).

measures/querysets.py
```diff
@@ -265,23 +281,37 @@ def with_reference_price_string(self):
...
                     Case(
                         When(
-                            condition_measurement__measurement_unit__code__isnull=True,
+                            Q(condition_measurement=None)
+                            | Q(
+                                condition_measurement__measurement_unit=None,
+                            )
+                            | Q(
+                                condition_measurement__measurement_unit__abbreviation=None,
+                            ),
```
stuaxo, line 295: Maybe these should these be using `...__isnull=True` instead of `...=None` ?

stuaxo (approved):

measures/querysets.py
gabelton: Defo! I thought I'd updated all of those. Thanks!

PR #545, opened 2022-04-29
WIP - Move Geographical Areas question to separate page in the Measures Creation Wizard 

LaurenQurashi:
# TP-??? Your PR title here

## What
This PR is not finished!

This PR moves the Geographical Area question from the details step, to its own new step just after the details step. The work has been completed to move this question out, however, there's a particular test causing me grief on this. 

the `test_measure_form_wizard_finish` test checks that the submission of a measure has worked. In particular,  within this test, there's one assert that checks for a 302 status code. On master, this passes, however in this particular branch, the test fails as the status code comes back as a 200. If you simply just change the assert to check for a 200, it'll fail on a rerun as it comes back as a 302. This is the weirdest thing I've ever seen. I'm sure there's something flakey about it, but it seems consistent with reruns that if the assert is looking for one code, the response will provide the other 🤯

I've also tried just deleting the test, but then the next assert within the test fails as for the geographical area step, the `valid_user_client.post()` response doesn't return a URL like it does with the others. I've added in the geographical area step's data in the Wizard data object just like all the others, but maybe there's something extra I need to set up in order for one to come back. 

sorry for leaving this to someone else while I'm on holiday! :(


eadpearce (approved):
lgtm!

PR #547, opened 2022-04-29
TP2000 275 fix duty validation

gabelton:
# TP2000-275 fix duty validation

## Why
The duty validation is not working on either commodities or conditions steps of the measure create wizard. This is because the `MeasureCreateWizard.get_form_initial` method isn't passing measure_start_date variable through to the formsets on these steps, so the `validate_duties` method never gets called.

## What
- Overrides get_form_kwargs instead of get_form_initial
- Separates the formset used on the measure create wizard conditions step from that used on the measure edit page and creates a mixin for shared logic
- Adds more view tests and updates form tests


measures/tests/test_forms.py
```diff
@@ -285,4 +418,37 @@ def test_measure_forms_conditions_clears_unneeded_certificate(date_ranges):
...
+    )
+    form_expects_certificate.is_valid()
+    assert form_expects_certificate.cleaned_data["required_certificate"] == certificate
+
+    form_expects_no_certificate = forms.MeasureConditionsWizardStepForm(
+        dict(data, **{"condition_code": code_without_certificate.pk}),
+        prefix="",
+        measure_start_date=date_ranges.normal,
+    )
+    form_expects_no_certificate.is_valid()
```
stuaxo, line 453: ^ while we're here is it worth asserting on the value of is_valid, to ensure our test is running as expected ?

stuaxo (approved):

PR #548, opened 2022-05-05
Geo areas form and measure edit form very slow 

gabelton:
# TP2000-299 Geo areas form and measure edit form very slow 

## Why

Loading the geo areas step of the wizard or the measure edit form currently takes ~8 seconds. With this change it takes about 1 or 2 secs

## What

Adds annotation to geo area queries, so that we no longer have to use the `structure_description` property, which runs over 300 queries when calling `descriptions.last()`


eadpearce (approved):
"14 queries in 30.37 ms" 🤯

PR #549, opened 2022-05-06
Tp2000 278  WorkBasket list and detail views

paulpepper-trade:
# TP2000-278 WorkBasket list and detail views

## Why

There's no provision to allow WorkBaskets, other than the one that is currently in editing state, to be viewed and reviewed. This is a problem both for:

- Managing the operations process, which requires tariff manages and engineers accurately communicating WorkBasket status.
- QA and peer review of changes in non-editing workbaskets

## What

Added workbasket list view with specific workbasket selection to view details. The details view provides top-level info to assist with operations tasks (workbasket state changes, exports, etc) and the changes assists QA and peer reviewing.

- Requires migrations? No


gabelton (commented):
Looks good! Will need to have a reread, but could you add a couple of screenshots?

gabelton (approved):
Is it worth creating a separate ticket for adding some view tests? In the meantime I guess @SteveCN7 can try to do some manual user testing in uat.

PR #555, opened 2022-05-10
Add ARCHIVED Transaction partition.

stuaxo:
Archived workbaskets were getting checked by the business rules, adding a lot of extra work.

This PR attempts to solve the problem by creating ARCHIVED Transactions, and making sure that when workbaskets are set to archived their transactions are too.

This PR also updates the method that defrags workbaskets on publishing to use bulk_update, which should be a use less queries.

apply_transaction_order:
- Rename to move_to_end_of_partition.
- Now updates partition (as this didn't make sense as a separate operation.
- Takes partition, not partition update schemea.
- Use bulk_update.

preorder_negative_transactions:
- Use bulk_update

approved_transactions:
- != DRAFT was also including ARCHIVED transactions, filter them out.

# TP-??? Your PR title here


stuaxo:
> This looks like a lot of work and is adding complexity – why not just exclude transactions from archived workbaskets in https://github.com/uktrade/tamato/blob/master/checks/tasks.py#L149?

You're right that excluding workbaskets fixes this, I'm going to push a separate PR for it.
Initially we tried this, but tripped up on one of the APIs and it looked like it wasn't working - probably as we were new to them and fixing them during a time critical period.

This does expose some complexity, I think it's the right thing to do, but I need to finish off the original partitions ADR (and hopefully I don't ret-con too much into it!), and then do a small ADR for this.

paulpepper-trade (commented):
> ## Why
> 
> Archived workbaskets were getting checked by the business rules, adding a lot of extra work.

Has the why changed since?

> You're right that excluding workbaskets fixes this, I'm going to push a separate PR for it.
> Initially we tried this, but tripped up on one of the APIs and it looked like it wasn't working - probably as we were new to them and fixing them during a time critical period.
> 
> This does expose some complexity, I think it's the right thing to do, but I need to finish off the original partitions ADR (and hopefully I don't ret-con too much into it!), and then do a small ADR for this.

Business rules are performed when particular actions occur (for instance, when a workbasket changes from one state to another), so perhaps this could be solved by filtering on specific events? (more or less the same thing as Simon's remark).

Are there other things this PR is attempting to fix though?

stuaxo:
> > ## Why
> > Archived workbaskets were getting checked by the business rules, adding a lot of extra work.
> 
> Has the why changed since...?
> 
Good spot!   I believe it has (and I need to write it up) - if anything I should repurpose the ticket behind this and make a new one for the small fix.

> Business rules are performed when particular actions occur (for instance, when a workbasket changes from one state to another), so perhaps this could be solved by filtering on specific events? (more or less the same thing as Simon's remark).
> 
Think we can chat more on how that would work, the workbasket filter thing seems like a start.

> Are there other things this PR is attempting to fix though?

1. - There's a conceptual / neatness thing (I should write the first ADR for partitioning, then one for this as a follow up).

3.  - There's a performance improvement around moving transitions from draft to the end of another partition, this could be split out it's a pre-requisite to a tweak around how "pre-emptive transactions work".

stuaxo:
> This looks like a lot of work and is adding complexity ...

The refactorings I originally put in here (since separated out and merged) were probably more important than adding ARCHIVED, and since they were separated out and merged not much complexity remains.

So even though I'm not as wedded to this as I was, I'll put some thoughts down to get them out of my head :)


One thing transaction partitions are - is a denormalisation of workbasket status (in effect grouping certain statuses together).

In this world, EDITING/SENT/ERRORED workbaskets are all kinds of editing state, while REVISION/SEED are frozen (ideally immutable), after editing - these groups are reflected by the seed partitions.
ARCHIVED workbaskets are is a new, 3rd kind of status, outside of the normal user workflow.

> – why not just exclude transactions from archived workbaskets in https://github.com/uktrade/tamato/blob/master/checks/tasks.py#L149?

Transaction partitions enabled most checks to be achieved in the direction of workbasket->transaction, and not the other way round, untangling a bunch of code.

By having to check if the workbasket is ARCHIVED, that isn't as doable any more, adding ARCHIVED (or some better name) enables that again.

It certainly is possible to put the check in there, though it's not the only place - having an a partition just for this stuff keeps it separate.

stuaxo:
Closing as this is (even less!) unlikely to get merged without me being here.

PR #556, opened 2022-05-11
TP2000-305 Multiple objects returned error on measure edit form

gabelton:
# TP2000-305 Multiple objects returned error on measure edit form

## Why

The measure edit form is broken because the queryset returned by `with_description_string` can contain duplicates. Incidentally this also means that the dropdown displays aren't alphabetised. 

## What

Using `order_by`, as we do already on the measure create geo areas step, implicitly makes the queryset distinct. I've also excluded geo_areas with null value description fields on their description objects, as these are not active descriptions and create blank space on the dropdown.


geo_areas/util.py
```diff
@@ -4,14 +4,8 @@
 def with_description_string(qs):
     """If a description object has no description value, return an empty string,
     else return description value."""
-    return qs.annotate(
```
stuaxo, line 7: This breaks the convention of the other "with_" methods[1] that add an annotation by filtering as well, consider breaking the filter out from the annotate.

The docstring could probably mention that this annotates.

[1]
(OT: I apologise for calling the OG one with_ when the convention should probably be just annotate_)

stuaxo (approved):
Approved in advance, just a couple of small things - ace query building :)

PR #557, opened 2022-05-12
TP2000-306 Tests failing when MeasureAction codes with > 3 digits were generated

stuaxo:
Side effect:  
Users should **unset** `DJANGO_SETTINGS_MODULE=settings.dev` and ensure they have `ENV=dev` in their `.env`, appologies, this is the opposite of a previous PR, but that broke some of the things this fixes.

Test fixes
=======
Various test fixes, not quite sure how I triggered all these bugs ... but this was blocking me from running tests.

MeasureActionFactory code too long:

Running tests can run the MeasureActionFactory > 1000 times, even though there are only a few MeasureActions in any one test, this caused the sequence to hit 1000 and an illegal code with 4 digits was generated, this is fixed by wrapping round every 1000 iteations.

Upgrade pytest and friends.
Pytest and pytest-* dependencies are all upgraded.
pytest-bdd was only upgrade to 4.1.0 since 5.x caused an error in the bdd tests "step not found".

Fix bug in manage.py
An earlier fix I made, required using DJANGO_SETTINGS_MODULE, but that actually broke tests, this undoes that by making manage.py aware of ENV=dev

Fix logging bug in util/common/xml
Drive-by fix for a logging issue here, where transactions were meant to be logged, but weren't

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,7 @@ run: collectstatic migrate
 test-fast:
 	@echo
 	@echo "> Running tests..."
-	@echo ${PYTHON} -m pytest -x -n=auto --dist=loadfile --pdb
-	@${PYTHON} -m pytest -x -n=auto --dist=loadfile -s --pdb
+	@${PYTHON} -m pytest -x -n=auto --dist=loadfile
 
 test:
 	@echo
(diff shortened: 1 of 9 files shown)


gabelton (approved):
Nice one.

PR #559, opened 2022-05-13
TP2000 140 the edit measures journey doesn't support editing measure conditions

gabelton:
# TP2000-140 the edit measures journey doesn't support editing measure conditions

## Why

The current logic for editing existing conditions on a measure via the edit form involves deleting all existing conditions on the measure and then creating a set of new conditions with the user's changes applied. This creates more transactions than necessary and means that conditions are created with an `update_status` of "CREATE" when they should show "UPDATE". Similarly condition components are always created when sometimes an existing component should be updated.

## What

- Moves the condition creation logic out of `MeasureForm` and into `MeasureUpdate` view, adding logic to differentiate between an update or a create when adding conditions.
- Moves `diff_components` out of `Measure` model and into measures/util.py, updating it to handle ConditionComponent as well as MeasureComponent
- Updates view tests to check for transactions and `update_status`


measures/patterns.py
```diff
@@ -299,16 +302,21 @@ def create_condition_and_components(
                 "condition_measurement",
             ),
         )
+        if data.get("version_group"):
+            condition.version_group = data.get("version_group")
```
eadpearce, line 306: instead of using an if statement would this work? `condition.version_group = data.get("version_group", None)`

gabelton: I don't think so because we never want to set version_group to None and it's not always in the data

measures/util.py
```diff
@@ -15,3 +21,55 @@ def convert_eur_to_gbp(amount: str, eur_gbp_conversion_rate: float) -> str:
         / 100
     )
     return f"{converted_amount:.3f}"
+
+
+def diff_components(
```
eadpearce, line 26: can these be keyword args instead? would make it a bit more explicit which thing is which when the function is called. and it wouldn't matter if the order is different

gabelton: I've updated. Is that what you meant? Don't know if I can add defaults for the others

measures/views.py
```diff
@@ -375,17 +377,87 @@ def get_context_data(self, **kwargs):
...
+        """
+        formset = self.get_context_data()["conditions_formset"]
+        excluded_sids = []
+        conditions_data = []
+        workbasket = WorkBasket.current(self.request)
+        existing_conditions = obj.conditions.approved_up_to_transaction(
+            workbasket.get_current_transaction(self.request),
+        )
+
+        for f in formset.forms:
```
stuaxo, line 404: I guess the is_valid is a sanity check ?

measures/views.py
```diff
@@ -375,17 +377,87 @@ def get_context_data(self, **kwargs):
...
+                start=1,
+            ):
+                # Create conditions and measure condition components, using instance as `dependent_measure`
+                measure_creation_pattern.create_condition_and_components(
+                    condition_data,
+                    component_sequence_number,
+                    obj,
+                    parser,
+                    workbasket,
+                )
```
stuaxo, line 455: ^ Should this be inside MeaureCreationPattern itself ?

It already has some code to create some kinds of measure conditions, so maybe this should be there too ?

measures/views.py
```diff
@@ -375,17 +377,87 @@ def get_context_data(self, **kwargs):
...
+
     def get_result_object(self, form):
         obj = super().get_result_object(form)
         form.instance = obj
+        self.create_conditions(obj)
```
stuaxo, line 460: ^ This isn't not going to create more conditions if you keep resubmitting/refreshing the form is it ?

gabelton: It definitely isn't called when refreshing the measure edit form page

measures/views.py
stuaxo: Doh, mis-read this, LGTM.

PR #561, opened 2022-05-16
TP2000-309 GeographicalArea.MultipleObjectsReturned error on measure edit form

gabelton:
# TP2000-309 GeographicalArea.MultipleObjectsReturned error on measure edit form

## Why

Currently `with_description_string` returns a queryset with duplicate geographical areas where a geographical area has more than one associated description. The old logic which used `structure_description` and was [replaced in this PR](https://github.com/uktrade/tamato/pull/548) accounted for this by calling `descriptions.last()` but this was really expensive in terms of SQL queries and made the page too slow. 

## What

- Renames `with_description_string` to `with_latest_description_string` and adds filtering logic to return only the latest descriptions by `validity_start_date`. (I know we don't normally annotate and filter in `with` methods, but I think the annotation by itself isn't something we want anywhere)
- Adds test for this multiple description scenario
- Updates edit form geo area querysets to match create wizard querysets

### DIFF
--- a/geo_areas/forms.py
+++ b/geo_areas/forms.py
@@ -7,7 +7,7 @@ from common.forms import CreateDescriptionForm
 from common.forms import delete_form_for
 from geo_areas.models import GeographicalArea
 from geo_areas.models import GeographicalAreaDescription
-from geo_areas.util import with_description_string
+from geo_areas.util import with_latest_description_string
 from geo_areas.validators import AreaCode
 
 
@@ -75,7 +75,7 @@ class GeographicalAreaFormMixin(forms.Form):
         self.transaction = tx
         super().__init__(*args, **kwargs)
 
-        self.fields["geo_group"].queryset = with_description_string(
+        self.fields["geo_group"].queryset = with_latest_description_string(
             GeographicalArea.objects.filter(
                 area_code=AreaCode.GROUP,
             )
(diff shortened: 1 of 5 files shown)


geo_areas/forms.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, *args, **kwargs):
         # self.fields[
         #     "geo_group_exclusions"
         # ].queryset = GeographicalArea.objects.approved_up_to_transaction(tx)
-        self.fields["geo_area"].queryset = with_description_string(
```
stuaxo, line 93: ^ some commented code still here.

geo_areas/util.py
```diff
@@ -1,11 +1,15 @@
...
-def with_description_string(qs):
+def with_latest_description_string(qs):
     """Returns a queryset annotated with description object's description field
     value."""
-    return qs.annotate(
-        description=models.F(
-            "descriptions__description",
-        ),
+    return (
+        qs.annotate(latest_date=models.Max("descriptions__validity_start"))
```
paulpepper-trade, line 8: Because `latest_date` is related to the description start date, could the annotated name better reflect this? `latest_date_of_description` or similar (less wordy!).

geo_areas/forms.py
gabelton: I think that predates me and might be worth leaving in, as I believe it will be uncommented as part of the geo area exclusions epic

PR #562, opened 2022-05-16
Select up to 2 countries/regions for geographical area

eadpearce:
# TP2000-298 Select up to 2 countries/regions for geographical area

## Why
* Make measure creation process faster for creating measures with similar commodity codes for multiple countries

## What
* Adds a nested GeoAreaFormSet to MeasureGeographicalAreaForm
* GeographicalAreaMixin and GeographicalAreaSelect have been removed
* Moves form fields out of the GeographicalAreaMixin into the MeasureGeographicalAreaForm
* Adds a custom template for the geographical area step of the create measure wizard
* Lists the countries added on the review screen

Screenshot 2022-05-16 at 15 06 10

## Checklist
- Requires migrations? No


measures/forms.py
```diff
@@ -620,35 +621,154 @@ def clean(self):
...
+                    )
+                    .erga_omnes()
+                    .get()
+                )
+            ]
+            cleaned_data["geo_area_exclusions"] = erga_omnes_exclusions
+
+        self.fields["geo_area_type"].initial = geo_area_type
+
+        # Don't try to validate the whole form when user clicks add or delete on the country subform
```
gabelton, line 747: Is this logic not already covered by the `FormSet` class ?

measures/forms.py
```diff
@@ -620,35 +621,154 @@ def clean(self):
...
+            .with_latest_links("descriptions")
+            .prefetch_related("descriptions")
+            .order_by("descriptions__description"),
+            # descriptions__description" should make this implicitly distinct()
+        )
+        for field in ["geo_area"]:
+            self.fields[field].label_from_instance = lambda obj: obj.description
+
+
+GeoAreaFormSet = formset_factory(
```
gabelton, line 653: Nice! I should probably have used this factory before now...

measures/forms.py
```diff
@@ -620,35 +621,154 @@ def clean(self):
...
+            cleaned_data["geo_area_exclusions"] = geo_group_exclusions
+
+        if geo_area_type == self.GeoAreaType.COUNTRY:
+            if not geo_area_list and not subform_submit:
+                raise ValidationError("One or more countries or regions is required.")
+            cleaned_data["geo_area_list"] = geo_area_list
+
+        self.fields["geo_group"].initial = geo_group.pk if geo_group else None
+
+        if not cleaned_data["geo_area_list"] and not subform_submit:
```
gabelton, line 769: It feels like these three validation errors are kind of similar. Is there any way we could combine into one generic error message?

Related question: does / should this design support the creation of a measure with one group and one single country? i.e. European Union and Iceland (I think Iceland isn't in the EU?)

measures/forms.py
gabelton: If not, then might be worth a test for when ADD / DELETE is in data

measures/forms.py
SteveCN7: In answer to the related question, not at this stage, in this scenario it would probably be better to create a new geographical area (i imagine there is one for EU and Iceland)
One to keep as an improvement if it is needed

eadpearce: They're all kinda necessary because the validation changes depending on which radio button is selected so it's kind of nested validation. Might be able to simplify it when I come to adding exclusions.

EU is actually in this country/region list so yes that is possible (at least with the database dump I have loaded locally)

measures/forms.py
eadpearce: I might have covered that in the form wizard test I updated but I will double check

eadpearce: WAIT A MINUTE. You're right. I think I can grab this logic from the nested formset so I don't have to repeat it here

gabelton (approved):
Not a blocker but perhaps you could parametrize some of the form tests and add docstrings

PR #563, opened 2022-05-16
filter archived workbaskets from business rule checks.

stuaxo:
Archived workbaskets were being included in business rule checks.

This updates requires_check to exclude archived workbaskets by default.

Not done here - is hooking up the `ignore_archived` flag to give control of this through the whole stack, but could be if desired.

### DIFF
--- a/checks/tasks.py
+++ b/checks/tasks.py
@@ -10,6 +10,7 @@ from common.models.trackedmodel import TrackedModel
 from common.models.transactions import Transaction
 from common.models.transactions import TransactionPartition
 from common.models.utils import override_current_transaction
+from workbaskets.validators import WorkflowStatus
 
 logger = logging.getLogger(__name__)
 
--- a/checks/tests/test_tasks.py
+++ b/checks/tests/test_tasks.py
@@ -3,6 +3,7 @@ from itertools import zip_longest
 from unittest import mock
 
 import pytest
+from pytest_django.asserts import assertQuerysetEqual  # type: ignore
 
 from checks import tasks
 from checks.models import TransactionCheck
@@ -10,6 +11,7 @@ from checks.tests import factories
 from checks.tests.util import assert_requires_update
 from common.models.transactions import TransactionPartition
 from common.tests import factories as common_factories
+from workbaskets.validators import WorkflowStatus
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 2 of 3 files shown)


checks/tasks.py
```diff
@@ -156,6 +157,7 @@ def update_checks(self):
             ),
         )
         .filter(partition=TransactionPartition.DRAFT)
+        .exclude(workbasket__status=WorkflowStatus.ARCHIVED)
```
gabelton, line 160: I think I'm missing something, but doesn't this make the optional filtering out of archived in checks/querysets.py redundant? Or is it possible to have archived transactions within a workbasket that isn't in archived state

checks/tests/test_tasks.py
```diff
@@ -3,13 +3,15 @@
 from unittest import mock
 
 import pytest
+from pytest_django.asserts import assertQuerysetEqual  # type: ignore
```
gabelton, line 6: Why do you need type: ignore here?

checks/tests/test_tasks.py
```diff
@@ -178,3 +180,49 @@ def test_detecting_of_transactions_to_update():
...
+    assert_requires_update(revision_update, True)
+
+    # Set transactions workbasket status to ARCHIVED and verify
+    # requires_update returns transaction checks pointing to them
+    # only if include_archived is True
+    draft_update.transaction.workbasket.status = WorkflowStatus.ARCHIVED
+    revision_update.transaction.workbasket.status = WorkflowStatus.ARCHIVED
+
+    draft_update.transaction.workbasket.save()
+    revision_update.transaction.workbasket.save()
```
gabelton, line 214: Is it overkill to do an `assert_requires_update( , False)` on these two transactions?

gabelton:
Can this PR be closed now ? https://github.com/uktrade/tamato/pull/555

checks/tasks.py
simonwo: Yes, I had the same question.

checks/tests/test_tasks.py
simonwo: Because the function is dynamically generated by the library, and hence doesn't appear to the type hinter

checks/tasks.py
stuaxo: Good catch, this is from the first go I had at this, I will remove it.

checks/tests/test_tasks.py
stuaxo: Seems like pytest_django should ship the some file for this.

checks/tests/test_tasks.py
stuaxo: Might have got the wrong end of the stick, will simplify the test first...

stuaxo:
> Can this PR be closed now ? #555

I think it is the right thing, but I should write an ADR (and maybe be told it still isn't and then close it).

checks/tests/test_tasks.py
stuaxo: Simplified the test, see what you think now.

PR #564, opened 2022-05-17
Add created_at and updated_at timestamps.

paulpepper-trade:
# TP2000-314

## Why

In an ops support environment it isn't possible to see when checks were performed.

## What

Add timestamps to `TransactionCheck` and `TrackedModelCheck` models by inheriting from `common.models.mixins.TimestampedMixin`.

## Checklist
- Requires migrations? Yes


checks/migrations/0002_auto_20220517_1331.py
```diff
@@ -0,0 +1,44 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name="trackedmodelcheck",
+            name="created_at",
+            field=models.DateTimeField(
+                auto_now_add=True,
+                default=datetime.datetime(1970, 1, 1, 0, 0),
+            ),
+            preserve_default=False,
```
gabelton, line 23: I don't quite understand the docs on [preserve_default](https://docs.djangoproject.com/en/4.0/ref/migration-operations/#addfield) . If we inspect a check after this migration will it have a created_at value of 1970 or None ?

paulpepper-trade: I'm not sure either, but let me have a punt... does it perhaps mean that other, future migrations can change this value? I think that's what's meant mainly because of "... or if it is temporary and just for **this** migration (False)" - my emboldening for emphasis.

`TimestampedMixin.created_at` and `TimestampedMixin.updated_at` leave their `DateTimeField.null` attr set to the default value, `False`. They'll have a value set to epoch start.

PR #565, opened 2022-05-18
Enable CELERYD_POOL_RESTARTS so we can restart celery workers.

stuaxo:
This should let use restart the workers, which will help with getting things unstuck.


settings/common.py
```diff
@@ -348,6 +348,7 @@
 CELERY_ACCEPT_CONTENT = ["json"]
 CELERY_TASK_SERIALIZER = "json"
 CELERY_TIMEZONE = TIME_ZONE
+CELERYD_POOL_RESTARTS = True  # Restart worker if it dies
```
gabelton, line 351: --> CELERY_POOL_RESTARTS

gabelton (approved):

PR #566, opened 2022-05-18
Celery: Use -0 fair on the worker, which is better when there is a mi…

stuaxo:
Adds -0fair, it also seems the older PR used an old version of the worker setting.

Testing locally always succeeds on restarting the worker making testing tricky.


TomMacca (approved):
Looks good to me Stuart

PR #568, opened 2022-05-20
TP2000 242 build create certificate form

LaurenQurashi:
# TP2000-242 Build create certificate form

## Why
Currently, the Tariff Managers don't have the ability to make new certificates, however we think that's a bit silly. This PR adds a Create Certificates form, and a link on the home page, to allow TM's to make new certificates. 

## What
A new form has been added to the tool, that allows TM's to create their own certificates. It's a very similar process to footnotes, just with certificates instead. Once a certificate has been created, you get a nice looking success message like you do with footnotes, but beyond that nothing else has changed. 

here's a photo for you to enjoy: 

Screenshot 2022-05-20 at 13 49 08

Screenshot 2022-05-20 at 13 49 33

!! psa to anyone reviewing this - please let me know if I've missed out on any tests. I've copied the setup for footnotes, but it feels a bit little. If there's any other tests I could add, or any types of tests I've missed, please let me know!!

## Checklist
- Requires migrations? - no

### DIFF
--- a/certificates/forms.py
+++ b/certificates/forms.py
@@ -68,15 +68,11 @@ class CertificateCreateForm(ValidityPeriodForm):
     def save(self, commit=True):
         instance = super(CertificateCreateForm, self).save(commit=False)
 
-        workbasket = WorkBasket.current(self.request)
-        tx = None
-        if workbasket:
-            tx = workbasket.transactions.order_by("order").last()
-
+        current_transaction = WorkBasket.get_current_transaction(self.request)
         instance.sid = get_next_id(
             models.Certificate.objects.filter(
                 certificate_type__sid=instance.certificate_type.sid,
-            ).approved_up_to_transaction(tx),
+            ).approved_up_to_transaction(current_transaction),
             instance._meta.get_field("sid"),
             max_len=3,
         )
--- a/certificates/forms.py
+++ b/certificates/forms.py
@@ -52,21 +52,13 @@ class CertificateCreateForm(ValidityPeriodForm):
             Submit("submit", "Save"),
         )
 
-    def clean(self):
-        cleaned_data = super().clean()
-
-        if self.errors:
-            return cleaned_data
-
-        cleaned_data["certificate_description"] = models.CertificateDescription(
-            description=cleaned_data["description"],
-            validity_start=cleaned_data["valid_between"].lower,
-        )
-
-        return cleaned_data
-
     def save(self, commit=True):
         instance = super(CertificateCreateForm, self).save(commit=False)
+
+        self.cleaned_data["certificate_description"] = models.CertificateDescription(
+            description=self.cleaned_data["description"],
+            validity_start=self.cleaned_data["valid_between"].lower,
+        )
 
         current_transaction = WorkBasket.get_current_transaction(self.request)
         instance.sid = get_next_id(
--- a/certificates/forms.py
+++ b/certificates/forms.py
@@ -22,14 +22,14 @@ class CertificateCreateForm(ValidityPeriodForm):
 
     certificate_type = forms.ModelChoiceField(
         label="Certificate type",
-        help_text="Some sort of help text will go here",
+        help_text="Selecting the right certificate type will determine whether it can be associated with measures, commodity codes, or both",
         queryset=models.CertificateType.objects.latest_approved(),
         empty_label="Select a certificate type",
     )
 
     description = forms.CharField(
         label="Certificate description",
-        help_text="Some sort of help text will go here.",
+        help_text="You may enter HTML formatting if required. See the guide below for more information.",
         widget=forms.Textarea,
     )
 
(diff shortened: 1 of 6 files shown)


gabelton (commented):
Looks really good. I think perhaps you could add some view tests? Maybe you could look at test_views.py in additional_codes and footnotes for some comparison

certificates/forms.py
```diff
@@ -10,11 +10,88 @@
...
 from common.forms import ValidityPeriodForm
 from common.forms import delete_form_for
+from common.util import get_next_id
+from workbaskets.models import WorkBasket
+
+
+class CertificateCreateForm(ValidityPeriodForm):
+    """The form for creating a new certificate."""
+
+    certificate_type = forms.ModelChoiceField(
```
gabelton, line 23: I can't remember what the design for this bit looked like. Did we expect AutoCompleteField to be used here?

certificates/forms.py
```diff
@@ -10,11 +10,88 @@
...
+        label="Certificate type",
+        help_text="Selecting the right certificate type will determine whether it can be associated with measures, commodity codes, or both",
```
gabelton, line 25: Out of scope for this ticket, but it would be cool to add some docstrings and / or links to data manual in the certificate models file, as I'd kind of forgotten the relevance of some of them and the different places they're used

certificates/forms.py
```diff
@@ -10,11 +10,88 @@
...
+            Submit("submit", "Save"),
+        )
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if self.errors:
+            return cleaned_data
+
+        cleaned_data["certificate_description"] = models.CertificateDescription(
```
gabelton, line 61: It would be good if the view test or another form test could check that this description is created and saved. I'm struggling to read the second form test - does that test this logic?

certificates/forms.py
```diff
@@ -10,11 +10,88 @@
...
+            description=cleaned_data["description"],
+            validity_start=cleaned_data["valid_between"].lower,
+        )
+
+        return cleaned_data
+
+    def save(self, commit=True):
+        instance = super(CertificateCreateForm, self).save(commit=False)
+
+        workbasket = WorkBasket.current(self.request)
```
gabelton, line 71: I would use `WorkBasket.get_current_transaction(self.request)` here

certificates/forms.py
```diff
@@ -10,11 +10,88 @@
...
+        tx = None
+        if workbasket:
+            tx = workbasket.transactions.order_by("order").last()
+
+        instance.sid = get_next_id(
```
gabelton, line 76: Do we definitely need to use this method? Does it not get auto-generated?

certificates/tests/test_forms.py
```diff
@@ -0,0 +1,76 @@
...
+from common.tests.util import validity_period_post_data
+from common.util import TaricDateRange
+
+pytestmark = pytest.mark.django_db
+
+
+def test_form_save_creates_new_certificate_sid_and_certificate_type_sid_combo(
+    session_with_workbasket,
+):
+    """Tests that two certificates of the same type are created with different
```
gabelton, line 22: Perhaps you could add an assertion that the new sid == "002" ?

certificates/forms.py
gabelton: @SteveCN7 Do we have any tickets / epics that cover this?

SteveCN7: https://uktrade.github.io/tariff-data-manual/documentation/data-structures/certificates.html#certificates

https://uktrade.github.io/tariff-data-manual/documentation/data-structures/measures.html#measures

SteveCN7: I'm inclined not to explore this much further really as I think the text will make users make sure that they're using the correct certificate type.

certificates/views.py
```diff
@@ -67,6 +70,38 @@ class CertificateList(CertificateMixin, TamatoListView):
...
+        self.object.update_type = UpdateType.CREATE
+        self.object.transaction = transaction
+        self.object.save()
+
+        description = form.cleaned_data["certificate_description"]
+        description.described_certificate = self.object
+        description.update_type = UpdateType.CREATE
+        description.transaction = transaction
+        description.save()
+        return HttpResponseRedirect(self.get_success_url())
```
eadpearce, line 93: Instead of calling the redirect here I think you should just be able to do `super().form_valid()` because the redirect logic is already in the DraftCreateView class

certificates/tests/test_forms.py
```diff
@@ -0,0 +1,70 @@
...
+        "description": "A participation certificate",
+    }
+    form = forms.CertificateCreateForm(
+        data=certificate_b_data,
+        request=session_with_workbasket,
+    )
+    certificate_b = form.save(commit=False)
+
+    assert certificate_a.certificate_type == certificate_b.certificate_type
+    assert certificate_a.sid != certificate_b.sid
```
gabelton, line 43: Do you think you could do one more test checking that when you create a certificate with a different type it doesn't increment the sid? i.e. check that the two certificates of different type both have an sid of 001

gabelton (approved):

PR #569, opened 2022-05-24
Add task information to logs from celery task by using celerys logger.

stuaxo:
Use the celery logger in celery tasks, adding task information to the log messages.

This adds useful information, such as the task status, which is handy for debugging.
Celery itself also provides some functionality to output these.

During local development celery logs output in the terminal the celery worker is in, and the log level is controllable from the commandline.

### DIFF
--- a/checks/tasks.py
+++ b/checks/tasks.py
@@ -1,7 +1,7 @@
-import logging
 from itertools import cycle
 
 from celery import group
+from celery.utils.log import get_task_logger
 
 from checks.checks import applicable_to
 from checks.models import TransactionCheck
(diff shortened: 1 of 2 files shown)


checks/tasks.py
```diff
@@ -10,9 +10,8 @@
 from common.models.transactions import Transaction
 from common.models.transactions import TransactionPartition
 from common.models.utils import override_current_transaction
-from workbaskets.validators import WorkflowStatus
 
-logger = logging.getLogger(__name__)
```
paulpepper-trade, line 15: Perhaps it's worth pointing out somewhere that the Celery task logger adds task-related information to the log messages? Even if it's just in the headline info of this PR?

paulpepper-trade (approved):

PR #572, opened 2022-05-24
TP2000-339 - Log transaction summaries including partiton, workbasket, workbasket status.

stuaxo:
Improve logging for transactions outputting a summary transaction and workbasket pk and status.
    
Info is fetched only during stringification so cost is only incurred if logging is enabled.

In the example below it is easy to see that transactions in archived workbaskets are being checked..
image

### DIFF
--- a/checks/tasks.py
+++ b/checks/tasks.py
@@ -65,7 +65,7 @@ def is_transaction_check_complete(check_id: int) -> bool:
 
     if check.completed:
         check.successful = not check.model_checks.filter(successful=False).exists()
-        logger.info("Completed checking transaction %s", check.transaction.id)
+        logger.info("Completed checking %s", check.transaction.summary)
 
     check.save()
     return check.completed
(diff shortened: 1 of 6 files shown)


stuaxo:
Probably going for a more compact format, with  pk, status  next to each other:

image

It strikes me looking at this, that these fields work just like "identifying_fields" in TrackedModel, maybe Transaction and Workbasket should have `identifying_fields = ("partition", "order")` and `identifying_fields = ("pk", "status")`  - though that's a little off topic.

checks/tasks.py
```diff
@@ -108,15 +108,14 @@ def check_transaction(self, transaction_id: int):
     check, model_ids = setup_or_resume_transaction_check(transaction)
     if check.completed and not any(model_ids):
         logger.debug(
-            "Skipping check of transaction %s "
-            "because an up-to-date check already exists",
-            transaction_id,
+            "Skipping check of %s because an up-to-date check already exists",
```
paulpepper-trade, line 111: Do we prefer f-strings for string formatting?

paulpepper-trade (approved):

checks/tasks.py
stuaxo: In general, but the python logging module is "special" and uses the original %s fomat so that the formatting can be avoided if you aren't logging.

stuaxo: (which is why this has the whole rigmarole implementing these lazy string thingies)

PR #573, opened 2022-05-25
Docker tweaks

paulpepper-trade:
# Docker tweaks

## Why

* Docker build currently fails on broken APT dependencies.
* Postgres docker image is misaligned with prod.

## What

* Fix broken docker build.
* Align Postgres docker image with prod.

## Checklist
- Requires migrations? No

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM node:lts-buster-slim AS jsdeps
 
-RUN apt-get update -y
+RUN apt-get update -y --fix-missing
 RUN apt-get install -y g++ build-essential python3 libmagic1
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM node:lts-buster-slim AS jsdeps
 
-RUN apt-get update -y
+RUN apt-get update -y --fix-missing
 RUN apt-get install -y g++ build-essential python3 libmagic1
```
stuaxo, line 4: Not a blocker -

Is it worth combining the update and install lines with `&&` ?

stuaxo (approved):

PR #576, opened 2022-05-26
TP2000-296 MultipleObjectsReturned error on AdditionalCode detail view

gabelton:
# TP2000-296 MultipleObjectsReturned error on AdditionalCode detail view

## Why

We're seeing additional codes being created with the same sid, which breaks the detail view and leads to unique identifying fields violations.

## What

This updates the AdditionalCode create form to stop filtering by additional code type and to just always generate the highest  sid possible. We don't need to worry about the uniqueness of the code and three digit combination because this is given to us. 

 Requires migrations? No


additional_codes/forms.py
```diff
@@ -139,9 +139,9 @@ def save(self, commit=True):
         tx = WorkBasket.get_current_transaction(self.request)
 
         highest_sid = (
-            models.AdditionalCode.objects.filter(type__sid=instance.type.sid)
-            .approved_up_to_transaction(tx)
-            .aggregate(Max("sid"))["sid__max"]
+            models.AdditionalCode.objects.approved_up_to_transaction(tx).aggregate(
```
paulpepper-trade, line 142: Might this be better placed on a `AdditionalCodeQuerySet` or a class method of `AdditionalCode`?

paulpepper-trade (approved):

PR #587, opened 2022-06-01
Allow importing workbaskets of any status (so we can have ARCHIVED wo…

stuaxo:
We need to import ARCHIVED workbaskets, allow any valid workbasket status.


paulpepper-trade (approved):
LGTM

PR #589, opened 2022-06-06
Exclude checks app models from sqlite dump.

paulpepper-trade:
# TP2000-349 Sqlite failing to upload to S3

## Why

Attempts to include models in the checks app is causing a check constraint violation. `checks` model instances shouldn't be included in the Sqlite dump since it isn't part of the public data set and they can reference transactions that are not in a published state.

## What

Provide a new settings list attribute, `SQLITE_EXCLUDED_APPS`, that makes provision for excluding domain apps and models from the Sqlite export. Include the `checks` app in this new list attribute.

## Checklist
- Requires migrations? No


TomMacca:
Makes sense Paul, approved

PR #591, opened 2022-06-08
is_truthy returned True when passed None, this was unexpected and cau…

stuaxo:
Some of our settings were defaulting to On, not Off.

ENABLE_DJANGO_DEBUG_TOOLBAR and CELERY_TASK_ALWAYS_EAGER for instance.


paulpepper-trade (approved):
Nice.

stuaxo:
Holding this back until Toms work goes into the system, as ALWAYS_EAGER was somehow made commcode imports start working locally, and don't want to block that.

PR #592, opened 2022-06-09
Add some comments to the business rule checker.

stuaxo:
Add some comments to the business rule checks as I try and grok them.

Broke my own rule of not adding slightly redundant comments in one bit, but it's extra reassurance that what's happening isn't magic :magic_wand:


gabelton (approved):
Is it worth adding a test for the new caching logic?

stuaxo:
> Is it worth adding a test for the new caching logic?

Good catch - I meant to do this, and forgot, will do this before merging.

PR #594, opened 2022-06-11
Optimisations around moving transactions.

stuaxo:
Optimise apply_transaction_order and rename to move_to_end_of_partiton:

- Make it use bulk_update to massively reduce query count
- Make this a generalised "move transactions to end of any partition".
- Make it take a partition, instead of partition scheme, since at this point the partition is frozen.
In future this may allow us to remove the hack of using negative "pre-emptive" transactions.

Optimise preorder_negative_transactions by using bulk_update.

(Slightly) future proof  Transaction.approved by instead of checking != DRAFT, checking ! partition <= REVISION

Update tests for apply_transaction_order to work with move_to_end_of_partiton and work with partitions not partition schemes.

These fixes were extracted from the "archived transactions" PR - this are all the optimisations and cleanups that do not involve adding archived transactions.

Generalising and speeding up "moving transactions to the end of a partition" is a win, apart from the speed up on saving - it opens the way to a cleaner implementation of "pre-emptive partitions" - that doesn't involve negative transaction ids.

### DIFF
--- a/common/models/transactions.py
+++ b/common/models/transactions.py
@@ -23,16 +23,28 @@ class TransactionPartition(models.IntegerChoices):
     Transactions are partitioned into global groupings based on type.
 
     Within a partition, a transactions order applies, to obtain a global
-    ordering a order_by("partition", "order") must be used.
+    ordering call .order_by("partition", "order").
 
     The numbers chosen increment by "era" of transaction, starting
-    from approved transactions SEED_FILE and REVISION
+    from approved transactions SEED_FILE, REVISION
     then ending with DRAFT.
+
+    This system enables a simple filter for "approved" transactions, by checking
+    for <= HIGHEST_APPROVED_PARTITION, this is currently set to REVISION,
+    which is implemented in TransactionQuerySet.approved().
     """
 
     SEED_FILE = 1, "Seed"
     REVISION = 2, "Revision"
     DRAFT = 3, "Draft"
+
+    @classmethod
+    def get_highest_approved_partition(cls):
+        """Return the highest approved partition, this is used by
+        TransactionQuerySet.approved() to filter approved partitions using a
+        comparison operator as opposed to "in"."""
+        # This is a function, not a const as duplicate are not allowed in an Enum.
+        return cls.REVISION
 
     @classmethod
     def approved_partitions(cls):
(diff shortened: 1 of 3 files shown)


common/models/transactions.py
```diff
@@ -81,31 +92,34 @@ def preorder_negative_transactions(self) -> None:
         """
         if self.count() and self.order_by("order").first().order < 0:
             order = PREEMPTIVE_TRANSACTION_SEED
+            transactions = self.order_by("order").all()
             for tx in self.order_by("order").all():
                 order += 1
                 tx.order = order
-                tx.save()
+
+            type(self).objects.bulk_update(transactions, ["order"])
```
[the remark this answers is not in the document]
stuaxo: OK, fixed and added unit tests to appease the unit test gods (who were surely frowning on us before).

common/models/transactions.py
```diff
@@ -79,33 +90,36 @@ def preorder_negative_transactions(self) -> None:
...
-        existing_tx = self.model.objects.filter(
-            partition=partition_scheme.get_approved_partition(),
-        ).last()
+        # Ensure order of the transactions in this query to start at end of the partition.
+        existing_tx = (
+            self.model.objects.filter(
+                partition=partition,
+            )
+            .exclude(pk__in=self.values_list("pk", flat=True))
+            .last()
```
gabelton, line 117: We know that this queryset will be ordered by `order` field right?

common/tests/test_transactions.py
```diff
@@ -38,7 +39,7 @@ def test_tracked_models_are_sorted_correctly():
     order, i.e. ordered first by transaction order and then in record code
     order."""
     transaction_first = factories.TransactionFactory(
-        partition=partition_scheme.get_partition("PROPOSED"),
```
gabelton, line 41: We haven't had "PROPOSED" for a while have we? Was this just not affecting the test?

gabelton (approved):

common/models/transactions.py
stuaxo: Hmm... I think being explicit here may not be a bad thing, as far as I know two "order"s that are the same aren't an extra cost anyway.

PR #596, opened 2022-06-13
dump_workbaskets:  add options to disable splitting, and set max size.   add benchmarking.   fix auto-envelope-id.

stuaxo:
Output how long it takes to serialize envelopes.

- Add option to disable envelope splitting by size.
- Add option to specify the maximum envelope size.
- Add option to use the next available envelope id (not super useful yet as we don't use envelope upload)

<pre><font color="#5FD700">❯</font> <font color="#00AA00">python</font> <u style="text-decoration-style:single">manage.py</u> dump_transactions 000313 313 --disable-split-by-size
DIT000313.xml ✅  XML valid.  98 transactions, serialized in 17.21 seconds using 277202 bytes.

</pre>

### DIFF
--- a/common/serializers.py
+++ b/common/serializers.py
@@ -199,12 +199,19 @@ class EnvelopeSerializer:
         self,
         output: IO,
         envelope_id: int,
-        transaction_counter: Counter = counter_generator(),
         message_counter: Counter = counter_generator(),
         max_envelope_size: Optional[int] = None,
         format: str = "xml",
         newline: bool = False,
     ) -> None:
+        """
+        :param output: The output stream to write to.
+        :param envelope_id: The id of the envelope.
+        :param message_counter: A counter for the message ids.
+        :param max_envelope_size: The maximum size of an envelope, if None then no limit.
+        :param format: Format to serialize to, defaults to xml.
+        :param newline: Whether to add a newline after the envelope.
+        """
         self.output = output
         self.message_counter = message_counter
         self.envelope_id = envelope_id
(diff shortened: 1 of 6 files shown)


common/serializers.py
```diff
@@ -219,6 +227,9 @@ def __init__(
             )
         self.format = format
         self.newline = newline
+        self.depth = 0
+        self.render_started_at = None
+        self.render_time = None
```
paulpepper-trade, line 232: Could we move timing capture out of the serializer and into the caller - i.e. would higher level timings be sufficient for timing analysis. It would avoid adding attrs and functionality that's incidental to the purpose of this class, which are also spread across the serializer hierarchy.

exporter/management/commands/dump_transactions.py
```diff
@@ -56,6 +56,21 @@ def add_arguments(self, parser):
...
+        parser.add_argument(
+            "--max-envelope-size",
+            help=f"Set the maximum envelope size in bytes, defaults to settings.EXPORTER_MAXIMUM_ENVELOPE_SIZE [{settings.EXPORTER_MAXIMUM_ENVELOPE_SIZE}].",
+            type=int,
+            default=settings.EXPORTER_MAXIMUM_ENVELOPE_SIZE,
+            action="store",
+        )
+
+        parser.add_argument(
+            "--disable-split-by-size",
```
paulpepper-trade, line 68: Perhaps it's sufficient to just name this `--disable-splitting` if that's the intention - don't split the envelope, no matter what?

stuaxo: We might split at workbasket boundries in the future, though I guess we can work out if they need their own flags later on.

stuaxo:
@SteveCN7 it certainly sounds like a good idea, I think we have everything in the system to try this from our end already.

PR #598, opened 2022-06-14
workbasket_list:  add status.   add ability to list all approved statuses.

stuaxo:
Allow outputting all approved workbaskets
Add status column to compact workbasket list output.


paulpepper-trade (approved):
LGTM

PR #600, opened 2022-06-17
TP2000-225 Users cannot find some countries from geo area autocomplete or search

gabelton:
# TP2000-225 Users cannot find some countries from geo area autocomplete or search

## Why

Countries that have been updated (i.e. where more than one version exists) aren't showing when searching by description. The countries that aren't displaying in search appear in the queryset generated by`GeographicalArea.objects.filter(descriptions__isnull=True)`, even though calling `.descriptions` on the model clearly shows that descriptions are attached. The query includes the SQL `INNER JOIN "geo_areas_geographicalareadescription" ON ("geo_areas_geographicalarea"."trackedmodel_ptr_id" = "geo_areas_geographicalareadescription"."described_geographicalarea_id")`. Where a country doesn't display in search the description's described_geographicalarea_id does not match the geo area trackedmodel_ptr_id.

## What

I've created a draft adding assertions to a test Simon wrote as part of his work to fix reverse relations. The last two are failing but I think all of them should be passing and wanted to get some opinions before going further with it.

### DIFF
--- a/common/tests/test_models.py
+++ b/common/tests/test_models.py
@@ -277,6 +277,13 @@ def test_new_version_retains_related_objects(sample_model):
         sample_model.transaction.workbasket,
     )
     assert new_model.descriptions.get() == description
+    assert sample_model.descriptions.get() == description
+    assert type(sample_model).objects.filter(descriptions__isnull=False).exists()
+    assert (
+        type(sample_model).objects.filter(descriptions__isnull=False).first()
+        == new_model
+    )
+    assert not type(sample_model).objects.filter(descriptions__isnull=True).exists()
 
 
 def test_current_as_of(sample_model):


common/tests/test_models.py
```diff
@@ -277,6 +277,13 @@ def test_new_version_retains_related_objects(sample_model):
         sample_model.transaction.workbasket,
     )
     assert new_model.descriptions.get() == description
+    assert sample_model.descriptions.get() == description
+    assert type(sample_model).objects.filter(descriptions__isnull=False).exists()
```
paulpepper-trade, line 281: Would a queryset count be preferable and more complete here?
`assert type(sample_model).objects.filter(descriptions__isnull=False).count() == 2`
I'm not 100% certain, but I think that's right?

PR #604, opened 2022-06-22
TP2000-366 Edit workbaskets list page

eadpearce:
# TP2000-366 Edit workbaskets list page

## Why

* Implementing new design

## What

* Reuses the workbasket list page and makes a few modifications according to the new design
* Querystring used to change current workbasket when they navigate to the new homepage from the workbasket list
* Querystring used to alter breadcrumbs on the homepage to reflect where the user came from
* Links aren't displayed for published workbaskets because workbaskets with this status can't be set as the current workbasket

Screenshot 2022-06-22 at 16 53 37

### DIFF
--- a/common/jinja2/common/my-workbasket.jinja
+++ b/common/jinja2/common/my-workbasket.jinja
@@ -3,7 +3,6 @@
 
 {% block page_title %}Workbasket {{ request.session.workbasket.id }}{% if request.session.workbasket.title %} - {{ request.session.workbasket.title }}{% endif %}{% endblock %}
 {% block page_subtitle %}What would you like to do?{% endblock %}
-{% set page_title %}Workbasket {{ request.session.workbasket.id }}{% endset %}
 
 {% block breadcrumb %}
   {% if request.GET.edit %}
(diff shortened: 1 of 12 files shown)


SteveCN7:
Looks good Edie - I think the filters design might be a little out - Interesting about published and we might need to think about how to look those up in the design. I'll check if I've given you the right figma version, the filters at the top should match the workbasket status' in DJANGO

eadpearce:
@SteveCN7 They do match (although there are some in the system that are missing from design) but there is some preexisting backend logic that prevents a published workbasket from being selected so maybe we will want to remove this filter option

SteveCN7:
I'm so bad with GitHub Commenting by the way apologies for any faux pas

> @SteveCN7 They do match (although there are some in the system that are missing from design) but there is some preexisting backend logic that prevents a published workbasket from being selected so maybe we will want to remove this filter option
> 
> Edit: I just checked again and there isn't a "reviewed" status. The ones we have are: Archived, Editing, Proposed, Approved, Sent, Published, Errored.

The ones we have are: Archived, Editing, Proposed, Approved, Sent, Published, Errored. - This is correct and in the correct order as well :-)

PR #605, opened 2022-06-23
TP2000 365 workbasket landing form

LaurenQurashi:
# TP-2000-365 Build the new TAP landing page design

## What
This PR introduces the new landing page for TAP. When a user navigates to the site, they will now be greeted with this form that asks the user what they want to do. Eventually, the user will have 3 radio options to choose from, but for now, it only has one option which is "Edit an existing workbasket". Once the user has submitted the choice, they will be redirected to the Select workbasket page.

Screenshot 2022-06-24 at 17 08 35

Due to this new 'workbasket actions' form replacing the index page, I've renamed the old index page to 'dashboard' and updated all links, redirects, and tests that rely on the old name for the home page (aka they still go to the same place they used to, the name of the place changed).

## Why
This is being introduced as part of wider work to enable TM's to manage multiple workbaskets at the same time. This page allows them to be efficiently directed to the right place for selecting a workbasket to work on, as soon as they arrive at TAP.

## Checklist
- Requires migrations? no


common/forms.py
```diff
@@ -153,6 +154,30 @@ def get_bound_field(self, form, field_name):
         return super().get_bound_field(form, field_name)
 
 
+class WorkbasketActionForm(forms.Form):
+    class Meta:
```
eadpearce, line 158: if the form isn't a ModelForm you don't need this part

eadpearce (changes requested):
looks good! just a small comment

PR #613, opened 2022-06-29
TP2000-377 Add footnotes tab to measure detail page

eadpearce:
# TP2000-377 Add footnotes tab to measure detail page

## Why

* Users need to be able to see measures' associated footnotes on the measue detail page

## What

* Adds an extra tab to display a measure's footnotes

### DIFF
--- a/measures/jinja2/includes/measures/tabs/footnotes.jinja
+++ b/measures/jinja2/includes/measures/tabs/footnotes.jinja
@@ -2,23 +2,21 @@
 
 <h2 class="govuk-heading-l">Footnotes</h2>
 
-{% set has_footnotes = false %}
-{% set table_rows = [] %}
-{% for footnote in object.footnotes.all() %}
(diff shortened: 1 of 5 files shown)


measures/jinja2/includes/measures/tabs/footnotes.jinja
```diff
@@ -0,0 +1,24 @@
...
+{% endfor %}
+  {{ govukTable({
+  "head": [
+      {"text": "SID"},
+      {"text": "Description"},
+  ],
+  "rows": table_rows
+  }) }}
+
+{% if has_footnotes %}
```
LaurenQurashi, line 22: This isn't coming up for measures with no footnotes. It's just showing an empty table.

eadpearce: OH I just copied this from another template. and now that I check that template... you're right it doesn't actually do anything. I'll fix

PR #620, opened 2022-07-08
TP2000-386 get_current_memberships uses latest_approved instead of approved_up_to_transaction

gabelton:
# TP2000-386 get_current_memberships uses latest_approved instead of approved_up_to_transaction

## Why

At present newly created memberships don't display in the memberships tab when viewing a geo area. Separately we also don't see new geo areas in dropdown menus on the measure create or edit forms.

## What

- Have `get_current_memberships` use the `current` queryset method
- Pass in current transaction to geo area forms, so that latest editing workbasket changes are visible to the user

### DIFF
--- a/geo_areas/models.py
+++ b/geo_areas/models.py
@@ -94,7 +94,7 @@ class GeographicalArea(TrackedModel, ValidityMixin, DescribedMixin):
             GeographicalMembership.objects.filter(
                 Q(geo_group__sid=self.sid) | Q(member__sid=self.sid),
             )
-            .latest_approved()
+            .current()
             .select_related("member", "geo_group")
         )
 
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -221,6 +221,7 @@ class CountryRegionForm(forms.Form):
     )
 
     def __init__(self, *args, **kwargs):
+        # Depending on whether it is used a part of a formset or independently, transaction will be passed directly into kwargs or as part of form_kwargs
         tx = kwargs.pop("transaction", None)
         form_kwargs = kwargs.pop("form_kwargs", None)
         if form_kwargs and not tx:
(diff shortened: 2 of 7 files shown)


geo_areas/models.py
```diff
@@ -94,7 +94,7 @@ def get_current_memberships(self):
             GeographicalMembership.objects.filter(
                 Q(geo_group__sid=self.sid) | Q(member__sid=self.sid),
             )
-            .latest_approved()
+            .current()
```
paulpepper-trade, line 97: Yah - just what this new-ish method is for.

measures/forms.py
```diff
@@ -223,7 +222,9 @@ class CountryRegionForm(forms.Form):
 
     def __init__(self, *args, **kwargs):
         tx = kwargs.pop("transaction", None)
-        self.transaction = tx
+        form_kwargs = kwargs.pop("form_kwargs", None)
+        if form_kwargs and not tx:
```
paulpepper-trade, line 226: Why are there two ways of passing `transaction` to this `__init__()`? Does it need rationalising/refactoring,  a comment, a TODO, or?

paulpepper-trade (approved):

measures/forms.py
gabelton: I've added a comment, but it is a bit of a hack. Open to other suggestions but it was more fiddly than expected passing kwargs down the chain of nested forms. @eadpearce I defer to you as master of formception, if you have any other suggestions...

paulpepper-trade: I assume the `transaction` is necessary in order to pass into `TrackedModelQuerySet.approved_up_to_transaction()` or similar? If that's so, then `TrackedModelQuerySet.current()` may ultimately be the answer - `WorkBasket.get_current_transaction()` is used to set the current transaction via Middleware, but it can be overridden.
But I know we use the param-passing in various parts of the app, so there's perhaps a bit of a wider refactor looming there.

PR #622, opened 2022-07-12
Revert "TP2000-225 Fix geo area search"

gabelton:
Reverts uktrade/tamato#617

Lots of countries are missing from search (e.g. Erga Omnes 1011, North Macedonia) and the detail view returns a multiple objects returned for countries with multiple descriptions. I think filtering by transaction order works better (my bad for suggesting to change that), but, looking at it again, I'm not sure we want to be changing the main queryset in GeoAreaMixin and using `objects.all()`, as I think this is causing the multiple objects returned errored. Needs a bit more thought and probably we should separate out the filtering logic and annotation logic from with_latest_description_string.

### DIFF
--- a/geo_areas/tests/test_util.py
+++ b/geo_areas/tests/test_util.py
@@ -36,22 +36,3 @@ def test_with_latest_description_multiple_descriptions():
 
     assert qs.count() == 1
     assert later_description.description == qs.first().description
-
-
-def test_with_latest_description_multiple_descriptions_same_date():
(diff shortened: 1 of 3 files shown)


eadpearce (approved):

paulpepper-trade:
Just missed my chance to check before this was approved and merged, but is this a straightforward git-revert of commit for [TP2000-225-geo-area-search2](https://github.com/uktrade/tamato/pull/617)?

PR #624, opened 2022-07-15
TP2000-402 500 when trying to create erga omnes measure with phytosanitary group exclusion

gabelton:
# TP2000-402 500 when trying to create erga omnes measure with phytosanitary group exclusion

## Why

The [erga omnes 1011](https://www.manage-trade-tariffs.trade.gov.uk/geographical-areas/400/) group shows the latest version of Anguilla as one of its members. It was originally created, however, with an earlier version of Anguilla. The current version of Anguilla belongs to [this group](https://www.manage-trade-tariffs.trade.gov.uk/geographical-areas/514/#memberships). When we try to create a measure with erga omnes and exclude this group from it we get an error because the system sees that the latest version of Anguilla doesn't belong to erga omnes and complains that we cannot exclude a country from a group that it doesn't belong to.

We should instead adopt the same logic as `get_current_memberships` and compare using `sid` instead of `pk`.

## What

Updates assertion in `MeasureCreationPattern.create_measure_excluded_geographical_areas` and adds test for bug scenario

### DIFF
--- a/measures/patterns.py
+++ b/measures/patterns.py
@@ -228,17 +228,13 @@ class MeasureCreationPattern:
                 m.member
                 for m in GeographicalMembership.objects.as_at(
                     measure.valid_between.lower,
-                )
-                .filter(
+                ).filter(
                     geo_group=measure.geographical_area,
                 )
-                .all()
             )
-            for membership in (
-                GeographicalMembership.objects.as_at(measure.valid_between.lower)
-                .filter(geo_group=exclusion)
-                .all()
-            ):
+            for membership in GeographicalMembership.objects.as_at(
+                measure.valid_between.lower,
+            ).filter(geo_group=exclusion):
                 member = membership.member
                 assert member.sid in list(
                     m.sid for m in measure_origins
--- a/measures/patterns.py
+++ b/measures/patterns.py
@@ -240,8 +240,8 @@ class MeasureCreationPattern:
                 .all()
             ):
                 member = membership.member
-                assert (
-                    member in measure_origins
+                assert member.sid in list(
+                    m.sid for m in measure_origins
                 ), f"{member.area_id} not in {list(x.area_id for x in measure_origins)}"
                 yield MeasureExcludedGeographicalArea.objects.create(
                     modified_measure=measure,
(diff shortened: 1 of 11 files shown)


measures/patterns.py
```diff
@@ -240,8 +240,8 @@ def create_measure_excluded_geographical_areas(
                 .all()
             ):
                 member = membership.member
-                assert (
-                    member in measure_origins
+                assert member.sid in list(
+                    m.sid for m in measure_origins
```
paulpepper-trade, line 244: Checking my understanding: this was previously raising an exception (because of incorrect use of foreign key filtering rather than SID, or equivalently version_group, filtering) preventing creation of `MeasureExcludedGeographicalArea` instance?
If that's ^^ right, then are there ever circumstances where a `GeographicalMembership` is subsequently 'Deleted'?

gabelton: That's right. Looking back through the uk-tariff-data-migration repo, I don't think we ever delete a membership. We only ever end-date them. I think that's ok because in MeasureCreationPattern we use `as_at`, though we might need to tweak `get_current_memberships` because I think that will return any memberships that haven't been deleted

measures/patterns.py
```diff
@@ -240,8 +240,8 @@ def create_measure_excluded_geographical_areas(
                 .all()
             ):
                 member = membership.member
-                assert (
-                    member in measure_origins
+                assert member.sid in list(
```
paulpepper-trade, line 243: Whilst in this area, perhaps it's worth removing those two spurious `.all()` method calls chained after the `.filter()` call?

measures/patterns.py
paulpepper-trade: Right, so in addition to filtering by current transaction (via `.current()`) also filter by valid between dates. It might be tricky to get those dates in the context of its use within a template, but perhaps that's for another ticket.

paulpepper-trade (approved):

measures/patterns.py
gabelton: I think it's for another ticket. I'm trying to think what the logic should be. e.g. Erga Omnes has a start date in 1958. If we used that start date to filter then I don't think we'd see any countries display in the memberships tab because all the erga omnes countries have a later start date. I think it's probably just fine to list all memberships past and present. The user can always click on the detail view, which makes it clear if that membership if that membership has been end-dated or not

PR #627, opened 2022-07-20
TP2000-286 multiple workbaskets

paulpepper-trade:
# TP2000-286 multiple workbaskets

## Why

- Tamato currently only offers a globally "active" workbasket, presenting a process bottleneck to users.
- A system administrator is required to set the "active" workbasket.
- There are opportunities to improve the process and UI around workbasket management and review.

## What

This PR addresses the "Why" above by:

- Allowing each user to concurrently make changes to their own, active workbasket (no more shared, active workbasket, though users can choose to work on a shared workbasket if they wish).
- Providing users with the ability to activate their own workbasket for editing, removing the need for an administrator at this step.
- Reorganising, extending and refining the workbasket UI views.

A user's active workbasket is maintained in their session and can, and normally will, be unique to that user.

## Checklist
- Requires migrations? No


stuaxo:
All looks good, just a couple of comments on tests, involuntarily said "ooh I like that" to myself as reading this.

workbaskets/tests/test_views.py
```diff
@@ -2,20 +2,48 @@
...
+    valid_user_api_client,
+):
+
+    # Post a form
+    create_url = reverse("workbaskets:workbasket-ui-create")
+
+    form_data = {
+        "title": "My new workbasket",
+        "reason": "Making a new workbasket",
+    }
```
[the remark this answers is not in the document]
eadpearce: Why do we want random numbers in here?

workbaskets/jinja2/includes/workbaskets/edit-workbasket-list.jinja
```diff
@@ -0,0 +1,33 @@
+{% from "components/button/macro.njk" import govukButton %}
+{% from "components/input/macro.njk" import govukInput %}
+{% from "components/table/macro.njk" import govukTable %}
+
+{%- set table_rows = [] -%}
+{% for workbasket in object_list %}
+  {%- set workbasket_linked_pk -%}
+    <a
```
paulpepper-trade, line 8: Raised on a branch that was merged into this one I think - use of POST when creating new user session data would be preferable to avoid mishap (shared links that change active workbasket), security issues (XSS) and proxy caching (POST requests shouldn't be cached).
It would be good to include rationale for the choice of GET in the view class that handles this.

workbaskets/views/ui.py
```diff
@@ -223,3 +276,175 @@ def download_envelope(request):
...
+    template_name = "workbaskets/review-workbasket.jinja"
+
+    def dispatch(self, request, *args, **kwargs):
+        workbasket_pk = request.GET.get("workbasket")
+
+        if workbasket_pk:
+            workbasket = WorkBasket.objects.get(pk=workbasket_pk)
+
+            if workbasket:
+                workbasket.save_to_session(request.session)
```
paulpepper-trade, line 330: A comment here would be useful to explain why GET is used in place of the convention of using POST to set server-side state.

workbaskets/views/ui.py
```diff
@@ -223,3 +276,175 @@ def download_envelope(request):
...
+
+@method_decorator(require_current_workbasket, name="dispatch")
+class ReviewWorkbasketView(TemplateResponseMixin, FormMixin, View):
+    template_name = "workbaskets/review-workbasket.jinja"
+
+    def dispatch(self, request, *args, **kwargs):
+        workbasket_pk = request.GET.get("workbasket")
+
+        if workbasket_pk:
+            workbasket = WorkBasket.objects.get(pk=workbasket_pk)
```
paulpepper-trade, line 327: An `edit` param is passed in on the URL but I don't see it being used here. Is it used elsewhere or should it perhaps be removed?

eadpearce: It's used in the template to control which breadcrumbs are displayed

workbaskets/tests/test_views.py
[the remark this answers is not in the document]
stuaxo: Not necessarily random numbers but faker would be fine.   Set strings has a small risk of collisions depending on how you do concurrently, if someone makes another test with title of "My new workbasket", so it - depending on the concurrency model used when testing.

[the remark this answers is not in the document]
stuaxo: Something with faker would probably be fine, or factory_boy, maybe using one of the Workbasket factories to generate a dict (though this last one is possibly a little heavy).

eadpearce (approved):

PR #628, opened 2022-07-20
TP2000-225 Users cannot find some countries from geo area autocomplete or search

gabelton:
# TP2000-225 Users cannot find some countries from geo area autocomplete or search

## Why

Countries that have been updated (i.e. where more than one version exists) aren't showing when searching by description. The countries that aren't displaying in search appear in the queryset generated byGeographicalArea.objects.filter(descriptions__isnull=True), even though calling .descriptions on the model clearly shows that descriptions are attached. The query includes the SQL INNER JOIN "geo_areas_geographicalareadescription" ON ("geo_areas_geographicalarea"."trackedmodel_ptr_id" = "geo_areas_geographicalareadescription"."described_geographicalarea_id"). Where a country doesn't display in search the description's described_geographicalarea_id does not match the geo area trackedmodel_ptr_id.

## What

- Adds with_description_string annotation method, using subquery to annotate latest description to geo areas
- Update geo area list view and forms using geo area querysets


geo_areas/util.py
```diff
@@ -1,16 +1,18 @@
...
-    return (
-        qs.annotate(latest_description_date=models.Max("descriptions__validity_start"))
-        .filter(descriptions__validity_start=models.F("latest_description_date"))
-        .annotate(
-            description=models.F(
-                "descriptions__description",
-            ),
-        )
+
+def with_current_description(qs):
```
paulpepper-trade, line 7: Is it preferable to locate this as a method on `GeographicalAreaQuerySet`?

paulpepper-trade (changes requested):

paulpepper-trade (approved):

PR #629, opened 2022-07-26
TP2000-434 QA6 should use approved_up_to_transaction

gabelton:
# TP2000-434 QA6 should use approved_up_to_transaction

## Why

It's not currently possible to update a QuotaAssociation and change its sub_quota_relation_type without QA6 seeing the old version as clashing with the new one.
## What

Updates QA6 to use `approved_up_to_transaction` and adds test for bug scenario


paulpepper-trade (approved):
LGTM

PR #630, opened 2022-07-26
TP2000-433: Fix missing pk in review workbasket url

eadpearce:
# TP2000-433: Fix missing pk in review workbasket url

## Why

* In changing some of the urls we needed to add the workbasket pk to the url instead of the view getting it from the session and in this view that was missed

## What

* Adds the pk to the url
* Adds a test to check it works

### DIFF
--- a/workbaskets/tests/test_views.py
+++ b/workbaskets/tests/test_views.py
@@ -322,8 +322,8 @@ def test_review_workbasket_redirects(
     workbasket = factories.WorkBasketFactory.create(
         status=WorkflowStatus.EDITING,
     )
-    with factories.TransactionFactory.create(workbasket=workbasket):
-        factories.FootnoteTypeFactory.create_batch(30)
+    with workbasket.new_transaction() as tx:
+        factories.FootnoteTypeFactory.create_batch(30, transaction=tx)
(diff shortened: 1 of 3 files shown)


paulpepper-trade (approved):

workbaskets/tests/test_views.py
```diff
@@ -303,3 +303,35 @@ def test_select_workbasket_page_200(valid_user_client):
...
+)
+def test_review_workbasket_redirects(
+    form_action,
+    url_name,
+    valid_user_client,
+):
+    workbasket = factories.WorkBasketFactory.create(
+        status=WorkflowStatus.EDITING,
+    )
+    with factories.TransactionFactory.create(workbasket=workbasket):
```
gabelton, line 325: if you want to be more explicit you could do:
`with workbasket.new_transaction() as tx:` then create the footnote type with transaction=tx

gabelton (approved):

workbaskets/tests/test_views.py
eadpearce: oh cool. i was looking at some other tests to see how to do this. that way reads nicer

PR #632, opened 2022-07-27
TP2000-436: Fix use of index url that no longer exists

eadpearce:
# TP2000-436: Fix use of index url that no longer exists

## Why

* "index" url is gone. We use the workbasket main menu instead

## What

* Replaces any remaining uses of "index" in templates and makes the main menu link into a reusable template

### DIFF
--- a/workbaskets/tests/test_views.py
+++ b/workbaskets/tests/test_views.py
@@ -335,3 +335,12 @@ def test_review_workbasket_redirects(
 
     elif form_action == "page-next":
         assert "?page=3" in response.url
+
+
+def test_delete_changes_confirm_200(valid_user_client, session_workbasket):
+    url = reverse(
+        "workbaskets:workbasket-ui-delete-changes-done",
+        kwargs={"pk": session_workbasket.pk},
+    )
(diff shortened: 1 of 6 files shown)


workbaskets/tests/test_views.py
```diff
@@ -335,3 +335,12 @@ def test_review_workbasket_redirects(
...
         assert "?page=3" in response.url
+
+
+def test_delete_changes_confirm_200(valid_user_client, session_workbasket):
+    url = reverse(
+        "workbaskets:workbasket-ui-delete-changes-done",
+        kwargs={"pk": session_workbasket.pk},
+    )
+    response = valid_user_client.get(url)
+    assert response.status_code == 200
```
gabelton, line 346: It is a 200 we want here isn't it? it's not meant to do a 302 redirect?

gabelton (approved):

workbaskets/tests/test_views.py
eadpearce: Nah this is the success page after it's already done the redirect. It's just a plain template view

PR #634, opened 2022-07-28
TP2000-411 ON10 doesn't support creating an origin for a later period 

gabelton:
# TP2000-411 ON10 doesn't support creating an origin for a later period 

## Why

We want to be able to end date a measure with a quota attached. We then want to be able to copy that measure for a later start date with different exclusions. Our current implementation of ON10 and ME119 prevents this because they see the origins / measures for a later period as not contained by the origins / measures for the earlier period. Each measure should be contained by one origin, but not by every origin.

## What

Update ON10 and ME119 to reflect above logic and add scenario tests for both.


measures/business_rules.py
```diff
@@ -459,52 +457,32 @@ class ME119(ValidityPeriodContained):
...
-        return (
-            contained.with_validity_field()
-            .filter(**self.extra_filters)
-            .exclude(
-                **{
-                    f"{contained.model.validity_field_name}__contained_by": valid_between,
-                },
+        Loop over these and raise a violation if a) the measure validity period
+        is not contained by any of the origins or b) the validity_period is
+        contained by more than one origin.
```
simonwo, line 467: Does b) hold for all historical data too? Or does this rule need a "not before" property on it?

measures/business_rules.py
```diff
@@ -459,52 +457,32 @@ class ME119(ValidityPeriodContained):
...
+        """
+        with override_current_transaction(self.transaction):
+            contained_measure = measure.get_versions().current().get()
+            origins = QuotaOrderNumberOrigin.objects.current().filter(
+                order_number__order_number=measure.order_number.order_number,
             )
```
simonwo, line 473: If you wanted to, you could also use `measure.follow_path("order_number__origins")`

measures/tests/test_business_rules.py
```diff
@@ -736,6 +737,53 @@ def test_ME119(date_ranges):
...
+        order_number__origin__valid_between=valid_between,
+        valid_between=valid_between,
+    )
+    later_origin = factories.QuotaOrderNumberOriginFactory.create(
+        order_number=measure.order_number,
+        valid_between=TaricDateRange(date(2021, 1, 1), date(2021, 1, 31)),
+    )
+    from quotas.business_rules import ON10
+
+    ON10(later_origin.transaction).validate(later_origin)
```
simonwo, line 759: Should this be in the quotas business rule test file?

quotas/business_rules.py
```diff
@@ -124,6 +126,33 @@ class ON10(ValidityPeriodContained):
...
+                )
+                .exists()
+            ):
+                return True
+
+        return False
+
+    def validate(self, model):
+        with override_current_transaction(self.transaction):
+            if self.violating_models(model).exists() and not self.valid_origin(model):
```
simonwo, line 153: What does `valid_origin` check for? It seems to return `True` if any of the measures on the origin is contained by it? So I think this is finding origins that don't contain all of their measures and excluding any that don't contain any measures? What about if origin X contains measure A and origin Y contains measure B, but they don't overlap with each other? Won't `valid_origin` return `True` for that and it'll still fail?

measures/business_rules.py
gabelton: It's got `@only_applicable_after("2007-12-31")`, but I'm running a script now to check if that's the right date or if there are exceptions

measures/tests/test_business_rules.py
gabelton: I guess it tests both ON10 and ME119 and could go in either. Or did you mean that I should have the same test in both files?

measures/business_rules.py
gabelton: I think maybe this rule was buggy before because, when l checked the output of follow_path, it was returning the GeographicalArea objects rather than the QuotaOrderNumberOrigin objects themselves

measures/business_rules.py
```diff
@@ -456,6 +457,72 @@ class ME119(ValidityPeriodContained):
...
+                        m.member for m in origin.geographical_area.memberships.current()
+                    ]:
+                        origin_sids.append(a.sid)
+                else:
+                    origin_sids.append(origin.geographical_area.sid)
+
+            # Check that the geographical area sid is included in the list of origin area sids
+            for sid in area_sids:
+                if sid not in origin_sids:
+                    raise self.violation(measure)
```
simonwo, line 524: This looks like ajob for `sets`. E.g. put these things into a set by doing `area_sids = area_sids & …` and then comapre them with `if any(area_sids - origin_sids):`

measures/tests/test_business_rules.py
simonwo: I guess I'm saying that importing business rules from the quotas module as part of measure tests feels unhygenic and is tying the modules closer together… the tests should be testing things in their own module only?

quotas/business_rules.py
```diff
@@ -124,6 +126,30 @@ class ON10(ValidityPeriodContained):
...
+                contained_count = 0
+                for origin in origins:
+                    if validity_range_contains_range(
+                        origin.valid_between,
+                        measure.valid_between,
+                    ):
+                        contained_count += 1
+
+                if contained_count == 0:
+                    raise self.violation(order_number_origin)
```
simonwo, line 151: This doesn't look quite right, because you raise that the violation is with `order_number_origin` even if a different origin was at fault.

But anyway, you are checking all of the origins on the measure. You should only be checking `order_number_origin` because that was the one passed into the rule. So I don't think you need the origins loop – just check the one origin against all the measures. The reason why is that this rule will only be called if the origin itself has changed – so checking all the other origins is redundant as they haven't changed. ME119 exists to do this the other way around for when the measure changes.

quotas/tests/test_business_rules.py
```diff
@@ -164,6 +166,25 @@ def test_ON10(date_ranges):
...
+    )
+    later_origin = factories.QuotaOrderNumberOriginFactory.create(
+        order_number=measure.order_number,
+        valid_between=TaricDateRange(date(2021, 1, 1), date(2021, 1, 31)),
+    )
+
+    business_rules.ON10(later_origin.transaction).validate(later_origin)
+    from measures.business_rules import ME119
+
+    ME119(later_origin.transaction).validate(measure)
```
simonwo, line 185: Again, I think it's bad for circular referencing that we are importing stuff from measures into the quotas module. `measures` is dependent on `quotas` because a measure contains a quota, but that other way around is not true. So 'quotas' module should not be depending on 'measures'.

simonwo (changes requested):

quotas/business_rules.py
gabelton: This entire rule now seems unnecessary to me. I want to be able to create an order number origin that spans a new time period, one not yet covered by any measures. In that scenario, if I just loop over the measures, not the origins, `contained_count` will be 0. ME119 takes care of the scenario where a measure is created with a quota order number but no origin and ON5 covers me against overlapping origins for the same quota.

simonwo: I guess the rule is mainly here for if the origin date gets updated, and then no longer spans the measures it used to span. How to handle that one is indeed tricky. Maybe you do need this rule, but instead you just loop over the measures and call ME119 on each of them??

measures/business_rules.py
```diff
@@ -456,6 +457,72 @@ class ME119(ValidityPeriodContained):
     # This checks the same thing as ON10 from the other side of the relation
     container_field_name = "order_number__quotaordernumberorigin"
```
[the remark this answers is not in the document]
gabelton: I've removed it because it was returning the geo areas, rather than the origins themselves

simonwo (approved):

PR #635, opened 2022-07-29
TP2000-432 QuotaAssociation objects have been created with the same main and sub quota

gabelton:
# TP2000-432 QuotaAssociation objects have been created with the same main and sub quota

## Why

It shouldn't be possible to create an association where the same definition is both the parent and the child.

It shouldn't be possible to create overlapping definitions for the same quota order number.

Therefore, we should formalise the above as business rules.

## What

Adds two new rules, tests, and updates model business rules lists

### DIFF
--- a/quotas/business_rules.py
+++ b/quotas/business_rules.py
@@ -250,6 +250,24 @@ class QuotaBlockingPeriodMustReferToANonDeletedQuotaDefinition(
         return quota_definition.quotablocking_set.model
 
 
+class OverlappingQuotaDefinition(BusinessRule):
+    """There may be no overlap in time of two quota definitions with the same
+    quota order number id."""
+
+    def validate(self, quota_definition):
+        if (
+            type(quota_definition)
+            .objects.approved_up_to_transaction(quota_definition.transaction)
+            .filter(
+                order_number=quota_definition.order_number,
+                valid_between__overlap=quota_definition.valid_between,
+            )
+            .exclude(sid=quota_definition.sid)
+            .exists()
+        ):
+            raise self.violation(quota_definition)
+
+
 class QA1(UniqueIdentifyingFields):
     """The association between two quota definitions must be unique."""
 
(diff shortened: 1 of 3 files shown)


quotas/business_rules.py
```diff
@@ -250,6 +250,24 @@ def get_relation_model(self, quota_definition):
         return quota_definition.quotablocking_set.model
 
 
+class OverlappingQuotaDefinition(BusinessRule):
```
paulpepper-trade, line 253: Do we distinguish between TARIC business rules and Tamato's own, non-TARIC biz rules in any way? If not, is there value in doing so?

quotas/business_rules.py
```diff
@@ -250,6 +250,24 @@ def get_relation_model(self, quota_definition):
...
+    """There may be no overlap in time of two quota definitions with the same
+    quota order number id."""
+
+    def validate(self, quota_definition):
```
paulpepper-trade, line 257: Perhaps worth adding some type hinting to the params of these `validate` methods.

paulpepper-trade (approved):

quotas/business_rules.py
gabelton: The only distinction I'm aware of is that TARIC rules follow the alphanumeric "ME32" format and our own rules take the form of a description of the violation. There may be exceptions to that, but I think that's the general principle

quotas/business_rules.py
gabelton: I think the issue is that we get a circular import error when specifying QuotaDefinition, which I suppose is why the other rules in this file don't use it

paulpepper-trade: Could we use forward references?
https://peps.python.org/pep-0484/#forward-references

quotas/business_rules.py
SteveCN7: I wonder if there is scope in instituting our own business rules similar to how we have with the true TARIC spec ones? Even if we store this somewhere accessible - food for thought

PR #636, opened 2022-07-29
TP2000-446 Implement a guard against double form submits

gabelton:
# TP2000-446 Implement a guard against double form submits

## Why

Double clicking a form submit button leads to transactions being created with duplicate composite_key values and causes a 500 integrity error.

## What

Add the govuk prevent double click attribute across submit buttons


stuaxo (approved):
:+1: :+1: :+1:

PR #637, opened 2022-08-01
TP2000-401: Add test for chunk_taric

eadpearce:
# TP2000-401: Add test for chunk_taric

## Why

* We need to improve test coverage of the importer

## What

* This test covers previously uncovered lines of code


importer/tests/test_chunker.py
```diff
@@ -115,3 +121,14 @@ def test_filter_transaction_records_negative(
...
     assert transaction is None
+
+
+def test_chunk_taric():
+    with open(f"{TEST_FILES_PATH}/goods.xml", "rb") as f:
+        content = f.read()
+    taric_file = SimpleUploadedFile("goods.xml", content, content_type="text/xml")
+    batch = factories.ImportBatchFactory.create()
+    chunk_taric(taric_file, batch)
+    assert ImporterXMLChunk.objects.count()
```
gabelton, line 132: For sanity checking, might be worth an assertion that there are no chunk objects prior to chunk_taric being called

importer/tests/test_chunker.py
```diff
@@ -115,3 +121,14 @@ def test_filter_transaction_records_negative(
...
+    chunk = ImporterXMLChunk.objects.first()
+    assert chunk.chunk_text
```
gabelton, line 134: Do we know what the text should be ? can we add an assertion about it?

importer/tests/test_chunker.py
```diff
@@ -115,3 +121,14 @@ def test_filter_transaction_records_negative(
     )
 
     assert transaction is None
+
+
+def test_chunk_taric():
```
gabelton, line 126: Could you add a docstring?

gabelton (approved):

importer/tests/test_chunker.py
eadpearce: it should be the contents of the original xml file but it gets spaces removed, attributes reordered... so i gave up trying to test it

PR #638, opened 2022-08-01
TP2000-247 Reverse related aggregated duties fix

paulpepper-trade:
# TP2000-247 Reverse related aggregated duties fix

## Why

The duties associated with measure's commodity is presented to the user as a 'duty sentence'. The duty sentence applied to a measure via an annotation on `MeasureQuerySet`. Each duty sentence is composed of an aggregation of the current version (as at the version of the associated measure) of one or more `MeaureComponents`.

The annotation that contains the aggregation is incorrect because:

- A `MeasureComponent`'s foreign key to the 'current' `Measure` instance is used rather than correctly using the Measure's version_group.
- The annotation repeats for as many times are there are versions of the annotated `Measure`.

## What

`DutySentenceMixin.with_duty_sentence()` has now been replaced by a non-grouped aggregation in `ComponentQuerySet.duty_sentence()`. The new queryset is now used to create an `objects` `ModelManager` instance on each of `MeasureComponent` and `MeasureConditionComponent`.

Additionally, a `duty_sentence` property has been added to `Measure` and `MeasureCondition`, each of which calls `ComponentQuerySet.duty_sentence()` to retrieve the correct 'current' duty sentence for a specific version of `Measure` / `MeasureCondition`.

Solutions using a Common Table Expression (CTE) -based approaches were tested, pushing more of the duty sentence construction down to the database level (requiring fewer calls to the database), but various performance issues were encountered - there's no doubt scope for revisiting and optimisation efforts here, should there be a need.

## Checklist
- Requires migrations? No


measures/querysets.py
```diff
@@ -16,27 +17,36 @@
...
+
+    def duty_sentence(
+        self,
+        component_parent: Union["measures.Measure", "measures.MeasureCondition"],
+    ):
         """
-        Annotates the query set with a human-readable string that represents the
-        aggregation of all of the linked components into a single duty sentence.
+        Returns the human-readable "duty sentence" for a Measure or
+        MeasureComponent instance as a string. The returned string value is a
```
gabelton, line 38: Should this be MeasureCondition rather than MeasureComponent?

measures/querysets.py
```diff
@@ -81,83 +91,99 @@ def with_duty_sentence(self) -> QuerySet:
...
+            component_parent.transaction,
+        )
+        if not component_qs:
+            return ""
+        latest_transaction_id = component_qs.aggregate(
+            latest_transaction_id=Max(
+                "transaction_id",
+            ),
+        ).get("latest_transaction_id")
+        component_qs = component_qs.filter(transaction_id=latest_transaction_id)
```
gabelton, line 107: Can we assume that in the case of a compound duty sentence the various components will each have the same transaction id?

gabelton (commented):
This looks great and so much cleaner than using `with_duty_sentence` all over the place. Is it worth adding a queryset test for the original bug? i.e. update a measure so that it would've had a duplicated duty sentence string before this change

measures/querysets.py
paulpepper-trade: We talked through this on another (side-) channel, but for completeness here: scrutinising [`measures.util.diff_components()`](https://github.com/uktrade/tamato/blob/dc86b22bc9a00a03ae397027b09d52104fb08176/measures/util.py#L28-L29) I think we decided that components will have the same transaction ID. I'll tighten up the signature on that function, particularly around use of `transaction`, in order to narrow down opportunities for incorrect use.

paulpepper-trade:
> This looks great and so much cleaner than using `with_duty_sentence` all over the place. Is it worth adding a queryset test for the original bug? i.e. update a measure so that it would've had a duplicated duty sentence string before this change

Yes, I've just added a unit test that constructs an object (`Measure` and `MeasureCondition`) history, with a variety of attached and changing components. The test also time travels backwards to ensure correct component association and duty sentence construction.

gabelton (approved):
Happy for this to be merged. Do you want me to create a separate ticket to update diff_components and add some tests?

PR #641, opened 2022-08-03
TP2000-311 Envelope schema expects a six-digit number to be present, the TARIC file only contains five

gabelton:
# TP2000-311 Envelope schema expects a six-digit number to be present, the TARIC file only contains five

## Why

Currently we have to manually update the envelope id on files sent to us by the EU before they can be imported
## What

Updates envelope xml schema to expect a 5 digit id.

### DIFF
--- a/commodities/tests/test_forms.py
+++ b/commodities/tests/test_forms.py
@@ -1,0 +1,21 @@
+from os import path
+
+from django.core.files.uploadedfile import SimpleUploadedFile
+
+from commodities import forms
+
+TEST_FILES_PATH = path.join(path.dirname(__file__), "test_files")
+
+
+def test_import_form_invalid_envelope_id():
+    upload_file = open(f"{TEST_FILES_PATH}/valid.xml", "rb")
+    file_data = {
+        "taric_file": SimpleUploadedFile(
+            upload_file.name,
+            upload_file.read(),
+            content_type="text/xml",
+        ),
+    }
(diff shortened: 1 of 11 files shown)


paulpepper-trade (approved):

commodities/tests/test_forms.py
```diff
@@ -0,0 +1,39 @@
+from os import path
+
+from django.core.files.uploadedfile import SimpleUploadedFile
+
+from commodities import forms
+
+TEST_FILES_PATH = path.join(path.dirname(__file__), "test_files")
+
+
+def test_import_form_valid_envelope_id():
```
stuaxo, line 10: Always makes me feel funny not closing files explicitly (and can get you at weird times, you are running PyPy, or doing loads of things in parallel or just using windows)

Would be good to chuck this in a with statement - I *think* you can pass `SimpleUploadFile` the file handle itself, which might avoid a copy somewhere... it's only a tiny thing but they all add up when we have loads of unit tests.

```python
    with open(f"{TEST_FILES_PATH}/valid.xml", "rb") as upload_file:
        file_data = {
            "taric_file": SimpleUploadedFile(
                upload_file.name,
                upload_file,
                content_type="text/xml",
            ),
        }
        form = forms.CommodityImportForm({}, file_data)
```

Another scheme could be to make a fixture like `valid_xml` - it could do something like open the file once somewhere, grab the content and give a BytesIO ... but might be too much hassle :)

PR #645, opened 2022-08-08
Add importer chunker tests

gabelton:
# Add importer chunker tests

## Why

Lots of chunker methods are currently untested.
## What

Adds some chunker tests


importer/tests/test_chunker.py
```diff
@@ -135,3 +144,180 @@ def test_chunk_taric():
...
+@mock.patch("importer.chunker.get_record_code")
+def test_write_transaction_to_chunk_record_code_not_in_tree(
+    get_record_code,
+    envelope_commodity,
+    taric_schema_tags,
+    record_group,
+):
+    """Test that write_transaction_to_chunk returns None when record_code not
+    found in dependency tree."""
+    get_record_code.return_value = "rubbish"
```
stuaxo, line 158: 😅

importer/chunker.py
```diff
@@ -192,6 +193,22 @@ def rewrite_comm_codes(batch: models.ImportBatch, envelope_id: str, record_code=
...
+    return max(
+        code.text for code in transaction.findall("*/*/*/ns2:record.code", nsmap)
+    )
+
+
+def get_chapter_heading(transaction: ET.Element) -> str:
+    item_ids = transaction.findall(
+        "*/*/*/*/ns2:goods.nomenclature.item.id",
+        nsmap,
+    )
```
stuaxo, line 206: This could probably use `.find()` since only one item_id is used.

It's probably worth checking there aren't any other places in this PR or the codebase that use .find_all() when .find() would do, since we spend a lot of our time parsing XML.

importer/tests/test_chunker.py
```diff
@@ -135,3 +144,180 @@ def test_chunk_taric():
...
+    ],
+)
+def test_sort_commodity_codes(
+    taric_schema_tags,
+    record_group,
+    commodity_1_kwargs,
+    commodity_2_kwargs,
+):
+    """Test that sort_commodity_codes sorts xml elements by item_id, indent,
+    suffix, and transaction order."""
```
stuaxo, line 290: Ace :100:

importer/tests/test_chunker.py
```diff
@@ -5,12 +5,21 @@
 from unittest import mock
```
stuaxo, line 5: Is it possible to update the docstring in `test_filter_transaction_records_positive` ?   It needs a little more context... is it a test for filter_snippet_transaction, filter_transaction_records ?

importer/tests/test_chunker.py
```diff
@@ -135,3 +144,180 @@ def test_chunk_taric():
...
+
+@mock.patch("importer.chunker.get_chunk")
+@mock.patch("importer.chunker.close_chunk")
+def test_write_transaction_to_chunk_exceed_max_file_size(
+    close_chunk,
+    get_chunk,
+    envelope_commodity,
+    taric_schema_tags,
+    record_group,
+):
```
stuaxo, line 226: The important side effects a developer calling the chunking API, is that the data is written to multiple files and that the files they are written to are closed.
(also outside this test, that the ImportChunk models are populated).

Using mocking to check the file is closed (and a new one opened) might improve this test - (mocking our own APIs can be necessary, but it's better to do it at 3rd party API boundaries if possible).

Checking whether close_chunk is called works as a sanity check that the API is the same, but doesn't let us know that is does what's expected.

It might be possible to load the XML that was chunked and verify some bit of data is there that you expect.

common/tests/util.py
```diff
@@ -471,15 +472,15 @@ def assert_many_records_match(
     assert expected_data == imported_data
 
 
-def generate_test_import_xml(obj: dict) -> BytesIO:
+def generate_test_import_xml(objects: List[dict], transaction_id: int = 0) -> BytesIO:
```
stuaxo, line 475: ^ Not a blocker, but probably can use `Sequence[Dict]` here.

common/tests/util.py
```diff
@@ -471,15 +472,15 @@ def assert_many_records_match(
...
     last_transaction = Transaction.objects.last()
     next_transaction_id = (last_transaction.order if last_transaction else 0) + 1
     xml = render_to_string(
         template_name="workbaskets/taric/transaction_detail.xml",
         context={
             "envelope_id": next_transaction_id,
-            "tracked_models": [obj],
-            "transaction_id": next_transaction_id,
+            "tracked_models": objects,
+            "transaction_id": transaction_id if transaction_id else next_transaction_id,
```
stuaxo, line 483: It feels a bit unusual that explicitly passing in `0` for transaction_id will make it default to `next_transaction_id`

It might be worth using  `transaction_id: Optional[int] = None` and making this check something like -

`"transaction_id": transaction_id next_transaction_id if transaction_id is None else transaction_id`

importer/chunker.py
gabelton: From searching, it looks like all the other instances of `findall()` are looping over the results, so probs used correctly in those scenarios.

importer/tests/test_chunker.py
gabelton: My thinking for this was just that `get_chunk` and `close_chunk` already have decent tests, so I could assume that they do what's expected and only needed to check that they are in fact being called with the expected arguments by `write_transaction_to_chunk`

importer/chunker.py
gabelton: Actually I think this was findall() for a reason. I'm not sure what that reason is yet, but this test fails when I run the whole suite and passes when I just test the module

importer/chunker.py
```diff
@@ -192,6 +193,22 @@ def rewrite_comm_codes(batch: models.ImportBatch, envelope_id: str, record_code=
         close_chunk(chunk, batch, key)
 
 
+def get_record_code(transaction: ET.Element) -> str:
+    return max(
+        code.text for code in transaction.findall("*/*/*/ns2:record.code", nsmap)
+    )
+
+
+def get_chapter_heading(transaction: ET.Element) -> str:
```
dougmills-DIT, line 202: can you confirm, the original code would return "00" if item_ids are empty - i cant see how this behaviour is still present in the refactoring.

dougmills-DIT (commented):
Only one thing I saw that I could not resolve by looking at the code

importer/chunker.py
dougmills-DIT: my bad - was below comments

dougmills-DIT (approved):
Cant find anything wrong, looks good to me

PR #649, opened 2022-08-10
TP2000-469 GA18 should use sid instead of PK

gabelton:
# TP2000-469 GA18 should use sid instead of PK

## Why

GA18 isn't firing when two overlapping memberships have been created with different versions of the same member area
## What

- Makes GA18 filter on `sid`, rather than `pk` and adds test for bug scenario


geo_areas/business_rules.py
```diff
@@ -168,8 +168,8 @@ def validate(self, membership):
         if (
             type(membership)
             .objects.filter(
-                geo_group=membership.geo_group,
-                member=membership.member,
+                geo_group__sid=membership.geo_group.sid,
```
paulpepper-trade, line 171: It might be preferable to use `version_group` in place of `sid` here as this kind of filtering was one of the original reasons for its introduction, apparently - allowing identifying fields reuse and more performant(?). I suppose it's more generically applicable across all of our `TrackedModel` subclasses too.

paulpepper-trade (approved):

geo_areas/business_rules.py
stuaxo: I think we may want both... and to have this as a generic thing we do...

I guess the general rule is along the lines of narrow down to where both version groups overlap and then go from there.

gabelton: I guess if we want `sid` potentially to be reusable then we shouldn't use that here. I've swapped in `version_group` instead. I agree that it would be good to try to make comparison on `version_group` the default, rather than `pk`, but I think it's outside the scope of this PR. I can create a ticket for it.

gabelton: https://uktrade.atlassian.net/browse/TP2000-470

paulpepper-trade: Yep, I don't know if we've ever had to reuse identifying fields / SIDs, so perhaps that initial purpose could now be considered redundant?

PR #650, opened 2022-08-11
ADR on adding content checksums to TrackedModels.

stuaxo:
For the business rule checks it's useful to have a way to know if a previous check is still valid.

This ADR propsoses using checksums of the editable content of trackedmodels as the mechanism.


paulpepper-trade (approved):

docs/source/adr/0016-content-checksums-for-tracked-models.rst
```diff
@@ -0,0 +1,53 @@
...
+Status
+------
+
+Proposed
+
+
+Context
+-------
+
+After running the business rules it is useful to store the results of checks against models, and have a mechanism to check if the result is still valid to avoid re-running the check in future.
```
paulpepper-trade, line 16: We talked about what can invalidate a check after it has completed. We know that after a check has been run, then newly published tariff data can cause some types of check to become invalid - an ME32 check against a Measure, for instance. So hashing doesn't help avoid the need to run a check again in those circumstances.
It would be worth pointing out that caveat here.

stuaxo: Yep, this is really important - I've updated the PR description to cover this, and will update the ADR to reflect this (date validity checkers will be their own adr)

docs/source/adr/0016-content-checksums-for-tracked-models.rst
```diff
@@ -0,0 +1,53 @@
...
+By limiting the scope to business rule checks this gives time for the code to bed in, and makes it easier to change things if there are issues.
+
+
+Decision
+--------
+
+The fields to be hashed are contained in `copyable_fields` but since this doesn't communicate the intent used here, this will be proxied as `mutable_fields`, which is explicitly about providing which fields can be hashed to check equivilence.
+
+
+TrackedModel will gain an API named get_content_hash() that returns a sha256 hash.
```
gabelton, line 38: Is it possible to be more specfic than "API"? Would it be a class_method? As I read it currently, it sounds a bit like the hash will be attached to the tracked model, which contradicts what's written on L28. I know that's not the intent, but just be good to clarify

docs/source/adr/0016-content-checksums-for-tracked-models.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Iterates the fields in `mutable_fields` and building a dict keyed by field name, the value contains the location __module__ and __name__ of the field.
+
+If the field is not TrackedModel then the value includes the sha256 of their content, for foreign keys that are TrackedModels the value is obtained by calling get_content_hash()
+
+
+Consequences
+------------
+
+This provides a mechanism to store the results of business rule checks, only becomming invalid when the TrackedModels they reference change.
```
gabelton, line 50: Will part of this work involve creating a list of rules to which this kind of caching will be applicable?

docs/source/adr/0016-content-checksums-for-tracked-models.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Proposed
+
+
+Context
+-------
+
+After running the business rules it is useful to store the results of checks against models, and have a mechanism to check if the result is still valid to avoid re-running the check in future.
+
+This ADR adds a mechanims to checksum the user-editable data to check if to versions of a Tracked Model are equivilent.
```
gabelton, line 18: --> mechanism
--> equivalent

docs/source/adr/0016-content-checksums-for-tracked-models.rst
```diff
@@ -0,0 +1,53 @@
...
+
+In the simplest sense this means hashing fields that exclude the PK.
```
gabelton, line 20: What's meant by "fields that exclude the PK" ?

docs/source/adr/0016-content-checksums-for-tracked-models.rst
```diff
@@ -0,0 +1,53 @@
...
+
+
+TAP already has a mechanism to get the user editable fields, the attribute `.copyable_fields`, which provides a starting point.
+
+Unlike pythons default hashing, it is nessacary to distinguish between two different types that produce the same hash - so the string that gets passed the hashing function ("hashable_string"), will include the module and class name.
```
gabelton, line 25: --> necessary

gabelton, line 25: is it worth adding a link to docs on python's default hashing?

docs/source/adr/0016-content-checksums-for-tracked-models.rst
```diff
@@ -0,0 +1,53 @@
...
+
+
+The hash generated in this ADR will not currently be attached to TrackedModel - thought it may be desirable in future.
```
gabelton, line 28: --> though

docs/source/adr/0016-content-checksums-for-tracked-models.rst
gabelton, line 50: --> becoming

gabelton (approved):
Really cool stuff! I would love an example involving a current business rule, if at all possible, just so that I can visualise it / walk it through in my head, but looks great apart from that. Please excuse the pedantic spelling comments...

docs/source/adr/0016-content-checksums-for-tracked-models.rst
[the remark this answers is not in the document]
stuaxo: Saved to it's own field I guess.

docs/source/adr/0016-content-checksums-for-tracked-models.rst
stuaxo: Good point, in I guess I meant a method - e.g. .content_hash(), vs something we save forever in a field.

docs/source/adr/0016-content-checksums-for-tracked-models.rst
stuaxo: Yes, though it doesn't need to be exhaustive - as it can be extended if needed later.

PR #651, opened 2022-08-11
TP2000-449 add volumes business rule

gabelton:
# TP2000-449 add volumes business rule

## Why

HMRC have flagged that several recent steel quota definitions have been created with `volume` and `initial_volume` values that do not match. Doug and Tom's investigation has also revealed past instances of this, which shouldn't have been allowed either.

## What

Adds a rule preventing the creation of a `QuotaDefinition` with different `volume` and `initial_volume` values, unless that definition is used as the `main_quota` in a `QuotaAssociation`.

This rule implicitly enforces that, in order for a quota definition to be created with different `volume` and `initial_volume` values, an association will have to be created in the same or earlier transaction with that definition as its `main_quota`

### DIFF
--- a/quotas/business_rules.py
+++ b/quotas/business_rules.py
@@ -268,6 +268,20 @@ class OverlappingQuotaDefinition(BusinessRule):
             raise self.violation(quota_definition)
 
 
+class VolumeAndInitialVolumeMustMatch(BusinessRule):
+    """Unless it is the main quota in a quota association, a definition's volume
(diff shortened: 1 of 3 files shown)


stuaxo:
I wonder if we should prefix all our internal business rules with something, TAP or even TAMATO ?

It would be handy when looking at big lists of them.

stuaxo (approved):
Nice !  :+1:

PR #652, opened 2022-08-11
TP2000-278 Allow for workbaskets in the TAP page to be selected for inspection

gabelton:
# TP2000-278 Allow for workbaskets in the TAP page to be selected for inspection

## Why

The WorkBasketList view was removed as part of multi workbasket work. `SelectWorkbasketView` now lives at the `workbaskets` url, but this doesn't provide a view of PUBLISHED or ARCHIVED baskets. This is useful for TMs (and devs) when comparing the changes in a basket to what's visible in TAP.
## What

Restores `WorkBasketList` view, as added by @paulpepper-trade in [this PR](https://github.com/uktrade/tamato/pull/549) and adds a couple of view tests. The jinja is already there.

### DIFF
--- a/workbaskets/urls.py
+++ b/workbaskets/urls.py
@@ -56,6 +56,11 @@ ui_patterns = [
         ui_views.WorkBasketDeleteChangesDone.as_view(),
         name="workbasket-ui-delete-changes-done",
     ),
+    path(
+        "list",
+        ui_views.WorkBasketList.as_view(),
+        name="workbasket-ui-list-all",
+    ),
 ]
 
 urlpatterns = [
--- a/workbaskets/urls.py
+++ b/workbaskets/urls.py
@@ -57,7 +57,7 @@ ui_patterns = [
         name="workbasket-ui-delete-changes-done",
     ),
     path(
-        "list",
+        "list-all",
         ui_views.WorkBasketList.as_view(),
         name="workbasket-ui-list-all",
     ),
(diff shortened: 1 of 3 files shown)


workbaskets/views/ui.py
```diff
@@ -375,3 +375,17 @@ def form_valid(self, form):
         store.add_items(to_add)
         store.remove_items(to_remove)
         return super().form_valid(form)
+
+
+class WorkBasketList(WithPaginationListView):
```
paulpepper-trade, line 380: Oh, did this ListView get completely blitz in the last batch of changes?

paulpepper-trade (approved):
LGTM

workbaskets/urls.py
```diff
@@ -56,6 +56,11 @@
         ui_views.WorkBasketDeleteChangesDone.as_view(),
         name="workbasket-ui-delete-changes-done",
     ),
+    path(
+        "list",
```
paulpepper-trade, line 60: As there's another list view off the /workbaskets/ path, would it be useful to say in the path how this is different? Something like /workbaskets/list-all/ ?

workbaskets/views/ui.py
gabelton: Think it was removed here https://github.com/uktrade/tamato/pull/627/files#diff-0157b097c7d70c161db821e27bac9800815d2fc7275791b0544ca066efbdbf3cL41

PR #653, opened 2022-08-12
TP2000-278 allow for workbaskets in the tap page to be selected for inspection (pt. 2)

gabelton:
# TP2000-278 allow for workbaskets in the tap page to be selected for inspection (pt. 2)

## Why

When the user is on the workbaskets/list-all page and clicks on a workbasket they are taken to the WorkBasketDetail view which gives them the option to edit the basket. We don't want this as the point of the list-all endpoint is to give a view of published and archived baskets, which shouldn't be editable.

## What

- Restores the old WorkBasketDetail view as WorkBasketChanges
- Update workbasket list.jinja to use new list-all endpoint instead of list


paulpepper-trade (approved):
LGTM

PR #654, opened 2022-08-15
TP2000-215 add defusedxml package to protect against xml parsing vulnerabilities

gabelton:
# TP2000-215 add defusedxml package to protect against xml parsing vulnerabilities

## Why

defusedxml comes with a `DTDForbidden` exception. Disabling doctype declarations (DTD) is the safest way to prevent XML eXternal Entity injection (XXE) and denial of service (DOS) attacks. 

## What

Makes our lxml `parse()` and `fromstring()` methods disable DTD by default

My implementation is borrowed from the deprecated [lxml module](https://github.com/tiran/defusedxml/blob/main/defusedxml/lxml.py) in defusedxml . I didn't copy the forbid_entities logic because, if we disable DTD, we don't need it and also I think lxml already has built in protection against lots of different attacks. (I originally copied tests from defusedxml, using their test data and found that an `XMLSyntaxError` was raised in all cases where the DTD exception wasn't)


requirements.txt
```diff
@@ -8,6 +8,7 @@ boto3==1.18.51
 celery[redis]>=5.2.2
 coverage[toml]==5.4
 crispy-forms-gds @ git+https://github.com/uktrade/crispy-forms-gds.git@b50168d0e23ffacbdd30e7819ec8f9a08e055c8e
+defusedxml==0.7.1
```
dougmills-DIT, line 11: I suggest dependencies are not clamped to a bugfix version, but rather a minor version e.g. 0.7.*

dougmills-DIT:
I would suggest an explicit unit test example for billion laughs / quadratic blowup (kinda the same thing / relies on the same expansion mechanic) to illustrate the protection

dougmills-DIT (approved):
looks good!

PR #655, opened 2022-08-17
TP2000-448 north macedonia doesnt appear in search or dropdowns

gabelton:
# TP2000-448 north macedonia doesnt appear in search or dropdowns

## Why

Countries with multiple current descriptions, like North Macedonia, are currently only searchable by the description that was created first. It should be possible to search by both descriptions, but the most recent description should be the one that appears in dropdowns.
## What

- Change `with_current_description` --> `with_latest_description`, ordering by validity_start date
-  Add `with_current_descriptions` queryset method, annotating a queryset with all current descriptions as one concatenated string


stuaxo (approved):
LGTM

PR #657, opened 2022-08-18
TP2000-478 500 error when user not logged in tries to create a new workbasket

gabelton:
# TP2000-478 500 error when user not logged in tries to create a new workbasket

## Why

When running the app locally a user who hasn't logged in will get a 500 error when trying to create a new workbasket.
## What

Updates `WorkBasketCreate.form_valid` to redirect to login, if no username is provided


workbaskets/views/ui.py
```diff
@@ -61,6 +61,8 @@ class WorkBasketCreate(CreateView):
     form_class = forms.WorkbasketCreateForm
 
     def form_valid(self, form):
+        if not self.request.user.username:
```
paulpepper-trade, line 64: `User.is_authenticated` would normally be the preferred way to establish whether the current user is logged in:  https://docs.djangoproject.com/en/4.1/ref/contrib/auth/#django.contrib.auth.models.User.is_authenticated
Isn't that possible here?

paulpepper-trade, line 64: I'm also wondering why an Anonymous user (one without a "workbaskets.add_workbasket" permission) isn't / wouldn't be redirected to the login page: https://docs.djangoproject.com/en/4.1/topics/auth/default/#the-permission-required-decorator.

gabelton: I'll add a separate test for that

paulpepper-trade (approved):

PR #661, opened 2022-08-24
TP2000 465 Provide a view for commodity codes using existing view designs

LaurenQurashi:
# TP2000-465 Provide a view for commodity codes using existing view designs.
Hello! I'm really sorry that I didn't manage to finish this! the tests played me more than they should have! 

In total I have: 
- Added a find commodities link to the home page 
- Added a commodity list page with a working filter
- Added a commodity detail page with working parent links
- Made sure the breadcrumbs work, as there was an auto breadcrumbs coming from somewhere deep in the template layers but it was adding find and edit but we just want find 
- Added tests for the List view.

I need someone to: 
- Probably re do all of this as I make poor decisions lol 
- Add tests for the details view 
- Maybe tests for the filters? there's a test for filters in the views test.. and the filter is it's own thing so I'd imagine there would be tests for that.. but maybe that needs tests?
- finish the `test_commodity_list_filter` test in the commodities/test_views.py as I copied and pasted it from geo areas and it's got half the stuff in it still. 

Thank you and sorry! 

Screenshot 2022-08-30 at 10 25 04
Screenshot 2022-08-30 at 10 25 25


commodities/urls.py
```diff
@@ -13,14 +13,24 @@
 
 urlpatterns = [
     path("api/", include(api_router.urls)),
+    path(
```
LaurenQurashi, line 16: Here I decided not to use the `get_ui_paths()` function as it's pluralisation while making the URLs wouldn't work for Commodities - it would make it Commoditie. I have however amended the URLs in the file to follow the same naming conventions the function would bring, in order to match the rest of the apps

commodities/forms.py
```diff
@@ -1,16 +1,48 @@
...
+def get_mime_type(file):
+    """Get MIME by reading the header of the file."""
+    initial_pos = file.tell()
+    file.seek(0)
+    mime_type = magic.from_buffer(file.read(1024), mime=True)
+    file.seek(initial_pos)
+    return mime_type
+
 
-class CommodityImportForm(ImportForm):
```
gabelton, line 13: Don't think this needs to be changed

commodities/filters.py
```diff
@@ -15,3 +22,15 @@ def search_queryset(self, queryset, search_term):
             return queryset.filter(item_id__startswith=search_term)
 
         return super().search_queryset(queryset, search_term)
+
```
LaurenQurashi, line 25: Made this filter to filter for commodity code, and whether it's active or not

commodities/forms.py
```diff
@@ -10,6 +15,30 @@
 from importer.forms import ImportForm
 
 
+class CommodityFilterForm(forms.Form):
```
LaurenQurashi, line 18: Made this nice filter form to negotiate the layout. Some filters did it this way, some filters just had stuff in the filter itself? not sure what's the best way

commodities/jinja2/commodities/detail.jinja
```diff
@@ -0,0 +1,43 @@
+{% extends "layouts/layout.jinja" %}
```
LaurenQurashi, line 1: Details page template, shows a nice little details tab. I went off Quotas as there's no figma for all this, but there's more tabs to come according to steve so I think it's fine to have just a lonesome tab for now

commodities/jinja2/includes/commodities/tabs/core_data.jinja
```diff
@@ -0,0 +1,88 @@
+{% set parent_sid %}
```
LaurenQurashi, line 1: Detail page table

commodities/tests/test_views.py
```diff
@@ -6,22 +6,27 @@
...
 from common.tests.factories import ImportBatchFactory
 
+pytestmark = pytest.mark.django_db
+
 TEST_FILES_PATH = path.join(path.dirname(__file__), "test_files")
 
 
 def test_commodities_import_200(valid_user_client):
-    url = reverse("commodities-import")
+    url = reverse("commodity-ui-import")
```
LaurenQurashi, line 20: Here I thought I'd re name the commodities URLs so that they matched the project more? might have been unnecessary but here we are... this is just changing the name for it's tests

commodities/tests/test_views.py
```diff
@@ -40,11 +45,61 @@ def test_commodities_import_success_redirect(mock_save, valid_user_client):
...
+    GoodsNomenclatureFactory.create_batch(3)
+    view = CommodityList()
+    qs = view.get_queryset()
+    ordered_commodities = qs.order_by("item_id")
+
+    assert qs.first().item_id == ordered_commodities[0].item_id
+    assert qs.last().item_id == ordered_commodities.last().item_id
+
+
+def test_commodity_list_filter(search_terms, valid_user_client):
```
LaurenQurashi, line 93: Please note this may look like a complete test but it is infact a copy and paste from Geo areas that got half chopped up before Stuart's leaving games 😬

commodities/views.py
```diff
@@ -53,3 +56,23 @@ def form_valid(self, form):
 
 class CommodityImportSuccessView(TemplateView):
     template_name = "commodities/import-success.jinja"
+
+
```
LaurenQurashi, line 60: views.. ya know

common/jinja2/layouts/commodity_list.jinja
```diff
@@ -0,0 +1,50 @@
+{% extends "layouts/layout.jinja" %}
```
LaurenQurashi, line 1: If my memory serves me well, I made this duplicate as the original hard codes find and edit in the title and bread crumbs so I just did this? thinking back this could be completely unnecessary too so see how it fits and if this was a poor choice. It could just need a different page title passed in but I'm not sure at this very moment

pii-ner-exclude.txt
```diff
@@ -1182,4 +1182,9 @@ commodity_1_kwargs
 goods_nomenclature_indent.xml
 ElementTree
 403s
-importer.forms.UploadTaricForm.save
\ No newline at end of file
+importer.forms.UploadTaricForm.save
```
LaurenQurashi, line 1185: You know how it goes

commodities/views.py
```diff
@@ -53,3 +56,22 @@ def form_valid(self, form):
...
+class CommodityMixin:
+    model = GoodsNomenclature
+
+    def get_queryset(self):
+        return GoodsNomenclature.objects.current()
+
+
+class CommodityList(CommodityMixin, TamatoListView):
```
paulpepper-trade, line 68: Does this ListView require a reference to the current workbasket? If not, then `common.views.WithPaginationListView` may be preferable to `TamatoListView`.

commodities/filters.py
```diff
@@ -15,3 +22,19 @@ def search_queryset(self, queryset, search_term):
...
+
+class CommodityFilter(TamatoFilter, ActiveStateMixin):
```
paulpepper-trade, line 27: Shouldn't the mixin come before the base class?

commodities/jinja2/commodities/detail.jinja
```diff
@@ -0,0 +1,43 @@
...
+  {{ govukBreadcrumbs({
+    "items": [
+      {"text": "Home", "href": url("home")},
+      {"text": "Find commodity codes", "href": url("commodity-ui-list")},
+      {"text": page_title}
+    ]
+  }) }}
+{% endblock %}
+
+{% set tabs = {
```
paulpepper-trade, line 22: What's the rationale for having a tabbed view containing only a single tab? Perhaps this is a late-in-the-day question about the design, but it would at least be interesting to know.

commodities/urls.py
```diff
@@ -14,13 +14,23 @@
 urlpatterns = [
     path("api/", include(api_router.urls)),
     path(
-        "commodities/import/",
+        "commodities/",
```
paulpepper-trade, line 17: I think 'common.paths.get_ui_paths()` would correctly pick up the non-importer paths, so long as the Views are set correctly. There are only a couple here and we have a ticket to refactor `get_ui_paths`, so I don't have a strong feeling either way about using it here...

commodities/views.py
```diff
@@ -53,3 +56,22 @@ def form_valid(self, form):
...
+    template_name = "commodities/list.jinja"
+    filterset_class = CommodityFilter
+
+    def get_queryset(self):
+        return GoodsNomenclature.objects.current().order_by("item_id")
```
paulpepper-trade, line 73: Would it be preferable to place the ordering on the `GoodsNomenclature`'s `Meta.ordering` attribute? I can't immediately think of a downside to doing so.

paulpepper-trade (approved):

commodities/jinja2/commodities/detail.jinja
SteveCN7: Ooohh -  I can answer this, there will be more tabs later, the design will grow to have a few extra tabs

paulpepper-trade: I suppose if other tabs are to be added imminently and it matches the layout of our other find and filter pages... I think Quota details must have had a single tab its entire life.

additional_codes/tests/bdd/test_browse_additional_codes.py
```diff
@@ -22,8 +24,9 @@ def additional_code_list(additional_code_search, additional_code_X000):
     results = additional_code_search.json()["results"]
     assert len(results) == 1
     result = results[0]
-    assert (
-        result["label"]
-        == f"{additional_code_X000} - {additional_code_X000.get_description(transaction=additional_code_X000.transaction).description}"
-    )
+    with override_current_transaction(additional_code_X000.transaction):
```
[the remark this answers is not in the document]
gabelton: I think we didn't need it until I refactored `get_descriptions` to use current transaction, but agree that we should think of a better solution to setting it in tests. Some tests, like `test_set_transaction` assume that no current transaction is set, so we'd need to think about what to do there too.

commodities/views.py
gabelton: I think it would be preferable. The issue I'm encountering is that the ordering gets overridden when using `current`. I think it's the annotation done by `approved_up_to_transaction` but not 100% sure

paulpepper-trade: I've ran into the same issue: https://github.com/uktrade/tamato/blob/0909983e8f8e9f5d0efef9e71c6a85365bdea63c/measures/tests/test_views.py#L473
I don't see where `approved_up_to_transaction` is overriding the ordering.
I'd like to know what's going on, but putting the `order_by()` in the view here seems reasonable anyway.

paulpepper-trade (approved):

PR #671, opened 2022-08-31
Tp2000 494  django management command update

paulpepper-trade:
# TP2000-494 Django management command updates

## Why

Tamato is missing a management command to straight-forwardly get workbasket info for packaging purposes.
A previous management command `set_active_workbasket` is no longer valid since the introduction of multi-workbasket capability.

## What

Add a Django management command `workbasket_info` and enhance workbasket info output.
Remove the `set_active_workbasket` Django management command.

## Checklist
- Requires migrations? No


gabelton (approved):
Looks good! What happens if the user forgets to add a PK argument?

paulpepper-trade:
> Looks good! What happens if the user forgets to add a PK argument?

A `DoesNotExist` exception is raised, which isn't so pretty but I'd expect a dev user to know what it means. I add an update to catch it and improve the output if you prefer?

paulpepper-trade:
> Looks good! What happens if the user forgets to add a PK argument?

I've added more graceful handling of that case. Thanks Gabe!

gabelton:
Brill, thanks!

PR #672, opened 2022-08-31
TP-495  sync_run_checks command throws 500

paulpepper-trade:
# TP-495 sync_run_checks command throws 500

## Why

Django management commands `sync_run_checks` and `set_workbasket_status` don't gracefully handle incorrect workbasket identifier values.

## What

Factor out the behaviour to get a workbasket by its primary key, gracefully failing if no workbasket is found. Apply the refactored code to the Django management commands  `sync_run_checks` and `set_workbasket_status`.

## Checklist
- Requires migrations? No

### DIFF
--- a/workbaskets/management/util.py
+++ b/workbaskets/management/util.py
@@ -1,5 +1,7 @@
 import enum
 
+from workbaskets.models import WorkBasket
+
 
 class WorkBasketOutputFormat(enum.Enum):
     READABLE = 1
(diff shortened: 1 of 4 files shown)


workbaskets/management/util.py
```diff
@@ -77,3 +79,20 @@ def output_workbaskets(self, workbaskets, show_transaction_info, output_format):
 
         for w in workbaskets:
             self.output_workbasket(w, show_transaction_info, output_format)
+
+    def get_workbasket_or_exit(self, workbasket_pk):
```
gabelton, line 83: nice!

gabelton (approved):

PR #676, opened 2022-09-08
TP2000-506: Add urls for commodities and related objects

eadpearce:
# TP2000-506: Add urls for commodities and related objects

## Why

* Users should be able to click through to see a new commodity code in their workbasket

## What

* Modifies the get_url method on commodity models to link to the commodity detail page

### DIFF
--- a/commodities/models/orm.py
+++ b/commodities/models/orm.py
@@ -179,12 +179,6 @@ class GoodsNomenclatureIndent(TrackedModel, ValidityStartMixin):
         item_id = self.indented_goods_nomenclature.item_id
         return self.indent == 0 and item_id[2:] == "00000000"
 
-    def get_url(self):
-        return reverse(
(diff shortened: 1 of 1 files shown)


commodities/models/orm.py
```diff
@@ -175,6 +179,12 @@ def is_root(self) -> bool:
         item_id = self.indented_goods_nomenclature.item_id
         return self.indent == 0 and item_id[2:] == "00000000"
 
+    def get_url(self):
```
gabelton, line 182: Do we want Indent and Origin objects to link to a GoodsNomenclature object?

eadpearce: What else should we link to?

eadpearce: Then that's outside the scope of this ticket

gabelton (approved):

PR #678, opened 2022-09-09
TP2000-441 build a solution to allow users to review business rule violations

gabelton:
# TP2000-441 build a solution to allow users to review business rule violations

## Why

Currently tariff managers need to ask a dev to run business rules and see the results of these checks. This is time that could be better spent doing development work.
## What

- Adds a list view to allow users to see a list of business rule violations for a particular workbasket
- Adds a functional `run_business_rules` view to allow tariff managers to trigger a business rule check task from the workbasket review page.
- Adds `TimestampedMixin` to `TrackedModelCheck`
- Adds `rule_check_task_id` to `WorkBasket` so that we can save a rule check task id against a workbasket
- Implements an MVP design for viewing and correcting business rule errors. (We still need to build, amongst other things, the detail view for an individual business rule violation and the tabs design on the workbasket review page.)

- Requires migrations? Yes

Wren's design: https://www.figma.com/proto/RQEp14sSbH1MevMbrBc5HJ/Tamato---Aug01?node-id=1526%3A13464&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=27%3A2350

Screenshot 2022-09-09 at 15 00 22

Screenshot 2022-09-09 at 15 01 08


common/tests/util.py
```diff
@@ -640,7 +639,12 @@ def now(self):
 
     @property
     def datetime_now(self):
-        return datetime.now(tz=UTC).replace(hour=0, minute=0, second=0, microsecond=0)
+        return datetime.now(timezone("Europe/London")).replace(
```
eadpearce, line 642: is it worth saving the string "Europe/London" as a constant here or nah?

workbaskets/jinja2/workbaskets/summary-workbasket.jinja
```diff
@@ -4,6 +4,33 @@
 {% from "includes/workbaskets/navigation.jinja" import navigation %}
 
 {% set page_title %}Workbasket {{ workbasket.id if workbasket else request.session.workbasket.id }} - Summary{% endset %}
+{% set business_rules_button %} <a href="{{ url('workbaskets:workbasket-run-business-rules')}}" class="govuk-button govuk-button--secondary govuk-!-margin-left-5"">Run business rules</a> {% endset %}
+{% set rule_check_status %}
+  {% if rule_check_in_progress %}
+    <h2 class="govuk-heading-s">
```
eadpearce, line 10: maybe not for this PR but might be cool to make this into a reusable macro where you just pass in the message to be displayed

eadpearce (approved):

workbaskets/jinja2/workbaskets/summary-workbasket.jinja
gabelton: Couldn't think of the best way to make it a macro, but I've made it a bit less repetitive at least

PR #681, opened 2022-09-13
TP2000-510: Add business rule violation detail page

eadpearce:
# TP2000-510: Add business rule violation detail page

## Why

* Enable users to see more detail on a particular rule violation and remediate

## What

* Adds a detail view for a single business rule violation in a workbasket

### DIFF
--- a/workbaskets/jinja2/workbaskets/violation_detail.jinja
+++ b/workbaskets/jinja2/workbaskets/violation_detail.jinja
@@ -1,0 +1,42 @@
+{% extends "layouts/layout.jinja" %}
+
+{% from "components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
+{% from "components/table/macro.njk" import govukTable %}
+{% from "includes/workbaskets/navigation.jinja" import navigation %}
+
+{% set page_title %}Workbasket {{ workbasket.id if workbasket else request.session.workbasket.id }} - Rule violation
+details
+{% endset %}
+
+{% block content %}
+<h1 class="govuk-heading-xl govuk-!-margin-bottom-3">
+  {{ page_title }}
+</h1>
+{{ navigation(request, "violations") }}
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    {% set object_link %}
+    <a href="{{ object.model.get_url() }}" class="govuk-link">{{ object.model._meta.verbose_name.title() }}
+      {{ object.model.id }}</a>
+    {% endset %}
+    <h2 class="govuk-heading-m">Rule violation details - {{ object_link }}</h2>
+    <hr class="govuk-section-break govuk-section-break--s govuk-section-break--visible">
+
+    <h3 class="govuk-heading-s">Business rule ID:</h3>
+    <p class="govuk-body">{{ object.rule_code }}</p>
+    <h3 class="govuk-heading-s">Rule violation description:</h3>
+    <p class="govuk-body">{{ object.message }}</p>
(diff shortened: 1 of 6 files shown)


workbaskets/jinja2/workbaskets/violation_detail.jinja
```diff
@@ -0,0 +1,42 @@
...
+    <h2 class="govuk-heading-m">Rule violation details - {{ object_link }}</h2>
+    <hr class="govuk-section-break govuk-section-break--s govuk-section-break--visible">
+
+    <h3 class="govuk-heading-s">Business rule ID:</h3>
+    <p class="govuk-body">{{ object.rule_code }}</p>
+    <h3 class="govuk-heading-s">Rule violation description:</h3>
+    <p class="govuk-body">{{ object.message }}</p>
+    {% if object.solution %}
+    <h3 class="govuk-heading-s">Solution:</h3>
+    <p class="govuk-body">{{ object.solution }}</p>
```
eadpearce, line 32: placeholder attribute for when we eventually add content to help users resolve issues

gabelton: dumb question, but does the page not complain that the object doesn't have a solution attribute?

workbaskets/tests/test_views.py
```diff
@@ -598,3 +598,32 @@ def test_workbasket_violations(valid_user_client, session_workbasket):
...
+        "error_count": session_workbasket.tracked_model_check_errors.count(),
+    }
+    session.save()
+    response = valid_user_client.get(url)
+    assert response.status_code == 200
+
+    soup = BeautifulSoup(str(response.content), "html.parser")
+    paragraphs_text = [e.text for e in soup.select("p")]
+    assert check.rule_code in paragraphs_text
+    assert check.message in paragraphs_text
```
gabelton, line 629: Could add a check for solution, even if it's None now, so that it fails later?

workbaskets/urls.py
```diff
@@ -85,5 +85,10 @@
 
 urlpatterns = [
     path("workbaskets/", include(ui_patterns)),
+    path(
+        f"<wb_pk>/violations/<pk>/",
```
paulpepper-trade, line 89: Although it may not sit under `/workbasket/` it looks a bit odd hanging off the root path. Would something like `/business-rules/` or similar be preferable?

paulpepper-trade (approved):

workbaskets/jinja2/workbaskets/violation_detail.jinja
```diff
@@ -0,0 +1,52 @@
+{% extends "layouts/layout.jinja" %}
+
+{% from "components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
+{% from "components/table/macro.njk" import govukTable %}
+{% from "includes/workbaskets/navigation.jinja" import navigation %}
+
+{% set page_title %}Workbasket {{ workbasket.id if workbasket else request.session.workbasket.id }} - Rule violation
```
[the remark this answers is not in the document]
eadpearce: omg that's what those do! ty!

workbaskets/urls.py
eadpearce: OH this is a mistake. it should be under /workbaskets/

workbaskets/jinja2/workbaskets/violation_detail.jinja
eadpearce: Nope! I was actually expecting it to but it didn't.

PR #682, opened 2022-09-14
TP2000-510: Display object SID on violation detail page

eadpearce:
# TP2000-510: Display object SID on violation detail page

## Why

* We want to show the SID not the ID for the object the rule violation relates to

## What

* Displays the SID if available, the ID if not

### DIFF
--- a/workbaskets/jinja2/workbaskets/violation_detail.jinja
+++ b/workbaskets/jinja2/workbaskets/violation_detail.jinja
@@ -28,7 +28,7 @@ details
   <div class="govuk-grid-column-two-thirds">
     {% set object_link %}
     <a href="{{ object.model.get_url() }}" class="govuk-link">{{ object.model._meta.verbose_name.title() }}
-      {{ object.model.id }}</a>
+      {{ object.model.sid or object.model.id }}</a>
     {% endset %}
     <h2 class="govuk-heading-m">Rule violation details - {{ object_link }}</h2>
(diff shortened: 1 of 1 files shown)


workbaskets/jinja2/workbaskets/violation_detail.jinja
```diff
@@ -28,7 +28,7 @@ details
   <div class="govuk-grid-column-two-thirds">
     {% set object_link %}
     <a href="{{ object.model.get_url() }}" class="govuk-link">{{ object.model._meta.verbose_name.title() }}
-      {{ object.model.id }}</a>
+      {{ object.model.sid or object.model.id }}</a>
```
paulpepper-trade, line 31: Seems reasonable as an interim display option, though...

This will display the object's SID, if it has one. Elsewhere in the application we have `object._meta.verbose_name.title()`, which more generally gives a stringified version of `identifying_fields` - SIDs are just a special system-generated type of these. Non-SID identifying fields are confusing, but then a PK is kind of incorrect (because they point to a specific version of a TrackedModel). It's an issue we've never addressed, so perhaps its time has come.

paulpepper-trade (approved):

workbaskets/jinja2/workbaskets/violation_detail.jinja
eadpearce: this displays `object._meta.verbose_name.title()` and `object.model.sid or object.model.id`. is `verbose_name` supposed to display the SID/identifying_fields?

PR #684, opened 2022-09-16
TP2000-523 Get ME32 clashes

paulpepper-trade:
# TP2000-523 ME32 clashes Django command

## Why

Measures that are the source of ME32 clashes are not always easy to identify.

## What

This PR provides a Django management command to show ME32 clashing measures, if there are any, given a measure SID. 

## Checklist
- Requires migrations? No


workbaskets/management/commands/me32_clashes.py
```diff
@@ -0,0 +1,45 @@
...
+
+class Command(WorkBasketCommandMixin, BaseCommand):
+    help = "Show clashes for a given measure, if any exist."
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument("MEASURE_SID", type=int)
+
+    def handle(self, *args: Any, **options: Any) -> Optional[str]:
+        sid = int(options["MEASURE_SID"])
+        measure = Measure.objects.filter(sid=sid).last()
```
gabelton, line 23: Should this be `objects.latest_approved()` ?

paulpepper-trade: > Should this be `objects.latest_approved()` ?

I don't think so, Gabe. `TrackedModelQuerySet.latest_approved()` relies upon the measure being published (`TrackedModel.is_current` being non-null, which only published have). I wondered about using `TrackedModelQuerySet.approved_up_to_transaction()`but that would require a valid transaction, which we don't have from a prompt.

I suppose we could take the workbasket ID as a parameter to the command in order to get the last transaction ID. I'm not sure that the result would be any more correct, though. What do you think?

gabelton: Did we say that measure sid is unique across workbaskets? From quickly looking at `MeasureCreationPattern` again I would say it is. In which case, the sid is probably enough as a parameter. Perhaps we'd want to explicitly order by transaction order, to check that we're getting the latest version of that measure?

paulpepper-trade: Yes, I think SIDs are unique across the DB instance. Good call on the transaction ordering - I've add that.

gabelton (approved):

PR #685, opened 2022-09-20
TP2000-513: Rule violations sorting

eadpearce:
# TP2000-513: Rule violations sorting

## Why

* We want users to be able to sort by certain fields in the rule violations list view

## What

* Adds clickable sort icons to columns on the violations page to sort by that field
* Clicking the sort icon/column heading a second time will toggle the sort order between ascending/descending

Screenshot 2022-09-20 at 11 37 46

## Checklist
- Requires migrations? No


common/jinja2/components/sort_icon.jinja
```diff
@@ -0,0 +1,10 @@
+<svg width="20" height="19" viewBox="0 0 20 19" fill="none" xmlns="http://www.w3.org/2000/svg">
```
paulpepper-trade, line 1: Would this be better located in a .svg file - akin to keeping CSS in stylesheets. I think [our webpack bundling](https://github.com/uktrade/tamato/blob/1be2498a2ea094b9fad04f8b26b5d6b208bb3bc0/webpack.config.js#L36) should handle that.

workbaskets/jinja2/workbaskets/violations.jinja
```diff
@@ -31,13 +31,39 @@
   ]) or "" }}
 {% endfor %}
 
+{% macro sort_by(sort_by, title) %}
```
paulpepper-trade, line 34: Indentation to help with readability?

workbaskets/views/ui.py
```diff
@@ -470,14 +469,41 @@ def get_context_data(self, **kwargs):
...
+    def get_context_data(self, **kwargs):
+        return super().get_context_data(workbasket=self.workbasket, **kwargs)
+
+    def get_queryset(self):
+        qs = TrackedModelCheck.objects.filter(
+            transaction_check__transaction__workbasket=self.workbasket,
+            successful=False,
+        )
+        sort_by = self.request.GET.get("sort_by")
+        direction = self.request.GET.get("d")
```
paulpepper-trade, line 493: For readability and to self-doc this param, perhaps name it "direction" or "order".

workbaskets/views/ui.py
```diff
@@ -470,14 +469,41 @@ def get_context_data(self, **kwargs):
...
+        sort_by_fields = ["model", "date", "check_name"]
+        if sort_by and sort_by in sort_by_fields:
+            if sort_by == "date":
+                qs = qs.order_by("transaction_check__transaction__created_at")
+            elif sort_by == "model":
+                qs = qs.order_by("model__polymorphic_ctype")
+            else:
+                qs = qs.order_by(sort_by)
```
paulpepper-trade, line 501: Won't this 500 error if a non-sortable attribute value is given (unvalidated input)?

paulpepper-trade (changes requested):

workbaskets/views/ui.py
eadpearce: no `if sort_by and sort_by in sort_by_fields:`. there is a test for this too

paulpepper-trade (approved):

PR #686, opened 2022-09-22
TP2000-519: Fix measure condition link

eadpearce:
# TP2000-519: Fix measure condition link

## Why

* Measure condition get_url used the measure condition sid to create the url for the linked measure and this was causing a 404

## What

* Use the dependent_measure sid instead when generating the url

### DIFF
--- a/measures/models.py
+++ b/measures/models.py
@@ -1,7 +1,9 @@
 from datetime import date
+from typing import Optional
 from typing import Set
 
 from django.db import models
+from django.urls import reverse
 
 from common.business_rules import UniqueIdentifyingFields
 from common.business_rules import UpdateValidity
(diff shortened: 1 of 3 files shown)


measures/models.py
```diff
@@ -816,6 +818,14 @@ def condition_string(self) -> str:
     def duty_sentence(self) -> str:
         return MeasureConditionComponent.objects.duty_sentence(self)
 
+    def get_url(self) -> Optional[str]:
```
paulpepper-trade, line 821: Looking at [DescriptionMixin.get_url()](https://github.com/uktrade/tamato/blob/b94019215ba169776b601ce653ea7e5157b4d06d/common/models/mixins/description.py#L47), it handles list view URLs too. Is there also a case for that here, or is that already handled somehow, or do we never require that?

eadpearce: There's no point adding list view because we don't have a list view for measure conditions. We'd just end up using the measures list view which would probably not be helpful

paulpepper-trade (approved):

PR #688, opened 2022-09-26
TP2000-515 Newly created conditions don't display in the two measure filter displays, only when going in to the details.

gabelton:
# TP2000-515 Newly created conditions don't display in the two measure filter displays, only when going in to the details.

## Why

Newly created measures show in the review tab, but their conditions are missing from the conditions column.
## What

Use `current` instead of `latest_approved` for footnotes and conditions in jinja, so that newly created objects are visible

Screenshot 2022-09-26 at 13 31 12

### DIFF
--- a/workbaskets/tests/test_views.py
+++ b/workbaskets/tests/test_views.py
@@ -531,7 +531,7 @@ def test_workbasket_measures_review_conditions(valid_user_client):
     )
     factories.MeasureFactory.create_batch(5)
     certificate = factories.CertificateFactory.create()
-    tx = transaction = workbasket.new_transaction()
+    tx = workbasket.new_transaction()
     measure = factories.MeasureFactory.create(transaction=tx)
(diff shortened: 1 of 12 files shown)


workbaskets/tests/test_views.py
```diff
@@ -525,6 +525,32 @@ def test_workbasket_measures_review_pagination(
     assert measure_sids.issubset({str(m.sid) for m in workbasket_measures})
 
 
+def test_workbasket_measures_review_conditions(valid_user_client):
+    workbasket = factories.WorkBasketFactory.create(
+        status=WorkflowStatus.EDITING,
+    )
+    factories.MeasureFactory.create_batch(5)
+    certificate = factories.CertificateFactory.create()
+    tx = transaction = workbasket.new_transaction()
```
paulpepper-trade, line 534: Extra `transaction` assignment typo?

paulpepper-trade (approved):

PR #689, opened 2022-09-27
TP2000-531 - Prevent zombie rule checks

paulpepper-trade:
# TP2000-531 - Prevent zombie rule checks

## Why

Workbaskets are showing running checks without users knowingly starting them. The tasks don't seem to complete and there is no sign of the associated celery task.

## What

The root cause of the Why (above) isn't immediately obvious given the information available to us in the production environment. This PR is an attempt to remedy that and possibly fix the issue along the way by doing:

- Initiate rule check execution via a POST HTTP request in place of the current use of GET.
- Use Celery’s `CELERY_TASK_TRACK_STARTED` to allow acquiring task start info.
- Preserve Celery tasks across restarts using `CELERY_RESULT_PERSISTENT`.
- Add logging around new check / task creation.
- Prevent accidental multiple clicks via `govukButton`'s `preventDoubleClick` attribute and related debouce mechanism. 

## Checklist
- Requires migrations? No


gabelton (approved):
All really smart changes. Great to have those celery settings added. What's the thinking behind changing from GET to POST? (Not implying that we shouldn't, just wasn't sure of the rationale)

paulpepper-trade:
> What's the thinking behind changing from GET to POST? (Not implying that we shouldn't, just wasn't sure of the rationale)
Thanks, Gabe. A POST is the way to go when creating, updating or deleting server-side state such as is involved in running rules checks. A POST helps avoid mishap (a POST is more difficult to accidentally generate than a GET - form submission versus simply visiting a URL), is more secure (XSS) and they're manged differently by proxies for this kind update/change (state-changing POST requests shouldn't be cached).

PR #690, opened 2022-09-30
Tops 645 nig2 parent lookup update

dougmills-DIT:
# TOPS-645

## background
NIG2 and the wording surrounding it is a bit of a grey area, we have however observed data coming from the EU that gave us insights that the interpretation we have differs from the interpretation the EU has. This is causing issues when we are getting data from the EU. 

The rule states: "The validity period of the goods nomenclature must be within the validity period of the product line above in the hierarchy."

## currently
NIG2 will fail if the "current" parent does not encapsulate the valid between date period, so if a child good nomenclature is valid between 1/1/20202 and 1/2/2022 and the parent is valid between 1/1/2022 and 15/1/2022 (even if there is another parent that would be next in the chain with a valid period of 1/1/20202 and 1/2/2022 or more) the validation will fail (as of today) note : the rule is looking at historical dates!

## proposed
a child is valid if it has valid parents (note plural) from today forward, not looking at the start date of the child but from todays date forward, looking at all valid parents (which we are defining as the grandparents children, where the item id is less than or equal to the child's item id, or where the item ids are equal but the suffix is less than the child's suffix). We then join all the parents valid between dates where they overlap to create a valid date range that the child valid range can be compared against - if the child's valid range falls within the joint parents range - the rule passes

### DIFF
--- a/commodities/business_rules.py
+++ b/commodities/business_rules.py
@@ -84,7 +84,10 @@ class NIG2(BusinessRule):
             self.transaction,
         ).valid_between
 
-        if not child_validity.upper_inf and child_validity.upper < datetime.today().date():
+        if (
+            not child_validity.upper_inf
+            and child_validity.upper < datetime.today().date()
+        ):
             return True
 
         multi_parent_validity = None
@@ -108,7 +111,6 @@ class NIG2(BusinessRule):
 
         if child_start_date < datetime.today().date():
             child_start_date = datetime.today().date()
-
 
         child_validity = TaricDateRange(child_start_date, child_validity.upper)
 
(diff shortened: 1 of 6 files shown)


TomMacca:
@gabelton Would you mind having a look at this when you get a moment, sometime when you're back? You had some great points (e.g. about child dates using todays dates) in our last convo, do you think the latest updates here redress them? Thanks Gabe :)

commodities/business_rules.py
```diff
@@ -49,6 +51,71 @@ def parent_spans_child(self, parent, child) -> bool:
         ).valid_between
         return validity_range_contains_range(parent_validity, child_validity)
 
+    def parent_spans_childs_future(self, parent, child) -> bool:
```
gabelton, line 54: Can this method be deleted now? I don't see it being called

TomMacca (approved):

commodities/business_rules.py
TomMacca: Yep good catch

PR #691, opened 2022-09-30
TP2000-534 CertificateDoesNotExist error on measure/create/conditions step

gabelton:
# TP2000-534 CertificateDoesNotExist error on measure/create/conditions step

## Why

It shouldn't be possible to create measures with deleted certificates or certificates currently in a different, EDITING workbasket.

## What

Updates `MeasureConditionsFormMixin.required_certificate` to use current queryset.

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -272,7 +272,7 @@ class MeasureConditionsFormMixin(forms.ModelForm):
     )
     required_certificate = AutoCompleteField(
         label="Certificate, licence or document",
-        queryset=Certificate.objects.all(),
+        queryset=Certificate.objects.current(),
         required=False,
     )
     action = forms.ModelChoiceField(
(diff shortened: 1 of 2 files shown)


measures/forms.py
```diff
@@ -272,7 +272,7 @@ class Meta:
     )
     required_certificate = AutoCompleteField(
         label="Certificate, licence or document",
-        queryset=Certificate.objects.all(),
```
paulpepper-trade, line 275: It's probably worth us doing an audit of our queryset instances to ensure the correct instances are being made available!

paulpepper-trade (approved):
LGMT.

measures/forms.py
gabelton: Definitely. I think there's a ticket like that in the tech backlog to replace all `approved_up_to_transaction` with `current` where applicable.

PR #696, opened 2022-10-06
TP2000-537 Restrict dump transactions to checked workbaskets

paulpepper-trade:
# TP2000-537 Restrict dumps to checked workbaskets

## Why

Normally, it's expected that `dump_transactions` is only called on workbaskets that have had a complete and successful business rule check - i.e. those that are error-free. However, it's currently possible to accidentally export an unchecked workbasket leading to subsequence processing problems when using exported envelope. 

## What

This PR restricts dumping workbaskets to those that are known to have been checked. A flag, `--force-unchecked-rules`, has been added to allow explicit override of this behaviour for more uncommon situations. 

## Checklist
- Requires migrations? No

### DIFF
--- a/exporter/management/commands/dump_transactions.py
+++ b/exporter/management/commands/dump_transactions.py
@@ -72,6 +72,13 @@ class Command(BaseCommand):
             action="store_true",
         )
 
+        parser.add_argument(
+            "--force-unchecked-rules",
+            help="Force export of workbaskets that have not had complete and successful business rule checks.",
+            default=False,
+            action="store_true",
+        )
+
     @atomic
     def handle(self, *args, **options):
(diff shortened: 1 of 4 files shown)


gabelton:
Looks v clean and makes sense to me. 

Would be good to expose the `submit_for_approval` method to tariff managers, as that would make the workflow a bit smoother for devs doing the packaging. I know that currently I tend to package an EDITING basket, once I see its corresponding Jira ticket in the packaging column, rather than moving it through to PROPOSED. I think this is all taking us in the right direction, though, as moving a basket to PROPOSED after CDS have already ingested it is obviously not right.

gabelton (approved):

PR #698, opened 2022-10-13
TP2000-482 Real edits not new updates in workbaskets

paulpepper-trade:
# TP2000-482 Real edits not new updates

## Why

Currently making edits to entities in a workbasket (with EDITING status) creates new UPDATE instances - multiple versions of the same TARIC object. This can lead to users running into editing difficulties due to the interleaved, non-atomic nature of these objects. For instance, changing a measure UPDATE instance results in a new UPDATE, causing difficulties when trying to find and remove business rule breaking instances. 

## What

While a workbasket is in EDITING status, its TARIC objects have not been published and can therefore be edited directly - i.e. updating the instance's row in the database rather than creating a new row, representing a new version.

This PR implements real edits on TARIC objects in EDITING workbaskets This requires:

- Adding new URL routes for editing TARIC UPDATE and CREATE objects.
- Support distinguishing between creating TARIC CREATE instances and editing TARIC CREATEs.
- Adding new View classes for the real edits.
- Adding new Form classes for the real edits.
- Fix unit tests that break as a result of new versions not being created when editing objects in EDITING workbaskets.

## Checklist
- Requires migrations? No


common/models/trackedmodel.py
```diff
@@ -636,6 +636,17 @@ def get_url(self, action: str = "detail") -> Optional[str]:
         if action not in ["list", "create"]:
             kwargs = self.get_identifying_fields()
         try:
+            if (
+                action == "edit"
+                and self.transaction.workbasket.status == WorkflowStatus.EDITING
+            ):
+                # Edits in WorkBaskets that are in EDITING state get real
+                # changes via DB updates, not newly created UPDATE instances.
+                if self.update_type == UpdateType.CREATE:
```
gabelton, line 645: Why do we need to distinguish between an edit for an object with update_type=CREATE and one with update_type=UPDATE ? Isn't it the same operation from a Django perspective and the same journey from a user perspective (i.e. visiting the edit form view) ?

paulpepper-trade: I think there's a difference. When editing an object whose `update_type` is `CREATE`, the user should be able to edit all fields as they would when first creating that object. However, when editing an `UPDATE` object, not all fields are editable in the same way. For instance, an additional code's ID and type are immutable once it has been created (and published), so the edit form for the UPDATE should prevent those fields from changing.

additional_codes/views.py
```diff
@@ -161,10 +195,24 @@ def get_initial(self):
     template_name = "common/create_description.jinja"
 
 
+class AdditionalCodeDescriptionEditCreate(
```
gabelton, line 198: Could you add docstrings like the one you added to `CertificateCreate` ? I find it really helpful just to be able to check exactly what update_type each object is going to be created with

gabelton:
Is it worth adding some view / form tests? I'd imagine you could create something generic like we have for testing the delete endpoints. But would be good to just to check that things are being created in the correct transaction, the transaction count isn't being altered, and objects have the right update_type

additional_codes/tests/test_views.py
```diff
@@ -49,6 +50,48 @@ def test_additional_code_create_form(use_create_form, new_data, expected_valid):
...
+    to update the end date day, month, and year to an earlier date.
+
+    We expect a later end date to fail because the validity period extends
+    beyond that of the footnote type. We test end date, rather than start_date
+    because it is not possible to edit the start date through the view without
+    separately updating the description start date beforehand.
+    """
+    # AdditionalCodeType instance must be published for the AdditionCode edit
+    # forms to validate correctly.
+    approved_wb = factories.ApprovedWorkBasketFactory.create()
```
gabelton, line 72: You might be able to use the `approved_workbasket` fixture, to save a line or two

additional_codes/tests/test_views.py
```diff
@@ -49,6 +50,48 @@ def test_additional_code_create_form(use_create_form, new_data, expected_valid):
...
+@pytest.mark.parametrize(
+    ("data_changes", "expected_valid"),
+    (
+        ({}, True),
+        ({"code": ""}, False),
+    ),
+)
+def test_additional_code_edit_create_view(data_changes, expected_valid, use_edit_view):
+    """
+    Tests that footnote update view allows an empty dict and that it is possible
```
gabelton, line 62: --> additional code

additional_codes/tests/test_views.py
```diff
@@ -49,6 +50,48 @@ def test_additional_code_create_form(use_create_form, new_data, expected_valid):
...
+    to update the end date day, month, and year to an earlier date.
+
+    We expect a later end date to fail because the validity period extends
+    beyond that of the footnote type. We test end date, rather than start_date
```
gabelton, line 66: see above

additional_codes/views.py
```diff
@@ -214,6 +222,8 @@ class AdditionalCodeDescriptionUpdate(
     TrackedModelDetailMixin,
     CreateTaricUpdateView,
 ):
+    """UI endpoint for AdditionalCodeDescription CREATE instances."""
```
gabelton, line 225: Should this be UPDATE instances?

gabelton (approved):

PR #700, opened 2022-10-20
TP2000-551 Map current workbasket state transitions

gabelton:
…cture docs, including wb state diagram

# TP2000-551 Map current workbasket state transitions

## Why

The docs are missing info about the workbaskets module and workbaskets as a concept more generally.
## What

Adds some info to our docs, including a state map diagram, a glossary entry, and some info in the architecture section


paulpepper-trade:
Did you have a newer version of the (current) state diagram, @gabelton ? I know it'll be replaced with another, future version when WorkBasket.status is tweaked.

paulpepper-trade (approved):

PR #702, opened 2022-10-24
Remove NER check. Too many false positives

eadpearce:
## Why

* Too many false positives
* Wastes developer time adding lines to the exclude file
* Not adding any value

## What

* Remove the Named Entity Recognition check from pre-commit hooks
* Remove the NER exclude file


paulpepper-trade:
In light of the [acceptance that the PII-NER checker needs to go back into discovery](https://ditdigitalteam.slack.com/archives/C24S29EN9/p1669281976950609?thread_ts=1669222752.396279&cid=C24S29EN9) (and the various woes that have been flagged as reason for that), I think this can probably be merged into our main line.

PR #703, opened 2022-10-25
TP2000-571 Cannot import QuotaDefinition with description greater than 500 characters

gabelton:
# TP2000-571 Cannot import QuotaDefinition with description greater than 500 characters

## Why

Some files from the EU contain quota definitions with descriptions longer than 500 chars
## What

- Increase the character limit from 500 to 1000 for both importers
- Ensure that two importers use different xsd files in validation
- Ensure that non-commodities importer accepts files with envelope ids of 5 or 6 digits


TomMacca (approved):
Looks great & good test!

PR #704, opened 2022-10-26
Added ON4 business rule and updated ME6

dougmills-DIT:
updated ME6 business rule to prevent nulls

# TP2000-558

## Why
ON4 was missing and has caused issues historically
ME6 (based on taric spec) A measure must be linked to a good id, but we have 300k of unlinked measures. This is an attempt to stop this happening going forward, then we can tidy up the historic issues

## What
Added ON4 business rule
updated ME6 business rule to prevent nulls

### DIFF
--- a/quotas/business_rules.py
+++ b/quotas/business_rules.py
@@ -62,6 +62,19 @@ class ON2(BusinessRule):
             raise self.violation(order_number)
 
 
+class ON4(BusinessRule):
+    """The referenced geographical area must exist."""
+
+    def validate(self, order_number):
+        if (
+            order_number.origins.approved_up_to_transaction(
+                order_number.transaction,
+            ).count()
+            == 0
+        ):
+            raise self.violation(order_number)
+
+
 class ON5(BusinessRule):
     """There may be no overlap in time of two quota order number origins with
     the same quota order number SID and geographical area id."""
--- a/quotas/tests/test_business_rules.py
+++ b/quotas/tests/test_business_rules.py
@@ -54,6 +54,26 @@ def test_ON2(date_ranges, existing_range, new_range, ranges_overlap):
         business_rules.ON2(order_number.transaction).validate(order_number)
 
 
+def test_ON4_pass(date_ranges, approved_transaction, unapproved_transaction):
+    order_number = factories.QuotaOrderNumberFactory.create(
+        valid_between=date_ranges.normal,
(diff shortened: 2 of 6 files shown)


quotas/business_rules.py
```diff
@@ -62,6 +62,19 @@ def validate(self, order_number):
             raise self.violation(order_number)
 
 
+class ON4(BusinessRule):
```
gabelton, line 65: Could this inherit from `MustExist` or `MustExistNotNull` ?

quotas/tests/test_business_rules.py
```diff
@@ -54,6 +54,26 @@ def test_ON2(date_ranges, existing_range, new_range, ranges_overlap):
...
+    )
+
+    assert business_rules.ON4(order_number.transaction).validate(order_number) is None
+
+
+def test_ON4_fail(date_ranges, approved_transaction, unapproved_transaction):
+    order_number = factories.QuotaOrderNumberFactory.create(
+        valid_between=date_ranges.normal,
+        transaction=approved_transaction,
+        origin=None,
```
gabelton, line 70: I think this should be `origins=[]` or something similar because origin isn't a field on `QuotaOrderNumber`

quotas/business_rules.py
dougmills-DIT: This is a special case where the relationship is a many to many. I didn't think that this would work the same way, as the must Exist not null and must exist relies on a 1 to many relationship from what i can see.

quotas/tests/test_business_rules.py
dougmills-DIT: The factory has an origin property, that is added as part of the construct - setting it to null overwrites this and does not add one.

dougmills-DIT:
> Do we know what the thinking was behind the original `MustExist` class? That one explicitly accepts None values. Why was that desirable?

in the taric spec there are statements that say the related object exists, and some that define the related object as must exist - I believe this is the defining difference between allowed nulls and disallowed nulls in checking business rules. 

MustExist only checks if its populated, the record link must be a valid one FK wise

quotas/tests/test_business_rules.py
gabelton: Oh sorry, my bad!

gabelton (approved):

PR #707, opened 2022-10-31
TP2000-508 multiple measures delete

LaurenQurashi:
# TP-2000-508 Multiple Measures Delete 

## Why

In order for TM's to delete multiple measures in one go, I have added the functionality and templates to allow TM's to select the measures they wish to delete from the list view, confirm that they want to delete these measures and delete them.

## What
- Amend the List page to show checkboxes next to each item, make them selectable, and show their start and end dates. 
- Create a MultipleMeasuresDelete view which is responsible for showing the measures the users have chosen on a review page, and deleting those measures. 
- Added templates to accommodate the designs. 
- Added tests for the SelectableObjectsForm as it was missing them.

Screenshot 2022-11-25 at 14 33 43
Added checkboxes to each item, along with start and end dates for the measures on the list page. 

Screenshot 2022-11-25 at 14 34 51
Added Delete multiple measures "review" page, where users can see which measures they've selected to mass delete in the Details component and delete functionality to remove the measures. 

Wider issues noted that are to be separate tickets: 
- No measures should be shown on the page without being filtered first. We should add a new page with just the filter on, which redirects to a new page showing the filtered list (considered the easiest approach with how the filtering is set up) 
- Confirmation of deletion pages should be added to all

## Checklist
- Requires migrations? - No

### DIFF
--- a/measures/jinja2/includes/measures/list.jinja
+++ b/measures/jinja2/includes/measures/list.jinja
@@ -1,41 +1,92 @@
 {% from 'macros/create_link.jinja' import create_link %} 
 {% from "includes/measures/conditions.jinja" import conditions_list %}
+{% from "components/button/macro.njk" import govukButton %}
 
-{% set table_rows = [] %}
-{% for measure in object_list %}
-  {% set measure_link -%}
-    <a class="govuk-link govuk-!-font-weight-bold" href="{{ measure.get_url() }}">{{measure.sid}}</a>
-  {%- endset %}
-  {% set footnotes %}
-    {% for footnote in measure.footnotes.all() %}
-        {{ create_link(footnote.get_url(), footnote|string) }}{{ ", " if not loop.last else "" }}
-    {% endfor %}
-  {% endset %}
-  {{ table_rows.append([
-    {"html": measure_link},
-    {"text": measure.measure_type.sid ~ " - " ~ measure.measure_type.description},
-    {"text": measure.goods_nomenclature.item_id|wordwrap(2)|replace("\n", " ") if measure.goods_nomenclature else '-', "classes": "govuk-!-width-one-eighth"},
-    {"text": measure.duty_sentence if measure.duty_sentence else '-'},
-    {"html": create_link(url("additional_code-ui-detail", kwargs={"sid": measure.additional_code.sid}), measure.additional_code.type.sid ~ measure.additional_code.code) if measure.additional_code else '-'},
-    {"html": create_link(url("geo_area-ui-detail", kwargs={"sid": measure.geographical_area.sid}), measure.geographical_area.area_id ~ " - " ~ measure.geographical_area.get_description().description) if measure.geographical_area else '-'},
-    {"text": create_link(measure.order_number.get_url(), measure.order_number.order_number) if measure.order_number else '-'},
-    {"text": footnotes if measure.footnotes.all() else '-'},
-    {"text": conditions_list(measure) if measure.conditions.latest_approved() else "-", "classes": "govuk-!-width-one-quarter"},
-  ]) or "" }}
-{% endfor %}
+{# Sets out checkbox #}
+{% macro checkbox_item(field) -%}
+  {% set id = field.id_for_label %}
+  {% set name = field.name %}
+  {% set checked = "checked" if field.value() else "" %}
+  <div class="govuk-form-group">
+    <fieldset class="govuk-fieldset">
+      <div class="govuk-checkboxes govuk-checkboxes--small" data-module="govuk-checkboxes">
+        <div class="govuk-checkboxes__item">
+          <input class="govuk-checkboxes__input" id="{{ id }}" name="{{ name }}" type="checkbox" {{ checked }} data-check-trackedmodel>
+          <label class="govuk-label govuk-checkboxes__label" for="{{ id }}"></label>
+        </div>
+      </div>
+    </fieldset>
+  </div>
+{%- endmacro %}
 
-{{ govukTable({
-  "head": [
-    {"text": "ID"},
-    {"text": "Type"},
-    {"text": "Commodity code"},
-    {"text": "Duties"},
-    {"text": "Additional code"},
-    {"text": "Geographical area"},
-    {"text": "Quota"},
-    {"text": "Footnote"},
-    {"text": "Conditions"},
-  ],
-  "rows": table_rows,
-  "classes": "govuk-table-m"
-}) }}
+{# Sets out checking all checkbox #}
+{% set checkbox_check_all -%}
+  <div class="govuk-form-group">
+    <fieldset class="govuk-fieldset">
+      <div class="govuk-checkboxes govuk-checkboxes--small">
+        <div class="govuk-checkboxes__item">
+          <input class="govuk-checkboxes__input" id="select-all" name="select-all" type="checkbox" data-check-all="data-check-trackedmodel">
+          <label class="govuk-label govuk-checkboxes__label govuk-!-padding-right-0" for="selected"></label>
+        </div>
+      </div>
+    </fieldset>
+  </div>
+{%- endset %}
+
+{# sets out form #}
+<form method="post">
+  <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrf_token }}">
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      {% if form.fields %}
+        {% set table_rows = [] %}
+        {% for field in form %}
+          {% set checkbox = checkbox_item(field) %}
+          {% set measure = field.field.obj %}
+            {% set measure_link -%}
+              <a class="govuk-link govuk-!-font-weight-bold" href="{{ measure.get_url() }}">{{measure.sid}}</a>
+            {%- endset %}
+            {% set footnotes %}
+              {% for footnote in measure.footnotes.all() %}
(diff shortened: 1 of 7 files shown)


gabelton:
Could you add some screenshots?

measures/views.py
```diff
@@ -537,3 +577,62 @@ class MeasureDelete(
...
     form_class = forms.MeasureDeleteForm
     success_path = "list"
+
+
+class MeasureMultipleDelete(TemplateView, ListView):
+    """UI for user review and deletion of multiple Measures."""
+
+    template_name = "measures/delete-multiple-measures.jinja"
+
+    def _workbasket(self):
```
gabelton, line 587: Do you need this? Could you use `WorkBasket.current()` instead?

measures/views.py
```diff
@@ -537,3 +577,62 @@ class MeasureDelete(
 ):
     form_class = forms.MeasureDeleteForm
     success_path = "list"
+
+
+class MeasureMultipleDelete(TemplateView, ListView):
```
gabelton, line 582: Lots of this is borrowed from `WorkBasketDeleteChanges`. Could you add a mixin that both classes can inherit from?

measures/tests/test_views.py
```diff
@@ -113,6 +113,47 @@ def test_measure_delete(use_delete_form):
...
+
+    url = reverse("measure-ui-delete-multiple")
+    client.force_login(valid_user)
+    session = client.session
+    session["workbasket"] = {
+        "id": session_workbasket.pk,
+        "status": session_workbasket.status,
+        "title": session_workbasket.title,
+    }
+    session.update(
```
gabelton, line 132: could you update the workbasket at the same time as updating DELETE_MEASURE_SELECTIONS?

measures/tests/test_views.py
```diff
@@ -113,6 +113,47 @@ def test_measure_delete(use_delete_form):
...
+        {
+            "DELETE_MEASURE_SELECTIONS": {
+                measure_1.pk: True,
+                measure_2.pk: True,
+                measure_3.pk: True,
+            },
+        },
+    )
+    session.save()
```
gabelton, line 141: Do you definitely need to save it? (You probably do. I know sessions are fussy, especially in unit tests)

measures/tests/test_views.py
```diff
@@ -113,6 +113,47 @@ def test_measure_delete(use_delete_form):
     use_delete_form(factories.MeasureFactory())
 
 
+def test_multiple_measure_delete(client, valid_user, session_workbasket):
```
gabelton, line 116: nice test

measures/jinja2/includes/measures/list.jinja
```diff
@@ -1,41 +1,92 @@
...
-    {"text": measure.goods_nomenclature.item_id|wordwrap(2)|replace("\n", " ") if measure.goods_nomenclature else '-', "classes": "govuk-!-width-one-eighth"},
-    {"text": measure.duty_sentence if measure.duty_sentence else '-'},
-    {"html": create_link(url("additional_code-ui-detail", kwargs={"sid": measure.additional_code.sid}), measure.additional_code.type.sid ~ measure.additional_code.code) if measure.additional_code else '-'},
-    {"html": create_link(url("geo_area-ui-detail", kwargs={"sid": measure.geographical_area.sid}), measure.geographical_area.area_id ~ " - " ~ measure.geographical_area.get_description().description) if measure.geographical_area else '-'},
-    {"text": create_link(measure.order_number.get_url(), measure.order_number.order_number) if measure.order_number else '-'},
-    {"text": footnotes if measure.footnotes.all() else '-'},
-    {"text": conditions_list(measure) if measure.conditions.latest_approved() else "-", "classes": "govuk-!-width-one-quarter"},
-  ]) or "" }}
-{% endfor %}
+{# Sets out checkbox #}
```
LaurenQurashi, line 5: Okay so this is just a nice template update that adds checkboxes to each item etc

measures/jinja2/measures/delete-multiple-measures.jinja
```diff
@@ -0,0 +1,138 @@
+{% extends "layouts/layout.jinja" %}
```
LaurenQurashi, line 1: This is the template for making the Delete Measures "review" sort of page. I added breadcrumbs manually, not sure if that's the right thing to do but the automatic implementation that pre existed would skip out the Find and Edit measures link which was a bit off.

measures/views.py
```diff
@@ -59,12 +65,46 @@ def get_queryset(self):
...
 
 
-class MeasureList(MeasureMixin, TamatoListView):
+class MeasureList(MeasureMixin, FormView, TamatoListView):
     """UI endpoint for viewing and filtering Measures."""
 
-    paginator_class = MeasurePaginator
     template_name = "measures/list.jinja"
     filterset_class = MeasureFilter
+    form_class = SelectableObjectsForm
```
gabelton, line 73: Is this form worth unit testing now that it's being used in a couple of different places?

measures/tests/test_views.py
```diff
@@ -114,6 +114,109 @@ def test_measure_delete(use_delete_form):
     use_delete_form(factories.MeasureFactory())
 
 
+def test_multiple_measure_delete_functionality(client, valid_user, session_workbasket):
```
LaurenQurashi, line 117: This is cute I worked hard on this, this is a test that checks that the functionality of the view grabs measures from the session (emulating they'd been selected in list view) then makes a post to trigger the deleting. The test checks that the page redirects, clears the session once it's done deleting, and has changed the update type to Delete which is type 2.

measures/tests/test_views.py
```diff
@@ -114,6 +114,105 @@ def test_measure_delete(use_delete_form):
...
+
+    # on success, the page redirects to the list page
+    assert response.status_code == 302
+    assert client.session["DELETE_MEASURE_SELECTIONS"] == {}
+    for measure in workbasket_measures:
+        # check that the update type is delete which is 2
+        assert measure.update_type == 2
+
+
+def test_multiple_measure_delete_template(client, valid_user, session_workbasket):
```
LaurenQurashi, line 153: similar sort of test to above but checks that the stuff in the template is the same as what's stored on the measures in the session, and doesn't include ones that haven't been selected.

workbaskets/tests/test_forms.py
```diff
@@ -68,3 +71,54 @@ def test_workbasket_create_form_invalid_data():
     assert not form.is_valid()
     assert "This field is required." in form.errors["title"]
     assert "This field is required." in form.errors["reason"]
+
+
+def test_selectable_objects_form():
+    """Test that the SelectableObjectsForm creates a SelectableObjectField for
```
LaurenQurashi, line 77: This test made my brain hurt so I've added comments. let me know if you think it's too confusing.

workbaskets/tests/test_forms.py
```diff
@@ -68,3 +71,54 @@ def test_workbasket_create_form_invalid_data():
...
+    # create the object for us, so we just need to check it matches what we expect.
+    assert form.cleaned_data_no_prefix == {
+        measure_pks[0]: True,
+        measure_pks[1]: True,
+        measure_pks[2]: True,
+        measure_pks[3]: False,
+        measure_pks[4]: False,
+    }
+    # Check that the fields are of SelectableObjectField type
+    for item in measures:
```
gabelton, line 123: Could you combine this with the loop above rather than looping over measures twice?

gabelton: Actually there are three loops which look like they could be combined into one

measures/jinja2/measures/delete-multiple-measures.jinja
```diff
@@ -0,0 +1,138 @@
...
+    <div class="govuk-grid-column-two-thirds">
+      <p class="govuk-body"> You are about to delete your selected measures.</p>
+      {{ govukWarningText({
+        "text": "You should only delete a measure if end dating the measure isn't appropriate.",
+        "iconFallbackText": "Warning"
+      }) }}
+      <p class="govuk-body govuk-!-font-weight-bold" style="padding-left: 45px;">Check your selected measures meet this criteria before pressing "Delete Measures" as it could prevent the correct measure from being used.</p>
+      <div class="govuk-inset-text">
+        Your deleted measures will be added to your work basket as items.
+        <br>
```
eadpearce, line 44: Instead of breaks here could these two sentences be in separate p tags?

measures/tests/test_views.py
```diff
@@ -114,6 +114,115 @@ def test_measure_delete(use_delete_form):
...
+    url = reverse("measure-ui-delete-multiple")
+    response = client.get(url)
+
+    assert response.status_code == 200
+
+    soup = BeautifulSoup(str(response.content), "html.parser")
+
+    # grab the whole measure objects for our pk's we've got in the session, so we can compare attributes.
+    selected_measures = Measure.objects.filter(
+        pk__in=[key for key in session["DELETE_MEASURE_SELECTIONS"].items()],
```
eadpearce, line 197: You can just do `session["DELETE_MEASURE_SELECTIONS"].keys()` here without the list comprehension

measures/jinja2/measures/delete-multiple-measures.jinja
LaurenQurashi: idk what is wrong with me, thank you Edie

gabelton (approved):

PR #709, opened 2022-11-02
TP2000-585 Remove workbasket submit endpoint

gabelton:
# TP2000-585 Remove workbasket submit endpoint

## Why

We don't use this endpoint in the UI anymore and whatever we build as part of the automation epic will look so different that it's not worth keeping this one around as a blueprint.

## What

Removes the endpoint and the unused detail.jinja that called it. Also removes related unit tests.


workbaskets/views/ui.py
```diff
@@ -371,7 +339,7 @@ def run_business_rules(self):
 
     def get_success_url(self):
         form_action = self.request.POST.get("form-action")
-        if form_action in ("publish-all", "remove-selected"):
+        if form_action in ("remove-selected"):
```
paulpepper-trade, line 342: I guess this could now just be an equality check.
```suggestion
        if form_action == "remove-selected":
```

paulpepper-trade (approved):
LGTM.

PR #710, opened 2022-11-02
TP2000-583 ME40 gives false positives

gabelton:
# TP2000-583 ME40 gives false positives

## Why

ME40 gives false positives when run against deleted measures.
## What

Adds `skip_when_deleted` decorator to ME40


paulpepper-trade (approved):

dougmills-DIT:
I know its a pain but it needs a test around it

PR #711, opened 2022-11-03
TP2000-583 me40 gives false positives (pt. 2)

gabelton:
# TP2000-583 me40 gives false positives (pt. 2)

## Why

ME119 is now giving false positives
## What

Adds `skip_when_deleted` decorator to ME119 and adds some tests for that decorator and `skip_when_not_deleted`

The tests are not an ideal solution, but I spent quite a bit of time trying different things without success, so created this ticket to refactor at some point https://uktrade.atlassian.net/browse/TP2000-591

### DIFF
--- a/common/tests/test_business_rules.py
+++ b/common/tests/test_business_rules.py
@@ -12,6 +12,8 @@ from common.business_rules import NoBlankDescription
 from common.business_rules import NoOverlapping
 from common.business_rules import PreventDeleteIfInUse
 from common.business_rules import UniqueIdentifyingFields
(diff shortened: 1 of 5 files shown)


dougmills-DIT (approved):
A test for the business rules effected would be preferable (so that if the decorator is removed - a test fails against ME199) - I normally address tests directly against a bug raised but it looks good. Maybe add something to backlog - i would happily write the tests

gabelton:
I've added some specific tests for ME40 and ME119. Let me know if there's anything else you'd like adding and I can create a ticket.

PR #712, opened 2022-11-03
TP2000-569 cds error transition doesnt handle transaction state

gabelton:
# TP2000-569 cds error transition doesnt handle transaction state

## Why

The `WorkBasket.cds_error()` transition method correctly reverses the changes made by `submit_for_approval` as far as `version_group` is concerned. It does not, however, reverse the effects of `save_drafts`, so transactions remain in the wrong partition (REVISION).

## What

Add a `move_to_draft` method, to effectively reverse the changes made by `save_to_draft`.


common/models/transactions.py
```diff
@@ -175,6 +184,48 @@ def save_drafts(self, partition_scheme):
 
         self.move_to_end_of_partition(approved_partition)
 
+    def revert_current_version(self):
+        """Set current_version to previous version or None on contained tracked
```
paulpepper-trade, line 188: I'm unfamiliar with what "contained" means in the context of TrackedModels and this queryset. I assume it to mean transactions that are not in an approved partition. Perhaps a bit more explanation in the docstring?

common/models/transactions.py
```diff
@@ -175,6 +184,48 @@ def save_drafts(self, partition_scheme):
...
+            else:
+                version_group.current_version = versions.first()
+            version_group.save()
+
+    @atomic
+    def move_to_draft(self):
+        """
+        Save SEED_FILE or REVISION transactions as DRAFT.
+
+        Set current_version to previous version or None on contained tracked
```
paulpepper-trade, line 208: "contained" as previous comment.

common/models/transactions.py
```diff
@@ -175,6 +184,48 @@ def save_drafts(self, partition_scheme):
...
+        model version groups.
+        """
+        if self.unapproved().exists():
+            pks = self.values_list("pk")
+            msg = f"One or more Transactions was already in the DRAFT partition: {pks}"
+            logger.error(msg)
+            raise TransactionsAlreadyInDraft(msg)
+
+        if self.first() is None:
```
paulpepper-trade, line 217: Perhaps more conventional to use `QuerySet.exists()`?
Should be conditional block be come at the top of the method?

paulpepper-trade (approved):

common/models/transactions.py
gabelton: I guess its my own metaphor. I think transactions "containing" a TrackedModel and a workbasket containing transactions. There's probably a better way of describing it

PR #715, opened 2022-11-09
TP2000-604 Newly created Commodity Codes imported into a workbasket are not visible in the Measures journey

gabelton:
# TP2000-604 Newly created Commodity Codes imported into a workbasket are not visible in the Measures journey

## Why

The commodities dropdown in the measure create wizard doesn't show commodities with a start date in the future.
## What

Adds a `as_at_and_beyond` queryset method and uses this instead of `as_at` in `GoodsNomenclatureViewset`

### DIFF
--- a/commodities/tests/test_views.py
+++ b/commodities/tests/test_views.py
@@ -1,3 +1,4 @@
+import json
 from os import path
 from unittest.mock import patch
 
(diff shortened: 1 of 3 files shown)


paulpepper-trade (approved):
LGTM.

PR #716, opened 2022-11-14
TP2000-399 fix missing current versions

dougmills-DIT:
# relates to issues raised in TP2000-399 : fix underlying missing current version fix

## Why
This PR creates a migration that updates the current version where the current version is missing, allowing valid existing records to be updated correctly, where as before a new record would be created and result in business rule violations either at TAP or CDS

## What
Add migration to update current version programatically for all missing current versions (around 150 records)
Add tests to verify the migration works correctly

See: https://uktrade.atlassian.net/browse/TP2000-399


common/tests/test_migrations.py
```diff
@@ -0,0 +1,76 @@
...
+from common.validators import UpdateType
+
+
+@pytest.mark.django_db()
+def test_missing_current_version_fix(migrator):
+    migrator.reset()
+
+    """Ensures that the initial migration works."""
+    # migrator.apply_initial_migration(("measures", "0010_add_requires_to_action_and_accepts_to_condition_code"))
+    # migrator.apply_initial_migration(("workbaskets", "0005_workbasket_rule_check_task_id"))
```
gabelton, line 16: Do you still need these comments?

common/tests/test_migrations.py
```diff
@@ -0,0 +1,76 @@
...
+    version_group = version_group_class.objects.create()
+    workbasket = workbasket_class.objects.create(
+        title=f"xxx {datetime.time}",
+        reason="some reason",
+        author=user,
+    )
+    transaction = transaction_class.objects.create(
+        workbasket=workbasket,
+        order=1,
+        partition=2,
```
gabelton, line 43: Could you use TransactionPartition.REVISION to make this slightly more explicit?

requirements.txt
```diff
@@ -26,6 +26,7 @@ django-rest-polymorphic==0.1.9
 django-sequences==2.6
 django-staff-sso-client==3.1.1
 django-storages==1.11.1
+django-test-migrations==1.2.0
```
gabelton, line 29: Been wanting an excuse to check out this library...

gabelton (approved):
Where do you think these are coming from? I guess workbaskets must be getting created in a particular status and not transitioned through proposed to approved, then force published?

common/tests/test_migrations.py
dougmills-DIT: apologies, merged before reading this - will create another PR with this change

common/tests/test_migrations.py
dougmills-DIT: will remove in tidy up PR to follow

PR #724, opened 2022-11-15
Edit multiple measures

eadpearce:
# TP-2000-553: Edit multiple measures

## Why

* Users want to be able to edit many measures at a time with the same data e.g. end dates

## What

* Measures in the measures list page are selectable and can be edited by clicking the Edit measures button
* Reuses code from the measure create wizard and the current measure edit page to allow edits of multiple measures in a multi-step form
* Steps displayed in the form depends on which are checked in the first step

## Checklist
- Requires migrations? ❌

Screenshot 2022-11-15 at 14 52 40

Screenshot 2022-11-15 at 14 51 43


measures/conditions.py
```diff
@@ -0,0 +1,61 @@
...
+        return constants.ADDITIONAL_CODE in cleaned_data.get("fields_to_edit")
+
+
+def show_step_conditions(wizard):
+    cleaned_data = wizard.get_cleaned_data_for_step(constants.START)
+    if cleaned_data:
+        return constants.CONDITIONS in cleaned_data.get("fields_to_edit")
+
+
+def show_step_footnotes(wizard):
```
gabelton, line 46: Could these be refactored into one generic method?

measures/helpers.py
```diff
@@ -0,0 +1,183 @@
...
+from measures.models import FootnoteAssociationMeasure
+from measures.models import MeasureComponent
+from measures.models import MeasureCondition
+from measures.models import MeasureConditionComponent
+from measures.parsers import DutySentenceParser
+from measures.patterns import MeasureCreationPattern
+from measures.util import diff_components
+
+
+def update_measure(instance, transaction, current_workbasket, cleaned_data, defaults):
```
gabelton, line 12: I'm wondering if these methods should live on a class, perhaps a MeasureUpdatePattern or MeasureEditPattern, in patterns.py. I think they're all part of a similar process

measures/conditions.py
gabelton: After refactoring it might be worth considering whether these belong in their own file or whether they should sit on/above a class somewhere

eadpearce: Agree it'd be nice to be able to parametrise these with the different step names but due to how formtools is expecting them to be passed to the view (as a dictionary in urls.py) I don't think this is an easy fix. Maybe a future tech debt ticket?

measures/helpers.py
eadpearce: Yeah that would be good. I'll see if I can get that done before I go on leave at the end of the week

measures/conditions.py
gabelton: Ah yeah that's a bit awkward. I feel like there's a way to do it with lambdas, but might be worth creating ticket, as you say

measures/jinja2/measures/confirm-edit-multiple.jinja
```diff
@@ -0,0 +1,55 @@
+{% extends "layouts/confirm.jinja" %}
+{% from "components/breadcrumbs.jinja" import breadcrumbs %}
+
+{% set page_title = "Updated measures" %}
+
+{% macro panel_title(edited_measures) %}
```
dougmills-DIT, line 6: In my experience (and checking best practice for Django) logic in views should be kept to a minimum, can this be pre-populated from the model and simply displayed? the general rule is to keep views dumb, controllers thin and models fat

dougmills-DIT: agreed to let this one slide in tech chat :)

measures/views.py
```diff
@@ -59,12 +117,201 @@ def get_queryset(self):
...
+        edited_measures = []
+
+        footnote_pks = [
+            item["footnote"].pk
+            for item in cleaned_data.get("formset-footnotes", [])
+            if not item.get("DELETE")
+        ]
+        conditions_data = cleaned_data.get("formset-conditions", [])
+        for condition in conditions_data:
+            condition["update_type"] = UpdateType.CREATE
```
gabelton, line 282: What happens when we want to update existing conditions on the measure?

eadpearce: This form currently only supports adding additional conditions.

SteveCN7: I've been thinking about this as a separate ticket - and we need to support editing/removing geo area exclusions - I wonder if a small interim addition would be a "Remove" button to remove Geo Areas and Measure Conditions for now. And then (in the same form) to re-enter the correct information

gabelton: I'm not sure that the ability to add additional conditions is useful without the ability to remove or reorder the existing ones

measures/helpers.py
```diff
@@ -0,0 +1,183 @@
...
+            # Creating components in the same transaction as the new version
+            # of the measure minimises number of transaction and groups the
+            # creation of measure and related objects in the same
+            # transaction.
+            transaction,
+            MeasureComponent,
+            "component_measure",
+        )
+
+    return instance.new_version(current_workbasket)
```
gabelton, line 48: You need to pass the the fields you want to update to this method. i.e. `instance.new_version(current_workbasket, **cleaned_data)`

gabelton: You'll see:
`TypeError: Measure() got an unexpected keyword argument 'fields_to_edit'`
and then I imagine it will complain about M2M fields like conditions and duties, so you'll need to find a way to filter those out before passing the dict

PR #725, opened 2022-11-17
Tops 681 update to OverlappingQuotaDefinition Rule

dougmills-DIT:
# TOPS-681

## Why
 * OverlappingQuotaDefinition correctly filtered deleted models, but did not check latest version of model was current.
 * No test to verify if a Quota definition previously deleted would pass through the rule check 
 * A production update example caused this issue to prevent an update

## What
 * Updates the OverlappingQuotaDefinition business rule to check only current version of model

## Checklist
- Requires migrations? NO
- Requires dependency updates? NO

Links to relevant material
See: This change will unblock this ticket : https://uktrade.atlassian.net/browse/TOPS-831
See : Original request that identified the issue : https://uktrade.atlassian.net/browse/TOPS-681

### DIFF
--- a/quotas/business_rules.py
+++ b/quotas/business_rules.py
@@ -285,22 +285,26 @@ class OverlappingQuotaDefinition(BusinessRule):
     quota order number id."""
 
     def validate(self, quota_definition):
-        if (
+
+        potential_quota_definition_matches = (
             type(quota_definition)
-            .objects.approved_up_to_transaction(quota_definition.transaction)
+            .objects.approved_up_to_transaction(
+                quota_definition.transaction,
+            )
             .filter(
                 order_number=quota_definition.order_number,
                 valid_between__overlap=quota_definition.valid_between,
             )
-            .exclude(
-                sid=quota_definition.sid,
-                update_type=2,
-                trackedmodel_ptr_id=quota_definition.trackedmodel_ptr_id,
-            )
-            .exclude(update_type=2)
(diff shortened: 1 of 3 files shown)


dougmills-DIT:
> There's a `latest_deleted` queryset method that might be useful here

Its more that it find the create / update before the delete update - and fails on that

quotas/business_rules.py
```diff
@@ -285,17 +285,26 @@ class OverlappingQuotaDefinition(BusinessRule):
...
             .exclude(sid=quota_definition.sid)
-            .exists()
-        ):
-            raise self.violation(quota_definition)
+        )
+
+        if potential_quota_definition_matches.exists():
+            for potential_quota_definition in potential_quota_definition_matches:
+                if (
+                    potential_quota_definition
```
gabelton, line 304: In the normal course of things (running rules against an EDITING basket) will the current_version be set on the deleted object? I thought current version was set later (when moving basket from PROPOSED to APPROVED)

dougmills-DIT: yes, proposed > approved does set the current_version 

We are excluding the current quota definition here : 

.exclude(sid=quota_definition.sid)

it only looks for matches that are not the current SID, and the problem is that when (in history) you have a create / update and a delete, the rule would fail on the create / update, and not check if it was current version. thats why we need this update.

gabelton (approved):

PR #727, opened 2022-11-21
TP Changes so rhe application & tests can run locally via docker

a-gleeson:
## Why
Previously the set up was not working 100% to set up local development so this PR addresses testing & developing with docker locally

## What

Add make commands for docker commands, added test docker-compose that can run the tests with the correct permissions and configuration.
Make commands have been added to run commonly used commands


TomMacca:
Hey @a-gleeson , got TAP running sucessfully using `docker-compose up` so well done! I went through the readme you wrote in order to test the other steps and had a few problems. Thought I'd share them incase you're seeing them on other machines:

1. `make first-use` generates this error:  
`make: *** No rule to make target `first-use'.  Stop.`
2. `make docker-test` generates this error: 
```
bin/sh: -c: line 0: unexpected EOF while looking for matching `"'
/bin/sh: -c: line 1: syntax error: unexpected end of file
make: *** [docker-test] Error 2
```
3. `make docker-pytest` generates this error:
```
PermissionError: [Errno 13] Permission denied: '/home/tamato/app/.coverage'
ERROR: 1
make: *** [docker-pytest] Error 1
```

Dockerfile
```diff
@@ -12,39 +12,51 @@ FROM python:3.8-buster
...
 
-ENV DJANGO_SETTINGS_MODULE "settings"
+ARG ENV="prod"
+ENV ENV="${ENV}" \
+    PYTHONUNBUFFERED=1 \
+    PYTHONDONTWRITEBYTECODE=1\
+    PATH="${PATH}:/home/tamato/.local/bin" 
+
+# don't run as root
+RUN groupadd -g 1000 tamato && \
```
paulpepper-trade, line 22: Will this require a `tamato` user and group creating on the host environment?

Dockerfile
```diff
@@ -12,39 +12,51 @@ FROM python:3.8-buster
...
+    useradd -u 1000 -g tamato -m tamato
+
+# WORKDIR /home/tamato/app
```
paulpepper-trade, line 25: Dead comment?

Dockerfile
```diff
@@ -12,39 +12,51 @@ FROM python:3.8-buster
...
+RUN chown -R 1000:1000 /app
 
 # install python dependencies
-COPY requirements.txt ./
-RUN pip install -U pip && \
-    pip install -r requirements.txt --no-warn-script-location
+COPY requirements.txt requirements-dev.txt /app/
+RUN pip install --upgrade pip
+# Only install dev requirements in dev do not want to expose build tools in production
+# RUN  if [ "${ENV}" == "dev" ]; then \
```
paulpepper-trade, line 41: Dead comment?

Makefile
```diff
@@ -1,12 +1,21 @@
+COMPOSE_LOCAL=docker-compose 
+DOCKER=docker
 PROJECT=tamato
 DEV=true
 
+#default run can be replaced with exec
+DOCKER_RUN?=run --rm
+#db import file name
+DUMP_FILE?=tamato_db_seed.sql
```
paulpepper-trade, line 9: Could we just call the dump file tamato_db.sql?

Makefile
```diff
@@ -1,12 +1,21 @@
+COMPOSE_LOCAL=docker-compose 
+DOCKER=docker
```
paulpepper-trade, line 2: Any reason why these commands are referenced as variable?

Makefile
```diff
@@ -108,3 +125,89 @@ build-docs html:
...
+## docker-db-dump: Run db dump to import data into Docker
+docker-db-dump: docker-up-db
+	@echo "> Running db dump in docker..."
+	@cat ${DUMP_FILE} | ${COMPOSE_LOCAL} exec -T db psql -U postgres 
+
+## docker-first-use: Run application for first time in Docker 
+docker-first-use: docker-clean docker-build docker-down docker-db-dump docker-migrate docker-superuser docker-up 
+
+## docker-migrations: Run django makemigrations in Docker
+docker-migrations:
```
paulpepper-trade, line 168: `docker-makemigrations` to be clear?

paulpepper-trade (approved):
Vey nice :)

Dockerfile
a-gleeson: nope this is just in the container

Dockerfile
a-gleeson: broken code although we would want to add this in later so a production container does not install dev dependicies in the future. Not needed right now

Makefile
a-gleeson: if you need to change the compose file you are using you can edit it here and it is updated for all commands

Dockerfile
paulpepper-trade: Ah, everything is copied into the container rather than volume mapping the source files.

a-gleeson: although you still get live code updates so no need to rebuild the image!

PR #728, opened 2022-11-21
TP2000-608 Unit tests for api list views

a-gleeson:
# TP2000-608  Unit tests for api list views

## Why
Previously the ReadOnlyModelViewSet views did not have any units tests, this story ensures that this coverage has been addressed

## What
Unit testing API view sets

### DIFF
--- a/additional_codes/urls.py
+++ b/additional_codes/urls.py
@@ -14,7 +14,6 @@ api_router.register(
 api_router.register(
     r"additional_code_types",
     views.AdditionalCodeTypeViewSet,
-    basename="additionalcodetype",
 )
 
 detail = "<sid:sid>"
--- a/additional_codes/urls.py
+++ b/additional_codes/urls.py
@@ -11,7 +11,11 @@ api_router.register(
     views.AdditionalCodeViewSet,
     basename="additionalcode",
 )
-api_router.register(r"additional_code_types", views.AdditionalCodeTypeViewSet)
+api_router.register(
+    r"additional_code_types",
+    views.AdditionalCodeTypeViewSet,
+    basename="additionalcodetype",
+)
 
 detail = "<sid:sid>"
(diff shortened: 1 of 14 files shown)


workbaskets/tests/test_views.py
```diff
@@ -51,7 +51,10 @@ def test_workbasket_create_user_not_logged_in_dev_sso_disabled(client, settings)
     settings.ENV = "dev"
     settings.SSO_ENABLED = False
     settings.LOGIN_URL = reverse("login")
-    settings.MIDDLEWARE.remove("authbroker_client.middleware.ProtectAllViewsMiddleware")
+    if "authbroker_client.middleware.ProtectAllViewsMiddleware" in settings.MIDDLEWARE:
```
gabelton, line 54: Why is this change needed?

additional_codes/urls.py
```diff
@@ -11,7 +11,11 @@
     views.AdditionalCodeViewSet,
     basename="additionalcode",
 )
-api_router.register(r"additional_code_types", views.AdditionalCodeTypeViewSet)
+api_router.register(
+    r"additional_code_types",
+    views.AdditionalCodeTypeViewSet,
+    basename="additionalcodetype",
```
gabelton, line 17: Do we need to add basename args? It gets the same name from the view queryset, I think

workbaskets/tests/test_views.py
a-gleeson: When running tests locally this wasn't in the test middleware and would throw an exception. So I just added a check so it wouldn't fail.

additional_codes/urls.py
a-gleeson: Not needed, the ones I added are now removed

gabelton (approved):

PR #729, opened 2022-11-23
change test so it passes all the time

dougmills-DIT:
# No Ticket, small PR to fix test: 

## Why
checks/tests/test_querysets.py::test_current_queryset_returns_correct_results[check of approved transaction with greater approved check transaction] will intermittently fail, roughly 50% of the time. 

## What
* The test would fail 50% of the time (and has presumably for all of its existence) since the parameterised test would set up 2 transactions with the same order number. Incrementing the order number of the second created transaction in the test fixes the issue and the increate in coverage indicates here is additional code that is now being covered that was not before. 

## Checklist
Requires migrations? no
Requires dependency updates? no

### DIFF
--- a/checks/tests/test_querysets.py
+++ b/checks/tests/test_querysets.py
@@ -49,7 +49,7 @@ def test_current_queryset_returns_correct_results(
             approved=True,
             order=head_order + 1,
         )
-    assert Transaction.objects.all().approved.last() == latest
+    assert Transaction.objects.all().approved().last() == latest
 
     assert_current(check, expect_current)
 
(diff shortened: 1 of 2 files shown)


gabelton:
Why does this fix work?

paulpepper-trade:
This should use the PR template - Jira ticket (for tracking) with a description of what problem it addresses and how it solves it. It's useful for looking back at our commit history to understand main line changes.

dougmills-DIT:
@paulpepper-trade @gabelton I have updated the description now to reflect the logic / reasoning for the change and why it works. Essentially this test checks what is last() in transactions and this is based on order value. The test created a data scenario that allowed 2 records to have the same order value and would order intermittently correctly / incorrectly for what the test was expecting. Adding an additional +1 to the order value of the second transaction correctly presented the order and allowed the test to succeed consistently.

gabelton (approved):
Nice one!

PR #731, opened 2022-11-23
TP2000-618 MTPO Tariff Manager user journey - Create notification subsystem

gabelton:
# TP2000-618 MTPO Tariff Manager user journey - Create notification subsystem

## Why

Currently a developer will email CDS to let them know that a file has been uploaded to sharepoint and CDS will email back if that file has been ingested or rejected. We want to automate these three email scenarios, using govuknotify, and free up devs (and CDS) to do more interesting things.
## What

- Adds notifications app (including NotifiedUser and NotificationLog models with associated admin views)
- Adds a basic task for sending emails via python client

TODO
- need to create templates in govuk notify
- need to determine where and when these notifications will be sent. (this will have to wait until other parts of the automation journey are complete. And so it may be worthwhile to merge these changes, rather than waiting for this.)

- Requires migrations?  run `python manage.py migrate notifications`
- Requires dependency updates? adds govuk notify python client to requirements.txt


pii-ner-exclude.txt
```diff
@@ -1250,3 +1250,8 @@ Implementing RegulationLoadingReport
 loading_report
 Queued
 LoadingReport
+NotifiedUser
+NotificationLog
+template_id
+template_id"
+GOVUK Notify
```
paulpepper-trade, line 1257: I think this has been removed from our shared branch after pulling in latest changes from master.

paulpepper-trade (approved):
LGTM.

PR #733, opened 2022-11-25
update ME32 rule violation message to include violating measures

dougmills-DIT:
# TOPS-839
* I can only imagine how frustrating it is for a TM to have to ask a DE / Dev to find out which measure is overlapping when faced with an ME32 message but no other information. 

## Why
 * All errors should give enough information for a TM to process and update the workbasket in order to resolve a rule violation. This is the first attempt to expose key information that will make TMs work more autonomous 

## What
 * Really simple change, adding the measure SIDs to the end of the violation message
 * Example : There may be no overlap in time with other measure occurrences with a goods code in the same nomenclature hierarchy which references the same measure type, geo area, order number, additional code and reduction indicator. Clash with Measure SID 20198035. 

## Checklist
- Requires migrations? No

Links to relevant material
Ticket : TOPS-839 (https://uktrade.atlassian.net/browse/TOPS-839)

### DIFF
--- a/measures/business_rules.py
+++ b/measures/business_rules.py
@@ -361,15 +361,13 @@ class ME32(BusinessRule):
         if measure.goods_nomenclature is None:
             return
 
-        if self.clashing_measures(measure).exists():
-            message = (
-                "There may be no overlap in time with other measure occurrences with a goods code in the same "
-                "nomenclature hierarchy which references the same measure type, geo area, order number, "
-                "additional code and reduction indicator."
-            )
+        clashing_measures = self.clashing_measures(measure)
 
-            for clashing_measure in self.clashing_measures(measure):
-                message += f"Overlap detected with Measure SID {clashing_measure.sid}. "
+        if clashing_measures.exists():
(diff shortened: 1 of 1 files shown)


TomMacca:
Looks good to me and makes sense. All for better error messages in the UI.

measures/business_rules.py
```diff
@@ -362,7 +362,16 @@ def validate(self, measure):
             return
 
         if self.clashing_measures(measure).exists():
-            raise self.violation(measure)
+            message = (
```
gabelton, line 365: Could you use self.__doc__ ?

TomMacca (approved):

measures/business_rules.py
dougmills-DIT: did this : message = self.violation(measure).default_message() 

Thoughts?

dougmills-DIT: the default_message() method only grabs the first paragraph, so a little better than __doc__

gabelton: agreed

gabelton (approved):

PR #734, opened 2022-11-28
Remove unnecessary H1s on confirmation pages

dalecannon:
# TP2000-385 Multiple H1s on confirm create pages

## Why
Several confirmation pages contain multiple H1 tags. Such pages should only contain one, that being the title within a panel component confirming an action has been successfully completed. 

## What
Removes unnecessary H1 page title and confirmation span.

## Checklist
- Requires migrations? No

Screenshot 2022-11-28 at 10 00 57

### DIFF
--- a/common/jinja2/common/confirm_create_description.jinja
+++ b/common/jinja2/common/confirm_create_description.jinja
@@ -20,9 +20,6 @@
 {% block content %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <span class="govuk-caption-l">Confirmation</span>
-      <h1 class="govuk-heading-xl">{{ page_title }}</h1>
-
       {{ govukPanel({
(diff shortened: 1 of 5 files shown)


common/jinja2/common/confirm_create_description.jinja
```diff
@@ -20,9 +20,6 @@
 {% block content %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <span class="govuk-caption-l">Confirmation</span>
-      <h1 class="govuk-heading-xl">{{ page_title }}</h1>
```
eadpearce, line 24: Is this content okay to remove rather than changing HTML element type? To me it looks like it was extraneous anyway but want to make sure it has been okay'd with content people.

dalecannon: I've just double-checked and the content can be removed.

PR #737, opened 2022-12-02
Tops 815 add back missing measures

dougmills-DIT:
# TOPS-815 adding back three missing measures (TAP data only) 
workbasket 545 (and envelope ID 220185) had these measures in when sent to CDS, but were subsequently removed before committing to TAP. This update adds these back in so TMs can change them going forward. 

## Why
 * Adding back three measures, we need to make changes to 
 * Aligning to what we have sent across to CDS and CI 
 * Using a migration as this seems more controlled than hacking at console, plus its testable 

## What
 * Uses a migration to add back data to the database that was incorrectly removed historically

## Checklist
- Requires migrations? yes
--->

Links to relevant material
See: Original workbasket ticket : https://uktrade.atlassian.net/browse/TOPS-707
See: Fix ticket : https://uktrade.atlassian.net/browse/TOPS-815


dougmills-DIT:
note: tested against production data and migrated correctly, also ran business rules on workbasket and no issues found

conftest.py
```diff
@@ -128,6 +129,21 @@ def spanning_dates(request, date_ranges):
     )
 
 
+def setup_content_types(apps):
```
gabelton, line 132: Does this need the @pytest.fixture decorator?

measures/migrations/0012_add_back_three_missing_measures_already_published.py
```diff
@@ -0,0 +1,145 @@
...
+    # Guard clauses
+    if not WorkBasket.objects.filter(id=target_workbasket_id).exists():
+        return
+
+    target_work_basket = WorkBasket.objects.get(id=target_workbasket_id)
+
+    if not GeographicalArea.objects.filter(sid=146).exists():
+        return
+
+    geographical_area_canada = GeographicalArea.objects.get(sid=146)  # Canada
```
gabelton, line 30: Should these queries use `latest_approved` ?

measures/tests/test_migrations.py
```diff
@@ -0,0 +1,132 @@
...
+    target_workbasket_id = 545
+
+    measurement_class = old_state.apps.get_model("measures", "Measure")
+
+    assert measurement_class.objects.filter(sid=20194965).exists() is False
+    assert measurement_class.objects.filter(sid=20194966).exists() is False
+    assert measurement_class.objects.filter(sid=20194967).exists() is False
+
+    # mock up workbasket
+    new_work_basket = ApprovedWorkBasketFactory.create(id=target_workbasket_id).save()
```
gabelton, line 39: Does this migration target the original already-approved workbasket? If so, might be worth making some assertions about `current_version` and transaction partition (checking things are in REVISION, not DRAFT)

conftest.py
dougmills-DIT: No, it is technically a fixture, but wont accept the argument apps if its decorated

measures/tests/test_migrations.py
dougmills-DIT: Good point, will verify

measures/migrations/0012_add_back_three_missing_measures_already_published.py
dougmills-DIT: actually yes, - didn't work originally because of the way i loaded the classes but you are right. Will add. I have already checked and each only returns one result anyway but since I can use it now - will update.

measures/tests/test_migrations.py
dougmills-DIT: so the current version is correct for the three measures (as latest version), and the partition is in REVISION(2)
I might add an assert in the test so its clear to all

dougmills-DIT: ok, added checks to the test, resolve if you are happy :)

measures/migrations/0012_add_back_three_missing_measures_already_published.py
dougmills-DIT: updated now, resolve if you are happy

gabelton (approved):

PR #738, opened 2022-12-02
TP2000-628 Add WorkBasket transition APPROVED -->  EDITING

gabelton:
TP2000-628 Add WorkBasket transition APPROVED -->  EDITING

## Why

Currently when a workbasket is approved there is no way to restore it to EDITING status and make further changes prior to sending to CDS. Looking ahead to packaging automation (and potentially to HMRC review capabiltiy) It will be important to have this option.

## What

Adds method and updates unit test


a-gleeson (approved):
Looks good to me!

PR #741, opened 2022-12-06
TP2000-599 Quota details tab

eadpearce:
# TP2000-599 Quota details tab

## Why

* Show users more quota data

## What

* Adds an extra tab to the quota detail page for definition details

## Checklist
- Requires migrations? ✅

Screenshot 2022-12-06 at 16 12 52


quotas/views.py
```diff
@@ -83,9 +84,21 @@ class QuotaList(QuotaMixin, TamatoListView):
...
+class QuotaDetail(QuotaMixin, TrackedModelDetailView, WithCurrentWorkBasket):
     template_name = "quotas/detail.jinja"
 
+    def get_context_data(self, *args, **kwargs):
+        latest_definition = (
+            self.object.definitions.approved_up_to_transaction(
+                self.workbasket.current_transaction,
+            )
+            .order_by("valid_between")
+            .last()
```
gabelton, line 96: Is this logic right? Where a 2022 quota and a 2023 quota exist, it will get the future definition, rather than the current one. Possibly worth a test or a docstring or comment to make explicit the desired logic

eadpearce: I have no idea. I don't know which one is correct to display in this view

eadpearce: Just talked through this in planning. I will update this so we filter for whichever is active for the current date. If there isn't one, we get the next one in the future. If that one doesn't exist, we just leave this section blank.

gabelton: Sounds good. You should be able to use `as_at(date.today())` for the first bit and `not_yet_in_effect()` for the future definitions

eadpearce: Oh god I was building the queries with `Q(valid_between__upper_inf=True) | Q(valid_between__contains=current_date)`. This looks much simpler

gabelton (approved):

PR #743, opened 2022-12-06
TOPS-745 fix description causing NIG12

TomMacca:
# TOPS-747 fix description causing NIG12 business rule failure

## What
This PR creates a migration that edits the description date on a particular commodity code description (as the EU have done). 

Tests have been written for this migration so recommend running them. 

Ticket link:
https://uktrade.atlassian.net/jira/software/projects/TOPS/boards/221?selectedIssue=TOPS-745&text=745


commodities/migrations/0012_TOPS_745_description_date_fix.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    if GoodsNomenclatureDescription.objects.filter(
+        trackedmodel_ptr_id=10008934,
+    ).exists():
+        gn_desc = GoodsNomenclatureDescription.objects.get(
+            trackedmodel_ptr_id=10008934,
+        )
+        gn_desc.validity_start = datetime.date(2022, 1, 6)
+        gn_desc.save()
```
gabelton, line 29: Do we definitely want to call save() directly? Shouldn't we use `new_version` ?

commodities/migrations/0012_TOPS_745_description_date_fix.py
```diff
@@ -0,0 +1,41 @@
...
+    This is a manual change replicating the goods nomenclature description
+    period change that takes place in TGB22197 but is missed by the importer
+    """
+
+    GoodsNomenclatureDescription = apps.get_model(
+        "commodities",
+        "GoodsNomenclatureDescription",
+    )
+
+    if GoodsNomenclatureDescription.objects.filter(
```
gabelton, line 22: Can you use `objects.latest_approved()` within a migration?

commodities/tests/test_migrations.py
```diff
@@ -0,0 +1,96 @@
...
+        "commodities",
+        "GoodsNomenclatureDescription",
+    )
+    GoodsNomenclature = old_state.apps.get_model("commodities", "GoodsNomenclature")
+    Transaction = old_state.apps.get_model("common", "Transaction")
+    Workbasket = old_state.apps.get_model("workbaskets", "WorkBasket")
+    VersionGroup = old_state.apps.get_model("common", "VersionGroup")
+
+    ApprovedWorkBasketFactory.create().save()
+    workbasket = Workbasket.objects.last()
```
gabelton, line 30: Don't think you need to call `save()` on top of create. Also you could assign the result of create to `workbasket` and save a line

commodities/migrations/0012_TOPS_745_description_date_fix.py
dougmills-DIT: actually this is a problem for a couple of reasons. Firstly its possible to have a TAP model change in the future that will cause a historical migration to blow up, so using  apps.get_model() is a lot safer, but it only returns a basic django ORM object - and methods like latest_approved() is not available. Tom and I went back and forth with this issue but arrived at using apps.get_model() being the best solution (note I will update my historic migrations also to line up with this)

now I understand a little more i can see why apps.get_model() exists and is used in migrations.

gabelton: That makes sense. Probs worth some more assertions about version / transaction in the test in that case

commodities/migrations/0012_TOPS_745_description_date_fix.py
TomMacca: @gabelton what we're trying to do is retrospectively fix some data that was not imported correctly in the past so we're not making a new workbasket or transaction for this (which new_version needs I think) - does that answer the q?

TomMacca: I might be misunderstanding you though

commodities/migrations/0012_TOPS_745_description_date_fix.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+        original_workbasket = Workbasket.objects.get(
+            id=238)
+
+        goodsnomenclature_current_version = GoodsNomenclature.objects.get(
+            trackedmodel_ptr_id=10008944,
+        )
+
+        my_new_transaction = Transaction.objects.create(
```
gabelton, line 54: Guessing `WorkBasket.new_transaction()` isn't available to you within a migration?

TomMacca: Yeah I think not Gabe, the apps.get_model() returns a much more limited class

gabelton (approved):

PR #747, opened 2022-12-14
TP2000-600 Quota definitions list

eadpearce:
# TP2000-600 Quota definitions list

## Why

* Users need to be able to see more information about quotas

## What

* This view shows details for current and historic quota definitions for a given quota order number

Screenshot 2022-12-14 at 13 59 54
Screenshot 2022-12-14 at 14 00 05


quotas/views.py
```diff
@@ -98,6 +100,23 @@ def get_context_data(self, *args, **kwargs):
         )
 
 
+class QuotaDefinitions(SortingMixin, ListView):
```
gabelton, line 103: Could we name this QuotaDefinitionList to be more specific. Might potentially be worth renaming QuotaList to QuotaOrderNumberList as well

quotas/views.py
```diff
@@ -98,6 +100,23 @@ def get_context_data(self, *args, **kwargs):
...
+    template_name = "quotas/definitions.jinja"
+    model = models.QuotaDefinition
+    sort_by_fields = ["sid", "valid_between"]
+
+    def get_queryset(self):
+        self.queryset = models.QuotaDefinition.objects.filter(order_number=self.quota)
```
gabelton, line 109: Should this be `objects.current.filter` ? Might be worth adding some kind of view qs test to check that old versions aren't being returned

common/views.py
```diff
@@ -306,3 +306,45 @@ def form_valid(self, form):
             return self.form_invalid(form)
 
         return FormMixin.form_valid(self, form)
+
+
+class SortingMixin:
```
gabelton, line 311: niceeee

quotas/views.py
eadpearce: Can't easily rename QuotaList at the moment as it's tied up in the `get_ui_paths` function

gabelton: Ah fair enough. Maybe just worth adding a short docstring to QuotaList to make it clear that it's returning QuotaOrderNumber objects?

eadpearce: Yeah that works. I'll add one

gabelton (approved):

PR #748, opened 2022-12-15
update setup content types to be full fixture, and only load available apps

dougmills-DIT:
Updates to conftest fixture setup_content_types to be a full fixture. Also update to only load available apps and not fail

# TP2000-660

## Why
previous work I created setup_content_types but issues have occurred since relating to this work, so have updated the fixture to be an actual fixture and added a check to verify apps are available before loading them. 

## What
improvements to fixture setup_content_types

## Checklist
- Requires migrations? No

See: [[Description](https://example.com/...)](https://uktrade.atlassian.net/browse/TP2000-660)


a-gleeson:
LGTM! Pulled the branch down and tests are passing locally for me now!

a-gleeson (approved):

PR #749, opened 2022-12-15
TP2000-654 Create 403 error page

dalecannon:
# TP2000-654 Create 403 error page

## Why
When a user attempts to access a part of the service to which they don't have the required access rights, a generic 403 error response is displayed without guidance on how to rectify the problem or raise an issue. 

## What
Adds custom 403 error page with a link to a form where the user can request access or raise an issue. 

## Checklist
- Requires migrations? No

Before:
Screenshot 2022-12-15 at 17 47 51

After:
Screenshot 2022-12-16 at 12 10 31

### DIFF
--- a/common/jinja2/common/403.jinja
+++ b/common/jinja2/common/403.jinja
@@ -1,6 +1,6 @@
 {% extends "layouts/layout.jinja" %}
 
-{% set page_title = "Sorry, there is a problem with access to the service" %}
+{% set page_title = "You do not have access to this part of the service" %}
 
 {% block breadcrumb %}{% endblock %}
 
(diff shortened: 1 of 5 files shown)


gabelton:
Where do users currently see a 403? Could you add a test for at least one view in the "wild" to check that a user without permissions sees this page?

common/jinja2/common/403.jinja
```diff
@@ -0,0 +1,22 @@
...
+{% block breadcrumb %}{% endblock %}
+
+{% block content %}
+<div class="govuk-width-container">
+  <main class="govuk-main-wrapper govuk-main-wrapper--l" id="main-content" role="main">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">Sorry, there is a problem with access to the service</h1>
+        <p class="govuk-body">
+          <a class="govuk-link" href="https://forms.office.com/Pages/DesignPageV2.aspx?subpage=design&FormId=7Beij6oz-0atlt_mgAa7hnv6Jn6EmylOnTuPOq1EO2NUMjBNRllURlk0R0dFS1FHQkVBMFhNWjROVi4u&Token=3e454d078ea64c8fbbfcf336325d0427">Contact the TAP team</a> to change your access rights.
```
gabelton, line 14: Is this link right? And should it contain a token hard-coded?

common/jinja2/common/403.jinja
```diff
@@ -0,0 +1,22 @@
...
+{% set page_title = "Sorry, there is a problem with access to the service" %}
+
+{% block breadcrumb %}{% endblock %}
+
+{% block content %}
+<div class="govuk-width-container">
+  <main class="govuk-main-wrapper govuk-main-wrapper--l" id="main-content" role="main">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">Sorry, there is a problem with access to the service</h1>
```
gabelton, line 12: I find this message a bit weird. It makes it sound like a bug on our end, when most likely a 403 is what we intend the user to see because they don't have the right perms.

common/jinja2/common/403.jinja
```diff
@@ -0,0 +1,22 @@
...
+        <p class="govuk-body">
+          <a class="govuk-link" href="https://forms.office.com/Pages/DesignPageV2.aspx?subpage=design&FormId=7Beij6oz-0atlt_mgAa7hnv6Jn6EmylOnTuPOq1EO2NUMjBNRllURlk0R0dFS1FHQkVBMFhNWjROVi4u&Token=3e454d078ea64c8fbbfcf336325d0427">Contact the TAP team</a> to change your access rights.
+        </p>
+        <p class="govuk-body">Please describe what you were doing before seeing this error message.</p>
```
gabelton, line 16: I feel the same way about this as the "problem with access to the service" message above. We're inviting the user to write steps, as if we're trying to replicate a bug, but it probably isn't a bug. Someone will just have to add (or not add) a perm in the admin

urls.py
```diff
@@ -36,6 +36,8 @@
     path("admin/", admin.site.urls),
 ]
 
+handler403 = "common.views.handler403"
```
gabelton, line 39: How does the app know to redirect to this ? Do you have any docs that you followed for this approach? I might be missing something obvious

gabelton (approved):

common/jinja2/common/403.jinja
```diff
@@ -1,6 +1,6 @@
 {% extends "layouts/layout.jinja" %}
 
-{% set page_title = "Sorry, there is a problem with access to the service" %}
+{% set page_title = "You do not have access to this part of the service" %}
```
gabelton, line 3: :thumbsup:

PR #751, opened 2022-12-19
Tp2000 611 send wb to packaging queue

a-gleeson:
# TP2000-611 

This PR:
    - Correctly manage new form action in workbasket detail
    - Added Package workbasket create view with form 
    - Packaged workbasket create view contains appropriate error handling and State tranisition for a workbasket.
    - Correctly style rule check summary page.
    - Added an SVG provided by our UI designer.
    - Added view unit tests.


publishing/views.py
```diff
@@ -60,3 +72,92 @@ def post(self, request, *args, **kwargs):
         envelopes."""
         # TODO: manage post actions.
         return super().post()
+
+
+class PackagedWorkbasketCreateView(PermissionRequiredMixin, CreateView):
+    """UI endpoint for creating packaged workbaskets."""
+
+    permission_required = "workbaskets.add_workbasket"
```
gabelton, line 80: Is this the right perm? Shouldn't it be "publishing.add_packaged_workbasket" or similar?

publishing/views.py
```diff
@@ -60,3 +72,92 @@ def post(self, request, *args, **kwargs):
...
+
+    def __init__(self):
+        super().__init__()
+        self.logger = logging.getLogger(type(self).__name__)
+
+    @property
+    def workbasket(self):
+        return WorkBasket.current(self.request)
+
+    def form_valid(self, form):
```
gabelton, line 92: Might be worth a docstring and/or some comments

publishing/views.py
```diff
@@ -60,3 +72,92 @@ def post(self, request, *args, **kwargs):
...
+        kwargs["request"] = self.request
+        return kwargs
+
+
+class PackagedWorkbasketConfirmCreate(DetailView):
+    template_name = "publishing/confirm_create.jinja"
+    model = PackagedWorkBasket
+
+    def get_queryset(self):
+        """Return all items that are awaiting processing or are actively being
```
gabelton, line 159: Is this docstring accurate?

publishing/views.py
a-gleeson: supposed to be common.tracked model I can update this thanks

publishing/views.py
```diff
@@ -133,3 +144,95 @@ def post(self, request, *args, **kwargs):
         envelopes."""
         # TODO: manage post actions.
         return super().post()
+
+
+class PackagedWorkbasketCreateView(PermissionRequiredMixin, CreateView):
+    """UI endpoint for creating packaged workbaskets."""
+
+    permission_required = "common.add_trackedmodel"
```
gabelton, line 152: I'm not sure this is right either because PackagedWorkBasket doesn't inherit from TrackedModel. I think it will need to be something like "publishing.add_packagedworkbasket". Test doc string will also need updating

common/jinja2/layouts/form.jinja
```diff
@@ -10,7 +10,9 @@
 
 {% block content %}
   <h1 class="govuk-heading-xl">{{ page_title }}</h1>
-
+  {% if page_label %}
```
paulpepper-trade, line 13: There isn't much to structure here, but indents would still be good style.

publishing/jinja2/includes/warning_help.jinja
```diff
@@ -0,0 +1,6 @@
+{% from "components/warning-text/macro.njk" import govukWarningText %}
+
+{{ govukWarningText({
```
paulpepper-trade, line 3: I don't think you need a separate template with this warning. Would the GDS crispy forms library allow you to include this warning in your Layout? https://crispy-forms-gds.readthedocs.io/en/latest/components/warning.html

publishing/views.py
```diff
@@ -133,3 +144,95 @@ def post(self, request, *args, **kwargs):
...
+            wb.submit_for_approval()
+            wb.save()
+        except ValidationError as err:
+            self.logger.error(
+                "Error: %s \n Redirecting to work basket %s summary",
+                err.message,
+                self.workbasket.id,
+            )
+            return redirect(
+                reverse(
```
paulpepper-trade, line 180: I think `redirect()` may allow you to provide the view name and args without need for `reverse()`.

publishing/views.py
```diff
@@ -133,3 +144,95 @@ def post(self, request, *args, **kwargs):
...
+    def __init__(self):
+        super().__init__()
+        self.logger = logging.getLogger(type(self).__name__)
+
+    @property
+    def workbasket(self):
+        """Current Workbasket in session."""
+        return WorkBasket.current(self.request)
+
+    def form_valid(self, form):
```
paulpepper-trade, line 165: There's more than one `save()` in this method. I think it may need to be decorated with `django.db.transaction.atomic` to ensure it either all succeeds or all fails.

publishing/views.py
```diff
@@ -133,3 +144,95 @@ def post(self, request, *args, **kwargs):
...
+            queued_wb = PackagedWorkBasket.objects.create(
+                workbasket=wb, **form.cleaned_data
+            )
+        except PackagedWorkBasketDuplication as err:
+            self.logger.error(
+                "Error: %s \n Redirecting to packaged work basket queue",
+                err,
+            )
+            return redirect(
+                reverse(
```
paulpepper-trade, line 200: Ditch the `reverse`?

publishing/forms.py
```diff
@@ -0,0 +1,76 @@
...
+
+    jira_url = forms.URLField(
+        label="Tops Jira",
+        help_text="Insert Tops Jira ticket link",
+        widget=forms.TextInput,
+        required=True,
+    )
+
+    def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop("request", None)
```
paulpepper-trade, line 50: I know this pattern is used in other parts of TAP, but is `self.request` ever referenced by the form? if not, then it should probably be removed.

publishing/views.py
```diff
@@ -133,3 +144,95 @@ def post(self, request, *args, **kwargs):
...
+        return redirect(
+            reverse(
+                "publishing:packaged-workbasket-queue-confirm-create",
+                kwargs={"pk": queued_wb.pk},
+            ),
+        )
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs["request"] = self.request
```
paulpepper-trade, line 226: Needed?

paulpepper-trade (changes requested):

publishing/jinja2/includes/warning_help.jinja
a-gleeson: I'll try that, was just thinking it was good to have template logic in jinja

a-gleeson: Made changes although template logic is now in forms.py, is having template logic here suitable?

paulpepper-trade (approved):

gabelton (approved):

PR #753, opened 2022-12-19
TP2000-651 combine proposed approved and sent statuses into queued

gabelton:
TP2000-651 combine proposed approved and sent statuses into queued

## Why

We identified that as part of current manual packaging process devs often transition a workbasket through from EDITING -> PROPOSED -> APPROVED -> SENT in one go. This suggested that we ought to simplify the state machine and consolidate PROPOSED, APPROVED, and SENT into a new QUEUED status, within which the PackagedWorkBasket would pass through its own state transitions.

## What

Updates current references to PROPOSED to either EDITING or QUEUED, as appropriate
Updates APPROVED and SENT to QUEUED.
Move transition method logic into `queue` method
Update unit tests


workbaskets/models.py
```diff
@@ -369,13 +369,22 @@ def archive(self):
     def unarchive(self):
         """Restore a workbasket to an in use state."""
 
+    def approve(self, user: int, scheme_name: str):
```
paulpepper-trade, line 372: A docstring would probably be useful here, or perhaps incorporate something into the name of the function that indicates partition scheme changes?

workbaskets/models.py
```diff
@@ -386,59 +395,23 @@ def submit_for_approval(self):
...
         field=status,
-        source=WorkflowStatus.APPROVED,
+        source=WorkflowStatus.QUEUED,
         target=WorkflowStatus.EDITING,
-        custom={"label": "Restore an approved workbasket to an in use state."},
+        custom={"label": "Restore a queued workbasket to an in use state."},
     )
-    def unapprove(self):
-        """A workbasket that has not yet been sent to CDS requires further
+    def unqueue(self):
```
paulpepper-trade, line 406: Did we consider dequeue() for this? I'm agnostic about which is best.

paulpepper-trade (approved):

publishing/views.py
```diff
@@ -180,11 +180,11 @@ def form_valid(self, form):
             )
             return redirect(
                 "workbaskets:workbasket-ui-detail",
-                pk=self.workbasket.id
+                pk=self.workbasket.id,
             )
 
-        wb.approve(self.request.user, settings.TRANSACTION_SCHEMA)
-        wb.save()
+        # wb.approve(self.request.user, settings.TRANSACTION_SCHEMA)
```
a-gleeson, line 186: could these lines be removed?

a-gleeson (approved):
LGTM!

PR #755, opened 2022-12-21
TP2000-648 Add measures tab

eadpearce:
# TP2000-648 Add measures tab

## Why

* Users need to be able to find measures related to a quota order number

## What

* Adds a measures tab to the quota order number detail page that shows currently active measures
* Adds a link to the find and edit measures page that filters by the quota order number

Screenshot 2022-12-21 at 15 34 45

### DIFF
--- a/quotas/jinja2/includes/quotas/actions.jinja
+++ b/quotas/jinja2/includes/quotas/actions.jinja
@@ -2,10 +2,11 @@
   <div class="app-related-items" role="complementary">
       <h2 class="govuk-heading-s" id="subsection-title">Actions</h2>
       <ul class="govuk-list govuk-!-font-size-16">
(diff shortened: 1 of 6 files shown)


TomMacca:
with `current_definition = definitions.as_at(date.today()).first()` why is `first()` necessary? If there are more than one objects returned - what are they ordered by i.e. what will `first()` return here?

eadpearce:
@TomMacca `definitions.as_at(date.today())` returns a queryset but we only want one. The queryset should only contain one because there should only be one quota definition period active at a time or else this is a business rule violation

TomMacca (approved):

TomMacca:
@eadpearce that makes sense but if it should only return one why is `first()` used as well? Just in case it returns more than one? I wonder how first() orders the objects, by tracked model number?

eadpearce:
@TomMacca `.as_at()` returns a queryset which is treated like a list. The template only wants one object. I presume the objects are ordered by pk by default

PR #757, opened 2022-12-21
TP2000-641 create govuk notify templates

gabelton:
# TP2000-641 create govuk notify templates

## Why

We want to be able to notify users when a packaged workbasket has reached the top of the queue, then again when CDS has rejected or accepted an envelope. To do this we need will use govuk notify and its template system, dynamically populating templates for those three scenarios, rather than manually drafting an email, as we do at present.

## What

Adds personalisation params to `send_emails` task
Makes notify methods on `PackagedWorkBasket` call task
Adds LoadingReport storage options
Hooks notifications into queue management actions - accept / reject envelope, move queued instances, etc.
Implements an environment variable flag to allow disabling publishing system-specific notifications.
Enhancements and fixes implemented when encountering related parts of the publishing system.

Note that unit tests and file error handling still need work (another ticket), but getting this into PR now to minimise diverging branches causing merge conflicts and trial on a staging service.

See [lite's implementation](https://github.com/uktrade/lite-api/blob/master/api/cases/notify.py) for comparison.

## Checklist
- Requires migrations? yes


publishing/models.py
```diff
@@ -481,42 +626,84 @@ def refresh_from_db(self, using=None, fields=None):
...
+        (see `publishing.tasks.create_xml_envelope_file()`).
+        """
+
+        personalisation = {
+            "envelope_id": self.envelope.envelope_id,
+            "download_url": (
+                settings.BASE_SERVICE_URL + reverse("publishing:envelope-queue-ui-list")
+            ),
+            "theme": self.theme,
+            "eif": self.eif if self.eif else "Immediately",
```
a-gleeson, line 646: Should embargo be included in this personalisation?

publishing/models.py
```diff
@@ -481,42 +626,84 @@ def refresh_from_db(self, using=None, fields=None):
...
+    @skip_notifications_if_disabled
     def notify_processing_succeeded(self):
-        """TODO."""
+        """
+        Notify users that envelope processing has been succeeded for this.
 
+        instance - correctly ingested into HMRC systems.
+        """
+
+        if not settings.ENABLE_PACKAGING_NOTIFICATIONS:
```
a-gleeson, line 662: Does skip_notifications_if_disabled decorator not carry this check out now?

publishing/models.py
```diff
@@ -481,42 +626,84 @@ def refresh_from_db(self, using=None, fields=None):
...
+            personalisation=personalisation,
+        )
+
+    @skip_notifications_if_disabled
     def notify_processing_failed(self):
-        """TODO."""
+        """Notify users that envelope processing has been failed - HMRC systems
+        rejected this instances associated envelope file."""
+
+        if not settings.ENABLE_PACKAGING_NOTIFICATIONS:
```
a-gleeson, line 688: Still required?

publishing/models.py
```diff
@@ -164,22 +173,39 @@ def completed_processing_states(cls):
...
+    """Report associated with an attempt to load (process) a PackagedWorkBasket
+    instance."""
 
-    # TODO Change report_file to correct field for / s3 object reference.
-    report_file = models.FileField(
+    file = models.FileField(
         blank=True,
         null=True,
+        storage=LoadingReportStorage,
+        upload_to=report_bucket,
```
a-gleeson, line 189: Upload_to & file path generation can be removed as this will end up creating a duplicate path in the s3 bucket eg (staging/staging/...) 
In publishing/storages.py LoadingReportStorage.generate_filename generates the path too.

settings/common.py
```diff
@@ -312,6 +312,28 @@
 HMRC_STORAGE_BUCKET_NAME = os.environ.get("HMRC_STORAGE_BUCKET_NAME", "hmrc")
 HMRC_STORAGE_DIRECTORY = os.environ.get("HMRC_STORAGE_DIRECTORY", "tohmrc/staging/")
 
+# Loading reports AWS settings
+LOADING_REPORTS_BUCKET_NAME = os.environ.get(
+    "LOADING_REPORTS_BUCKET_NAME",
+    "loading-reports",
+)
+LOADING_REPORTS_S3_ACCESS_KEY_ID = os.environ.get(
```
a-gleeson, line 320: Are the auth details  & the s3 urk for loading reports going to be different from s3 for envelope?
Should the naming be more generic if so or should the env vars take from the same variable in .env?

publishing/tests/test_models.py
```diff
@@ -16,8 +22,18 @@ def test_create():
     """Test multiple PackagedWorkBasket instances creation is managed
     correctly."""
 
-    first_packaged_work_basket = factories.PackagedWorkBasketFactory()
-    second_packaged_work_basket = factories.PackagedWorkBasketFactory()
+    with patch(
```
a-gleeson, line 25: This mock is reused a lot is it possible to move this to conftest or does that cause an issue with the mocking?

a-gleeson (changes requested):
Looking good! Left a few comments

publishing/models.py
paulpepper-trade: Yes! Now sorted.

publishing/models.py
paulpepper-trade: Now sorted.

publishing/tests/test_models.py
paulpepper-trade: We have a fixure in `publishing/conftest.py` for this (which looks a lot like the code that you comment on), but the trickiness is with passing two instances of that fixture into the test. Maybe we could create a fixture factory for this, which would reduce the bulk of these context manager statements, but I haven't completely pieced that together just yet.

publishing/models.py
paulpepper-trade: We hardcoded embargo to None in the notification template. I have a vague recollection of a discussion around this and we have never issued an envelope with anything other than None as a value there. For consistency with current process, we're continuing to communicate that to HMRC. If there's ever a need to do something different, I think we said we'd implement embargo as a variable in our `PackagedWorkBasket` model and the associated form that collects that info.

publishing/models.py
paulpepper-trade: I think we need a round of work on our S3 storage capability here, bringing in the good work that you've done with envelope management and S3. It's kind of slipped between the cracks until now. I've created [a new ticket](https://uktrade.atlassian.net/browse/TP2000-695) to cover this off.

settings/common.py
paulpepper-trade: I think we should deal with this on [the S3-related ticket](https://uktrade.atlassian.net/browse/TP2000-695) that has been created for dealing with these issues.

a-gleeson (approved):
LGTM!

publishing/models.py
paulpepper-trade: As you've pointed out Anthoni, most of the work is in place to optionally populate embargo in the template and use the term 'None' if the model value is None.

paulpepper-trade: So now applying that to the Notify template.

PR #761, opened 2023-01-03
Create regulation ID using the current year.

paulpepper-trade:
# TP2000-671 - Year-sensitive Regulation unit test broken

## Why

The unit test test_regulation_create_form_invalid_part_value has a hard-coded date dependency, causing a test failure when the year differs to hard coded value.

## What

Make the test use the current year as per the other parts of the test.

## Checklist
- Requires migrations? No

### DIFF
--- a/regulations/tests/test_forms.py
+++ b/regulations/tests/test_forms.py
@@ -1,4 +1,5 @@
 import pytest
+from django.utils import timezone
 
 from common.tests import factories
 from regulations import forms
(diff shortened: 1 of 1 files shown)


regulations/tests/test_forms.py
```diff
@@ -57,7 +58,8 @@ def test_regulation_create_form_unapproved_and_not_draft(date_ranges, session_re
 def test_regulation_create_form_invalid_part_value(date_ranges, session_request):
     """Test that RegulationCreateForm excepts an IndexError when looking for an
     alphanumeric character after Z and raises a ValidationError."""
-    factories.RegulationFactory.create(regulation_id="C220001Z")
+    year = timezone.now().strftime("%y")
```
eadpearce, line 61: Can this be written as `timezone.now().year` instead? I feel like it's a bit easier to read

eadpearce: Ah nvm this would return the full year rather than just '23'

eadpearce (approved):

PR #769, opened 2023-01-09
TP2000-665 Corrects data fields on commodities details view

dalecannon:
# TP2000-665 Corrects data fields on commodities details view

## Why

- Commodities details view incorrectly displays commodity origin data as parent commodity data. 
- Commodities details view displays outdated indent values (TP2000-528).
- Data fields contain spelling errors.

## What

- Fixes parent commodity data.
- Shows only current indent value.
- Corrects spelling errors.

Before:
Screenshot 2023-01-09 at 14 25 16

After:
Screenshot 2023-01-09 at 14 10 54

### DIFF
--- a/commodities/views.py
+++ b/commodities/views.py
@@ -11,6 +11,7 @@ from commodities.filters import CommodityFilter
 from commodities.filters import GoodsNomenclatureFilterBackend
 from commodities.forms import CommodityImportForm
 from commodities.models import GoodsNomenclature
+from commodities.models.dc import get_chapter_collection
 from common.serializers import AutoCompleteSerializer
 from common.views import TrackedModelDetailView
 from common.views import WithPaginationListView
@@ -75,3 +76,17 @@ class CommodityList(CommodityMixin, WithPaginationListView):
 
 class CommodityDetail(CommodityMixin, TrackedModelDetailView):
     template_name = "commodities/detail.jinja"
+
+    def get_context_data(self, *args, **kwargs):
+        collection = get_chapter_collection(self.object)
+        tx = WorkBasket.get_current_transaction(self.request)
+        date = self.object.valid_between.upper
+        snapshot = collection.get_snapshot(tx, date)
+        commodity = snapshot.get_commodity(self.object, self.object.suffix)
+        parent = snapshot.get_parent(commodity)
+
+        return super().get_context_data(
+            parent=parent,
+            *args,
+            **kwargs,
+        )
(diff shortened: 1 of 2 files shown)


commodities/views.py
```diff
@@ -75,3 +76,17 @@ def get_queryset(self):
 
 class CommodityDetail(CommodityMixin, TrackedModelDetailView):
     template_name = "commodities/detail.jinja"
+
+    def get_context_data(self, *args, **kwargs):
+        collection = get_chapter_collection(self.object)
+        tx = WorkBasket.get_current_transaction(self.request)
```
paulpepper-trade, line 82: Wondering if there are circumstances where the transaction shouldn't be taken from the active workbasket. None come to mind for the context of editing a workbasket, as I expect is this view's use.

commodities/views.py
```diff
@@ -75,3 +76,17 @@ def get_queryset(self):
...
+        date = self.object.valid_between.upper
+        snapshot = collection.get_snapshot(tx, date)
+        commodity = snapshot.get_commodity(self.object, self.object.suffix)
+        parent = snapshot.get_parent(commodity)
+
+        return super().get_context_data(
+            parent=parent,
```
paulpepper-trade, line 89: `parent` shouldn't be passed into super's `get_context_data()` method. We'd normally get the context dictionary, set new dictionary items and then return the context dictionary.
```something like:
    def get_context_data(self, *args, **kwargs):
        context = super().get_context_data( *args, **kwargs)
        
        collection = get_chapter_collection(self.object)
        date = self.object.valid_between.upper
        snapshot = collection.get_snapshot(tx, date)
        commodity = snapshot.get_commodity(self.object, self.object.suffix)
        context["parent"] = snapshot.get_parent(commodity)
        
        return context
```

commodities/views.py
```diff
@@ -75,3 +76,17 @@ def get_queryset(self):
 
 class CommodityDetail(CommodityMixin, TrackedModelDetailView):
     template_name = "commodities/detail.jinja"
+
+    def get_context_data(self, *args, **kwargs):
+        collection = get_chapter_collection(self.object)
+        tx = WorkBasket.get_current_transaction(self.request)
+        date = self.object.valid_between.upper
```
paulpepper-trade, line 83: Similar to choice of transaction... In a more generalised commodity code inspection capability I suppose this could be set by the user, but that's for another ticket!

paulpepper-trade (changes requested):

paulpepper-trade (approved):

PR #774, opened 2023-01-11
TP2000-672 Add multiple measures edit end date form

LaurenQurashi:
# TP2000-672 Add multiple measures edit end date form

## Why
Due to the create measures form having some quirks, it's not a bright idea to make a multiple measures edit form based on it. We've decided as a team that it needs some extra thought and planning, along with some time spent polishing up the create form. As an interim, we've got this 'MVP' version allowing TM's to edit the end dates of multiple measures cause that would be a lifesaver as a minimum. 

## What
This is only an interim change, that's been implemented until the polished article can come out. This PR adds a button to the MeasureList page, next to the delete button to edit the end dates. This takes you to a new page, similar to the multiple delete review page, except it has an end date form on it. This form has no visible validation on it, as that's coming later in a separate ticket as we need to think about how we show errors to the users when making multiple edits at once. The code, however, has a little clause in it that will stop any measure from being altered if the start date is after the proposed end date. 

Here's the new button on the measures list page... 
Screenshot 2023-01-13 at 17 56 58

And here's the edit multiple measures end dates page ...
Screenshot 2023-01-13 at 17 54 13

## Checklist
- Requires migrations? - no

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -21,6 +21,7 @@ from certificates.models import Certificate
 from commodities.models import GoodsNomenclature
 from common.fields import AutoCompleteField
 from common.forms import BindNestedFormMixin
+from common.forms import DateInputFieldFixed
 from common.forms import FormSet
 from common.forms import RadioNested
 from common.forms import ValidityPeriodForm
@@ -1135,3 +1136,25 @@ class MeasureReviewForm(forms.Form):
 
 
 MeasureDeleteForm = delete_form_for(models.Measure)
+
+
+class MeasureEndDateForm(forms.Form):
+    end_date = DateInputFieldFixed(
+        label="End date",
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.helper = FormHelper(self)
+        self.helper.label_size = Size.SMALL
+        self.helper.legend_size = Size.SMALL
+        self.helper.layout = Layout(
+            "end_date",
+            Submit(
+                "submit",
+                "Save measure end dates",
+                data_module="govuk-button",
+                data_prevent_double_click="true",
+            ),
+        )
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1174,9 +1174,11 @@ class MeasureEndDateForm(forms.Form):
                 lower = measure.valid_between.lower
                 upper = datetime.date(year, month, day)
                 if lower > upper:
+                    formatted_lower = lower.strftime("%d/%m/%Y")
+                    formatted_upper = upper.strftime("%d/%m/%Y")
                     raise ValidationError(
                         f"The end date cannot be before the start date: "
-                        f"{lower} does not start before {upper}",
+                        f"Start date {formatted_lower} does not start before {formatted_upper}",
                     )
 
         return cleaned_data
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 
 from crispy_forms_gds.helper import FormHelper
@@ -1145,6 +1146,7 @@ class MeasureEndDateForm(forms.Form):
     )
 
     def __init__(self, *args, **kwargs):
+        self.selected_measures = kwargs.pop("selected_measures", None)
         super().__init__(*args, **kwargs)
 
         self.helper = FormHelper(self)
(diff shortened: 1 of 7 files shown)


measures/tests/test_views.py
```diff
@@ -132,7 +134,7 @@ def test_multiple_measure_delete_functionality(client, valid_user, session_workb
                 "status": session_workbasket.status,
                 "title": session_workbasket.title,
             },
-            "DELETE_MEASURE_SELECTIONS": {
+            "MULTIPLE_MEASURE_SELECTIONS": {
```
LaurenQurashi, line 137: Here I rename the session, as all the measures that are selected go into the same session, and are grabbed from it regardless of whether you pick edit or delete, so I've made it a more general name

measures/views.py
```diff
@@ -612,3 +620,60 @@ def post(self, request):
...
+
+    def get_context_data(self, **kwargs):
+        store_objects = self.get_queryset()
+        self.object_list = store_objects
+        context = super().get_context_data(**kwargs)
+
+        return context
+
+    def post(self, request):
+        if request.POST.get("submit", None) != "Save measure end dates":
```
LaurenQurashi, line 652: Not sure it's the best idea to base the if on the name of the button, but other than the date, that's all that comes through in the Post from the crispy form. If anyone knows how I can add attributes in the form and collect them in the post method, I'd love to know as really this should be some sort of dedicated variable to look out for.

measures/views.py
```diff
@@ -612,3 +620,60 @@ def post(self, request):
...
+        for obj in object_list:
+            # make a new version of the object with an update type of update, pass in the new end date.
+            end_year = int(request.POST["end_date_2"])
+            end_month = int(request.POST["end_date_1"])
+            end_day = int(request.POST["end_date_0"])
+
+            lower = obj.valid_between.lower
+            upper = datetime.date(end_year, end_month, end_day)
+
+            if upper >= lower:
```
LaurenQurashi, line 667: This is the interim validation to stop bad end dates from being added. I need to speak to Steve and Wren about how we show the users any errors when they may apply to some and not other measures they may have picked.

eadpearce: If there's a way to move this validation into the form instead that would make it easy to redisplay the page and show the errors (I'm not sure how difficult that would be though)

measures/jinja2/measures/edit.jinja
```diff
@@ -10,7 +10,7 @@
 {% from "components/breadcrumbs.jinja" import breadcrumbs %}
 
 
-{% set page_title = "Edit " ~ object._meta.verbose_name ~ " details" %}
+{% set page_title = "Edit" ~ object._meta.verbose_name ~ " details" %}
```
eadpearce, line 13: Does this still leave a space between 'Edit' and 'object._meta.verbose_name'?

measures/jinja2/measures/edit-multiple-measures-enddates.jinja
```diff
@@ -0,0 +1,116 @@
...
+  <div class="govuk-grid-row govuk-!-margin-bottom-5">
+    <div class="govuk-grid-column-two-thirds">
+      <p class="govuk-body"> You are about to edit the end date of your selected measures.</p>
+      {{ govukWarningText({
+        "text": "You are currently only able to change the end dates of multiple measures.",
+        "iconFallbackText": "Warning"
+      }) }}
+    
+      <div class="govuk-inset-text">
+        <p>Your edited measures will be added to your work basket as items.</p>
```
eadpearce, line 43: Content question: are we using 'work basket' or 'workbasket'? Should be consistent

measures/jinja2/measures/edit-multiple-measures-enddates.jinja
```diff
@@ -0,0 +1,116 @@
...
+      <h1 class="govuk-heading-xl">
+        Edit measures
+        <span class="govuk-caption-xl"></span>
+      </h1>
+    </div>
+  </div>
+
+  <div class="govuk-grid-row govuk-!-margin-bottom-5">
+    <div class="govuk-grid-column-two-thirds">
+      <p class="govuk-body"> You are about to edit the end date of your selected measures.</p>
```
eadpearce, line 36: 'end dates'?

measures/views.py
paulpepper-trade: I think this approach is fine, Lauren. It's one we use elsewhere in the application ([WorkBaskets for instance](https://github.com/uktrade/tamato/blob/eea1ab70864c8433e3108c1c6d29a9dcb23f2c6e/workbaskets/views/ui.py#L343). The submit control is just another input control, afterall.

measures/jinja2/measures/edit-multiple-measures-enddates.jinja
LaurenQurashi: Yeah that's a good point I actually made all these words up so I better show Wren and Matt

measures/views.py
paulpepper-trade: Is it something that couldl be handled in the [Form.clean](https://docs.djangoproject.com/en/4.1/ref/forms/api/#django.forms.Form.clean) method?

eadpearce: Yeah that was my idea. I think the object list just needs passing from the view to the form for it to do the validation. We do that elsewhere with the `get_form_kwargs` method

measures/forms.py
```diff
@@ -1135,3 +1137,46 @@ class MeasureReviewForm(forms.Form):
...
+                data_module="govuk-button",
+                data_prevent_double_click="true",
+            ),
+        )
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if "end_date" in cleaned_data:
+            for measure in self.selected_measures:
```
eadpearce, line 1169: 🙌

PR #776, opened 2023-01-16
TP2000-673 Sort by commodity code in Quota measures tab

dalecannon:
# TP2000-673 Make commodity codes sortable in Quota measures tab

## Why

- Tariff Managers have indicated that the ability to sort by commodity code would help them to compare the list of commodity codes attached against other data easier.

## What

- Make commodity code list sortable (in ascending or descending order) by user.

Before:
quota-measures-list-before

After:
sortable-comm-codes-quota-measures-tab

### DIFF
--- a/common/jinja2/components/sort_by.jinja
+++ b/common/jinja2/components/sort_by.jinja
@@ -1,12 +1,12 @@
 
-{% macro sort_by(request, sort_by, title, base_url) %}
+{% macro sort_by(request, sort_by, title, base_url, anchor="") %}
 
   {% if request.GET.sort_by == sort_by and request.GET.order == "desc" %}
-    <a class="govuk-link govuk-link--no-visited-state sort-icon--down" href="{{ base_url }}?sort_by={{ sort_by }}&order=asc">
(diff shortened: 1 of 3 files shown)


quotas/views.py
```diff
@@ -116,8 +117,14 @@ def get_context_data(self, *args, **kwargs):
             .first()
         )
 
-        context["measures"] = Measure.objects.filter(order_number=self.object).as_at(
-            date.today(),
+        order = self.get_ordering()
+        if not order:
+            order = "goods_nomenclature"
```
eadpearce, line 122: Does this mean comm code is the default ordering?

common/jinja2/components/sort_by.jinja
```diff
@@ -1,12 +1,12 @@
 
-{% macro sort_by(request, sort_by, title, base_url) %}
+{% macro sort_by(request, sort_by, title, base_url, anchor="") %}
```
eadpearce, line 2: 👍

eadpearce (approved):

quotas/views.py
eadpearce: Cool. I think the default ordering is by SID which doesn't seem all that useful

PR #777, opened 2023-01-17
TP2000-602: Get Trade Tariff API quota data

eadpearce:
# TP2000-602: Get tariffs API quota data

## Why

* Some data on quotas is not held in TAP so we get this data from the [Trade Tariff API](https://api.trade-tariff.service.gov.uk/reference.html#trade-tariff-public-api-v2)

## What

* Adds status and balance data for quotas and definition periods
* If the API is unreachable we omit this data from the page
* Adds a link to the quota on the Trade Tariff API in the sidebar
* Adds support for testing async functions with pytest

## Checklist
- Requires migrations? ❌ 
- Requires dependency updates? ✅ 

Screenshot 2023-01-17 at 11 51 11

Screenshot 2023-01-17 at 11 50 18

### DIFF
--- a/common/tariffs_api.py
+++ b/common/tariffs_api.py
@@ -31,16 +31,8 @@ async def async_get_all(urls):
         return await asyncio.gather(*[async_get(url, session) for url in urls])
 
 
-def get_quota_definitions_data(order_number, object_list):
-    """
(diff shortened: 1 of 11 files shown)


dougmills-DIT (approved):
looks good to me

PR #778, opened 2023-01-18
TP2000-693 Fix duplication in Quota measures tab

dalecannon:
# TP2000-693 Fix duplication in Quota measures tab

## Why
The measures queryset used in the measures tab on a quota detail view returns duplicated data for some quotas.

## What
Adds latest_approved() to the query to return only the current version of an object and not its historical updates.

Before:
before

After:
after


LaurenQurashi (approved):
Looks great! Nice one Dale 🎉

PR #780, opened 2023-01-19
TP2000-679 Add button to stop rule check

dalecannon:
# TP2000-679 Add button to stop rule check

## Why
In some circumstances, business rules are shown to be running on a workbasket when in fact they’re not. When this happens, users ask a member of the DDaT team to terminate these "ghost" rule checks on their behalf.

## What
Adds a button to allow users themselves to stop a rule check.

stop-rule-check-button

### DIFF
--- a/workbaskets/tests/test_views.py
+++ b/workbaskets/tests/test_views.py
@@ -485,6 +485,42 @@ def test_run_business_rules(check_workbasket, valid_user_client, session_workbas
     assert not session_workbasket.tracked_model_checks.exists()
 
 
+def test_terminate_rule_check(valid_user_client, session_workbasket):
+    session_workbasket.rule_check_task_id = 123
+
+    with session_workbasket.new_transaction() as transaction:
+        good = GoodsNomenclatureFactory.create(transaction=transaction)
+        check = TrackedModelCheckFactory.create(
+            transaction_check__transaction=transaction,
+            model=good,
+            successful=False,
+        )
+
+    session = valid_user_client.session
+    session["workbasket"] = {
+        "id": session_workbasket.pk,
+        "status": session_workbasket.status,
+        "title": session_workbasket.title,
+    }
+    session.save()
+
+    url = reverse(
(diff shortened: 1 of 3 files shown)


workbaskets/tests/test_views.py
```diff
@@ -485,6 +485,42 @@ def test_run_business_rules(check_workbasket, valid_user_client, session_workbas
...
+    url = reverse(
+        "workbaskets:workbasket-ui-detail",
+        kwargs={"pk": session_workbasket.id},
+    )
+    response = valid_user_client.post(
+        url,
+        {"form-action": "terminate-rule-check"},
+    )
+    assert response.status_code == 302
+    response_url = f"/workbaskets/{session_workbasket.pk}/"
```
paulpepper-trade, line 500: It'd be preferable to reverse [the URL's path name](https://github.com/uktrade/tamato/blob/01a4fb0d9003497cf6e9d62ae46456d2e71be448/workbaskets/urls.py#L45) here, which would avoid breaking the test should the path change in some way.

workbaskets/tests/test_views.py
```diff
@@ -485,6 +485,42 @@ def test_run_business_rules(check_workbasket, valid_user_client, session_workbas
...
+    assert response.url[: len(response_url)] == response_url
+
+    session_workbasket.refresh_from_db()
+
+    assert not session_workbasket.rule_check_task_id
+
```
paulpepper-trade, line 506: Maybe also check that the associated `TrackedModelCheck` has been deleted since you've gone to the trouble of adding one?

paulpepper-trade (approved):
A couple of comments, but otherwise looks good.

workbaskets/tests/test_views.py
dalecannon: I've since removed it because I didn't think it was needed. Unless I should add it back and check for its deletion?

paulpepper-trade: I don't think it's needed in this view test, so I'd leave as is now.

PR #786, opened 2023-01-24
Tp2000 656  - Implement envelope creation, storage and retrieval from S3

a-gleeson:
# TP2000-656 - Implement envelope creation, storage and retrieval from S3

## Why
We want to upload files that we want HMRC users to download and process. 

It's necessary to:
    Implement the subsystem and task that supports creating the Envelope within the model for reporting purposes.
    Generate the xml from the packaged workbaskets transactions 
    Upload the valid xml to a file system
    Read the file system when downloading the file
    Trigger delete envelope when packaged workbasket is removed from the top of a queue

## What

- updated publishing s3 environment variables for enveople & loading reports. Now in a single bucket with multiple directories
- set up s3 storage for envelope file
- added envelope model to publishing system that includes creation and deletion of envelopes
- integrated envelope creation with celery tasks
- integrated envelope file read to the download envelope view
- added tests, skipping current ones which do not work
- Updated envelope id value in hmrc envelope confirm pages
- refactored models structure to a folder
- updated intial.py migration
- added delete method to Envelope that deletes the s3 object too. 
- updated envelope foreign key to set to Null when an envelope is deleted

## Checklist
- Requires migrations? - yes

See: [Django s3 storage](https://django-storages.readthedocs.io/en/1.13.2/backends/amazon-S3.html)


publishing/models/envelope.py
```diff
@@ -0,0 +1,218 @@
...
+
+            serializer = MultiFileEnvelopeTransactionSerializer(
+                output_file_constructor,
+                envelope_id=self.envelope_id,
+                max_envelope_size=settings.EXPORTER_MAXIMUM_ENVELOPE_SIZE,
+            )
+
+            rendered_envelope = list(
+                serializer.split_render_transactions(transactions),
+            )[0]
```
paulpepper-trade, line 191: This looks as though it will silently throw away envelopes beyond the first one when the total size exceeds `EXPORTER_MAXIMUM_ENVELOPE_SIZE`, which will has the potential to create problems that are hard to trace.

I think a good first effort here, since we're not initially catering for multiple envelopes per workbasket, would be to impose no maximum size, and let our envelope consumer (HMRC) reject. The associated workbasket can then be reworked / decomposed into something that fits within consumer size constraints and resubmitted to the packaging queue.

Either that, or raise an exception if multiple envelopes are created. But I'm not sure what provision we have for dealing with that situation in the UI at the moment.

publishing/models/envelope.py
```diff
@@ -0,0 +1,218 @@
...
+            logger.info(f"rendered_envelope {rendered_envelope}")
+            envelope_file = rendered_envelope.output
+            if not rendered_envelope.transactions:
+                msg = f"{envelope_file.name}  is empty !"
+                logger.error(msg)
+                raise EnvelopeNoTransactions(msg)
```
paulpepper-trade, line 197: Do we have provision to deal with this exception in the UI?

publishing/models/envelope.py
```diff
@@ -0,0 +1,218 @@
...
+            # Transaction envelope data XML is valid, ready for upload to s3
+            else:
+                envelope_file.seek(0, os.SEEK_SET)
+                try:
+                    validate_envelope(envelope_file)
+                except etree.DocumentInvalid:
+                    logger.error(f"{envelope_file.name}  is Envelope invalid !")
```
paulpepper-trade, line 204: Won't this allow the envelope generation failure to go unnoticed in the UI? It so, then it would be better to let the original exception bubble up or catch it, add more info, and re-raise.

publishing/models/packaged_workbasket.py
```diff
@@ -1,17 +1,9 @@
...
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
-from django.db.models import F
-from django.db.models import Max
-from django.db.models import Q
-from django.db.models import QuerySet
-from django.db.models import TextChoices
-from django.db.models import Value
-from django.db.models.functions import Coalesce
```
paulpepper-trade, line 14: Maybe it's a matter of taste, but it may have been easier to leave these in place - it reduces the `models.` prefix clutter where these functions are used.

publishing/models/packaged_workbasket.py
```diff
@@ -519,7 +344,7 @@ def create_envelope_for_top(cls):
                 seconds_delay=1,
             )
         else:
-            logging.info(
+            logger.info(
```
paulpepper-trade, line 347: Good catch!

publishing/views.py
```diff
@@ -9,18 +9,17 @@
...
-from publishing.models import PackagedWorkBasketDuplication
-from publishing.models import PackagedWorkBasketInvalidCheckStatus
-from publishing.models import PackagedWorkBasketInvalidQueueOperation
-from publishing.models import ProcessingState
+from publishing.models.exceptions import PackagedWorkBasketDuplication
+from publishing.models.exceptions import PackagedWorkBasketInvalidCheckStatus
+from publishing.models.exceptions import PackagedWorkBasketInvalidQueueOperation
+from publishing.models.operational_status import OperationalStatus
+from publishing.models.packaged_workbasket import PackagedWorkBasket
+from publishing.models.state import ProcessingState
```
paulpepper-trade, line 22: Your taste in individual imports has changed earlier^^ ;-)

publishing/views.py
```diff
@@ -274,9 +267,11 @@ def transition_packaged_work_basket(self, packaged_work_basket):
...
     def get_context_data(self, **kwargs):
         data = super().get_context_data(**kwargs)
-        # TODO: Get the envelope ID when envelope management is done.
-        envelop_id = "220999"
-        data["message"] = f" Envelope ID {envelop_id} was accepted."
+
+        packaged_workbasket = self.get_object()
+        envelope = packaged_workbasket.envelope
+
+        data["message"] = f" Envelope ID {envelope.envelope_id} was accepted."
```
paulpepper-trade, line 274: Is that an unnecessary space / typo at the start of the message?

publishing/views.py
```diff
@@ -303,11 +298,13 @@ def transition_packaged_work_basket(self, packaged_work_basket):
...
-        # TODO: Get the envelope ID when envelope management is done.
-        envelop_id = "220999"
+
+        packaged_workbasket = self.get_object()
+        envelope = packaged_workbasket.envelope
+
         data[
             "message"
-        ] = f" Envelope ID {envelop_id} was rejected and queue was paused."
+        ] = f" Envelope ID {envelope.envelope_id} was rejected and queue was paused."
```
paulpepper-trade, line 307: Space typo?

paulpepper-trade (changes requested):

common/tests/factories.py
```diff
@@ -1243,6 +1243,29 @@ class Meta:
...
+    WorkBasket that contains transactions."""
+
+    workbasket = factory.SubFactory(QueuedWorkBasketFactory)
+
+
+class PublishedEnvelopeFactory(factory.django.DjangoModelFactory):
+    """Creates an Envelope instance."""
+
+    class Meta:
+        model = "publishing.Envelope"
```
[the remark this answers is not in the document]
a-gleeson: added models into init as suggested

publishing/migrations/0001_initial.py
```diff
@@ -41,7 +71,7 @@ class Migration(migrations.Migration):
                         blank=True,
                         null=True,
                         storage=publishing.storages.LoadingReportStorage,
-                        upload_to=publishing.models.report_bucket,
+                        upload_to="",
```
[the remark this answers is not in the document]
a-gleeson: migration had initially set this to blank, tested with it removed and it worked

publishing/models/envelope.py
```diff
@@ -0,0 +1,218 @@
...
+            .filter(
+                packagedworkbaskets__processing_state=ProcessingState.SUCCESSFULLY_PROCESSED,
+            )
+            .last()
+        )
+
+        if envelope is None:
+            # First envelope of the year.
+            now = datetime.today()
+            counter = 1
```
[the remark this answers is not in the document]
a-gleeson: Implemented and tested

publishing/models/envelope.py
```diff
@@ -0,0 +1,218 @@
...
+        else:
+            year = int(envelope.envelope_id[:2])
+            counter = int(envelope.envelope_id[2:]) + 1
```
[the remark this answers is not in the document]
a-gleeson: Implemented

publishing/models/envelope.py
a-gleeson: Removed for the team being but can be raised as something we should be aware of!

publishing/views.py
a-gleeson: Was for a Panel Title removed space

publishing/views.py
a-gleeson: Was for a Panel Title removed space

publishing/models/envelope.py
```diff
@@ -0,0 +1,218 @@
...
+        envelope_id = Envelope.next_envelope_id()
+        envelope = super().create(envelope_id=envelope_id, **kwargs)
+        envelope.upload_envelope(
+            packaged_work_basket.workbasket,
+        )
+        return envelope
+
+
+class EnvelopeQuerySet(models.QuerySet):
+    def envelopes_by_year(self, year: Optional[int] = None):
```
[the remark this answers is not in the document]
a-gleeson: updated as suggest :)

publishing/views.py
paulpepper-trade, line 22:
```suggestion
from publishing.models import PackagedWorkBasketDuplication
from publishing.models import PackagedWorkBasketInvalidCheckStatus
from publishing.models import PackagedWorkBasketInvalidQueueOperation
from publishing.models import OperationalStatus
from publishing.models import PackagedWorkBasket
from publishing.models import ProcessingState
```

sample.env
```diff
@@ -12,11 +12,16 @@ CELERY_BROKER_URL=redis://127.0.0.1:6379/1
 # S3 Bucket for HMRC envelope uploads.
 HMRC_STORAGE_BUCKET_NAME=hmrc
 HMRC_STORAGE_DIRECTORY=tohmrc/staging/
-LOADING_REPORTS_BUCKET_NAME=loading_reports
 AWS_ACCESS_KEY_ID=minio_access_key
 AWS_SECRET_ACCESS_KEY=minio_secret_key
 AWS_S3_ENDPOINT_URL=http://localhost:9003
 
+# S3 packaging automation.
+HMRC_PACKAGING_SEED_ENVELOPE_ID=230000
```
paulpepper-trade, line 20: If we don't need the year portion of this ID (as previous comment), then:
```suggestion
HMRC_PACKAGING_SEED_ENVELOPE_ID=0001
```

paulpepper-trade (changes requested):

publishing/admin.py
```diff
@@ -78,3 +131,11 @@ def has_delete_permission(self, request, obj=None):
 
 
 admin.site.register(OperationalStatus, OperationalStatusAdmin)
+
+admin.site.register(Envelope, EnvelopeAdmin)
+
+
+#   Envelope.objects.for_year()
```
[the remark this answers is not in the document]
a-gleeson: Will remove

publishing/management/commands/create_envelope_for_top.py
```diff
@@ -1,6 +1,6 @@
 from django.core.management import BaseCommand
 
-from publishing.models import PackagedWorkBasket
+from publishing.models.packaged_workbasket import PackagedWorkBasket
```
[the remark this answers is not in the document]
a-gleeson: Will make changes

publishing/jinja2/includes/packaged-workbasket-queue.jinja
```diff
@@ -109,7 +109,9 @@
       {# First row. #}
       {% if obj == currently_processing %}
         <span class="processing-state tamato-badge-light-green">CDS DOWNLOADED</span>
-      {% elif obj.processing_state == "AWAITING_PROCESSING" and not obj.envelope %}
+      {% elif obj.processing_state == "AWAITING_PROCESSING" and 
+        (not obj.envelope or not obj.envelope.xml_file)
```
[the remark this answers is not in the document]
a-gleeson: I was thinking that, more than happy to implement this

publishing/models/envelope.py
```diff
@@ -0,0 +1,225 @@
...
+            .filter(
+                packagedworkbaskets__processing_state=ProcessingState.SUCCESSFULLY_PROCESSED,
+            )
+            .last()
+        )
+
+        if envelope is None:
+            # First envelope of the year.
+            now = datetime.today()
+            counter = max(1, int(settings.HMRC_PACKAGING_SEED_ENVELOPE_ID[2:]))
```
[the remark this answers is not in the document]
a-gleeson: When the env was implemented the default was a string like "230000" should I update this to just the sufix of the ID?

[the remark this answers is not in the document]
paulpepper-trade: Yes, because we don't use the year part in the code, and always slice it off, then I think just the ID is needed.

sample.env
a-gleeson: Ah I missed this we can update that and the default in the settings

paulpepper-trade: Yeah, go for it.

paulpepper-trade (approved):

PR #794, opened 2023-01-30
Add API published flag with admin access

paulpepper-trade:
# Add API published flag with admin access.

## Why

There's no way of determining whether an envelope has been sent to the tariff-api service within Tamato. 

## What

This PR adds a flag, `Envelope.published_to_tariffs_api` to the Envelope model to show whether and when an envelope has been sent to the tariff-api service. This flag will be set manually, within the admin site by a member of staff on packaging duty, after they have pushed a published envelope to the production tariff-api service.

## Checklist
- Requires migrations? Yes


publishing/admin.py
```diff
@@ -108,14 +108,15 @@ class EnvelopeAdmin(admin.ModelAdmin):
         "envelope_id",
         "packagedworkbaskets_processing_state",
         "packagedworkbaskets_workbasket_id",
+        "published_to_tariffs_api",
         "deleted",
-        # TODO add user
```
a-gleeson, line 112: Thanks, this was a todo where I was thinking if it would be useful to include a user in the data. Although user for what I am not sure, would it be creator of envelope, accept/rejecter, etc

a-gleeson (approved):
LGTM!

publishing/admin.py
paulpepper-trade: Envelope creator should always be the system (auto-created on a background task). We don't capture who has packaged a workbasket (DIT-side) or started procssing an envelope (HMRC-side), so it's not something we can do right now. We may need to capture that info at some point in the future as you suggest though

PR #795, opened 2023-01-30
from issues identified when addressing TOPS-957

dougmills-DIT:
# TOPS-957
when there are two or more indents attached to a single goods nomenclature with the same validity_start date the query will randomly select one of the values, as the is no order_by clause

## Why
 * This bug impacted ME32 issues, as randomly if the above scenario exists it will flip between different indents

## What
 * modified the CommodityCollectionLoader._apply_filters method to get the correct / latest indent based on the trackedmodel_ptr_id

## Checklist
- Requires migrations? no

Links to relevant material
See: [tops ticket](https://uktrade.atlassian.net/browse/TOPS-957)
See: [original data change requirement tops ticket](https://uktrade.atlassian.net/browse/TOPS-946)


commodities/tests/test_commodities_collection.py
```diff
@@ -0,0 +1,61 @@
...
+            item_id="2903691900",
+            suffix=80,
+            indent__indent=3,
+        )
+
+        target = CommodityCollection([parent_good, child_good_1, child_good_2])
+        assert len(target.commodities) == 3
+
+    def test_init_invalid_arguments(self):
+        # with pytest.raises()
```
a-gleeson, line 40: is this init test missing pytest.raises?

commodities/tests/test_commodities_collection.py
```diff
@@ -0,0 +1,61 @@
...
+
+        target = CommodityCollection([Commodity(current)])
+        change_sequence.append(
+            CommodityChange(
+                collection=target,
+                update_type=UpdateType.DELETE,
+                current=Commodity(current),
+            ),
+        )
+        target.update(change_sequence)
```
a-gleeson, line 61: should we assert that the target is correctly updated? Ie/ removed from collection

a-gleeson (approved):
A couple comments on the tests but otehr than that LGTM!

commodities/tests/test_commodities_collection.py
dougmills-DIT: removed for now, I thought it would raise an exception but didn't, the initialiser is using Django model auto initialiser magic, added a TODO on the model to improve init arguments validation (maybe in post_init or something like that)

commodities/tests/test_commodities_collection.py
dougmills-DIT: good catch, was rushing and forgot to add an assert, fixed now

PR #800, opened 2023-01-31
TP2000-723: Multiple objects returned error fix

eadpearce:
# TP2000-723: Multiple objects returned error fix

## Why

* If there was more than one version of a quota order number doing QuotaOrderNumber.objects.get(sid=sid) would raise a MultipleObjectsReturned error

## What

* Make sure to use .current() to only get the most recent version of the quota order number


a-gleeson (approved):
LGTM! Cannot think of what test

PR #803, opened 2023-02-01
from issues identified when addressing TOPS-957

dougmills-DIT:
# TOPS-957
when there are two or more indents attached to a single goods nomenclature with the same validity_start date the query will randomly select one of the values, as the is no order_by clause

## Why
 * This bug impacted ME32 issues, as randomly if the above scenario exists it will flip between different indents

## What
 * modified the CommodityCollectionLoader._apply_filters method to get the correct / latest indent based on the trackedmodel_ptr_id

## Checklist
- Requires migrations? no

Links to relevant material
See: [tops ticket](https://uktrade.atlassian.net/browse/TOPS-957)
See: [original data change requirement tops ticket](https://uktrade.atlassian.net/browse/TOPS-946)


dougmills-DIT:
@paulpepper-trade we have tested this update, it preserves the correct indent and does not lock up rules checks

commodities/models/dc.py
```diff
@@ -692,6 +692,8 @@ class CommodityCollection(CommodityTreeBase):
     - see the docs to `CommodityCollection.get_snapshot` method for more details.
     """
 
+    # TODO : add validation to initializer, currently can be initialised with invalid arguments as long as its in a list
```
paulpepper-trade, line 695: Possibly best removed from here and placed in a Jira ticket?

paulpepper-trade (approved):

commodities/models/dc.py
dougmills-DIT: fair point

PR #808, opened 2023-02-03
adding gtm_auth config param

a-gleeson:
# TP2000-712 adding gtm_auth config param

## Why
The google analytics team had sent accross script logic to add to the project. 
Previous script logic was missing gtm_auth param.

## What
added config for gtm_auth and will handle when the configuartion is not present


common/jinja2/layouts/_generic.jinja
```diff
@@ -39,7 +40,12 @@
   {% if settings.GOOGLE_ANALYTICS_ID %}
     <!-- Google Tag Manager (noscript) -->
       <noscript>
-        <iframe src="https://www.googletagmanager.com/ns.html?id={{ settings.GOOGLE_ANALYTICS_ID }}"height="0" width="0" style="display:none;visibility:hidden"></iframe>
+        <iframe src="
+            https://www.googletagmanager.com/ns.html
+            ?id={{ settings.GOOGLE_ANALYTICS_ID }}
+            &gtm_cookies_win=x&gtm_preview=env-28
+            &{{ "gtm_auth" ~ settings.GTM_AUTH if settings.GTM_AUTH }}"
```
paulpepper-trade, line 47: Just to be sure, is gtm_auth optional? For instance, in staging or dev environments, does this have to be disabled?

a-gleeson: Hi Paul, yes gtm_auth is optional. I have also noticed that I should parameterise gtm_preview as this notes the environment we are producing data for. 
Some additional discussion with the GA team may be useful as previously our set up just fed data into GA and the hostname can be used to filter the data. If we can get away with removing gtm from the parameters in the URL then that can be less information shared publically (gtm params) when our web page is sending GA data from the client.

paulpepper-trade (approved):

.pre-commit-config.yaml
```diff
@@ -14,12 +14,12 @@ repos:
...
           ]
   - repo: https://github.com/pycqa/isort
-    rev: 5.10.1
+    rev: 5.12.0
     hooks:
       - id: isort
         args: [--force-single-line-imports]
   - repo: https://github.com/myint/docformatter.git
-    rev: v1.5.0
+    rev: v1.6.0.rc1
```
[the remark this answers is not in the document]
a-gleeson: I will need to revisit this, last week I was not able to commit as pre-commit was failing as it required an update. 
I can undo the pre-commit changes if it's the best option but will have to do it from github ui if I cannot fix the commiting issue on the older versions

paulpepper-trade (approved):

PR #809, opened 2023-02-07
TP2000-731: Persist measure selections across pages

eadpearce:
# TP2000-731: Persist measure selections across pages

## Why

* If measure search results span multiple pages, users need a way to select measures from all pages

## What

* Adds a view for ajax post requests from the frontend to update the session
* Extends the preexisting checkboxes JS to track checkbox state and uses ajax to sync this with the session state
* For when JS is not available there is an extra button at the bottom of the page that submits the selection and saves it to the session

Screenshot 2023-02-10 at 10 06 26


common/static/common/js/checkboxes.js
```diff
@@ -0,0 +1,78 @@
+
+class CheckBoxes {
```
paulpepper-trade, line 2: Does this offer a progressive enhancement?

paulpepper-trade (changes requested):

common/static/common/js/checkboxes.js
eadpearce: The checkboxes don't disappear when JS is disabled but it will revert back to previous functionality where checked items don't persist across pages

eadpearce: Added an extra button for no JS solution

measures/views.py
```diff
@@ -86,29 +89,44 @@ def paginator(self):
...
             url = reverse("measure-ui-edit-multiple-end-date")
+        elif form.data["form-action"] == "persist-selection":
+            # keep selections from other pages. only update newly selected/removed measures from the current page
+            selected_objects = {k: v for k, v in form.cleaned_data.items() if v}
+
+            # clear this page from the session
+            self.session_store.remove_items(form.cleaned_data)
+
+            # then add the selected items
+            selected_objects = {k: v for k, v in form.cleaned_data.items() if v}
```
a-gleeson, line 125: Is this repeated line required?

eadpearce: oh yeah oops. i can get rid of that

common/static/common/js/application.js
```diff
@@ -15,5 +15,5 @@ showHideCheckboxes();
 initAutocomplete(false);
 initStepNav();
 initAll();
-initCheckAllCheckboxes();
+initCheckboxes();
```
a-gleeson, line 18: Is this where this initialisation belongs? Would this not set up the event listeners on every page? 
Would that possibly effect page performance?

eadpearce: Here it checks if the script element exists on the page, otherwise it doesn't run it https://github.com/uktrade/tamato/pull/809/files/108cb51bb52ca6d6d1e57247be5d1dd17a51a3bd#diff-7ddd91f71ff729075d86689c0edde6ccaad476adc5942d4c88372200786750b5R105

common/static/common/js/checkboxes.js
```diff
@@ -0,0 +1,111 @@
...
```
a-gleeson, line 2: Should we rename this script? Checkbox is specific to measures so would not be reusable for other integrations unless parameterised?

common/static/common/js/application.js
a-gleeson: Thanks, that looks good! Happy with this. Would have to have a deeper dive but would this integration be useful to use htmx? https://htmx.org/attributes/hx-trigger/ . I think the integration with htmx would be a little more complicated in this case due to "multiple page"  persistence of checkboxies and noy required for this PR.

a-gleeson (approved):

common/static/common/js/checkboxes.js
eadpearce: Now that you've said that I've realised this has broken the checkboxes on the workbasket page. I'll go and fix that

a-gleeson: Good catch!

a-gleeson (approved):

paulpepper-trade (approved):

PR #810, opened 2023-02-07
TP2000-695 S3 loading reports integration

paulpepper-trade:
# TP2000-695 S3 loading reports integration

## Why

As part of the packaging automation work, TAP is required to save HMRC's loading reports to S3 and make them available to view and download.

## What

This PR primarily:

- Adds support for uploading a loading report file during the HMRC accept and reject envelope journeys.
- Formats the email notification emails with details of the loading report.
- Makes loading reports available to view, filter and download from within the Django admin site.
- Adds bulk update to enrol and un-enrol users for email notifcations in Django admin.

In addition to the title Jira ticket, this PR fixes issues described in the following tickets:

- [TP2000-732](https://uktrade.atlassian.net/browse/TP2000-732): include packager's description message in ready for processing message, correctly format.
- [TP2000-733](https://uktrade.atlassian.net/browse/TP2000-733): open Jira link in separate window.
- [TP2000-735](https://uktrade.atlassian.net/browse/TP2000-735): increase max comments to 1000 words.

This changeset adds the new variables `loading_report_message` and `description` to Notify templates and also removes the `link_to_file` variable. Because Notify requires all variables to be passed in to a template, the following deployment sequence is necessary:

1. Deploy this PR.
2. Update the Notify templates to add the new template variables, `loading_report_message` and `description`.
3. Update the Notify templates to remove the deleted template variable, `link_to_file`.
4. Deploy a follow-on PR that removes `link_to_file` from the set of variables passed into the Notify templates.

Once this PR has been merged and deployed, then [PR 811](https://github.com/uktrade/tamato/pull/811) can be deployed in order to remove the Notify template variable `link_to_file`.

Unit tests are included as part of the ticket [TP2000-720](https://uktrade.atlassian.net/browse/TP2000-720).

## Checklist
- Requires migrations? Yes

[TP2000-732]: https://uktrade.atlassian.net/browse/TP2000-732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


publishing/admin.py
```diff
@@ -166,6 +197,93 @@ def has_delete_permission(self, request, obj=None):
...
+
+        state = pwb.processing_state
+        if state == ProcessingState.SUCCESSFULLY_PROCESSED:
+            return "Accepted"
+        elif state == ProcessingState.FAILED_PROCESSING:
+            return "Rejected"
+
+        return f"Unexpected state"
+
+    def packaged_workbasket_id(self, obj):
```
a-gleeson, line 257: This function is repeated for envelope & loading report, should we maybe extract it so it can be reused?

publishing/admin.py
```diff
@@ -166,6 +197,93 @@ def has_delete_permission(self, request, obj=None):
...
+        if not pwb:
+            return "Missing packaged workbasket!"
+
+        pwb_url = reverse(
+            "admin:publishing_packagedworkbasket_change",
+            args=(pwb.pk,),
+        )
+        return mark_safe(f'<a href="{pwb_url}">{pwb.pk}</a>')
+
+    def workbasket_id(self, obj):
```
a-gleeson, line 268: Repeated in all 3 admin models, should we reuse?

publishing/models/packaged_workbasket.py
```diff
@@ -535,63 +532,69 @@ def notify_ready_for_processing(self):
...
+        `template_id` should be the ID of the Notify email template of either
+        the successfully processed email or failed processing.
         """
-
-        link_to_file = "None"
+        loading_report_message = "No loading report was provided."
         if self.loading_report.file:
-            f = self.loading_report.file.open("rb")
-            link_to_file = prepare_upload(f)
+            self.loading_report.file
```
a-gleeson, line 566: is this line needed?

publishing/views.py
```diff
@@ -229,13 +231,37 @@ def get(self, request, *args, **kwargs):
...
+
+    def get(self, request, *args, **kwargs):
+        loading_report = self.get_object()
+        file_name = (
+            loading_report.file_name if loading_report.file_name else "UNKNOWN_FILENAME"
+        )
+        content = loading_report.file.read()
+        response = HttpResponse(content)
+        response["content-length"] = len(content)
+        response["content-disposition"] = f'attachment; filename="{file_name}"'
```
a-gleeson, line 248: should we also include content type?

publishing/models/packaged_workbasket.py
paulpepper-trade: I expect it's a surplus copy + paste. Removed.

publishing/admin.py
paulpepper-trade: Now factored out into a mixin.

publishing/admin.py
paulpepper-trade: Now factored out into a mixin.

publishing/views.py
paulpepper-trade: I think there's a few reasons why we might not wish to - it's an optional header; the uploader's content-type shouldn't be trusted if it is provided; and there's always a chance that the file could be changed via other means (such as in Django admin) so it could become invalid / incorrect. We're not expecting to display the file inside a user's browser so I think it's less useful to capture the content type here. If we decide to load the report inside the user's browser, then we could employ content type sniffing and attach a content-type header at that point.
Do you think there may be cases where it's important to capture and store the content type?

a-gleeson (approved):
LGTM!

publishing/views.py
paulpepper-trade: Content-type header added, which is harmless. We had code in the importer that does MIME type sniffing, so that's now factored out into `common/util.py` and so is now more widely available.

PR #822, opened 2023-02-16
TP2000-682 Amend weak password policy

LaurenQurashi:
# TP2000-682

## Why
Thanks to a Security audit, it was Identified that when creating a username and password the old fashioned way without Staff SSO, the password manager would accept anything as a password. We don't use this functionality much, but it is in practice when staff SSO is not enabled in the project such as spinning up a local host. God knows how the auditors found that though. 

## What
This PR adds some simple settings to have a mandatory 12 character minimum for passwords, and checks the passwords against 20K odd similar passwords, under the recommendations of the pen test. 

## Checklist
- Requires migrations? - nah
- Requires dependency updates? - nope

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -23,6 +23,11 @@ SSO_ENABLED = is_truthy(os.environ.get("SSO_ENABLED", "true"))
 VCAP_SERVICES = json.loads(os.environ.get("VCAP_SERVICES", "{}"))
 VCAP_APPLICATION = json.loads(os.environ.get("VCAP_APPLICATION", "{}"))
 
+# -- Debug
+
+# Activates debugging
+DEBUG = is_truthy(os.environ.get("DEBUG", False))
+
 # -- Paths
 
 # Name of the project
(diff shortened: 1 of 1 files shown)


settings/common.py
```diff
@@ -172,6 +172,19 @@
 
 # -- Auth
 LOGIN_URL = reverse_lazy("login")
+
+AUTH_PASSWORD_VALIDATORS = [
+    {
+        'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
+        'OPTIONS': {
+            'min_length': 12,
```
paulpepper-trade, line 180: Will this force a minimum password length of 12 in dev environments? If so, then maybe including these validator could be predicated on 'DEBUG == False' or similar?

paulpepper-trade (commented):
> We don't use this functionality much, but it is in practice when staff SSO is not enabled in the project such as spinning up a local host. God knows how the auditors found that though.

I added [a comment to the Jira ticket](https://uktrade.atlassian.net/browse/TP2000-682?focusedCommentId=107973), which may explain this.

settings/common.py
LaurenQurashi: That's a good shout, I'll chuck an If in!

settings/common.py
```diff
@@ -172,6 +177,20 @@
 
 # -- Auth
 LOGIN_URL = reverse_lazy("login")
+
+if DEBUG is False:
```
paulpepper-trade, line 181: Perhaps a bit more pythonic:
```suggestion
if not DEBUG:
```

paulpepper-trade (approved):

PR #824, opened 2023-02-17
TP2000-743 Edit multiple measures start and end dates

eadpearce:
# TP2000-743 Edit multiple measures start and end dates

## Why

* Users want to be able to edit start dates as well as end dates on measures

## What

* Adds the end date form and a new start date form to a form wizard
* Users can select which fields they wish to edit at the start of the form

FireShot Capture 050 - Edit measures - Manage Trade Tariffs - localhost
Screenshot 2023-02-17 at 13 48 19
Screenshot 2023-02-17 at 13 48 32

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1218,6 +1218,7 @@ class MeasureStartDateForm(forms.Form):
 
                 upper = measure.valid_between.upper
                 lower = datetime.date(year, month, day)
+                # for an open-ended measure the end date can be None
                 if upper and lower > upper:
                     formatted_lower = lower.strftime("%d/%m/%Y")
                     formatted_upper = upper.strftime("%d/%m/%Y")
@@ -1234,7 +1235,7 @@ class MeasuresEditStartForm(forms.Form):
         choices=MeasureEditSteps.choices,
         widget=forms.CheckboxSelectMultiple,
         label="",
-        help_text="Select all that apply",
+        help_text="Select the fields you wish to edit",
     )
 
     def __init__(self, *args, **kwargs):
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1230,7 +1230,7 @@ class MeasureStartDateForm(forms.Form):
         return cleaned_data
 
 
-class MeasuresEditStartForm(forms.Form):
+class MeasuresEditFieldsForm(forms.Form):
     fields_to_edit = forms.MultipleChoiceField(
         choices=MeasureEditSteps.choices,
         widget=forms.CheckboxSelectMultiple,
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -34,6 +34,7 @@ from footnotes.models import Footnote
 from geo_areas.models import GeographicalArea
 from geo_areas.validators import AreaCode
 from measures import models
+from measures.constants import MeasureEditSteps
 from measures.parsers import DutySentenceParser
 from measures.patterns import MeasureCreationPattern
 from measures.util import diff_components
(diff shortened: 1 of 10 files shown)


measures/views.py
```diff
@@ -70,7 +72,31 @@ def get_queryset(self):
...
+    def measure_selections(self):
+        return [
+            SelectableObjectsForm.object_id_from_field_name(name)
+            for name in [*self.session_store.data]
+        ]
+
+
+class MeasureSelectionQuerysetMixin(MeasureSelectionMixin):
+    def get_queryset(self):
+        """Get the measures that are candidates for deletion."""
```
a-gleeson, line 95: Just for deletion or is it selected candiates?

measures/forms.py
```diff
@@ -1182,3 +1182,76 @@ def clean(self):
                     )
 
         return cleaned_data
+
+
+class MeasureStartDateForm(forms.Form):
```
a-gleeson, line 1187: Could the edit fields form be abstracted and reused? Or is it not worth the effort?

measures/views.py
```diff
@@ -163,6 +178,88 @@ def get_context_data(self, **kwargs: Any):
...
+        MeasureEditSteps.END_DATE: {
+            "title": "Edit the end date",
+            "link_text": "End date",
+        },
+    }
+
+    def get_template_names(self):
+        return self.templates.get(
+            self.steps.current,
+            "measures/edit-wizard-step.jinja",
```
a-gleeson, line 218: Should the default be the start/landing page for edits?

a-gleeson:
Do you think having the form start labeled as just "Start" is clear? When I started reviewing I got a little tripped up between start and start date. Maybe something like "landing" is good although that still may not be clear

measures/views.py
eadpearce: oop sorry this is a copypaste error. it's selection/editing

measures/forms.py
eadpearce: Yeah not worth it I think. The forms are going to be very slightly different

measures/views.py
eadpearce: Doesn't really matter. This is only a fallback in case it can't find the given key in the dict

eadpearce:
@a-gleeson Yeah that's a good point. I can change it to something like MeasuresEditFieldsForm ?

a-gleeson:
> @a-gleeson Yeah that's a good point. I can change it to something like MeasuresEditFieldsForm ?

Sounds good to me :)

measures/views.py
```diff
@@ -163,6 +178,88 @@ def get_context_data(self, **kwargs: Any):
         return context
 
 
+@method_decorator(require_current_workbasket, name="dispatch")
+class MeasureEditWizard(
+    MeasureSelectionQuerysetMixin,
+    NamedUrlSessionWizardView,
+):
```
paulpepper-trade, line 185: I think we need a docstring on this class to explain what it does and how it works. It's an important addition to the way that we manage measures through the UI and we should do our best to minimise the chances of running into those difficulties that we've encountered recently around this area.

paulpepper-trade (changes requested):

measures/views.py
paulpepper-trade: A docstring would also help us in reviewing this PR.

eadpearce: Done

paulpepper-trade (approved):
I think we'd benefit from more docs about how the wizard is used in this specific case, but it appears we may have a separate activity for that now.

@a-gleeson has given some good comments, so worth getting his approval too?

a-gleeson (approved):

eadpearce:
@paulpepper-trade Not sure why since everything I've done with the wizard is covered in the django form-tools docs

PR #825, opened 2023-02-17
TP2000-759 Fix werkzeug dependabot alerts

paulpepper-trade:
# TP2000-759 Fix werkzeug dependabot alerts

## Why

Dependabot has raised two alerts for the WerkZeug package used by Tamato. The auto-generated PR breaks Tamato's CI checks so it's necessary to make additional manual changes. 

## What

This PR:
Updates the WerkZeug package to a version that resolves the Dependabot alerts.
Resolves clashing requirements dependencies.
Manual testing in a staging environment (Training) didn't uncover any breaking changes to the Tamato.

## Checklist
- Requires migrations? No

### DIFF
--- a/requirements-dev-jupyter.txt
+++ b/requirements-dev-jupyter.txt
@@ -1,0 +1,4 @@
+-r requirements-dev.txt
+
+ipython
+jupyter
--- a/common/jinja2.py
+++ b/common/jinja2.py
@@ -12,7 +12,7 @@ from django.utils.safestring import mark_safe
 from django.utils.timezone import template_localtime
 from govuk_frontend_jinja.templates import Environment
 from govuk_frontend_jinja.templates import NunjucksExtension
-from jinja2 import Markup
+from jinja2.utils import markupsafe
 from webpack_loader.templatetags.webpack_loader import render_bundle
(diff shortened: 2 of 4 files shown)


requirements-dev-jupyter.txt
```diff
@@ -0,0 +1,4 @@
+-r requirements-dev.txt
```
a-gleeson, line 1: This is good to include! Although requirements-dev already pulls down our "development libraries" it isn't too long would it still make sense to have these dependencies in there?

a-gleeson (approved):

requirements-dev-jupyter.txt
paulpepper-trade: I put the Jupyter packages in a separate requirements file because not everyone uses it and it does pull in a lot of dependended-upon packages. I think that breaking things up in this way offers developers the choice - though I'd recommend Jupyter for all sorts of good reasons!

a-gleeson: This makes sense to me, no point pulling down the dependicies if they are not being used. This will help me advise making the same type of changes to the docker set up

PR #826, opened 2023-02-20
TP2000-758 Update quota order number page in create measure journey

dalecannon:
# TP2000-758 Update quota order number page in create measure journey

## Why
The quota order number page in the create measure journey requires small content amendment to make it consistent with the other pages. 

## What
- Amends title (adds optional to title, uses lowercase)
- Removes now redundant sentence in help text
- Extends unit test coverage

Before:
Screenshot 2023-02-20 at 10 35 35

After:
after-quota-order-number


paulpepper-trade (approved):
LGTM.

PR #827, opened 2023-02-23
Session expiry constants added

LaurenQurashi:
# TP-2000-683/686

## Why
The recent Pen Testing on TAP highlighted an issue with our session length and login/out controls. After doing some digging, we found that the login/out controls don't work for those using envs that have Staff SSO switched on. This is due to Staff SSO only signing a user out once a week (rumoured), due to the quantity of apps that it signs you into (G-Suite, all DIT apps, other stuff..). As we have no control over auth on Tap, we've decided to hide the logout controls for those envs that have SSO implemented. 

For the app's sessions, I've added some settings which reduce the length of the session, in order to combat security issues. I've added a session cookie age of 40 minutes, a setting to kill the session on browser close, and a setting to save every request. In combination, these session settings refresh a user's session cookie age every time they make a request, but after 40 minutes of being idle, the session cookie will go stale and be renewed on the next request again. Whenever a user closes the tab running tap, it'll also be killed. This should mitigate any security risks related to having an everlasting session /session cookie. 

## What
- Log in and out controls have been hidden behind an SSO_ENABLED flag. If staff sso is on, you wont see them. We need to keep this functionality though for scenarios where we need to log out on dev spinups that don't have sso enabled like localhost. 

SESSION_SAVE_EVERY_REQUEST 
has been set to true so that all data made in a request get the cookie age timer refreshes 

SESSION_EXPIRE_AT_BROWSER_CLOSE = True
has been set to kill the session when the browser is closed

SESSION_COOKIE_AGE = 40 * 60
has been set to expire the session cookie after 40 minutes.

## Checklist
- Requires migrations? no


eadpearce:
@LaurenQurashi are SESSION_SAVE_EVERY_REQUEST and SESSION_EXPIRE_AT_BROWSER_CLOSE settings used in the sso library? we don't need to do anything extra and the library just handles it?

common/tests/test_views.py
```diff
@@ -24,6 +25,34 @@ def test_index_displays_workbasket_action_form(valid_user_client):
     assert "Search the tariff" in page.select("label")[4].text
 
 
+def test_index_displays_login_buttons_correctly_SSO_off_logged_in(valid_user_client):
```
paulpepper-trade, line 28: Should this actually be named test_index_displays_**logout**_button_correctly_SSO_off_logged_in() since it's testing for the presence of a logout link?

common/tests/test_views.py
```diff
@@ -24,6 +25,34 @@ def test_index_displays_workbasket_action_form(valid_user_client):
...
+    settings.SSO_ENABLED = False
+    response = valid_user_client.get(reverse("home"))
+
+    assert response.status_code == 200
+
+    page = BeautifulSoup(str(response.content), "html.parser")
+    assert page.find_all("a", {"href": "/logout"})
+
+
+def test_index_displays_login_buttons_correctly_SSO_off_logged_out(client):
```
paulpepper-trade, line 38: This isn't testing for the display of any login buttons, so perhaps could be named to describe the "can't access home page if logged out and SSO disabled" behaviour.

common/tests/test_views.py
```diff
@@ -24,6 +25,34 @@ def test_index_displays_workbasket_action_form(valid_user_client):
...
+    settings.SSO_ENABLED = False
+    response = client.get(reverse("home"))
+
+    assert response.status_code == 302
```
paulpepper-trade, line 42: I suppose we know the Location value of where this redirects to. Is it worth checking that value in an assert?

paulpepper-trade (changes requested):

common/tests/test_views.py
```diff
@@ -24,6 +25,35 @@ def test_index_displays_workbasket_action_form(valid_user_client):
...
+    page = BeautifulSoup(str(response.content), "html.parser")
+    assert page.find_all("a", {"href": "/logout"})
+
+
+def test_index_redirects_to_login_page_logged_out_SSO_off(client):
+    settings.SSO_ENABLED = False
+    response = client.get(reverse("home"))
+
+    assert response.status_code == 302
+    assert response.url == "/auth/login/?next=%2F"
```
paulpepper-trade, line 43: Perhaps a little more robust:
```suggestion
    assert response.url.startswith(reverse('admin:login'))
```

paulpepper-trade (approved):

PR #829, opened 2023-02-24
TP2000-745 Add regulations to multiple measures edit form

eadpearce:
# TP2000-745 Add regulations to multiple edit form

## Why

* Users need to be able to edit the regulation for multiple measures

## What

* Adds an extra step to the multiple measure edits form wizard for the regulation

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1278,8 +1278,6 @@ class MeasureRegulationForm(forms.Form):
         self.selected_measures = kwargs.pop("selected_measures", None)
         super().__init__(*args, **kwargs)
 
-        self.fields["generating_regulation"].queryset = Regulation.objects.all()
-
         self.helper = FormHelper(self)
         self.helper.form_tag = False
         self.helper.legend_size = Size.SMALL
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1265,3 +1265,32 @@ class MeasuresEditFieldsForm(forms.Form):
                 data_prevent_double_click="true",
             ),
         )
+
+
+class MeasureRegulationForm(forms.Form):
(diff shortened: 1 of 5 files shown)


measures/forms.py
```diff
@@ -1254,3 +1254,32 @@ def __init__(self, *args, **kwargs):
...
+        label="Regulation ID",
+        help_text="Select the regulation which provides the legal basis for the measures.",
+        queryset=Regulation.objects.all(),
+    )
+
+    def __init__(self, *args, **kwargs):
+        self.selected_measures = kwargs.pop("selected_measures", None)
+        super().__init__(*args, **kwargs)
+
+        self.fields["generating_regulation"].queryset = Regulation.objects.all()
```
a-gleeson, line 1270: Why might the queryset for the "generating_regulation" field require the queryset set here and line 1263?

measures/views.py
```diff
@@ -270,6 +270,8 @@ def done(self, form_list, **kwargs):
                 cleaned_data["end_date"].month,
                 cleaned_data["end_date"].day,
             )
+        if cleaned_data.get("generating_regulation"):
```
a-gleeson, line 273: Could we remove line 260 and just use .get(,None) here as the default?

eadpearce: yep. thought I updated this but I guess I forgot

measures/forms.py
eadpearce: This is a holdover from when I was trying to figure out how to filter regulations according to the measure validity periods. I discussed it a bit with Doug and it ended up being way too complicated for any benefit it would add

a-gleeson (approved):
LGTM!

PR #831, opened 2023-02-27
TP2000-749: Add quota step to multiple measures edit form

eadpearce:
# TP2000-749: Add quota step to multiple measures edit form

## Why

* Users need to be able to edit the quota order number on multiple measures

## What

* Adds an extra step to the multiple measures edit wizard to allow editing of quota order number

Screenshot 2023-02-27 at 14 32 34
Screenshot 2023-02-27 at 14 32 56

### DIFF
--- a/measures/views.py
+++ b/measures/views.py
@@ -207,12 +207,11 @@ class MeasureEditWizard(
         (START, forms.MeasuresEditFieldsForm),
         (MeasureEditSteps.START_DATE, forms.MeasureStartDateForm),
         (MeasureEditSteps.END_DATE, forms.MeasureEndDateForm),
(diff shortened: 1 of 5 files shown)


measures/views.py
```diff
@@ -258,6 +261,7 @@ def done(self, form_list, **kwargs):
         workbasket = WorkBasket.current(self.request)
         new_start_date = None
         new_end_date = None
+        new_quota_order_number = None
```
a-gleeson, line 264: As line 277 .get has the default None do we need to set the variable default here?

eadpearce: oh yeah. can get rid of this

PR #834, opened 2023-03-01
TOPS-948 me32 issues

dougmills-DIT:
# TOPS-948 : 0703101900 is constantly failing rules for no reason

Note: Black updated a lot of files via the pre-commit hook which has created lots of changes not relevant to this PR

## Why
 * ME32 when looking at measures, would also include ARCHIVED workbasket changes 
 * No unit tests covering key complex parts of TAP 
 * Some unit tests used private methods in factoryboy that were not robust, and would fail if factories had multiple traits on a property due to the Maybe method providing multiple options and not a concrete factory

## What
 * explicitly excluded ARCHIVED workbaskets from the TrackedModelQuerySet : approved_up_to_transaction function (this is what did the trick) 
 * Updated the conftest : reference_nonexistent_record() to loop factory declarations and select the appropriate factory when presented with a Maybe object
 * Added tests to CommodityTreeSnapshot and MeasureSnapshot
 * Separated out ME32 tests into its own test file. 

## Checklist
- Requires migrations? No

Links to relevant material
See: [Jira Ticket](https://uktrade.atlassian.net/browse/TOPS-948)
Also see : [Another related Jira Ticket](https://uktrade.atlassian.net/browse/TP2000-134)


paulpepper-trade (approved):
It looks as though Black has been run over all the files in the repo, not just those that have changed as would happen with the pre-commit hooks (I think). It's a little difficult to see what should be reviewed!

I think the main things are the tests and the additional `exclude()` on the `TrackedModelQuerySet.approved_up_to_transaction()` which is nice, right? Apart from the excess formatting, LGTM.

PR #838, opened 2023-03-06
TP2000-773 Add measure condition details to create summary page

eadpearce:
# TP2000-773 Add measure condition details to create summary page

## Why

* Not enough information on the data that was entered in the create measure form was displayed on the summary page

## What

* Summary now shows measure condition data in line with what is displayed on the conditions tab on the measure detail page 

Screenshot 2023-03-06 at 13 52 24


paulpepper-trade (approved):
LGTM.

PR #839, opened 2023-03-06
Add validation to Workbasket Create form

LaurenQurashi:
# TP2000-698-input-validation

## Why
In order to prevent Cross Site Scripting, we need to add validation to the fields that accept user input, so anything the users can type into. Dropdowns, prepop, and choice fields aren't affected. This is what's been applied within each subapp: 

**Filters:** 
- The search on the side bar filters have had basic validation applied. 
- The code and ID fields on the Commodities and  Measures filters have validation applied.

**Quotas and Regulations:** 
- Required no changes due to either having no fields requiring validation, or validation applied on the model. 

**Footnotes:** 
- Required no changes as description requires HTML but is autoescaped by Django. 

**Measures:** 
- Create Measures form has had validation added to the fields that accept user input such as Duties etc. 

**Geographical Areas:** 
- Create description field had basic validation applied to it. 
- Description help details have been removed from the template. 

**Workbaskets:** 
- Create workbasket description field has basic validation applied to it. 
- Description help has been removed from the form. 
- Test added to create form

**Additional Codes:** 
- Number validation has been added to the Additional Code ID field. 
- Test added to the create form

**Certificates:** 
- Alphanumeric validation has been added to the certificate identifier field. 
- Test added to the create form

Additional Codes, Footnotes, and Certificates need to be able to take HTML in their descriptions, so no validation has been added to their description fields. These fields are safe however thanks to Edie's work on CSP headers, and Django autoescaping the text for us. This means if anyone tries executing script in the text fields, Django outs them when the script is rendered by showing the script as text - by escaping the characters so it's seen as a string. 

## Checklist
- Requires migrations? - nope
- Requires dependency updates? - nuhuh

### DIFF
--- a/common/tests/test_validators.py
+++ b/common/tests/test_validators.py
@@ -77,7 +77,7 @@ def test_alphanumeric_validator(value, expected_valid):
         ("Text When There Shouldn't be.", False),
         (1234, False),
         ("<Sketchy_Code></>", False),
-        (".,'()&£$%/@!", True),
+        (".,'()&£$%/@!-", True),
     ],
 )
 def test_symbol_validator(value, expected_valid):
--- a/common/tests/test_validators.py
+++ b/common/tests/test_validators.py
@@ -1,6 +1,7 @@
 import pytest
 
 from common.tests.util import check_validator
+from common.validators import AlphanumericValidator
 from common.validators import EnvelopeIdValidator
 from common.validators import NumberRangeValidator
 from common.validators import NumericSIDValidator
@@ -53,3 +54,17 @@ def test_numeric_sid_validator(value, expected_valid):
 )
 def test_envelope_id_validator(value, expected_valid):
     check_validator(EnvelopeIdValidator, value, expected_valid)
+
+
+@pytest.mark.parametrize(
+    ("value", "expected_valid"),
+    [
+        ("A Good Description", True),
+        ("A Good description with Numbers in 001", True),
+        (1234, True),
+        ("<Sketchy_Code></>", False),
+        ("40% of this is £10 & $15", True),
+    ],
+)
+def test_alphanumeric_validator(value, expected_valid):
+    check_validator(AlphanumericValidator, value, expected_valid)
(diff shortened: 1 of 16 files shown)


common/tests/test_validators.py
```diff
@@ -53,3 +56,43 @@ def test_numeric_sid_validator(value, expected_valid):
...
+)
+def test_symbol_validator(value, expected_valid):
+    check_validator(SymbolValidator, value, expected_valid)
+
+
+@pytest.mark.parametrize(
+    ("value", "expected_valid"),
+    [
+        ("Text When There Shouldn't be.", False),
+        (1234, True),
```
a-gleeson, line 92: The NumericValidator is applied to text fields should be testing numbers as strings?

common/validators.py
```diff
@@ -72,3 +72,16 @@ class ApplicabilityCode(models.IntegerChoices):
 
 
 EnvelopeIdValidator = RegexValidator(r"^(?P<year>\d\d)(?P<counter>\d{4})$")
+
+AlphanumericValidator = RegexValidator(
+    r"^[0-9A-Za-z\s.',\-]*$",
+    ValidationError("Only alphanumeric characters are allowed."),
+)
+NumericValidator = RegexValidator(
+    r"^[0-9\s]*$",
```
a-gleeson, line 81: should we be matching any whitespace characters in this regex? If there is whitespace inbetween digits does it just ignore them and also would any whitespace character be valid in this scenario?

common/validators.py
```diff
@@ -72,3 +72,16 @@ class ApplicabilityCode(models.IntegerChoices):
...
+    ValidationError("Only numbers are allowed."),
+)
+SymbolValidator = RegexValidator(
+    r"^[0-9A-Za-z\s.',()&£$%@!/\+-]*$",
+    ValidationError("Only symbols .,/()&£$@!+-% are allowed."),
```
a-gleeson, line 86: Checking this regex it complained about the forward slash / not being escaped by a backslash \. Although I do see the test for the symbol validator covers testing the forward slash so this should be okay?

common/tests/test_validators.py
eadpearce: We use it for some fields that have leading zeroes so have to be saved as strings not integers

PR #840, opened 2023-03-06
TP2000-764 Correct duty sentence parser to allow all available measurement units

dalecannon:
# TP2000-764 Correct duty sentence parser to allow all available measurement units

## Why
The duty sentence parser doesn't allow users to enter certain valid measurement units (for example "m2").

## What
- Adds EOF flag parser to measurement units without qualifiers


paulpepper-trade (approved):
LGTM!

PR #849, opened 2023-03-08
TP2000-795 Remove initial results from measures search page

eadpearce:
# TP2000-795 Remove initial results from measures search page

## Why

* On first load the measures search page was being loaded with no search params entered meaning it would fetch ALL measures from the database making the page extremely slow to load

## What

* Add a new page at /measures/search that just shows the form with the search filters. Users must submit the form with search terms before any measures are returned from the database

FireShot Capture 053 - Find and edit measures - Manage Trade Tariffs - localhost


measures/views.py
```diff
@@ -780,7 +800,7 @@ def post(self, request):
             )
         self.session_store.clear()
 
-        return redirect(reverse("measure-ui-list"))
+        return redirect(reverse("workbaskets:current-workbasket"))
```
a-gleeson, line 803: is this page or would the review measures list be more suitable? at workbaskets/current/review/
Not sure just wondering

a-gleeson (approved):

PR #854, opened 2023-03-13
TP2000-767  Prevent duplicate footnote selection

dalecannon:
# TP2000-767  Prevent duplicate footnote selection

## Why
Users should not be able to add the same footnote more than once in the footnotes form step of the create measure journey.

## What
- Displays an error if the same footnote is added multiple times
- Adds missing unit tests for the form

Screenshot 2023-03-13 at 15 45 50


measures/forms.py
```diff
@@ -1142,6 +1142,14 @@ def __init__(self, *args, **kwargs):
 class MeasureFootnotesFormSet(FormSet):
     form = MeasureFootnotesForm
 
+    def clean(self):
+        cleaned_data = super().cleaned_data
```
a-gleeson, line 1146: How about a doc string here to explain the function?

PR #856, opened 2023-03-14
formatting updates

dougmills-DIT:
# TP2000-761 - Crude interim fix for importer when handling missing goods nomenclature description periods on import

This fix is crude, and will be superseded by the correct separation of goods nomenclature description periods into a new table. In the interim this fix resolves a long standing issue with imports disregarding goods nomenclature description updates where the period is not also created in the TGB file. 

## Why
 * The current implementation if the importer would discard silently any record that was missing dependencies, in this case we have observed several instances where the goods nomenclature description was not applied to a workbasket from an import because the period was defined in another (previous) TGB file - that combined with the way descriptions (and periods) were implemented, it was not possible to reconcile the period and was discarded. 
 * expanding the reason the issue occured : TAP does not handle goods nomenclature description periods separately, but simply tacks on a date to the end of the description. The result is that no goods nomenclature description period SID data is recorded in TAP, and when a description update in a TGB file is encountered without an accompanying period, we cant lookup the period in the database, hence there is a missing dependency. 
 
## What
What is this PR doing, e.g. implementations, algorithms, etc.?
 *  Solution: I have added code to collect issues with imports (which will be used in the future for reporting, and possibly committed to the database) during the import process, there is a check to handle any import issues found, and a check to see if there is a method on the handler that corresponds to the issue to correct it. I have only added one such method at the moment, but the framework is there to add additional methods to perform similar operations. the method is called "create_missing_goods_nomenclature_description_period"and is found in the GoodsNomenclatureDescriptionHandler class, it's simple and just adds in a start date and removes the dependency from the description during import to allow the process to complete without dropping records. 
 * There are two edge cases that could be an issue : 
 * - When the end date of the goods nomenclature is before today, however if that's the case it does not matter since it will be historical
 * - there already exists a description with this date, which will cause a rule violation, also unlikely but possible. Importing the same TGB file the day after would resolve this issue. 
 * Unit tests cover this scenario (but we could do with a lot more around the importer)
 * Tested previously failed files, and they now work (injecting today's date as the descriptions validity_start date) 

## Checklist
- Requires migrations? No

Links to relevant material
See: [Ticket](https://uktrade.atlassian.net/browse/TP2000-761)


importer/handlers.py
```diff
@@ -444,6 +495,46 @@ def build(self) -> Set[str]:
         self.dependency_keys.add(self.key)
         return self.dependency_keys
 
+    def get_import_issues(self):
+        if not self.resolve_dependencies():
+            if not self.resolve_dependencies():
```
a-gleeson, line 500: double check?

importer/handlers.py
```diff
@@ -268,13 +290,22 @@ def _generate_dependency_keys(self) -> Set[str]:
...
+                tag=dependency.tag,
+                identifying_fields=self.identifying_fields,
+                data=self.data,
             )
+
+            # using this to allow the correct mapping to be used later when reporting back to user
+            self.dependency_key_mapping[key] = {
+                "tag": dependency.tag,
+                "identifying_fields": self.identifying_fields,
+                "data": self.data,
```
paulpepper-trade, line 304: Perhaps worth creating type (data class or named tuple) for this?

importer/handlers.py
```diff
@@ -304,6 +335,26 @@ def resolve_dependencies(self) -> bool:
...
         return True
 
+    def _get_missing_dependencies(self) -> list:
+        """
+        Search the cache for all object dependencies and collect the keys and
+        return them.
+
+        Previously found objects, which are dependent on the current object,
+        should be stored in the cache. This method loops over the current
+        objects dependencies and attempts to where there is no match it will
```
paulpepper-trade, line 345: "and attempts to..." missing words?

importer/handlers.py
```diff
@@ -304,6 +335,26 @@ def resolve_dependencies(self) -> bool:
...
+        return the matching key
+        """
+        dependencies = self.dependency_keys.copy()
```
paulpepper-trade, line 348: Maybe worth adding a class-level type annotation for the `dependency_keys` attribute.

importer/handlers.py
```diff
@@ -444,6 +495,46 @@ def build(self) -> Set[str]:
...
+                            missing_dependency_data["tag"],
+                            missing_dependency_data["identifying_fields"],
+                            key,
+                            dep_missing_details,
+                        ),
+                    )
+
+                    # raise exception
+
+                    # raise Exception(f'There are missing dependencies for {self.tag} : {dep_missing_details}')
```
paulpepper-trade, line 531: Redundant comments?

importer/taric.py
```diff
@@ -221,6 +221,7 @@ class EnvelopeError(ParserError):
 class EnvelopeParser(ElementParser):
     tag = Tag("envelope", prefix=ENVELOPE)
     transaction = TransactionParser(many=True)
+    issues = list()
```
paulpepper-trade, line 224: It looks as though nothing in this class accesses `issues`. Is it used? If it's used from outside the class then a docstring would be useful here to aid understanding those dependencies.

importer/tests/taric_xml_stream/test_goods_nomenclature_description_period_scenarios.py
```diff
@@ -0,0 +1,199 @@
...
+    settings,
+):
+    settings.CACHES = {
+        "default": {
+            "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        },
+    }
+
+    assert (
+        GoodsNomenclatureDescription.objects.all()
```
paulpepper-trade, line 67: You can ditch the `.all()`.

importer/tests/taric_xml_stream/test_goods_nomenclature_description_period_scenarios.py
```diff
@@ -0,0 +1,199 @@
...
+    taric.process_taric_xml_stream(
+        BytesIO(goods_description_with_period_create_xml_as_text.encode()),
+        WorkBasketFactory.create().id,
+        WorkflowStatus.EDITING,
+        get_partition_scheme(settings.TRANSACTION_SCHEMA),
+        valid_user.username,
+    )
+
+    assert (
+        GoodsNomenclatureDescription.objects.all()
```
paulpepper-trade, line 85: Ditch `.all()`.

importer/tests/taric_xml_stream/test_goods_nomenclature_description_period_scenarios.py
```diff
@@ -0,0 +1,199 @@
...
+        .filter(
+            described_goods_nomenclature__item_id="2903691100",
+            described_goods_nomenclature__suffix=80,
+        )
+        .count()
+        == 2
+    )
+
+    assert GoodsNomenclatureDescription.objects.all().filter(
```
paulpepper-trade, line 94: Ditch `.all()`.

paulpepper-trade (approved):

importer/handlers.py
dougmills-DIT: ooh, didn't spot that - good catch! will update

dougmills-DIT: updated

importer/handlers.py
```diff
@@ -226,6 +246,8 @@ def get_link_to_model_b_link(self, model, kwargs):
     links: Iterable[LinksType] = None
     serializer_class: Type[ModelSerializer] = None
     tag: str = None
+    dependency_key_mapping = {}
```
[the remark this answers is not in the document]
dougmills-DIT: added type (correctly called out code smell : data clump, fixed now

importer/handlers.py
dougmills-DIT: code smell data clump : updated now

importer/handlers.py
dougmills-DIT: updated desc, and added return value doc string.

importer/handlers.py
dougmills-DIT: Added done - can you check this is what you meant

importer/handlers.py
dougmills-DIT: please check this is what you were expecting

importer/handlers.py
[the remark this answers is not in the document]
dougmills-DIT: please check this is as you expect

paulpepper-trade (approved):

PR #857, opened 2023-03-14
TP2000-796  Progressively enhance "Add new" button on commodities/duties form

dalecannon:
# TP2000-796  Progressively enhance "Add new" button on commodities/duties form

## Why
When clicking the “Add new” button on the commodities / duties formset step of create measures journey, the formset is POSTed to the server each time a new form is added. This causes a page refresh and returns the browser viewport page to the top of the page.

## What
- Overrides default behaviour for the "Add new" button so that new forms are dynamically added client-side and scroll location is kept (if javascript is enabled).


common/static/common/js/addNewForm.js
```diff
@@ -0,0 +1,47 @@
...
+    let fieldset  = document.querySelector("fieldset");
+    let formset = fieldset.parentNode;
+    let newForm = fieldset.cloneNode(true);
+    
+    newForm.innerHTML = newForm.innerHTML.replaceAll("formset-0", "formset-" + numForms);
+    newForm.querySelector(".autocomplete").removeAttribute("data-original-value");
+
+    let formFields = newForm.querySelectorAll("input");
+    for (let field of formFields.values()) {
+      field.value = null;
```
paulpepper-trade, line 16: Perhaps a very small point... I'd generally choose to set an input value to the empty string in order to clear it. I don't know if it's more correct (whatever that might mean), but almost all examples seem to take the same approach.

paulpepper-trade (approved):
LGTM.

PR #858, opened 2023-03-14
TP2000-803  Apply "Add new" button enhancement to other formsets

dalecannon:
# TP2000-803  Apply "Add new" button enhancement to other formsets

## Why
When clicking the “Add new” button on conditions formset and footnotes formset steps of the create measures journey, the formset is POSTed to the server each time a new form is added. This causes a page refresh and returns the browser viewport page to the top of the page.

## What
- Applies the "Add new" button progressive enhancement to these formsets so that new forms are added dynamically client-side and keep scroll location (if Javascript is enabled)


a-gleeson (approved):
LGTM!

PR #859, opened 2023-03-14
Override SESSION_EXPIRE_AT_BROWSER_CLOSE and SESSION_COOKIE_AGE in dev.

paulpepper-trade:
# Override session and cookie settings for dev

## Why

In a dev environment, our new, more secure settings force logouts and session clearing.

## What

Override configuration settings in `settings/dev.py` to maintain session and cookie values for longer periods of time within a development environment. `settings/dev.py` is never used in production or staging environments, so relaxing these settings doesn't introduce any security risk.

## Checklist
- Requires migrations? No


a-gleeson (approved):
LGTM!

PR #860, opened 2023-03-14
TP2000-804 Add workbasket column to version control tabs

dalecannon:
# TP2000-804 Add workbasket column to version control tabs

## Why
The version control tab of an object shows the activity date of an update and changes to validity periods but doesn't provide the user with the workbasket number in which the update was made. 

## What
- Adds a new column titled "Workbasket" that details the workbasket number in which the update occurred and links to it.

## What
Screenshot 2023-03-14 at 13 29 26


common/jinja2/includes/common/tabs/version_control.jinja
```diff
@@ -3,9 +3,16 @@
 <h2 class="govuk-heading-l">Version control</h2>
 {% set table_rows = [] %}
 {% for object in object.version_group.versions.all() %}
+  {% set workbasket_link %}
```
a-gleeson, line 6: This section is repeated in the 4 version_control tab's could it be extracted into a function imported?

a-gleeson: have you commited the macro file?

a-gleeson: ah amazing good catch!

a-gleeson (approved):

PR #866, opened 2023-03-21
TP2000-814 Preserve footnote associations on measures edit

dalecannon:
# TP2000-814 Preserve footnote associations on measures edit

## Why
When a measure is updated through the multiple measures edit journey, footnote associations present on the measure aren't being preserved on the new version of the measure because the association is based on the old measure's pk.

## What
- Updates existing footnote associations with the new version of the measure. 
- Adds unit test

## 

Screenshot 2023-03-21 at 14 19 56

### DIFF
--- a/measures/tests/test_views.py
+++ b/measures/tests/test_views.py
@@ -1805,6 +1805,83 @@ def test_multiple_measure_edit_only_duties(
         assert measure.duty_sentence == duties
 
 
+def test_multiple_measure_edit_preserves_footnote_associations(
+    valid_user_client,
+    session_workbasket,
+):
+    """Tests that footnote associations are preserved in MeasureEditWizard."""
+
+    measure = factories.MeasureFactory.create()
+    footnote_association = factories.FootnoteAssociationMeasureFactory.create(
+        footnoted_measure=measure,
+    )
+    exp_footnote = measure.footnotes.first()
+
+    url = reverse("measure-ui-edit-multiple")
+    session = valid_user_client.session
+    session.update(
+        {
+            "workbasket": {
+                "id": session_workbasket.pk,
+                "status": session_workbasket.status,
+                "title": session_workbasket.title,
+            },
+            "MULTIPLE_MEASURE_SELECTIONS": {
+                measure.pk: 1,
+            },
+        },
+    )
+    session.save()
+
+    STEP_KEY = "measure_edit_wizard-current_step"
+
+    wizard_data = [
+        {
+            "data": {
+                STEP_KEY: START,
+                "start-fields_to_edit": [MeasureEditSteps.START_DATE],
+            },
+            "next_step": MeasureEditSteps.START_DATE,
+        },
+        {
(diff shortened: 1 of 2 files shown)


measures/tests/test_views.py
```diff
@@ -1805,6 +1805,83 @@ def test_multiple_measure_edit_only_duties(
...
+        response = valid_user_client.post(url, step_data["data"])
+        assert response.status_code == 302
+
+        assert response.url == reverse(
+            "measure-ui-edit-multiple",
+            kwargs={"step": step_data["next_step"]},
+        )
+
+    workbasket_measures = Measure.objects.filter(
+        trackedmodel_ptr__transaction__workbasket_id=session_workbasket.id,
```
paulpepper-trade, line 1873: I think you can simplify this
```suggestion
        transaction__workbasket=session_workbasket,
```

measures/tests/test_views.py
```diff
@@ -1805,6 +1805,83 @@ def test_multiple_measure_edit_only_duties(
...
+    valid_user_client,
+    session_workbasket,
+):
+    """Tests that footnote associations are preserved in MeasureEditWizard."""
+
+    measure = factories.MeasureFactory.create()
+    footnote_association = factories.FootnoteAssociationMeasureFactory.create(
+        footnoted_measure=measure,
+    )
+    exp_footnote = measure.footnotes.first()
```
paulpepper-trade, line 1818: What's the intent here with using `.first()`? `.approved_up_to_transaction()` perhaps?

measures/tests/test_views.py
```diff
@@ -1805,6 +1805,83 @@ def test_multiple_measure_edit_only_duties(
...
+        trackedmodel_ptr__transaction__workbasket_id=session_workbasket.id,
+    ).order_by("sid")
+
+    complete_response = valid_user_client.get(response.url)
+    # on success, the page redirects to the workbasket page
+    assert complete_response.status_code == 302
+    assert valid_user_client.session["MULTIPLE_MEASURE_SELECTIONS"] == {}
+    for measure in workbasket_measures:
+        assert measure.update_type == UpdateType.UPDATE
+        assert measure.footnotes.first() == exp_footnote
```
paulpepper-trade, line 1882: Same question about `.first()` intent.

paulpepper-trade (changes requested):

measures/tests/test_views.py
dalecannon: Using `.first()` here returned the first object in the queryset returned by `measures.footnotes`, which allowed for asserting that the footnote association for the old measure and the new version is the same. 

The test has been updated to check for footnote count and to check each footnote in the queryset, not just the first.

paulpepper-trade (approved):

measures/tests/test_views.py
dalecannon: Yes, that's right. 

The variable names have now been amended to reflect this more clearly.

PR #869, opened 2023-03-22
TP2000-526 Remove measures limit in workbasket review view

dalecannon:
# TP2000-526 Remove measures limit in workbasket review view

## Why

The 'Review measures' view on the workbasket page is limited to 1000 measures. It has been proposed that this limit can now be removed.

## What
- Removes paginator class that set the limit from ReviewMeasuresWorkbasketView


paulpepper-trade:
Do we know yet how the view performs with the default paginator?

PR #872, opened 2023-03-24
TP2000-829 Add DJANGO_AUTO_FIELD spec to settings

paulpepper-trade:
# TP2000-829 Add DJANGO_AUTO_FIELD spec to settings

## Why

[Since Django 3.2](https://docs.djangoproject.com/en/3.2/ref/settings/#std-setting-DEFAULT_AUTO_FIELD), if a primary key is not explicity specified on a model and no value is given for `DJANGO_AUTO_FIELD` in `settings.py`, then the following type of error is output by Django when its checks are executed:

```
importer.ImportBatch: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
	HINT: Configure the DEFAULT_AUTO_FIELD setting or the AppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
```

## What

This PR adds the `DEFAULT_AUTO_FIELD` setting to TAP's settings file, assigning it the default value, `"django.db.models.AutoField"`, that would have been created if Django had generated the project.

## Checklist
- Requires migrations? No

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -58,6 +58,11 @@ STATICFILES_FINDERS = [
 
 STATICFILES_STORAGE = "whitenoise.storage.CompressedStaticFilesStorage"
 
+
+# Auto field type specification required since Django 3.2.
+DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
+
+
 # -- Application
 
 DJANGO_CORE_APPS = [


a-gleeson (approved):
LGTM!

PR #874, opened 2023-03-28
TP2000-741 ME32 & ME16 rules checker failing

paulpepper-trade:
# TP2000-741 ME32 & ME16 rules checker failing

## Why

ME32 rules checks can fail when a commodity has an indent history. The ME32 rule check relies upon `CommodityCollectionLoader.load()` to load commodities and their indents, but it does not capture the indent history of each commodity and retains only the most recent associated indent. This causes the wrong indent to be associated with a commodity when the ME32 rule check iterates through measure snapshots - which in turn rely upon commodity snapshots to determine the commodity hierarchy, which relies upon correct commodity indents at each specific snapshot moment.

## What

This PR includes the changes:

- Refactor a `MeasureSnapshot.get_snapshot()` for improved readability.
- Preserve each commodity's indent history when loading the commodity collection.
- Apply the correct indent from each commodity's indent history when constructing the `ancestors` and `descendants` of a `CommodityTreeSnapshot` instance.
- Improved ME32 command line output.
- Improved docstring documentation.

## Checklist
- Requires migrations? No


commodities/models/dc.py
```diff
@@ -572,7 +606,12 @@ def _traverse(self):
         indents = [[]]
 
         for commodity in self.commodities:
-            indent = commodity.indent
+            if self.moment.date:
+                indent_obj = commodity.get_indent_at(self.moment.date)
+                indent = (indent_obj and indent_obj.indent) or commodity.indent
```
a-gleeson, line 611: Not sure if I am miss understanding indent but is this variable supposed to be set to a boolean or is it supposed to deefault to the value of commodity.indent? 
I guess if indent is a boolean this makes sense.

paulpepper-trade: Similar to, say, Bash, Python's boolean operators, `and` and `or`, evaluate the truthiness of their operands and return operand values as:

> The expression x and y first evaluates x; if x is false, its value is returned; otherwise, y is evaluated and the resulting value is returned.
> 
> The expression x or y first evaluates x; if x is true, its value is returned; otherwise, y is evaluated and the resulting value is returned.

https://docs.python.org/3/reference/expressions.html#boolean-operations

a-gleeson (approved):

a-gleeson (approved):

PR #876, opened 2023-03-31
TP2000-802  Add memberships to geo area editing

dalecannon:
# TP2000-802  Add membership creation to geo area editing

## Why
Users are unable to add countries or regions to geographical area groups to create new geographical memberships.

## What
- Extends `GeographicalAreaEditForm` to allow adding a single country or region as a member of a single area group
- Adds tests

##
When editing an area group:
Screenshot 2023-03-31 at 13 35 35

When editing a country or region:
Screenshot 2023-03-31 at 13 42 25

### DIFF
--- a/geo_areas/forms.py
+++ b/geo_areas/forms.py
@@ -152,7 +152,7 @@ class GeographicalMembershipValidityPeriodForm(forms.ModelForm):
         fields = ["valid_between"]
 
 
-class GeographicalMembershipEditForm(
+class GeographicalMembershipAddForm(
     BindNestedFormMixin,
     GeographicalMembershipValidityPeriodForm,
 ):
@@ -278,7 +278,7 @@ class GeographicalAreaEndDateForm(ValidityPeriodForm):
 
 class GeographicalAreaEditForm(
     GeographicalAreaEndDateForm,
-    GeographicalMembershipEditForm,
+    GeographicalMembershipAddForm,
 ):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


geo_areas/forms.py
```diff
@@ -34,25 +47,221 @@ class Meta:
...
-class GeographicalAreaEndDateForm(ValidityPeriodForm):
+class GeoAreaType(TextChoices):
+    COUNTRY = "COUNTRY", "Add a country"
+    REGION = "REGION", "Add a region"
+
+
+class GeoAreaCountryForm(forms.Form):
+    country = forms.ModelChoiceField(
+        label="Country",
+        queryset=GeographicalArea.objects.filter(area_code=AreaCode.COUNTRY),
```
eadpearce, line 58: If the queryset is being set in __init__ it might be worth leaving this as `None` and putting a comment just to avoid confusion

geo_areas/forms.py
```diff
@@ -34,25 +47,221 @@ class Meta:
...
+            .with_latest_description()
+            .as_at_today()
+            .order_by("description")
+        )
+        self.fields[
+            "region"
+        ].label_from_instance = lambda obj: f"{obj.area_id} - {obj.description}"
+
+
+class GeographicalMembershipValidityPeriodForm(forms.ModelForm):
```
eadpearce, line 99: Can the `ValidityPeriodForm` be used as a mixin here? https://github.com/uktrade/tamato/blob/master/common/forms.py#L343 It has all the validation in it already. It's just the content that might need overriding

eadpearce: Oh wait a min. I think I understand. `ValidityPeriodForm` is already being used for the geo area being edited and not the membership specifically

eadpearce (approved):

PR #881, opened 2023-04-14
TP2000-826  Prevent only description deletion

dalecannon:
# TP2000-826  Prevent only description deletion

## Why
When additional codes, certificates, footnotes and geo areas have only one description, TAP allows users to delete it, which results in a business rule violation. 

## What
- Hides delete link from description tab if only one description is available
- Adds `DescriptionDeleteMixin`, used by delete forms to prevent the only description from being deleted
- Styles the cancel link as a button
- Adds tests

## 
Screenshot 2023-04-14 at 09 58 47

### DIFF
--- a/common/jinja2/includes/common/tabs/descriptions.jinja
+++ b/common/jinja2/includes/common/tabs/descriptions.jinja
@@ -5,10 +5,10 @@
   <div class="govuk-grid-column-three-quarters">
     {% set description_rows = [] %}
     {% set descriptions = object.get_descriptions() %}
-    {% set is_last = descriptions|length == 1 %}
+    {% set is_only_description = descriptions|length == 1 %}
     {% for description in descriptions %}
       {% set edit_url = description.get_url("edit") %}
-      {% set delete_url = "" if is_last else description.get_url("delete") %}
+      {% set delete_url = "" if is_only_description else description.get_url("delete") %}
       {% set description_row = [
         {"text": "{:%d %b %Y}".format(description.validity_start)},
         {"text": description.description},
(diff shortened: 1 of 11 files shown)


common/jinja2/includes/common/tabs/descriptions.jinja
```diff
@@ -4,9 +4,11 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-three-quarters">
     {% set description_rows = [] %}
-    {% for description in object.get_descriptions() %}
+    {% set descriptions = object.get_descriptions() %}
+    {% set is_only_description = descriptions|length == 1 %}
+    {% for description in descriptions %}
       {% set edit_url = description.get_url("edit") %}
-      {% set delete_url = description.get_url("delete") %}
+      {% set delete_url = "" if is_only_description else description.get_url("delete") %}
```
paulpepper-trade, line 11: Should UI components be removed in this way or should they be disabled? It's a question for content or design, which you may have already checked?

common/views.py
```diff
@@ -335,6 +335,21 @@ def form_valid(self, form):
...
+    deleted."""
+
+    def form_valid(self, form):
+        described_object = self.object.get_described_object()
+        if described_object.get_descriptions().count() == 1:
+            form.add_error(
+                None,
+                "This description cannot be deleted because at least one description record is mandatory.",
+            )
+            return self.form_invalid(form)
```
paulpepper-trade, line 349: Is there a reason why this can't be placed in the form's `clean()` method?

paulpepper-trade (changes requested):

common/views.py
dalecannon: All the description delete forms are created by passing in a description model to `delete_form_for()` function, which is also used by other models to create delete forms.

common/jinja2/includes/common/tabs/descriptions.jinja
dalecannon: Yeah, I've just double checked that the delete link may be removed/hidden for the only available description.

paulpepper-trade (approved):

PR #882, opened 2023-04-17
TP2000-830  Expand validation for editing geo areas

dalecannon:
# TP2000-830  Expand validation for editing geo areas

## Why
Form validation for editing the end date of a geographical area needs to cover where the validity period of the geographical area would no longer span the validity period of a linked quota order number origin.

## What
- Adds validation to `GeographicalAreaEndDateForm`
- Adds test

~~Adds the ability to `BusinessRulesMixin` to run indirect business rules for `TrackedModelChangeViews`~~
~~Adds indirect business rule ON6 to geo area update view~~

##
Screenshot 2023-04-17 at 12 45 38

### DIFF
--- a/common/views.py
+++ b/common/views.py
@@ -262,7 +262,6 @@ class BusinessRulesMixin:
     """Check business rules on form_submission."""
 
     validate_business_rules: Tuple[Type[BusinessRule], ...] = tuple()
-    validate_indirect_business_rules: Tuple[Type[BusinessRule], ...] = tuple()
 
     def form_violates(self, form) -> bool:
         """
--- a/common/views.py
+++ b/common/views.py
@@ -33,6 +33,7 @@ from django.views.generic.edit import FormMixin
 from django_filters.views import FilterView
 from redis.exceptions import TimeoutError as RedisTimeoutError
 
+from checks.checks import IndirectBusinessRuleChecker
 from common import forms
 from common.business_rules import BusinessRule
 from common.business_rules import BusinessRuleViolation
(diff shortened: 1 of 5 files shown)


common/views.py
```diff
@@ -280,6 +282,14 @@ def form_violates(self, form) -> bool:
                 form.add_error(None, v.args[0])
                 violations = True
 
+        for rule in self.validate_indirect_business_rules:
+            checkers = IndirectBusinessRuleChecker.of(rule).checkers_for(self.object)
+            for checker in checkers:
+                (rule_passed, message) = checker.run(self.object)
```
paulpepper-trade, line 288: What performance impact does this have across forms for different models? Do we know that it's within acceptable bounds, whatever they may be?

paulpepper-trade (changes requested):

common/views.py
dalecannon: While this would only run indirect business rules specified in the view, how this would perform for other models' indirect rules is unknown and would need to be tested.

Until then, would it be better to opt for adding the required validation to the `clean()` method of `GeographicalAreaEndDateForm`, seeing as only this one extra check is needed for now?

SteveCN7: I think then, the work to move business rules on to a separate celery worker seems best to be honest

paulpepper-trade (approved):

common/views.py
paulpepper-trade: That wouldn't have made any difference to performance in this case, @SteveCN7 . Here the rules are being executed within the context of a web worker (when a user submits the form).

PR #884, opened 2023-04-17
TP2000 815 Comm code Import Page Redesign

LaurenQurashi:
# TP200-815

## Why
We've been squirrelling away at making improvements to the Importer. This PR provides new designs and a couple bells and whistles as part of the efforts to implement our V1 iteration. 

### Redesign of the current commodities importer: 
In production, we have this little sneaky page for uploading comm code files. Following the designs, I've added a button to the index page to allow easy access to the journey, I've also added a list page, made small design amendments to the existing upload page, and created a success page. 

### Addition of Authors and Status to the ImportBatch model: 
The designs required information about who uploaded a file, as well as the status of it. I've created an author field to record the user who uploaded it, as well as a status field with some transitions for keeping track of where a file is at. 

## To do: 
- Clean up the rough edges. It's a bit weird as there's a 'general' importer, and this taric one. There's not a lot of difference except one makes a workbasket for you and such but they're processed the same way. I didn't want to step on the toes of the general one, or change anything so I've been making new urls, new views, new filters etc. They're all a bit wonky in terms of naming and such but once there's some clarity on what can go and what needs to stay, we'll be able to make naming and file structure decisions. 

- Create an "uploading" panel. I've only been working on the happy path but my next task is to show when a file is taking ages to upload. This PR was getting big though so this feels like a nice place to stop and restart. 

- Create a Review Changes page. Again, seems good to put this in another PR.

## Checklist
- Requires migrations? - Yes
- Requires dependency updates? - No although there's a precommit update in here too 

--->


commodities/forms.py
```diff
@@ -31,10 +31,11 @@ def __init__(self, *args, **kwargs):
```
LaurenQurashi, line 31: This is just some tweaks to the original form to include some copy changes and so save an author and status to the import batch

commodities/views.py
```diff
@@ -17,6 +17,7 @@
 from common.views import TrackedModelDetailView
```
LaurenQurashi, line 17: This might be a dumb idea but I couldn't think of how else to do it.. In order to pass over the saved file name and workbasket id to my success page, I've saved it in session storage. The only other thing that did this was measures but that was mega confusing, so is this the right way to go about it?

common/tests/test_views.py
```diff
@@ -23,6 +23,7 @@ def test_index_displays_workbasket_action_form(valid_user_client):
     assert "Package Workbaskets" in page.select("label")[2].text
```
LaurenQurashi, line 23: Testing the home page shows my new button

common/views.py
```diff
@@ -64,6 +64,8 @@ def form_valid(self, form):
             return redirect(reverse("publishing:envelope-queue-ui-list"))
```
LaurenQurashi, line 64: This adds the actual functionality to the button

importer/filters.py
```diff
@@ -30,3 +30,19 @@ def filter_status(self, queryset, name, value):
     class Meta:
```
LaurenQurashi, line 30: I created a new filter so that I didn't mess with the old one

importer/forms.py
```diff
@@ -102,7 +102,9 @@ def __init__(self, *args, **kwargs):
```
LaurenQurashi, line 102: Initially I added this to the wrong form cause there's two (who would have thought) but I thought it might be sensible to keep since the models now require an author and a status so thought I'd keep it there but lemme know if this should go

importer/models.py
```diff
@@ -58,6 +87,42 @@ class ImportBatch(TimestampedMixin):
 
     objects = models.Manager.from_queryset(ImporterQuerySet)()
 
+    @transition(
```
LaurenQurashi, line 90: These transitions are cool, they're part of the Functional State Machine field thingy so it dictates what status comes before and after each one.. more stable than just text

importer/urls.py
```diff
@@ -13,4 +13,10 @@
         views.UploadTaricFileView.as_view(),
         name="import_batch-ui-create",
     ),
+    # new comm code importer url
```
LaurenQurashi, line 16: I didn't want to mess with the urls especially with the whole "There's two importers sort of " thing so for now I've made a new url and new other stuff but it does create a bit of a mess.. hence why I've also had to make comments on which one is the right one haha

measures/migrations/0014_action_pair_data_migration.py
```diff
@@ -26,6 +26,17 @@ def generate_action_pairs(apps, schema_editor):
 
     MeasureAction = apps.get_model("measures", "MeasureAction")
```
LaurenQurashi, line 27: There's a couple files in here which I have no clue about and I think it's Paul's work from a weird rebase of mine so I'll need to sort that out

PR #885, opened 2023-04-19
TP2000-844  Add suffix and indent columns to find commodities view

dalecannon:
# TP2000-844  Add suffix and indent columns to find commodities view

## Why
Both suffix and indent data is missing from the find commodities view. Displaying this data would help users more easily find a relevant commodity code. 

## What
- Adds suffix and indent columns to find commodities view
- Update test

##
Screenshot 2023-04-19 at 12 37 58

### DIFF
--- a/commodities/jinja2/includes/commodities/list.jinja
+++ b/commodities/jinja2/includes/commodities/list.jinja
@@ -8,7 +8,7 @@
     {{ table_rows.append([
         {"html": commodity_link},
         {"text": commodity.suffix},
-        {"text": commodity.indents.all().last().indent},
+        {"text": commodity.indents.filter(validity_start__lte=today).last().indent},
         {"text": commodity.get_description().description},
     ]) or "" }}
 {% endfor %}
--- a/commodities/jinja2/includes/commodities/list.jinja
+++ b/commodities/jinja2/includes/commodities/list.jinja
@@ -7,12 +7,16 @@
     {%- endset %}
     {{ table_rows.append([
         {"html": commodity_link},
+        {"text": commodity.suffix},
+        {"text": commodity.indents.all().last().indent},
         {"text": commodity.get_description().description},
     ]) or "" }}
 {% endfor %}
 {{ govukTable({
     "head": [
         {"text": "Commodity code"},
+        {"text": "Suffix"},
+        {"text": "Indent"},
         {"text": "Commodity description"},
 
     ],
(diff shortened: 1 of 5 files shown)


commodities/jinja2/includes/commodities/list.jinja
```diff
@@ -7,12 +7,16 @@
     {%- endset %}
     {{ table_rows.append([
         {"html": commodity_link},
+        {"text": commodity.suffix},
+        {"text": commodity.indents.all().last().indent},
```
paulpepper-trade, line 11: This view shows commodities as of today, so a new indent with a future start date (which would be `last()`) shouldn't appear here. I think the correct indent would have a `validity_start` such that `today >= indent.validity_start`. Something like this https://github.com/uktrade/tamato/blob/1b149520368c5bf83b05fcf86fbdb9b8d4363a67/commodities/models/dc.py#L184
I think that's the correct approach, but perhaps worth validating with Stephen.

SteveCN7: I think thats right - I wonder if we need to provide a "time dimension" to the search at some point, ie - show me the future data or the past data, not just the now? Something to consider

dalecannon: Yeah, that make sense. `indents.all().last().indent` was taken from the commodity detail view, so it'll need changing there too, right?

paulpepper-trade: Yes, it'll need a change there too.

commodities/views.py
```diff
@@ -78,17 +78,26 @@ class CommodityList(CommodityMixin, WithPaginationListView):
...
+
 
 class CommodityDetail(CommodityMixin, TrackedModelDetailView):
     template_name = "commodities/detail.jinja"
 
     def get_context_data(self, *args, **kwargs):
         context = super().get_context_data(*args, **kwargs)
 
+        context["indent_number"] = (
+            self.object.indents.filter(validity_start__lte=date.today()).last().indent
```
paulpepper-trade, line 94: You have a few duplicate expressions like this. Maybe worth factoring into a method with a docstring to explain the nature of indent selection?
```
class GoodsNomenclature(TrackedModel, ValidityMixin, DescribedMixin):

    ...

    def indent_as_of(self, date: date) -> GoodsNomenclatureIndent:
        indents = self.indents.filter(validity_start__lte=date)
        if not indents:
            return None
        return indents.last()
```

paulpepper-trade (approved):

PR #886, opened 2023-04-21
TP2000-849  Add active state filter to quota search

dalecannon:
# TP2000-849  Add active state filter to quota search

## Why
The 'search and filter' form on the 'find and edit quotas' view is missing options to filter by "active" and "terminated" quotas.

## What
- Adds `ActiveStateMixin` to `QuotaFilter`
- Adds test

##
Screenshot 2023-04-21 at 09 16 24

### DIFF
--- a/quotas/tests/test_views.py
+++ b/quotas/tests/test_views.py
@@ -140,6 +140,37 @@ def test_quota_list_view(view, url_pattern, valid_user_client):
     assert_model_view_renders(view, url_pattern, valid_user_client)
 
 
+@pytest.mark.parametrize(
+    ("search_filter", "search_term"),
+    [
+        ("active_state", "active"),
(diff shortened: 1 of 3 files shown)


quotas/tests/test_views.py
```diff
@@ -140,6 +140,37 @@ def test_quota_list_view(view, url_pattern, valid_user_client):
...
 
 
+@pytest.mark.parametrize(
+    ("search_filter", "search_term"),
+    [
+        ("active_state", "active"),
+        ("active_state", "terminated"),
+    ],
+)
+def test_quota_list_view_active_state_filter(
```
a-gleeson, line 150: Should we also include tests for an invalid search_term and/or search_filter?

a-gleeson (approved):

PR #887, opened 2023-04-21
TP2000-738 Display rule check progress

dalecannon:
# TP2000-738 Display rule check progress

## Why
Users are not able to see the progress of a rule check.

## What
- Adds `rule_check_progress` function to `WorkBasket` model to provide progress of a rule check for the workbasket
- Displays rule check progress on workbasket summary and application information pages.

##
Workbasket summary:
workbasket-summary

Application information:
app-info

### DIFF
--- a/workbaskets/models.py
+++ b/workbaskets/models.py
@@ -4,6 +4,7 @@ import logging
 from abc import ABCMeta
 from abc import abstractmethod
 from typing import Optional
+from typing import Tuple
 
 from celery.result import AsyncResult
 from django.conf import settings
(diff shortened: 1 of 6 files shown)


workbaskets/models.py
```diff
@@ -346,6 +347,27 @@ def rule_check_task_status(self):
             return None
         return task_result.status
 
+    def rule_check_progress(self) -> Tuple[int, int]:
```
a-gleeson, line 350: Should we have a unit test for this function where we create some complete checks & and some incomplete

a-gleeson (approved):

PR #889, opened 2023-04-25
TP2000-477 Prepopulate geo exclusions

eadpearce:
# TP2000-477 Prepopulate geo exclusions

## Why

* When a single measure needed to be edited its geo area exclusions were not prepopulated meaning the user would have to fill these in again even if editing a different field on the measure. This was time consuming and error prone

## What

* Updates the FormSet class and BindNestedForms mixin to properly add initial data to formsets
* Also updates the display of exclusions in the workbasket to show which area is excluded

Screenshot 2023-04-25 at 10 49 12
Screenshot 2023-04-25 at 10 49 28

### DIFF
--- a/common/forms.py
+++ b/common/forms.py
@@ -718,7 +718,7 @@ def formset_add_or_delete(prefixes, data):
     formset_data = {}
     for k, v in data.items():
         for prefix in prefixes:
-            if k in [f"{prefix}-ADD", f"{prefix}-DELETE"]:
+            if k.endswith("-ADD") or k.endswith("-DELETE"):
                 formset_data[k] = v
     if len(formset_data) > 0:
         return True
--- a/common/forms.py
+++ b/common/forms.py
@@ -46,7 +46,23 @@ class BindNestedFormMixin:
                     nested_forms = []
 
                     for form_class in form_list:
-                        bound_form = form_class(*args, **kwargs)
+                        if issubclass(form_class, FormSet):
+                            formset_kwargs = kwargs.copy()
+                            if "initial" in formset_kwargs:
+                                formset_kwargs.pop("initial")
+                            formset_initial = None
+                            if kwargs.get("initial"):
+                                formset_initial = kwargs.get("initial").get(
+                                    form_class.prefix,
+                                )
+
+                            bound_form = form_class(
+                                *args,
+                                **formset_kwargs,
+                                initial=formset_initial,
+                            )
+                        else:
+                            bound_form = form_class(*args, **kwargs)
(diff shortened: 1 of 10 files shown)


common/forms.py
```diff
@@ -552,8 +576,12 @@ def __init__(
                 # reinsert into data, with updated numbering
                 data[f"{self.prefix}-{i}-{field}"] = value
 
-        self.initial = list(formset_initial.values())
-        num_initial = len(self.initial)
+        self.initial = initial
```
dalecannon, line 579: `MeasureCommodityAndDutiesForm`, the problematic form in the failing test, doesn't seem to provide any values for `initial`. In this case, `self.initial` is always empty or `None`. This results in `INITIAL_FORMS = 0` and `TOTAL_FORMS = 0` based on `num_initial`, which causes a formset error to be raised because it doesn't meet the minimum number of forms.

eadpearce: Thanks so much! I think I've fixed it now!

common/forms.py
```diff
@@ -626,3 +661,65 @@ def formset_factory(
...
+    return output
+
+
+def formset_add_or_delete(prefixes, data):
+    """Find any {prefix}-ADD or {prefix}-DELETE in submitted data and return the
+    result as a boolean."""
+    formset_data = {}
+    for k, v in data.items():
+        for prefix in prefixes:
+            if k in [f"{prefix}-ADD", f"{prefix}-DELETE"]:
```
dalecannon, line 721: Forms marked for deletion in a formset are typically specified by their form number as in `{prefix}-{form_number}-DELETE`.  Perhaps it could check if `k.endswith(("-ADD", "-DELETE"))`.

measures/forms.py
```diff
@@ -61,6 +65,53 @@
...
 
 
+class GeoAreaInitialDataMixin:
+    @property
+    def formset_submitted(self):
+        return formset_add_or_delete(
+            [
+                GROUP_EXCLUSIONS_FORMSET_PREFIX,
+                GEO_GROUP_FORMSET_PREFIX,
+                COUNTRY_REGION_FORMSET_PREFIX,
```
dalecannon, line 75: Could`ERGA_OMNES_EXCLUSIONS_FORMSET_PREFIX`  also be added here?

common/forms.py
eadpearce: Good spot. I forgot that `-DELETE` keys need the index but `-ADD` keys don't

dalecannon (approved):

PR #890, opened 2023-04-26
TP2000-806 get envelope history

paulpepper-trade:
# TP2000-806 get envelope history

## Why

No provision exists in TAP to view and download a complete, versioned history of published envelopes. A partial view without envelope history is available via Django admin. However, that provision is incomplete (no historical access) and only a small number of admin users have the necessary access permissions.

## What

This PR adds two new views: 1) find processed envelopes, 2) envelope history (for an envelope ID). Unit tests for the views are also included.

### Find processed envelopes
image
A user with the necessary permissions (`view_envelope`) may view a filtered list of all envelopes that have been generated since automated packaging was introduced. A download link is provided with each published envelope within the list view.

### Envelope history
image
A user with the necessary permissions (`view_envelope`) may view the full history of an envelope, including those that were rejected for the viewed envelope ID. All historical envelope file instances can also be downloaded from this view (though note that only packaged evelopes created after envelope versioning was introduced will be downloadble).

## Checklist
- Requires migrations? No


publishing/models/envelope.py
```diff
@@ -124,6 +136,19 @@ def currently_processing(self):
             packagedworkbaskets__processing_state=ProcessingState.CURRENTLY_PROCESSING,
         )
 
+    def all_latest(self):
```
a-gleeson, line 139: I am not too clear on what this means by all latest? Is it all the "latest" because it doesn't include old envelopes that failed processing?

publishing/views.py
```diff
@@ -193,15 +198,11 @@ def _process_envelope(self, request, pk):
...
+    def download_response(self, envelope):
+        """Returns a Respond object with associated payload containing the
+        contents of `envelope.xml_file`."""
 
-    def get(self, request, *args, **kwargs):
-        packaged_workbasket = self.get_object()
-
-        envelope = packaged_workbasket.envelope
         file_content = envelope.xml_file.read()
         file_name = f"DIT{envelope.envelope_id}.xml"
```
a-gleeson, line 207: Could you use envelope.xml_file_name here?

publishing/models/envelope.py
paulpepper-trade: Good catch. It's redundant so I've deleted it! I must have had some purpose in mind for it.

publishing/models/envelope.py
```diff
@@ -152,6 +164,23 @@ def deconstruct(self):
         return name, path, args, kwargs
 
 
+def is_delete_marker(s3_object_version):
```
a-gleeson, line 167: Not sure but am I missing this usage?

publishing/models/envelope.py
```diff
@@ -229,6 +258,43 @@ def delete_envelope(self, **kwargs):
...
+        envelope_qs = (
+            Envelope.objects.filter(envelope_id=self.envelope_id)
+            .non_deleted()
+            .processed()
+            .order_by("-pk")
+        )
+        return envelope_qs
+
+    @property
+    def xml_file_exists(self) -> bool:
```
a-gleeson, line 274: Do we need to access s3 via the boto3 library here? Or could we access the file details by calling a function on `self.xml_file` ?
Checking the documentation : https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#model
An empty FileField has limited functionality and should return  `<FieldFile: None>` when we call `self.xml_file`  and it is deleted.

publishing/models/envelope.py
paulpepper-trade: It was a hangover from when this PR was attempting to use S3 versioning, which deals with deletions in a peculiar way, and which we don't have access to anyway. Usng the new versioning scheme it's become redundant and so I've now removed it.

publishing/models/envelope.py
paulpepper-trade: I implemented this with the intention of catching the case where the `FileField` is non-empty, but the accompanying S3 storage doesn't actually exist. That can happen when, say, a production snapshot is deployed to a staging service. This relatively low-level boto3 implementation also avoids pulling the whole file object from S3, making it much more efficient.

a-gleeson: I like the sound of that, makes sense to me :)

PR #892, opened 2023-04-27
TP2000-860 Sqlite dump and upload failing

paulpepper-trade:
# TP2000-860 Sqlite dump and upload failing

## Why

The scheduled sqlite dump of TAP data is failing due to a data migration failure. Migrations are run as part of the sqlite database creation process, before it has been populated with data. The migration failure causes the dump to fail.

## What

This PR guards against using non-existent data in the data migration `measures.migrations/0014_action_pair_data_migration.py`, preventing a failure within the Sqlite dump process.

## Checklist
- Requires migrations? No


a-gleeson (approved):
LGTM!

PR #894, opened 2023-04-28
TP2000-861 Fix footnote delete button on measure edit view

dalecannon:
# TP2000-861 Fix footnote delete button on measure edit view

## Why
On the measure edit view, the delete button for the footnote form redirects the user to /measures/<sid\>/**edi-footnotes**  causing a page not found (404) error.

## What
- Fix `self.path` so that "-footnotes/" is appended to an unsliced `path` (e.g /measures/<sid\>/edit).

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1406,7 +1406,7 @@ class MeasureUpdateFootnotesForm(MeasureFootnotesForm):
     def __init__(self, *args, **kwargs):
         path = kwargs.pop("path")
         if "edit" in path:
-            self.path = path[:-1] + "-footnotes/"
+            self.path = path + "-footnotes/"
 
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


measures/forms.py
```diff
@@ -1412,7 +1412,7 @@ class MeasureUpdateFootnotesForm(MeasureFootnotesForm):
     def __init__(self, *args, **kwargs):
         path = kwargs.pop("path")
         if "edit" in path:
-            self.path = path[:-1] + "-footnotes/"
+            self.path = path + "-footnotes/"
```
paulpepper-trade, line 1415: The original slice action, `path[:-1]`, would normally have been valid, since, by convention we, and Django generally, terminate paths with a `/` char. Is the issue here really that that terminating slash is missing, or is something else?

dalecannon: Yes, the path (`measures/963518/edit`) is missing a `/` char, which erroneously results in `measures/963518/edi-footnotes` here.

On further inspection, it appears that a terminating slash is missing from all URLs generated by `get_ui_paths` function.

PR #896, opened 2023-05-02
 Edit Measure Conditions Fix

a-gleeson:
# TP2000-856 Edit Measure Conditions Fix

## Why
When we automated the creation of the negative action code for use in the Create Measures journey, we removed the negative actions from the UI menu.
Unfortunately this has also removed them from the Edit Measures journey, meaning that a user can’t complete the journey as the required UI setting is missing.

## What

- Edit Measure condition form - field action - queryset updated to include all actions
- New validation for edit measure condition forms that checks that all fields are blank when a negative action code is selected.
- formset errors now appear in edit formset for conditions
- Formset validation now counts negative actions seperately so the form can successfully submit when editting negative actions too

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -450,8 +450,8 @@ class MeasureConditionsFormMixin(forms.ModelForm):
             self.add_error(
                 None,
                 ValidationError(
-                    "If you action code is negative you do not need to enter ‘reference price or quantity’, ‘certificate, licence or document’ or ‘duty’.",
-                ),  #
+                    "If the action code is negative you do not need to enter ‘reference price or quantity’, ‘certificate, licence or document’ or ‘duty’.",
+                ),
             )
 
         price_errored = False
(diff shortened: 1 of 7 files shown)


measures/forms.py
```diff
@@ -415,19 +420,40 @@ def conditions_clean(self, cleaned_data, measure_start_date):
...
                 ValidationError(
                     "For each condition you must complete either ‘reference price or quantity’ or ‘certificate, licence or document’.",
                 ),
             )
 
+        if is_negative_action_code and (price or certificate or applicable_duty):
+            self.add_error(
+                None,
+                ValidationError(
+                    "If you action code is negative you do not need to enter ‘reference price or quantity’, ‘certificate, licence or document’ or ‘duty’.",
```
eadpearce, line 453: "If you action code" should probably be "If the action code"

measures/forms.py
```diff
@@ -415,19 +420,40 @@ def conditions_clean(self, cleaned_data, measure_start_date):
...
+                ),  #
```
eadpearce, line 454: trailing comment mark?

a-gleeson: yup, removing

eadpearce (approved):

PR #897, opened 2023-05-02
TP2000-864 ME32 failing on multi-versioned indents

paulpepper-trade:
# TP2000-864 ME32 failing on multi-versioned indents

## Why

ME32 rule checks run against a measure whose attached commodity code has multiple indents within the same version group, can occasionally and randomly fail.

Currently when multiple versions of an indent exist against a commodity, the one selected within the `CommodityCollection` instance is non-deterministic. This is because no ordering is placed on the indents when querying the database.

## What

Apply an `order_by("transaction")` to the query that loads indents during `CommodityCollection` loading. This ensures the latest version of an indent is selected when many are available.

## Checklist
- Requires migrations? No


a-gleeson (approved):
LGTM!

PR #899, opened 2023-05-05
TP2000-859 Commodity hierarchy tab

eadpearce:
# TP2000-859 Commodity hierarchy tab

## Why

* It's helpful for users to be able to see where the commodity code they're looking at sits in the full tree

## What

* Uses CommodityTreeSnapshot to produce a visual display of the commodity tree for a given commodity code

Screenshot 2023-05-05 at 13 28 17

### DIFF
--- a/commodities/jinja2/commodities/hierarchy.jinja
+++ b/commodities/jinja2/commodities/hierarchy.jinja
@@ -2,39 +2,47 @@
 
 {% block content_class %}commodity-hierarchy{% endblock %}
 
+{% macro leaf_commodity(comm) -%}
+<p class="govuk-body {% if comm == this_commodity %}govuk-!-font-weight-bold{% endif %}"><a class="govuk-link" href="{{ url('commodity-ui-detail', args=[comm.sid]) }}">{{ comm.code }}</a> - {{ comm.description }} ({{ comm.suffix }})</p>
+{% endmacro %}
+
 {% block tab_content %}
 
 <h2 class="govuk-heading-l">Commodity hierarchy</h2>
 
+{% if snapshot.commodities|length == 1 %}
+  {{ leaf_commodity(snapshot.commodities[0]) }}
+{% else %}
+
 {% for comm in snapshot.commodities %}
 
-  {% if not loop.last %}
+  {% if loop.nextitem and loop.nextitem.indent > comm.indent %}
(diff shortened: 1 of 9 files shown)


commodities/views.py
```diff
@@ -148,3 +151,32 @@ def get_context_data(self, *args, **kwargs):
...
+        context = super().get_context_data(*args, **kwargs)
+
+        context["selected_tab"] = "hierarchy"
+
+        prefix = self.object.item_id[0:4]
+        commodities_collection = CommodityCollectionLoader(prefix=prefix).load()
+
+        for commodity in commodities_collection.commodities:
+            if commodity.item_id == self.object.item_id:
+                pass
```
dalecannon, line 169: I might be overlooking something, but this for loop doesn't look to be needed here.

commodities/jinja2/commodities/hierarchy.jinja
```diff
@@ -0,0 +1,40 @@
+{% extends "commodities/detail.jinja" %}
+
+{% block content_class %}commodity-hierarchy{% endblock %}
+
+{% block tab_content %}
+
+<h2 class="govuk-heading-l">Commodity hierarchy</h2>
+
```
dalecannon, line 8: Would there be value in adding content to say which HS heading the following hierarchy is a snapshot of?

eadpearce: Maybe? I'll ask Steve

commodities/views.py
eadpearce: Oh you're right. I just copied this from one of the tests but it's probably not needed

commodities/jinja2/commodities/hierarchy.jinja
SteveCN7: It might be worthwhile to add the chapter heading at the top as a H2 or H3 perhaps?
Ie 0100000000 is the Chapter Headings for Live animals
So maybe across the top just show that heading depending on the Chapter?

eadpearce: Gonna leave it as is for now and maybe add this in a follow-up ticket after getting it in front of users

SteveCN7: Works for me - once we've added the MODC tab, we can do a feedback session with users to look at enhancements we could make

dalecannon (approved):

PR #900, opened 2023-05-05
TP2000-870 Superuser rule check violation override

paulpepper-trade:
# TP2000-870 Superuser rule check violation override

## Why

TAP is generating false (incorrect) violations during rule checking around commodities with a specific indent arrangement.

At least one comm code (code: 0406901300, suffix: 80) has multiple indents with historic indent creation, all within the same transaction. When a measure is applied to the comm code, indent selection can be incorrect.

The root cause requires a fix. In the interim, a rule check override that is accessible only to users with superuser status needs to be made available.

## What

This PR implements a capability, only available to users with superuser status, to override failing rules checks.

## Checklist
- Requires migrations?
- Requires dependency updates?

### DIFF
--- a/measures/business_rules.py
+++ b/measures/business_rules.py
@@ -345,6 +345,17 @@ class ME32(BusinessRule):
     """
 
     def compile_query(self, measure):
+        """
+        Create a query that can be applied in a `MeasureQuerySet.filter()`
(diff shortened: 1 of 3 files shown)


a-gleeson (approved):
LGTM!

PR #906, opened 2023-05-12
TP2000 805  automate channel islands api

a-gleeson:
# TP2000-805 automate Crown Dependencies api

## Why

Currently the publishing to the Crown Dependencies is a manual process after HMRC accept the envelope in the packaging process. Publishing to the API could be automated after HMRC accept a packaged workbasket and the changes are published.

## What

- This PR implements an api publishing task in which publishes all packaged workbaskets in the SUCCESSFULLY_PROCESSED state.
- This will not publish envelopes that have been manually set and the date time set in the envelope.
- a CrownDependenciesEnvelope is created in the CURRENTLY_PUBLISHING state 
- the envelope is published to the TariffAPIClient.
- the api task will be recorded in CrownDependenciesPublishingTask when any envelopes require published
- whenever there is an error we save the detail to CrownDependenciesPublishingTask
- the task will be retried on CrownDependenciesException
- the task can be paused and it will not process the queue of packaged workbaskets
- CrownDependenciesPublishingOperationalStatus is the model that manages the pause
- Errors can automatically pause the queue for investigation
- Notifications are sent on publishing attempts successful or failed
- the api tasks will also handle any CrownDependenciesEnvelope's in the state CURRENTLY_PUBLISHING or FAILED_PUBLISHING.
- It is possible that the envelope publishing was successful so we will check if the CrownDependenciesEnvelope was already successful against the API.

image

## Checklist
- Requires migrations? yes

Links to relevant material
See: [Confluence](https://uktrade.atlassian.net/wiki/spaces/T/pages/3431825421/Automate+pushing+accepted+envelopes+to+channel+islands+API)


publishing/management/commands/publish_to_api.py
```diff
@@ -0,0 +1,59 @@
...
+        if incomplete:
+            self.stdout.write(
+                f"{incomplete.count()} envelope(s) not completed publishing:",
+            )
+            for i, crowndependencies in enumerate(incomplete, start=1):
+                self.stdout.write(
+                    f"{i}: {crowndependencies.packagedworkbaskets.last().envelope}",
+                )
+        self.stdout.write(
+            f"{unpublished.count()} envelope(s) ready to be published in the following order:",
```
eadpearce, line 46: if unpublished is None won't .count() error here?

publishing/models/crown_dependencies_envelope.py
```diff
@@ -0,0 +1,220 @@
...
+        default=None,
+    )
+
+    def __repr__(self) -> str:
+        return f'<CrownDependenciesEnvelope: id="{self.pk}", publishing_state={self.publishing_state}>'
+
+    def previous_envelope(self) -> CrownDependenciesEnvelopeQuerySet:
+        """Get the previous `CrownDependenciesEnvelope` by order of `pk`."""
+        try:
+            return CrownDependenciesEnvelope.objects.get(pk=self.pk - 1)
```
eadpearce, line 118: Would `CrownDependenciesEnvelope.objects.get(pk__lt=self.pk).last()` work instead? This would still work if a previous pk was skipped for whatever reason

publishing/models/crown_dependencies_envelope.py
```diff
@@ -0,0 +1,220 @@
...
+        except CrownDependenciesEnvelope.DoesNotExist:
+            return None
+
+    def can_publish(self) -> bool:
+        """Conditional check if the previous `CrownDependenciesEnvelope` has
+        been `SUCCESSFULLY_PUBLISHED`."""
+        previous_envelope = self.previous_envelope()
+        if (
```
eadpearce, line 126: You could do this instead. Bit neater
```
return (
        previous_envelope
        and previous_envelope.publishing_state
        == ApiPublishingState.SUCCESSFULLY_PUBLISHED
        or not previous_envelope
)
```

publishing/management/commands/publish_to_api.py
a-gleeson: good catch! It would

publishing/models/crown_dependencies_envelope.py
a-gleeson: nice idea, I think using the `lt` operator would be better in this case

publishing/models/crown_dependencies_envelope.py
a-gleeson: that's true, I could do that. Do you think I should maybe update the name of the function to "if_can_publish"?

eadpearce: if you're using the result in an if statement `if if_can_publish()` might read a bit weird

eadpearce (approved):

publishing/models/crown_dependencies_envelope.py
a-gleeson: that's true, it's unused now so I realised I could remove it

PR #910, opened 2023-05-17
TP2000-878 MODC actions

eadpearce:
# TP2000-878 MODC commodities actions

## Why

* Users need a way to further filter and edit the measures shown on the MODC tab 

## What

* Adds a link that passes the query params to the measures search page
* Adds an extra param to the measure list filters to return MODC

Screenshot 2023-05-17 at 12 50 29
Screenshot 2023-05-17 at 12 50 40

### DIFF
--- a/commodities/helpers.py
+++ b/commodities/helpers.py
@@ -1,0 +1,20 @@
+from commodities.models.dc import CommodityCollectionLoader
+from commodities.models.dc import CommodityTreeSnapshot
+from commodities.models.dc import SnapshotMoment
+from measures.snapshots import MeasureSnapshot
+
+
+def get_measures_on_declarable_commodities(transaction, item_id, date=None):
+    prefix = item_id[0:2]
+    commodities_collection = CommodityCollectionLoader(prefix=prefix).load()
+
+    moment = SnapshotMoment(transaction=transaction, date=date)
+    tree = CommodityTreeSnapshot(
+        commodities=commodities_collection.commodities,
+        moment=moment,
+    )
+    this_commodity = list(
+        filter(lambda c: c.item_id == item_id, tree.commodities),
+    )[0]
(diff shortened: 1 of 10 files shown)


commodities/jinja2/includes/commodities/actions.jinja
```diff
@@ -0,0 +1,8 @@
+<div class="govuk-grid-column-one-quarter">
```
a-gleeson, line 1: does the grid-layout maybe not belong here and should wrap the include statement in the measures-defined.jinja

commodities/helpers.py
```diff
@@ -0,0 +1,20 @@
+from commodities.models.dc import CommodityCollectionLoader
+from commodities.models.dc import CommodityTreeSnapshot
+from commodities.models.dc import SnapshotMoment
+from measures.snapshots import MeasureSnapshot
+
+
+def get_measures_on_declarable_commodities(transaction, item_id, date=None):
```
a-gleeson, line 7: docstring to maybe help explain this function?

commodities/views.py
```diff
@@ -151,6 +159,10 @@ def get_context_data(self, *args, **kwargs):
             GoodsNomenclature.objects.filter(sid=self.kwargs["sid"]).current().first()
```
a-gleeson, line 159: can this be replaced with self.commodity?

eadpearce: oh good spot. yes

commodities/jinja2/includes/commodities/actions.jinja
eadpearce: Yes probably because I remember messing up and nesting it in a column twice when I first added it haha

a-gleeson (approved):

PR #911, opened 2023-05-18
Tp2000 780  separate rule check worker

a-gleeson:
# TP2000-780  Separate rule check worker

## Why
- CPU usage on a single worker is exceeding over 100% in production so load should be reduced
- The rule check can execute multiple tasks and can be log running seperating out the worker ensure that these tasks will not hold up other tasks or vise versa

## What
- Celery routing updated to route to specific queue's standard & rule-check
- This routing is then set up locally using multiple docker instances each running a different queue
- Procfile updated to include new worker (will not run until scaled)
- Updating this page with infrastructure details https://readme.trade.gov.uk/docs/playbooks/trade-tariff-management.html

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -435,5 +435,4 @@ You may get a message like ``There are no running instances of this process.`` I
 
 .. code:: sh
 
-    cf scale <space> --process <process_name> -i <number_of_instances> -m 1G
-    
+    cf scale <space> --process <process_name> -i <number_of_instances> -m <memory>
--- a/README.rst
+++ b/README.rst
@@ -410,29 +410,3 @@ environment database with ``cf conduit tamato-dev-db -- psql``.
 
 .. |codecov| image:: https://codecov.io/gh/uktrade/tamato/branch/master/graph/badge.svg
    :target: https://codecov.io/gh/uktrade/tamato
-
-Scaling Celery workers in GOV.UK PaaS
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Then you need to login to the DIT GOV.UK PaaS:
-
-.. code:: sh
-
-    cf login --sso -s <space>
-
-Where ``<space>`` is one of ``tariffs-dev``, ``tariffs-staging``,
-``tariffs-training`` or ``tariffs-uat``.
-
-Once you are logged in, you can list the application hosted in the space and their resources with
-
-.. code:: sh
-
-    cf apps
-    cf scale <space>
-
-You may notice that the celery worker isn't running or running with the resources you'd like.
-You may get a message like ``There are no running instances of this process.`` In this case you should scale the worker with
-
-.. code:: sh
-
-    cf scale <space> --process <process_name> -i <number_of_instances> -m <memory> -d <disk>
--- a/README.rst
+++ b/README.rst
@@ -435,4 +435,4 @@ You may get a message like ``There are no running instances of this process.`` I
 
 .. code:: sh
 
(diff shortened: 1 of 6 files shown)


paulpepper-trade (changes requested):

settings/common.py
```diff
@@ -459,6 +459,16 @@
     },
 }
 
+CELERY_ROUTES = {
```
paulpepper-trade, line 462: Should we be using the new lowercase settings now? https://docs.celeryq.dev/en/stable/userguide/configuration.html?highlight=CELERY_ROUTES#new-lowercase-settings

settings/common.py
```diff
@@ -459,6 +459,16 @@
...
+    re.compile(r"(checks)\.tasks\..*"): {"queue": "rule-check"},
+    re.compile(
+        r"(exporter|importer|notifications|publishing|workbaskets)\.tasks\..*",
```
paulpepper-trade, line 465: As there are only two routes, each being the complement of the other, could this be expressed as a _not checks_ regex. Something like
```suggestion
        r"(?!checks)\.tasks\..*",
```

README.rst
```diff
@@ -407,3 +410,29 @@ environment database with ``cf conduit tamato-dev-db -- psql``.
...
+
+    cf apps
+    cf scale <space>
+
+You may notice that the celery worker isn't running or running with the resources you'd like.
+You may get a message like ``There are no running instances of this process.`` In this case you should scale the worker with
+
+.. code:: sh
+
+    cf scale <space> --process <process_name> -i <number_of_instances> -m <memory> -d <disk>
```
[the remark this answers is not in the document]
a-gleeson: I'll remove it, there is reference to it already. It's just a little vague.
This is taken from public documents from cloud foundry so the commit isn't harmful but I think it doesn't belong here.

common/celery.py
```diff
@@ -15,3 +16,7 @@
 
 # Load task modules from all registered Django app configs.
 app.autodiscover_tasks()
+
+# this *shouldn't* need to be applied directly but
+# automatically configured via ^^ config_from_object
```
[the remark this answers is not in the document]
a-gleeson: I need to set the celery routes in the line below. It should be automatically configured but instead I have to set it below

settings/common.py
a-gleeson: I'll try implement this

common/celery.py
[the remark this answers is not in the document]
paulpepper-trade: Could you perhaps clarify that in the comment?

settings/common.py
paulpepper-trade: It looks as though you can't mix old and new settings styles and I hadn't noticed the other celery settings that weren't shown in the diff! https://docs.celeryq.dev/en/stable/history/whatsnew-4.0.html#latentcall-django-admonition

paulpepper-trade (approved):

settings/common.py
```diff
@@ -459,6 +459,24 @@
     },
 }
 
+CELERY_ROUTES = {
```
paulpepper-trade, line 462:
```suggestion
CELERY_ROUTES = {
    # All rules check tasks are sent to the `rule-check` queue.
    re.compile("^(checks)\.tasks\..*$"): {
        "queue": "rule-check",
    },
    # All other, non-rules check tasks are sent to the `standard` queue.
    re.compile("^((?!checks).)*\.tasks\..*$"): {
        "queue": "standard",
    },
}
```

PR #912, opened 2023-05-18
TP2000-776/88  create envelope to publish

a-gleeson:
# TP2000-776 TP2000-788 Celery beats, retry and tests

## Why
To create production ready api publishing

## What
- Unit tests
- retry configuration
- Cron config for celery beat
- fixed sequence check when create API publish


sample.env
```diff
@@ -65,3 +65,10 @@ CLAM_AV_PASSWORD=password
 CLAM_AV_DOMAIN=test.com
 # Use the below if connecting to local ClamAV, see README for more info
 CLAM_USE_HTTP=True
+
+# Overwrite default cron for api publishing default "0 8-18/2 * * 1-5"
+CHANNEL_ISLANDS_API_CRON="*/5 * * * *"
```
dalecannon, line 70: It might also be helpful to say this cron expression is for every 5 minutes.

publishing/tests/test_models.py
```diff
@@ -451,3 +452,81 @@ def test_next_envelope_id(envelope_storage):
...
+    pwb.save()
+
+    PackagedWorkBasket.create_api_publishing_envelope()
+
+    pwb.refresh_from_db()
+
+    assert not pwb.tap_api_envelope
+
+
+def test_create_api_publishing_envelope_no_envelope_to_publish_2(
```
dalecannon, line 522: Perhaps a docstring could be added to help further differentiate it from the other test with the similar name?

PR #913, opened 2023-05-18
TP2000-777  Publish to API task

dalecannon:
# TP2000-777 Publish to API task

## Why
To implement a Celery task that manages publishing to the Tariff API.

## What
- Implements `publish_to_api()` task
- Adds state transition management for `TAPApiEnvelope`
- Adds `TariffAPIClient` to make POST request to API
- Adds `TariffAPIStubbed` and `TariffAPI` interface classes
- Adds Tariff API settings and updates sample.env
- Adds unit tests for `publish_to_api()`


publishing/interface.py
```diff
@@ -0,0 +1,66 @@
...
+    def post_envelope_staging(self, envelope: Envelope) -> Response:
+        """Upload envelope to Tariff API staging environment."""
+        return self.client.post_envelope_staging(envelope=envelope)
+
+    def post_envelope_production(self, envelope: Envelope) -> Response:
+        """Upload envelope to Tariff API production environment."""
+        return self.client.post_envelope_production(envelope=envelope)
+
+
+def get_tariff_api_interface() -> TariffAPIBase:
```
a-gleeson, line 57: should we maybe move this function to an init?
The client & interface.py could be moved in a subfolder called tariff_api

settings/common.py
```diff
@@ -431,6 +431,18 @@
 AWS_S3_SIGNATURE_VERSION = "s3v4"
 AWS_S3_REGION_NAME = "eu-west-2"
 
+# Tariff API envelope publishing automation
+TARIFF_API_INTERFACE = os.environ.get(
+    "TARIFF_API_INTERFACE",
+    "publishing.interface.TariffAPIStubbed",
```
a-gleeson, line 437: in common.py our default values should be what we use in production while our default values in sample.env can be what we would use locally

publishing/tasks.py
```diff
@@ -91,21 +92,91 @@ def schedule_create_xml_envelope_file(
...
+        if not envelope.can_publish():
+            logger.info(
+                f"Failed publishling to Tariff API: {envelope}. "
+                f"The previous envelope is unpublished",
+            )
+            return
+
+        pwb_envelope = envelope.packagedworkbaskets.last().envelope
+
+        if envelope.publishing_state in [
```
a-gleeson, line 171: I'd add a comment to this to help explain it

publishing/models/tap_api_envelope.py
```diff
@@ -165,3 +183,99 @@ class Meta:
 
     def __repr__(self):
         return f'<TAPApiEnvelope: id="{self.pk}", publishing_state={self.publishing_state}>'
+
+    def previous_envelope(self) -> ApiEnvelopeQuerySet:
+        """Get the previous `TAPApiEnvelope` by order of `pk`."""
+        return TAPApiEnvelope.objects.filter(pk__lt=self.pk).order_by("-pk").first()
```
a-gleeson, line 189: as you know you are getting the pervious pk would it be more efficent to do self.pk - 1 instead of an order_by

dalecannon: Yeah. I originally had a mind a scenario in which `TAPApiEnvelope`'s are deleted and the previous one might then not be found by decrementing. But a `TAPApiEnvelope` shouldn't be deleted anyway.

sample.env
```diff
@@ -29,6 +29,14 @@ SQLITE_STORAGE_BUCKET_NAME=sqlite
 MINIO_ACCESS_KEY=minio_access_key
 MINIO_SECRET_KEY=minio_secret_key
 
+# Tariff API envelope publishing automation
```
a-gleeson, line 32: maybe include default local value's in here

PR #920, opened 2023-05-24
TP2000-891 Multi-version + same transaction indent causes ME32

paulpepper-trade:
# TP2000-891 Multi-version + same transaction indent causes ME32

## Why

As per TARIC3, it's valid to have multiple instances of an indent that provide different start dates within the same transaction. This arrangement of indents in TAP data isn't catered for in commodity tree snapshot code, causing incorrect commodity extent values and therefore ancestor commodities within the tree.

## What

In addition to ordering indents by transaction when loading the commodity collection, also apply a secondary ordering of `start_date`. This ensures correct indent selection when constructing a commodity tree snapshot.

## Checklist
- Requires migrations? No

### DIFF
--- a/commodities/models/dc.py
+++ b/commodities/models/dc.py
@@ -584,7 +584,7 @@ class CommodityTreeSnapshot(CommodityTreeBase):
 
         NOTE: This methods is invoked by __post_init__ and should not be called directly.
 
-        The commodity collection is sorted on code, suffix and indent.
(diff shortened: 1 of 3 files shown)


eadpearce (commented):
tests?

paulpepper-trade:
> tests?

eadpearce (approved):

PR #922, opened 2023-05-26
Handle currently processing state

dalecannon:
# Handle currently processing state

## Why
An envelope becomes "stuck" in currently processing state if the previous publishing task fails to transition it to a new publishing state and/or set any published timestamps.

## What
- Extends API Client to include GET request for an envelope
- Handles API envelope stuck in currently processing state
- Adds tests

TP2000-777


a-gleeson (approved):
LGTM

a-gleeson (approved):

PR #923, opened 2023-05-26
TP2000-682  Add password policy validator

dalecannon:
# TP2000-682  Add password policy validator

## Why
Password validation in the admin (applies only to local, non-SSO created accounts) must take into account password policy.

## What
- Adds `PasswordPolicyValidator` that is used in addition to existing password validators.
- Adds test

### DIFF
--- a/common/validators.py
+++ b/common/validators.py
@@ -89,15 +89,16 @@ SymbolValidator = RegexValidator(
 
 class PasswordPolicyValidator:
     """Validate whether the password contains at least 1 capital letter, 1
-    number and a special character."""
+    lowercase letter, 1 number and a special character."""
 
-    HELP_TEXT = "Your password must contain at least 1 capital letter, 1 number and a special character."
+    HELP_TEXT = "Your password must contain at least 1 capital letter, 1 lowercase letter, 1 number and a special character."
 
     def validate(self, password, user=None):
         if (
             password.isalnum()
             or not any(c.isdigit() for c in password)
             or not any(c.isupper() for c in password)
+            or not any(c.islower() for c in password)
         ):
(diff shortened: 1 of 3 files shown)


common/validators.py
```diff
@@ -85,3 +85,24 @@ class ApplicabilityCode(models.IntegerChoices):
...
+        ):
+            raise ValidationError(
+                (
+                    "This password must contain at least 1 capital letter, 1 number and a special character."
+                ),
+                code="password_missing_characters",
+            )
+
+    def get_help_text(self):
+        return "Your password must contain at least 1 capital letter, 1 number and a special character."
```
paulpepper-trade, line 108: Could this string literal be factored out into, say, a class-level const since it's the same value as `ValidationError` message?

common/validators.py
```diff
@@ -85,3 +85,24 @@ class ApplicabilityCode(models.IntegerChoices):
     r"^[0-9A-Za-z\s.',()&£$%@!/\+-]*$",
     ValidationError("Only symbols .,/()&£$@!+-% are allowed."),
 )
+
+
+class PasswordPolicyValidator:
+    """Validate whether the password contains at least 1 capital letter, 1
```
paulpepper-trade, line 91: Does "at least 1 capital" imply that the validator should also check that there's at least one lowercase character too?

paulpepper-trade (approved):

common/validators.py
dalecannon: I omitted checking for at least one lowercase character only because the policy doesn't explicitly mention that it must contain one. The policy says to "use a password that is a minimum of X digits, 1 capital letter, 1 number and a special character".

PR #925, opened 2023-05-31
TP2000-781  Display publishing task in admin

dalecannon:
# TP2000-781  Display publishing task in admin

## Why
We need to be able to view the status and ID of an envelope's publishing task and to terminate it.

## What
- Adds `publishing_task_id` field to CrownDependenciesEnvelope model
- Adds `publishing_task_status()` to return the status of an envelope's publishing task
- Adds `terminate_publishing_task()` to terminate an envelope's publishing task
- Adds unit test

## Checklist
- Requires migrations? Yes


publishing/models/crown_dependencies_publishing_task.py
```diff
@@ -0,0 +1,68 @@
...
+
+    task_id = CharField(
+        max_length=50,
+        null=True,
+        blank=True,
+        unique=True,
+    )
+
+    @property
+    def task_status(self):
```
a-gleeson, line 25: add a return type?

a-gleeson (approved):

PR #927, opened 2023-06-05
TP2000-787  Toggle and pause publishing task

dalecannon:
# TP2000-787  Toggle and pause publishing task

## Why
We need functionality in place to enable/disable  and to pause/unpause publishing task.

## What
- Adds config setting `ENABLE_CROWN_DEPENDENCIES_PUBLISHING`
- Adds model `CrownDependenciesPublishingOperationalStatus` to control publishing task state
- Adds test


publishing/models/__init__.py
```diff
@@ -24,6 +27,7 @@
     "ApiPublishingState",
     "ProcessingState",
     "QueueState",
+    "CrownDependenciesPublishingState",
```
a-gleeson, line 30: Is this imported?

a-gleeson (approved):
Looks good to me!

PR #930, opened 2023-06-07
TP2000-894: Quota origin edits

eadpearce:
# TP2000-894: Quota origin edits

## Why

* Users need to be able to edit quota origins for some upcoming policy implementation

## What

* Adds an origin section to the quota edit form
* Allows editing of the origin validity, geographical area and exclusions
* Currently it's only possible to edit quotas with one origin, not multiple
* Also updates the quota detail UI to better represent quota origin data

FireShot Capture 059 - Edit quota details - Manage Trade Tariffs - localhost

Screenshot 2023-06-05 at 11 35 35
Screenshot 2023-06-05 at 11 35 47
Screenshot 2023-06-07 at 13 20 21

### DIFF
--- a/common/forms.py
+++ b/common/forms.py
@@ -16,8 +16,6 @@ from django import forms
 from django.contrib.postgres.forms.ranges import DateRangeField
 from django.core.exceptions import ValidationError
 from django.db.models import TextChoices
-from django.forms import TypedChoiceField
-from django.forms import formsets
 from django.forms.renderers import get_default_renderer
 from django.forms.utils import ErrorList
 from django.forms.widgets import Widget
@@ -26,7 +24,11 @@ from django.utils.safestring import mark_safe
 
 from common.util import TaricDateRange
 from common.util import get_model_indefinite_article
+from common.widgets import FormSetFieldWidget
 from common.widgets import RadioNestedWidget
+
+MESSAGE_FORM_MIXIN = "This field requires the form to use BindNestedFormMixin"
+MESSAGE_BIND_FORMS = "Nested forms must be instantiated with bind_nested_forms in the subclass's __init__"
 
 
 class BindNestedFormMixin:
@@ -35,6 +37,28 @@ class BindNestedFormMixin:
 
     in order to instantiate and validate nested forms.
     """
+
+    def get_nested_forms(self, form_class, args, kwargs):
+        nested_forms = []
+        if issubclass(form_class, FormSet):
+            formset_kwargs = kwargs.copy()
+            if "initial" in formset_kwargs:
+                formset_kwargs.pop("initial")
+            formset_initial = None
+            if kwargs.get("initial"):
+                formset_initial = kwargs.get("initial").get(
+                    form_class.prefix,
+                )
+
+            bound_form = form_class(
+                *args,
+                **formset_kwargs,
+                initial=formset_initial,
+            )
+        else:
+            bound_form = form_class(*args, **kwargs)
+        nested_forms.append(bound_form)
+        return nested_forms
 
     def bind_nested_forms(self, *args, **kwargs):
         if kwargs.get("instance"):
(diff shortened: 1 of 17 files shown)


common/forms.py
```diff
@@ -47,61 +69,62 @@ def bind_nested_forms(self, *args, **kwargs):
...
         return any(nested_formset_submit)
 
     def is_valid(self):
         return super().is_valid() and not self.formset_submit()
 
+    @staticmethod
+    def clean_form(form, cleaned_data):
+        if form.is_valid():
+            data = form.cleaned_data
+            if isinstance(form, FormSet):
```
a-gleeson, line 100: Just checking the purpose of this function is to clean the form but this isinstance check will change the way formset data is stored so we can have multiple forms stored in cleaned_data?

common/forms.py
```diff
@@ -168,7 +189,51 @@ def validate(self, value):
...
+                FormSet,
+            ), MESSAGE_BIND_FORMS
+            if not form.is_valid():
+                if isinstance(form, FormSet):
+                    if form.formset_action is not None:
+                        nested_formset_submit.append(True)
+                    for errors in form.errors:
+                        for e in errors.values():
+                            if e:
+                                raise ValidationError(e)
```
a-gleeson, line 225: If errors had multiple values would we only raise the first one? On the front end would we only see one error at a time?

common/templates/common/widgets/formset_field.html
```diff
@@ -0,0 +1,58 @@
...
+				{% endif %}
+			{% endif %}
+
+			{% for field in form %}
+				{% include "gds/field.html" %}
+			{% endfor %}
+		{% endwith %}
+
+		{% if nested_form.is_formset %}
+			{% if nested_form.initial|length == nested_form.max_num %}
```
a-gleeson, line 27: should we display non_form_errors for a formset here?

measures/forms.py
```diff
@@ -23,19 +22,24 @@
...
 from common.forms import ValidityPeriodForm
 from common.forms import delete_form_for
-from common.forms import formset_add_or_delete
 from common.forms import formset_factory
 from common.forms import unprefix_formset_data
 from common.util import validity_range_contains_range
 from common.validators import SymbolValidator
 from common.validators import UpdateType
 from footnotes.models import Footnote
+from geo_areas.constants import *
```
a-gleeson, line 35: maybe import like `from geo_areas import constants`?

quotas/forms.py
```diff
@@ -66,7 +77,45 @@ def __init__(self, *args, **kwargs):
...
+        label="",
+        queryset=GeographicalArea.objects.all(),  # modified in __init__
+        help_text="Select a country to be excluded:",
+        required=False,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["exclusion"].queryset = (
+            GeographicalArea.objects.current()
```
a-gleeson, line 91: this query set seems to be repeated, could it be extracted to the GeographicalArea queryset?

common/forms.py
```diff
@@ -641,14 +713,14 @@ def formset_factory(
     to the formset's attrs.
     """
     if min_num is None:
-        min_num = formsets.DEFAULT_MIN_NUM
+        min_num = forms.formsets.DEFAULT_MIN_NUM
```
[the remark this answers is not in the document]
eadpearce: It's from the import at the top of the file `from django import forms`

common/forms.py
```diff
@@ -714,14 +786,23 @@ def unprefix_formset_data(prefix, data):
     return output
 
 
-def formset_add_or_delete(prefixes, data):
+def formset_add_or_delete(data):
     """Find any {prefix}-ADD or {prefix}-DELETE in submitted data and return the
```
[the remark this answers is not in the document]
eadpearce: I think it still makes sense. It's just that we're not using the prefixes arg any more for the matching

common/templates/common/widgets/formset_field.html
eadpearce: iirc those errors get passed up to the parent form and will be shown in that form's non_form_errors instead

eadpearce: (this template is very similar to nested_radio.html)

measures/forms.py
eadpearce: I know this is bad practice but it takes up so many lines haha. No, I'll change it. Better to be explicit

common/forms.py
eadpearce: Yep yep. So the formset data starts out like:
```
"measure-conditions-formset-0-applicable_duty": "test1",
"measure-conditions-formset-1-applicable_duty": "test2",
"measure-conditions-formset-2-applicable_duty": "test3",
```
And ends up like:
```
{"applicable_duty": "test1"},
{"applicable_duty": "test2"},
{"applicable_duty": "test3"},
```

common/forms.py
eadpearce: Not sure. I think I need to add a test for this

common/forms.py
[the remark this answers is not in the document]
a-gleeson: thanks missed that

quotas/forms.py
eadpearce: Just tested and I can move it to the field definition so I'll delete this

common/templates/common/widgets/formset_field.html
a-gleeson: makes sense to me, I am also not sure we have any non form errors for any of these forms yet.

quotas/views.py
```diff
@@ -255,14 +258,84 @@ class QuotaUpdateMixin(
...
         object = super().get_result_object(form)
 
+        geo_area = form.cleaned_data["geographical_area"]
+        edited_origin = form.cleaned_data["existing_origin"]
+        form_exclusions = [
+            item["exclusion"]
+            for item in form.cleaned_data[QUOTA_ORIGIN_EXCLUSIONS_FORMSET_PREFIX]
+        ]
+
+        print("form_exclusions", form_exclusions)
```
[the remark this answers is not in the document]
eadpearce: Oops!

common/templates/common/widgets/formset_field.html
eadpearce: I think if we get business rules failures those will show up in the non_form_errors

PR #931, opened 2023-06-08
TP2000-894: Fix quota form exclusions prepopulate

eadpearce:
# TP2000-894: Fix quota form exclusions prepopulate

## Why

* Exclusions initial data was using the wrong model

## What

* Use the geographical area for form initial data instead of quota order number origin exclusion


a-gleeson (approved):
Good catch!

PR #932, opened 2023-06-09
TP2000-905  Remove newline characters from autocomplete results

dalecannon:
# TP2000-905  Remove newline characters from autocomplete results

## Why
Autocomplete doesn't handle results which contain new lines characters consistently. Results which contain new line characters (for example, the descriptions for commodity codes 8708942000 & 8414302030) are considered as a new query when selected, causing results to be repopulated again unnecessarily through an additional fetch request. 

If a user selects one of the results and submits a form before the fetch request has resolved, then the user's selection can sometimes be "forgotten" by the server. This intermittent issue can be seen occurring during the create a new measure (TP2000-823) and multiple measure edit journeys (TP2000-905). 

## What
- Adds a new function to remove new line characters from autocomplete results
- Fixes another issue with autocomplete where a previously selected result shows as undefined

## Examples
Problematic, superfluous request:
Screenshot 2023-06-09 at 12 50 41

Undefined error:
Screenshot 2023-06-09 at 12 53 05

### DIFF
--- a/common/static/common/js/autocomplete.js
+++ b/common/static/common/js/autocomplete.js
@@ -47,7 +47,7 @@ const autoCompleteElement = (element, includeNameAttr=true) => {
     templates: {
       inputValue: template,
       suggestion: template
-    },±
+    },
     onConfirm: value => {
       const autocomplete = document.querySelector(`#${hiddenInput.id}_autocomplete`);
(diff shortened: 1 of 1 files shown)


eadpearce (approved):
🔥

PR #934, opened 2023-06-13
TP2000-888  Add ME40 validation to create a new measure journey

dalecannon:
# TP2000-888  Add ME40 validation to create a new measure journey

## Why
Create a new measure journey lacks validation to prevent ME40 violations.

## What
- Fixes the displaying of duties under conditions on the review page
- Commodities and conditions forms check if duties are permitted for the selected measure type
- Review form checks if mandatory duties are missing for the selected measure type
- Updates and adds tests

#
Screenshot 2023-06-13 at 09 30 08

Screenshot 2023-06-13 at 09 25 33

Screenshot 2023-06-13 at 09 28 15

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1297,6 +1297,7 @@ class MeasureReviewForm(forms.Form):
             commodities_data=self.commodities_data,
             conditions_data=self.conditions_data,
         )
+        return super().clean()
 
 
 MeasureDeleteForm = delete_form_for(models.Measure)
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -391,7 +391,7 @@ class MeasureConditionsFormSet(MeasureConditionsBaseFormSet):
 class MeasureConditionsWizardStepForm(MeasureConditionsFormMixin):
     # override methods that use form kwargs
     def __init__(self, *args, **kwargs):
-        self.measure_start_date = kwargs.pop("measure_start_date")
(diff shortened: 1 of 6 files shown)


eadpearce (commented):
I've just realised we don't do this for any of the measures views but on some other models (geo areas, footnotes etc) we use views that inherit from TrackedModelChangeView which uses the BusinessRulesMixin allowing us to set validate_business_rules property on the class with a list of business rules to check against on form submission. I haven't investigated this myself but would it work to use this same pattern here? Or if it doesn't work with the form wizard, maybe we should find a way to rewrite the mixin in a way that does work. Then we could just reference the business rule instead of pulling out the logic that lives inside it

eadpearce (approved):

dalecannon:
> I've just realised we don't do this for any of the measures views but on some other models (geo areas, footnotes etc) we use views that inherit from TrackedModelChangeView which uses the BusinessRulesMixin allowing us to set validate_business_rules property on the class with a list of business rules to check against on form submission. I haven't investigated this myself but would it work to use this same pattern here? Or if it doesn't work with the form wizard, maybe we should find a way to rewrite the mixin in a way that does work. Then we could just reference the business rule instead of pulling out the logic that lives inside it

That’s an interesting idea and would certainly be preferable. To make use of the mixin would likely require some reworking in terms of passing the object (new measure) to be checked and making violation error messaging user friendly. But it sounds like it’s worth creating a follow-up ticket to investigate further.

PR #940, opened 2023-06-16
TP2000-911: Fix measure footnotes display

eadpearce:
# TP2000-911: Fix measure footnotes display

## Why

* What was displayed in the UI was not matching the latest version of the measure stored in the db

## What

* Use footnoteassociationmeasure_set.current() to get the most up to date version of footnotes via the footnote association for the measure

### DIFF
--- a/measures/jinja2/includes/measures/list.jinja
+++ b/measures/jinja2/includes/measures/list.jinja
@@ -49,7 +49,7 @@
               <a class="govuk-link govuk-!-font-weight-bold" href="{{ measure.get_url() }}">{{measure.sid}}</a>
             {%- endset %}
             {% set footnotes %}
-              {% for footnote in measure.footnoteassociationmeasure_set.all() %}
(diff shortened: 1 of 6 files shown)


measures/jinja2/includes/measures/list.jinja
```diff
@@ -49,7 +49,7 @@
               <a class="govuk-link govuk-!-font-weight-bold" href="{{ measure.get_url() }}">{{measure.sid}}</a>
             {%- endset %}
             {% set footnotes %}
-              {% for footnote in measure.footnotes.all() %}
+              {% for footnote in measure.footnoteassociationmeasure_set.all() %}
```
a-gleeson, line 52: Do we still want to use .all() here?

a-gleeson (approved):
LGTM!

measures/jinja2/includes/measures/list.jinja
eadpearce: oops missed one! thanks

PR #942, opened 2023-06-20
TP2000-879: Add commodity code footnotes

eadpearce:
# TP2000-879: Add commodity code footnotes

## Why

* Users need a way to add footnotes directly to a commodity code

## What

* Adds a link to a new add footnote form on the commodity detail pages

Footnotes tab
Add footnotes form

Footnote association create confirmation screen
Workbasket summary


commodities/views.py
```diff
@@ -216,10 +243,69 @@ def get_context_data(self, *args, **kwargs):
...
+    def form_valid(self, form):
+        self.object = self.get_result_object(form)
+        tx = self.object.transaction
+
+        for rule in self.validate_business_rules:
+            try:
+                with transaction.atomic():
+                    rule(tx).validate(self.object)
+            except BusinessRuleViolation as v:
+                form.add_error(None, v.args[0])
```
dalecannon, line 298: Would something like `transaction.set_rollback(True)` need to be added here to undo the object creation in the workbasket upon a business rule violation?

commodities/views.py
```diff
@@ -216,10 +243,69 @@ def get_context_data(self, *args, **kwargs):
...
+                return self.form_invalid(form)
+
+        return FormMixin.form_valid(self, form)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["commodity"] = self.commodity
```
dalecannon, line 305: I was looking into the `TransactionManagementError` you had come up against . I think it stems from a conflict in which version of the commodity is returned by the query (current() versus any latest version perhaps?). I found that there's no need to override `form_valid` in this view to prevent that error if this commodity context is unused/unneeded.

eadpearce: Yeah we need the commodity in the context and the form though. I don't know if I can find a way around this error

commodities/views.py
eadpearce: Yep and now I've added it it's broken again. I'm going to have to rewrite this view

commodities/urls.py
```diff
@@ -23,6 +23,11 @@
         views.CommodityDetail.as_view(),
         name="commodity-ui-detail",
     ),
+    path(
+        f"commodities/<sid>/footnotes/",
+        views.CommodityDetailFootnotes.as_view(),
+        name="commodity-ui-detail-footnotes",
```
dalecannon, line 29: Might it be worth adding a test for this new view too?

dalecannon (approved):

PR #943, opened 2023-06-20
Associate imports with workbasket

paulpepper-trade:
# Associate imports with workbasket

## Why

Imports in their various states must be able to reference the workbasket that tracked models are deposited into.

## What

This PR:

* Captures the Jira ticket ID in the upload form at the start of the upload process.
* Adds a workbasket association on ImportBatch.
* Adds an WorkBasketQuerySet filter to exclude workbaskets that have an associated actively importing or errored import.


importer/forms.py
```diff
@@ -138,8 +140,24 @@ class CommodityImportForm(ImportFormMixin, forms.Form):
...
+        max_length=255,
+        strip=True,
+        label="Tops/Jira number",
+        help_text=(
+            "Your Tops/Jira number is needed to associate your import's "
+            "workbasket with your Jira ticket. You can find this number at the "
+            "end of the web address for your Jira ticket. Your workbasket will "
+            "be given a unique number that may be different to your Tops/Jira "
+            "number. "
+        ),
```
dalecannon, line 160: Could we make use of `tops_jira_number_validator` here?

dalecannon (approved):

importer/forms.py
paulpepper-trade: Done.

PR #945, opened 2023-06-21
Updated ME67 and added additional unit tests 

dougmills-DIT:
added tests to verify the change not only passes existing tests but extends and verifies and passes additional scenarios that were incorrectly failed in production previously

# TP2000-914 Your PR title here

## Why
 * ME67 did not pass in valid scenarios.
 * Unit test coverage missed some data configurations that were valid.

## What
 * Updated the logic for ME67 to verify the date range of the change against the data
 * Added tests to cover additional scenarios, observed as incorrectly raising violations in production

## Checklist
- Requires migrations? NO
- Requires dependency updates? NO

Links to relevant material
See: [Description](https://uktrade.atlassian.net/browse/TP2000-914)

### DIFF
--- a/measures/business_rules.py
+++ b/measures/business_rules.py
@@ -1,6 +1,5 @@
 """Business rules for measures."""
 from datetime import date
-from datetime import timedelta
 from typing import Mapping
 from typing import Optional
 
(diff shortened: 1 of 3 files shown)


measures/business_rules.py
```diff
@@ -1231,30 +1232,42 @@ class ME66(ExclusionMembership):
...
+                valid_between__startswith__lte=measure.valid_between.lower,
             )
-            .filter(
-                geo_group__version_group=geo_group.version_group,
-                member__version_group=excluded.version_group,
-            )
-            .exists()
-        ):
+            & (
+                # because of the way that dates are queried, we need to add one day to the upper
```
paulpepper-trade, line 1260: Because the top of the date range is open - comparisons performed with less-than don't include the top value? Maybe include something to explain that in the comment?

paulpepper-trade (approved):

PR #953, opened 2023-06-27
TP2000-921 transition import status to succeeded and failed

paulpepper-trade:
# TP2000-921 transition import status to succeeded and failed

## Why

When the parser has completed its work it must, depending on its whether it has succeeded or failed, transition the related ImportBatch instance status to SUCCEEDED or FAILED.

## What

Upon the parser completing it correctly transitions the `ImportBatch.status` via the `succeeded()` and `failed()` transition functions. The workbasket associated with the import is also managed correctly, transitioning its state depending upon the outcome of the import. Info log messages have also been added to help with any debugging / investigation where necessary.

Comments also added to the parser code while exploring required work.

## Checklist
- Requires migrations? No


a-gleeson (approved):
LGTM! comments help you follow the steps

PR #954, opened 2023-06-28
TP2000-929  Fix quota form geo area member exclusions

dalecannon:
# TP2000-929  Fix quota form geo area member exclusions

## Why
The `get_all_members_of_geo_groups` function  (used when editing a quota order number origin to update any exclusions) doesn't use the latest version of `GeographicalMembership` queryset. This is causing an issue where an end-dated member of the passed in geo group is still ultimately added as an exclusion.

Also, the function doesn't guard against geo groups sharing members, sometimes resulting in a member/geo area being added multiple times as an exclusion.

When testing this PR, another issue had also been identified this time in the create measure journey concerning geo area member exclusions. `create_measure_excluded_geographical_areas` like `get_all_members_of_geo_groups` does not use the latest version of `GeographicalMembership`, causing a server 500 error on attempted measure creation.

## What
- Use current() on querysets to get the latest version of memberships
- Return all members as a set to ensure there are no duplicate geo areas
- Update test

##
Before:
Screenshot 2023-06-28 at 13 04 43

##
Screenshot 2023-06-28 at 14 56 41

### DIFF
--- a/geo_areas/utils.py
+++ b/geo_areas/utils.py
@@ -4,21 +4,21 @@ from geo_areas.validators import AreaCode
 
 
 def get_all_members_of_geo_groups(validity, geo_areas):
-    all_members = []
+    all_members = set()
     for geo_area in geo_areas:
         if geo_area.area_code == AreaCode.GROUP:
-            valid_memberships = [
-                member
-                for member in GeographicalMembership.objects.filter(
-                    geo_group=geo_area,
-                )
-                if date_ranges_overlap(member.valid_between, validity)
-            ]
-            origins = set(m.member for m in valid_memberships)
-            for membership in valid_memberships:
-                if membership.member.sid in [m.sid for m in origins]:
-                    all_members.append(membership.member)
+            all_members.update(
+                {
+                    membership.member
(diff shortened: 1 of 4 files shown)


geo_areas/utils.py
```diff
@@ -4,21 +4,21 @@
...
-                for member in GeographicalMembership.objects.filter(
-                    geo_group=geo_area,
-                )
-                if date_ranges_overlap(member.valid_between, validity)
-            ]
-            origins = set(m.member for m in valid_memberships)
-            for membership in valid_memberships:
-                if membership.member.sid in [m.sid for m in origins]:
-                    all_members.append(membership.member)
+            all_members.update(
```
paulpepper-trade, line 10: Does this `update()` always result in the correct member in the set? If so, could you say why (maybe in a code comment, too)?

paulpepper-trade (approved):

geo_areas/utils.py
dalecannon: I wouldn't imagine so as `update()` wouldn't overwrite the original item with a new items with the same hash.

PR #956, opened 2023-06-29
TP2000-936 Add validity period to quota order number autocomplete label

dalecannon:
# TP2000-936 Add validity period to quota order number autocomplete label

## Why
Autocomplete results for quota order numbers don't distinguish multiple versions of the same quota order numbers with different validity periods. 

## What
- Adds validity period to quota order number autocomplete label to help users identity the correct quota order number for their use

##
Before:
Screenshot 2023-06-29 at 11 03 51

##
After:
Screenshot 2023-06-29 at 10 48 05

### DIFF
--- a/quotas/models.py
+++ b/quotas/models.py
@@ -87,7 +87,7 @@ class QuotaOrderNumber(TrackedModel, ValidityMixin):
 
     @property
     def autocomplete_label(self):
-        return str(self)
+        return f"{self} ({self.valid_between.lower} - {self.valid_between.upper})"
 
     @property
     def is_origin_quota(self):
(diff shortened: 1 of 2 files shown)


quotas/models.py
```diff
@@ -87,7 +87,7 @@ def __str__(self):
 
     @property
     def autocomplete_label(self):
-        return str(self)
+        return f"{self} ({self.valid_between.lower} - {self.valid_between.upper})"
```
dougmills-DIT, line 90: Nice one, but no test?

dougmills-DIT (approved):

PR #959, opened 2023-06-30
Fix for envelope export, when a measurement unit is created.

dougmills-DIT:
There is a condition where a measurement may not be present in the XML if there is no associated measurement unit qualifier. 

This issue currently prevents an existing workbasket from correctly producing an envelope as it fails at the count of expected elements. 

This fix updates the logic used to count the expected elements and corrects the issue. Tests also added

## What
 * Added a method to conditionally count measurements if the measurement is in the workbasket, and there is an associated unit qualifier 
 * Updates tests to check correctly for this


dougmills-DIT:
tested against TOPS-1154 envelope and it passes validation now.

a-gleeson (approved):
LGTM!

PR #971, opened 2023-07-12
TP2000-919: Add ability to end date and delete definition periods

eadpearce:
# TP2000-919: Add ability to end date and delete definition periods

## Why

* Users need to be able to end date and delete future quota definition periods

## What

* Adds the required edit and delete forms
* Shows edit links next to definition periods
* Only shows the delete link for definition periods that are in the future and not yet active

Screenshot 2023-07-12 at 16 50 23
Screenshot 2023-07-12 at 16 50 47
Screenshot 2023-07-12 at 16 50 03


quotas/jinja2/quota-definitions/confirm-delete.jinja
```diff
@@ -0,0 +1,35 @@
...
+  {{ govukPanel({
+    "titleText": messages,
+    "text": "This change has been added to your workbasket",
+    "classes": "govuk-!-margin-bottom-7"
+  }) }}
+{% endblock %}
+
+
+{% block actions %}
+<li><a class="govuk-link" href="{{ url('quota-definitions', kwargs={'sid': object.sid}) }}">View other definition periods for this quota order number</a></li>
```
a-gleeson, line 32: Does this link take us to the page we can click the links for edit & delete?
When working on the create new origin form, it was fedback to me that if the user would most likely go back to editing quota's. It was suggested to add that as a primary button (changes in my PR).
Should we check if that's preferred on this confirmation screen too?

quotas/jinja2/quota-definitions/confirm-update.jinja
```diff
@@ -0,0 +1,20 @@
...
+  {{ breadcrumbs(request, [
+      {"text": "Find and edit quotas", "href": url("quota-ui-list")},
+      {"text": object.order_number._meta.verbose_name|capitalize ~ ": " ~ object.order_number|string, "href": object.order_number.get_url()},
+      {"text": "Quota ID: " ~ object.order_number.order_number ~ " - Data", "href": url("quota-definitions", args=[object.order_number.sid])},
+      {"text": page_title}
+    ])
+  }}
+{% endblock %}
+
+
```
a-gleeson, line 16: ditto comment on the other page. Would a user want to go back to the definitions page to edit more?

quotas/jinja2/quotas/tables/definitions.jinja
```diff
@@ -43,6 +43,14 @@
   }) }}
   {% endset %}
 
+  {% set actions_html -%}
+    <a href="{{ url('quota_definition-ui-edit', args=[object.sid])}}" class="govuk-link">Edit</a>
+    {# only show the delete link for future definitions that are not currently active #}
+    {% if object in object_list.as_at_today_and_beyond() and object not in object_list.as_at_today() %}
```
a-gleeson, line 49: This makes sense. Would it make sense to have a queryset that will return an object list of only beyond today but not today? I am just thinking of moving some logic out of templates

quotas/models.py
```diff
@@ -364,6 +364,53 @@ class Meta:
     def __str__(self):
         return str(self.sid)
 
+    # TODO: refactor this method as it's the same as in QuotaOrderNumberOrigin
```
a-gleeson, line 367: leftover comment? Or reminder to pull this into a mixin?

eadpearce: Actually when I made this comment it was the same but now it's diverged quite a bit so maybe not reusable

quotas/views.py
```diff
@@ -399,3 +401,83 @@ class QuotaOrderNumberOriginConfirmUpdate(
...
+
+class QuotaDefinitionDelete(
+    QuotaDefinitionUpdateMixin,
+    CreateTaricDeleteView,
+):
+    form_class = delete_form_for(models.QuotaDefinition)
+    template_name = "quota-definitions/delete.jinja"
+
+    def form_valid(self, form):
+        messages.success(
```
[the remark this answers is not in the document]
eadpearce: Yeah we pull it into the global env here https://github.com/uktrade/tamato/blob/master/common/jinja2.py#L128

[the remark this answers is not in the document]
eadpearce: This particular message isn't displayed in this view, it's shown on the confirm page that this one redirects to. That was the neatest way I could think of to pass the quota definition sid to the view since after you delete it the TrackedModelDetailView throws a 404 if you try to fetch it again

quotas/jinja2/quotas/tables/definitions.jinja
eadpearce: I thought about doing that but felt it was too specific? Not sure if it would be used for anything except quota definitions

quotas/jinja2/quota-definitions/confirm-delete.jinja
eadpearce: Oh I see. That looks good! I think we should use the same pattern here

a-gleeson: Sounds good to me, it was feedback from Patrick that users like there main action to be first!

quotas/jinja2/quotas/tables/definitions.jinja
a-gleeson: that's fair, I can't think of any reuse initially and maybe having all these different querysets could get hard to maintain

quotas/models.py
a-gleeson: Ah right! I hadn't checked the other function but makes sense to me!

a-gleeson (approved):
Just a couple existing comments. The PR looks good!

quotas/views.py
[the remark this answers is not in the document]
eadpearce: Yeah exactly. I don't think we use this pattern anywhere else which seems kind of weird. On the other delete forms I looked at we just redirect back to the object list page but I feel like it's useful to have that confirmation and like you said in the other comment, have those links so users can continue editing/deleting other things.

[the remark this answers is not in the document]
a-gleeson: I am surprised this hasn't come up before! 
I prefer this too because just redirecting back to the object list could confuse the user as it may not be clear if it was successful or not!

quotas/jinja2/quota-definitions/confirm-delete.jinja
eadpearce: I just stole the code from your PR to add a primary button to this page and the edit confirm page

PR #974, opened 2023-07-17
TP2000-960 Fix editing footnotes measure edit form

dalecannon:
# TP2000-960 Fix editing footnotes measure edit form

## Why
Sometimes footnotes on a measure aren't shown on the form when editing a single measure, incorrectly appearing as if the measure has no footnotes. 

In addition, when a single measure is edited to remove footnotes, the changes aren't reflected in the measure's Footnotes tab or in the results list on the Find and Edit Measures page. 

## What
- Filters by measure sid to return all footnotes associations for the measure to fix footnotes not being populated on form
- Removes footnote associations when footnotes are removed from a measure to fix viewing errors when a footnote association is updated
- Adds unit test


measures/tests/test_views.py
```diff
@@ -553,6 +553,41 @@ def test_measure_update_updates_footnote_association(measure_form, client, valid
...
+        footnoted_measure=measure,
+    ).associated_footnote
+
+    with override_current_transaction(Transaction.objects.last()):
+        assert measure.footnoteassociationmeasure_set.current().count() == 2
+
+    form_data = {k: v for k, v in measure_form.data.items() if v is not None}
+    session = valid_user_client.session
+    # Remove footnote2 from session to remove on form
+    session.update({f"instance_footnotes_{measure.sid}": [footnote1.pk]})
```
paulpepper-trade, line 574: Is it be possible to remove `footnote2` from the that session dict's list and if so, then would it perhaps express clearer intent and state to do so? I don't know precisely what'd be in the session at this point, so I'm curious to better understand that!

paulpepper-trade (approved):

PR #977, opened 2023-07-19
TP2000-896 Create measure condition action automatically

a-gleeson:
# TP2000-896 Create measure condition action automatically

## Why
from a previous story we removed the ability to select "negative" actions and we automatically generate them instead. 
Out of this there is an emerging requirement that we need to create a negative action when a certain action is selected for a given measure type.

## What

- When you create a measure with type 112 Autonomous tariff suspension with a condition that has an action code 01 you will see an automatically generated negaitve code 07 in the workbasket summary. Like the current measure type 115 automatic condition generation implementation.
 
image

- Note currently you will not see this in the create conitions form or the review screen this is pending work to refactor measure conditions

### DIFF
--- a/measures/tests/test_views.py
+++ b/measures/tests/test_views.py
@@ -1127,7 +1127,7 @@ def test_measure_form_wizard_create_measures(
         positive_action=action2,
         negative_action=action2_pair,
     )
-    # action with no pair
+    # action with no pair and action code 1 (will not create with pair if not measure type 112)
     action3 = factories.MeasureActionFactory.create(code="01")
 
     form_data = {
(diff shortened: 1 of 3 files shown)


measures/tests/test_views.py
```diff
@@ -1353,6 +1353,121 @@ def test_measure_form_wizard_create_measures(
...
+            },
+        ],
+        "formset-footnotes": [],
+    }
+
+    wizard = MeasureCreateWizard(request=mock_request)
+
+    # Create measures returns a list of created measures
+    measure_data = wizard.create_measures(form_data)
+    measures = Measure.objects.filter(goods_nomenclature__in=[commodity1])
```
paulpepper-trade, line 1418: Would the following work just as well and perhaps be (ever so) slightly clearer?

`measures = Measure.objects.filter(goods_nomenclature=commodity1)`

paulpepper-trade (approved):

measures/tests/test_views.py
a-gleeson: yes! This was left over from a copy paste from another test that had a list of commodity

PR #978, opened 2023-07-19
TP2000-949  Improve measure creation performance

dalecannon:
# TP2000-949  Improve measure creation performance

## Why
Users are experiencing performance-related issues when creating measures with 50+ commodities and duties.
The issue appears to be caused by a combination of how the duty sentencer parser is being used and how the form wizard fetches cleaned data for a step:

1) A new duty sentence parser is loaded each time to validate a duties field in the formset

2) The same duty sentence may be parsed multiple times

3) A new duty sentence parser is loaded for each measure created

4) `get_cleaned_data_for_step()` revalidates a form before returning the cleaned data (exacerbating 1). This is used both in `get_form_kwargs()` and on the summary page to populate the template. It can become a recursive call if a form's kwargs depends on the cleaned data of another form, as happens with commodities step fetching measure details cleaned data

5) `get_all_cleaned_data()` revalidates all forms before returning the cleaned data (exacerbating 1)

## What
- Validates a set of duties at formset level so as to initialise duty parser once and avoid parsing the same duty sentence multiple times

- Uses `selected_related()` on measurements query in the duty parser to reduce database hits

- Initialises duty parser outside of `measures_data` for loop in `create_measures` since the start date is the same for all measures being created

- Overrides `get_cleaned_data_for_step()` to store a step's cleaned data to avoid revalidating a form during multiple calls in the same request

- Overrides `get_all_cleaned_data()` to use already stored cleaned data from summary page (commodities, condition, measure details) to avoid revalidating those forms

- Adds and updates unit tests

##
Before: 70 commodities with duties, ~33k queries, ~2.5 minutes
Screenshot 2023-07-19 at 09 18 24
#
After: same 70 commodities with duties, ~1k queries, ~20 seconds
Screenshot 2023-07-19 at 09 26 36

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+from itertools import groupby
 
 from crispy_forms_gds.helper import FormHelper
 from crispy_forms_gds.layout import HTML
@@ -1174,6 +1175,9 @@ class MeasureCommodityAndDutiesForm(forms.Form):
 
     def clean(self):
         cleaned_data = super().clean()
+        # Associate duties with their form so that the
+        # formset may later add form errors for invalid duties
+        cleaned_data["form_prefix"] = int(self.prefix.rsplit("-", 1)[1])
         duties = cleaned_data.get("duties", "")
         if duties and self.measure_type and self.measure_type.components_not_permitted:
             raise ValidationError(
@@ -1218,15 +1222,22 @@ class MeasureCommodityAndDutiesFormSet(MeasureCommodityAndDutiesBaseFormSet):
             return
 
         cleaned_data = super().cleaned_data
-        duties = {data["duties"] for data in cleaned_data if "duties" in data}
+        data = tuple((data["duties"], data["form_prefix"]) for data in cleaned_data)
+        # Filter tuples(duty, form) for unique duties to avoid parsing the same duty more than once
+        duties = [next(group) for duty, group in groupby(data, key=lambda x: x[0])]
+
         duty_sentence_parser = DutySentenceParser.get(
             self.measure_start_date,
         )
-        for duty in duties:
+        for duty, form in duties:
             try:
                 duty_sentence_parser.parse(duty)
-            except ParseError:
-                raise ValidationError(f'"{duty}" is an invalid duty sentence')
+            except ParseError as error:
+                error_index = int(error.loc().split(":", 1)[1])
+                self.forms[form].add_error(
+                    "duties",
+                    f'"{duty[error_index:]}" is an invalid duty expression',
+                )
 
         return cleaned_data
 
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -274,7 +274,7 @@ class MeasureConditionsFormMixin(forms.ModelForm):
                 )
 
         if price and not price_errored:
-            parser = DutySentenceParser.get(measure_start_date)
+            parser = DutySentenceParser.create(measure_start_date)
             components = parser.parse(price)
             if len(components) > 1:
                 self.add_error(
@@ -1226,7 +1226,7 @@ class MeasureCommodityAndDutiesFormSet(MeasureCommodityAndDutiesBaseFormSet):
         # Filter tuples(duty, form) for unique duties to avoid parsing the same duty more than once
         duties = [next(group) for duty, group in groupby(data, key=lambda x: x[0])]
 
-        duty_sentence_parser = DutySentenceParser.get(
+        duty_sentence_parser = DutySentenceParser.create(
             self.measure_start_date,
         )
         for duty, form in duties:
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -14,6 +14,7 @@ from crispy_forms_gds.layout import Submit
 from django import forms
 from django.core.exceptions import ValidationError
 from django.urls import reverse
+from parsec import ParseError
 
 from additional_codes.models import AdditionalCode
 from certificates.models import Certificate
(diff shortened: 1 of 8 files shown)


measures/forms.py
```diff
@@ -1215,6 +1213,23 @@ def non_form_errors(self):
...
+        cleaned_data = super().cleaned_data
+        duties = {data["duties"] for data in cleaned_data if "duties" in data}
+        duty_sentence_parser = DutySentenceParser.get(
+            self.measure_start_date,
+        )
+        for duty in duties:
+            try:
+                duty_sentence_parser.parse(duty)
+            except ParseError:
+                raise ValidationError(f'"{duty}" is an invalid duty sentence')
```
a-gleeson, line 1229: This will raise a formset level error not tied to a form or field. As this is a field specific error should we raise it as such?

measures/views.py
```diff
@@ -607,6 +608,28 @@ def done(self, form_list, **kwargs):
 
         return render(self.request, "measures/confirm-create-multiple.jinja", context)
 
+    def get_all_cleaned_data(self):
```
a-gleeson, line 611: comments to explain why we have this function?

measures/views.py
```diff
@@ -607,6 +608,28 @@ def done(self, form_list, **kwargs):
...
+                    {
+                        f"formset-{form_key}": cleaned_data,
+                    },
+                )
+            else:
+                all_cleaned_data.update(cleaned_data)
+        return all_cleaned_data
+
+    def get_cleaned_data_for_step(self, step):
+        self.cleaned_data = getattr(self, "cleaned_data", {})
```
a-gleeson, line 626: comments to explain why we have this function?

measures/views.py
a-gleeson: a comment could also clear up why we need `if isinstance(cleaned_data, (tuple, list))` for formset data or that it's for that

measures/views.py
```diff
@@ -619,25 +642,40 @@ def get_context_data(self, form, **kwargs):
...
+            min_commodity_count = 0
+            measure_details = self.get_cleaned_data_for_step(self.MEASURE_DETAILS)
+            if measure_details:
+                measure_start_date = measure_details.get("valid_between").lower
+                measure_type = measure_type = measure_details.get("measure_type")
+                min_commodity_count = measure_details.get("min_commodity_count")
+            kwargs.update(
+                {
+                    "min_commodity_count": min_commodity_count,
+                    "measure_start_date": measure_start_date,
```
a-gleeson, line 656: would this update ever overwrite values with the default values set above?

measures/views.py
```diff
@@ -619,25 +642,40 @@ def get_context_data(self, form, **kwargs):
...
+                "measure_start_date": measure_start_date,
                 "measure_type": measure_type,
             }
 
         if step == self.SUMMARY:
-            measure_type = self.get_cleaned_data_for_step(self.MEASURE_DETAILS).get(
-                "measure_type",
+            measure_details = self.get_cleaned_data_for_step(self.MEASURE_DETAILS)
+            measure_type = (
+                measure_details.get("measure_type") if measure_details else None
```
a-gleeson, line 678: if measure details is a dictionary you could do ` measure_details.get("measure_type", None)` to set a default value

measures/forms.py
```diff
@@ -1215,6 +1213,23 @@ def non_form_errors(self):
 
         return self._non_form_errors
 
+    def clean(self):
+        if any(self.errors):
+            return
+
+        cleaned_data = super().cleaned_data
+        duties = {data["duties"] for data in cleaned_data if "duties" in data}
```
a-gleeson, line 1221: You could maybe try `.get("duties",None)` but it would leave Nones in the data possibly and not give the same desired result.

measures/views.py
dalecannon: The idea was to guard against Sentry errors triggered by `AttributeError` if `measure_details` is `None`, say, when a user skips ahead somehow or presses back in the browser after completing a journey.

measures/forms.py
```diff
@@ -1215,6 +1213,23 @@ def non_form_errors(self):
...
+        duty_sentence_parser = DutySentenceParser.get(
```
paulpepper-trade, line 1222: I know it hasn't been introduced in this PR, but `DutySentenceParser.get()` doesn't convey the creational behaviour. Maybe worth a naming refactor to `DutySentenceParser.create()` create?

measures/views.py
```diff
@@ -619,25 +658,40 @@ def get_context_data(self, form, **kwargs):
...
+                measure_start_date = measure_details.get("valid_between").lower
+                measure_type = measure_type = measure_details.get("measure_type")
+                min_commodity_count = measure_details.get("min_commodity_count")
+            kwargs.update(
+                {
+                    "min_commodity_count": min_commodity_count,
+                    "measure_start_date": measure_start_date,
+                },
+            )
+            kwargs["form_kwargs"] = {
```
paulpepper-trade, line 675: What's the distinction between elements that are direct members of `kwargs` and those of `kwargs["form_kwargs"]`. It looks a little unconventional, so a comment, perhaps in the function's docstring, would be useful.

paulpepper-trade (approved):
Still needs Anthoni's comments resolving, but otherwise, LGTM.

measures/views.py
dalecannon: The default values should only be used in the unlikely event `measure_details` is `None` to prevent any AttributeErrors which would occur if this step is reached before the first form is completed.

measures/views.py
a-gleeson: that's cool I missed it was actually checking measure_details is None

a-gleeson (approved):

PR #979, opened 2023-07-19
TP2000-962 Show report in goods review tab

paulpepper-trade:
# TP2000-962 Show report in goods review tab

## Why

Tariff managers would like to see the same content in the goods review tab as they see in an Excel goods report.

## What

Extract goods review content from GoodsReport source as performed for Excel reports and display the content on the workbasket's Goods Review tab.

## Checklist
- Requires migrations? No

### DIFF
--- a/workbaskets/views/ui.py
+++ b/workbaskets/views/ui.py
@@ -7,6 +7,7 @@ from celery.result import AsyncResult
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.mixins import PermissionRequiredMixin
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.paginator import Paginator
 from django.db.models import ProtectedError
 from django.db.transaction import atomic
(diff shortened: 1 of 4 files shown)


workbaskets/views/ui.py
```diff
@@ -297,95 +294,49 @@ def get_queryset(self):
...
+                [
+                    line.update_type.title(),
+                    line.record_name.title(),
+                    line.goods_nomenclature_item_id,
+                    line.suffix,
+                    line.validity_start_date,
+                    line.validity_end_date,
+                ]
+                for line in goods_report.report_lines
+            ]
```
dalecannon, line 338: Could the first setting of `context["report_lines"]` be removed since it's superseded right after?

dalecannon (approved):

workbaskets/views/ui.py
paulpepper-trade: Done! It was a hangover from my previous implementation approach.

PR #985, opened 2023-07-25
TP2000-970  Fix editing newly-created measures

dalecannon:
# TP2000-970  Fix editing newly-created measures

## Why
Follow-up fix to #984. `MeasureEditCreate` view was added to enable editing newly-created measures. However, the view inherits `EditTaricView` which changes an existing model instance without creating a new version (like a real edit). But the measure update form creates new versions (update records) of any measure components regardless. The outcome is that components aren't being edited in the same way as the measure itself, causing business rule violations about updates.

## What
- Inherit`CreateTaricUpdateView` instead of `EditTaricView` in edit views so UPDATE instances are created to prevent business rule violations

##
Before: business rule violations following an edit to duties on a newly-created measure:
Screenshot 2023-07-27 at 11 56 58
Screenshot 2023-07-27 at 11 57 34

After: update records following an edit to duties on newly-created measure:
Screenshot 2023-07-27 at 11 59 42

### DIFF
--- a/measures/tests/test_views.py
+++ b/measures/tests/test_views.py
@@ -1004,7 +1004,7 @@ def test_measure_edit_update_view(valid_user_client, erga_omnes):
     assert updated_measure.geographical_area == geo_area
 
 
-def test_measure_edit_create_view(valid_user_client, erga_omnes):
+def test_measure_edit_create_view(valid_user_client, duty_sentence_parser, erga_omnes):
     """Test that a measure CREATE instance can be edited."""
     measure = factories.MeasureFactory.create(
         update_type=UpdateType.CREATE,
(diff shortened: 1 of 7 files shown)


measures/tests/test_views.py
```diff
@@ -999,12 +999,12 @@ def test_measure_edit_update_view(valid_user_client, erga_omnes):
     response = valid_user_client.post(url, data=data)
     assert response.status_code == 302
 
-    measure.refresh_from_db()
-    assert measure.update_type == UpdateType.UPDATE
-    assert measure.geographical_area == geo_area
+    updated_measure = Measure.objects.filter(sid=measure.sid).last()
```
paulpepper-trade, line 1002: Would it be preferable to use `TrackedModelQuerySet.current()`, or related functionality, as is the convention in non-test code?

paulpepper-trade (approved):

PR #986, opened 2023-07-25
TP2000 926 Enforce a maximum of one active import per service instance

a-gleeson:
# TP2000-926 Enforce a maximum of one active import per service instance

## Why
The current parser capability relies on a shared caching resource (referred to as the nursery in the TAP codebase), which is not name-spaced between import processor. This is an issue when imports run in parallel, causing invalid objects to appear in an importer’s cache (nursery).

## What
- A seperate worker for importing has been set up to process importing tasks
- worker set up with a concurrency of 1 so tasks are only processed one at a time in the order they are recieved
- local celery worker set up

### DIFF
--- a/common/celery.py
+++ b/common/celery.py
@@ -20,3 +20,9 @@ app.autodiscover_tasks()
 # this should be automactically configured via ^^ config_from_object
 # but it isn't so here it's configured here
 app.conf.task_routes = settings.CELERY_ROUTES
+
+# turn on late ack so messages are only consumed from the broker on success
+app.conf.update(
+    task_acks_late=True,
+    broker_transport_options={"confirm_publish": True},
+)
--- a/common/celery.py
+++ b/common/celery.py
@@ -20,9 +20,3 @@ app.autodiscover_tasks()
 # this should be automactically configured via ^^ config_from_object
 # but it isn't so here it's configured here
 app.conf.task_routes = settings.CELERY_ROUTES
-
-# turn on late ack so messages are only consumed from the broker on success
-app.conf.update(
-    task_acks_late=True,
-    broker_transport_options={"confirm_publish": True},
-)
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -50,7 +50,7 @@ services:
     build:
       context: .
       args:
-        - "ENV=${ENV:-prod}"
+        - "ENV=${ENV:-prod}"
```
paulpepper-trade, line 53: Some sneaky white space crept in at the end of this line. I think there's no change to this file if it gets deleted.

common/celery.py
```diff
@@ -20,3 +20,9 @@
 # this should be automactically configured via ^^ config_from_object
 # but it isn't so here it's configured here
 app.conf.task_routes = settings.CELERY_ROUTES
+
+# turn on late ack so messages are only consumed from the broker on success
+app.conf.update(
+    task_acks_late=True,
+    broker_transport_options={"confirm_publish": True},
+)
```
paulpepper-trade, line 28: There may be some risk in applying retries this across all of our tasks. Could we roll this out to tasks gradually to allow us to observe and manage that risk? Maybe the importer's `importer.tasks.import_chunk()` and, say, the rule checker's `workbasket.tasks.call_check_workbasket_sync()` would be good, low-ish risk initial trial tasks?

docker-compose.yml
```diff
@@ -98,7 +113,7 @@ services:
       context: .
       args:
         - "ENV=${ENV:-prod}"
-    image: tamato
+    image: tamato
```
paulpepper-trade, line 116: Trailing white space.

paulpepper-trade (changes requested):

common/celery.py
a-gleeson: Hi, I will make these changes as it could be a risk putting it accross the board. I'll need to update the configuration a bit but shouldn't be an issue. This isn't exactly retry but instead more will replay the task it was in the process of processing.

paulpepper-trade (approved):
LGTM

PR #987, opened 2023-07-26
TP2000-252  Add ME88 validation to commodities form in create measure journey

dalecannon:
# TP2000-252  Add ME88 validation to commodities form in create measure journey

## Why
The commodities form in the create measure journey doesn't take into account the explosion level of the measure type.

## What
- Validates in the form's clean method whether the selected commodity code exceeds the explosion level of the selected measure type
- Adds unit test

##
Screenshot 2023-07-26 at 09 11 12

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1199,6 +1199,18 @@ class MeasureCommodityAndDutiesForm(forms.Form):
                 },
             )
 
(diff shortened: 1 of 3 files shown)


paulpepper-trade (approved):
LGTM

PR #992, opened 2023-07-31
TP2000-985 Fix failing geo area form test

dalecannon:
## Why
As of today there is a failing unit test: `FAILED geo_areas/tests/test_forms.py::test_geographical_membership_edit_form_invalid_end_date`

The test uses whatever is the current day’s date to set not only the validity period of the area group but also the end dates in the form. When the date is the 31st, the test fails because it attempts to add 1 to it, resulting in an invalid date of 32nd.

## What
- Use `date_range.earlier` and `date_range.later` fixture to avoid adding/subtracting whatever is the current date


paulpepper-trade (approved):
LGTM

PR #993, opened 2023-08-02
TP2000-987 - Add '0 goods items' support on importer list view

paulpepper-trade:
# TP2000-987 - Add '0 goods items' support on importer list view

## Why

When successfully importing a TARIC3 file containing goods-related elements, the import status shows SUCCEEDED, and the goods status column provides a navigation link to the related workbasket into which the goods objects have been saved.

However, when successfully importing TARIC file containing no goods-related elements, the import status correctly shows SUCCEEDED, but the goods status column is empty. This can leave some uncertainty about the status of the import.

## What

This PR:

* Factors out display decision logic into a view function.
* Populates the goods status column with "0 goods items" if an TARIC file is successfully imported but contains no goods-related elements.
* Adds unit tests to validate that correct goods status column contents are rendered on the goods importer list view.

image

## Checklist
- Requires migrations? No

### DIFF
--- a/importer/views.py
+++ b/importer/views.py
@@ -104,7 +104,37 @@ class CommodityImportListView(
         elif import_status == ImportBatchStatus.FAILED and workbasket_status == None:
             context["selected_link"] = "errored"
 
+        context["goods_status"] = self.goods_status
+
         return context
+
+    @classmethod
+    def goods_status(cls, import_batch: ImportBatchFilter) -> str:
+        """Returns the goods status of an ImportBatch instance, used to
+        determine rendered output of the goods status column in this list view's
+        template."""
+        workbasket = import_batch.workbasket
+
+        if (
+            import_batch.status == ImportBatchStatus.SUCCEEDED
+            and workbasket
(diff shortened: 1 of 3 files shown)


importer/views.py
```diff
@@ -104,8 +104,38 @@ def get_context_data(self, **kwargs):
...
+            and workbasket
+            and workbasket.status == WorkflowStatus.EDITING
+        ):
+            return "editable_goods"
+        elif (
+            import_batch.status == ImportBatchStatus.SUCCEEDED
+            and import_batch.taric_file
+            and (
+                not workbasket
+                or (workbasket and workbasket.status == WorkflowStatus.ARCHIVED)
```
a-gleeson, line 129: do we need to check workbasket here as this part of the OR should only check if `not workbasket` fails which means workbasket exists?

a-gleeson (approved):
LGTM!

importer/views.py
paulpepper-trade: Simplified!

PR #995, opened 2023-08-03
TP-2000-875 - Implement the improved Find and Edit measures design - Search page

CPrich905:
# TP-2000-875 - Implement the improved Find and Edit measures design - Search page
[Ticket](https://uktrade.atlassian.net/browse/TP2000-958)

## Why
User Research and Design ha been completed. This ticket is implementing those changes on the Measures Search page.

## What
Implementing the User Research & design of the attached ticket, in summary:
1. Reorganised the search and filter options, implemented some small wording changes to match the figma design on the ticket.
2. Adding a new filter section to contain `Include inherited measures for specific commodity code` and two new filters on a radio button: `Show active measures only` and `Only show measures in this current workbasket`
3. Disable the Include inherited measures for a specific commodity code if there is no commodity code.
4. Styling changes

Additionally:
- removed `filter-layout__filters` css styles following discussion with Edie Pearce.
- refactored `kwargs["data"]._mutable = True` to `data = kwargs["data"].copy()` in MeasureFilter

Search filters page, because `Specific commodity code` is empty, `Include inherited measures for specific commodity code` is disabled
Screenshot 2023-08-18 at 13 56 08

`Specific commodity code` has a valid commodity code, `Include inherited measures for specific commodity code` is enabled.
Screenshot 2023-08-18 at 13 56 35


common/jinja2/layouts/list.jinja
```diff
@@ -19,15 +19,15 @@
     {% endif %}
   </p>
 
-  <div class="filter-layout">
-    <div class="filter-layout__filters">
+  <div>
```
a-gleeson, line 22: do we require this extra div?

measures/filters.py
```diff
@@ -50,12 +59,12 @@ class MeasureTypeFilterBackend(TamatoFilterBackend):
...
-            if "start_date_modifier" not in kwargs["data"]:
-                kwargs["data"]["start_date_modifier"] = "exact"
-            if "end_date_modifier" not in kwargs["data"]:
-                kwargs["data"]["end_date_modifier"] = "exact"
-            kwargs["data"]._mutable = False
+            data = kwargs["data"].copy()
+            if "start_date_modifier" not in data:
+                data["start_date_modifier"] = "exact"
+            if "end_date_modifier" not in data:
+                data["end_date_modifier"] = "exact"
```
a-gleeson, line 66: might be monday brain but if we have changed to using `data` would we need to update `kwargs["data"]` at the end as we are still passing `kwargs` in the `super()` function and not passing in `data`.

common/jinja2/layouts/list.jinja
CPrich905: I'd left the extra div in as a wrapper for the header section, but technically no, we don't require it. I will remove.

measures/filters.py
```diff
@@ -86,17 +95,24 @@ def __init__(self, *args, **kwargs):
...
         method="commodity_modifier",
     )
 
+    measure_filters_modifier = ChoiceFilter(
+        label="",
+        widget=forms.RadioSelect,
+        method="measures_filter",
+        empty_label=None,
+        choices=ACTIVE_CURRENT_CHOICES,
+    )
```
a-gleeson, line 112: does this mean when searching we will not be able to search for non-active measures as one of these choices will have to be selected?

measures/filters.py
```diff
@@ -181,6 +197,15 @@ def __init__(self, *args, **kwargs):
         method="filter_end_date",
     )
 
+    certificates = AutoCompleteFilter(
+        label="Certificates",
+        field_name="certificates",
+        queryset=Certificate.objects.all(),
```
a-gleeson, line 203: I'd have to confirm which option but should we maybe be pulling down certificates with `.current`, `latest_approved` or `approved_up_to_transaction` instead of all? 
I would have to think but maybe we would want to pull all of the certificates but the current/ approved ones?

measures/filters.py
```diff
@@ -232,6 +257,37 @@ def filter_end_date(self, queryset, name, value):
...
+        if value:
+            modifier = self.data["measure_filters_modifier"]
+            if modifier == "active":
+                current_date = date.today()
+
+                filter_query = (
+                    Q(
+                        start_date__lt=current_date,
+                    )
+                    | Q(start_date=current_date)
```
a-gleeson, line 270: I think you can use `__lte `here as a single query

measures/filters.py
CPrich905: If I've understood the docs correctly then we shouldn't need to ([link to docs](https://docs.python.org/3/library/copy.html)), but also passing `data` in `super()` will throw an error as it's not always referenced. 
If I have misunderstood the docs, please let me know, likewise if you think it would be clearer/more robust to use `data = kwargs["data"].deepcopy()` instead

measures/filters.py
```diff
@@ -232,6 +257,37 @@ def filter_end_date(self, queryset, name, value):
...
+                    queryset.with_effective_valid_between()
+                    .annotate(start_date=StartDate("valid_between"))
+                    .annotate(end_date=EndDate("valid_between"))
+                    .filter(filter_query)
+                )
+
+            if modifier == "current":
+                queryset = WorkBasket.objects.filter(
+                    id=self.request.session["workbasket"].id,
+                )[0].measures
```
a-gleeson, line 287: You could use `Workbasket.current(self.request)` here.
Although I would maybe test if your current work basket was deleted in the background (another user/ other tab), just to validate there aren't any exceptions

measures/static/measures/js/conditionalDisablingFilters.js
```diff
@@ -0,0 +1,21 @@
+// conditionally disables modc filter based on goods_nomenclature_autocomplete filter having a value
+// TODO: refactor to allow passing two widget ids: 1. widget condition is based on 2. widget to be disabled.
+
+const setFilterDisabledToggle = () => {
```
a-gleeson, line 4: would having this as `initFilterDisabledToggleForComCode` be more suitable here as it inits the filter and it is commodity specific?

measures/filters.py
CPrich905: No, these options can be left empty which would also return non-active measures. While it wasn't a requirement for this ticket, I will ask Steve if a future update should include an option to filter for non-active measures.

measures/filters.py
CPrich905: Updated this, thanks!

measures/filters.py
CPrich905: Interesting - this throws a TypeError for me, though seems to work fine in other places throughout the codebase. I'll have a further poke about and see if I can make it work, as that does look a lot nicer.

measures/static/measures/js/conditionalDisablingFilters.js
CPrich905: Updated

measures/filters.py
a-gleeson: Would you be expecting `kwargs["data"]` to be updated with the new field?
So running this through my head as it is a shallow copy `kwargs["data"]` would not be updated here, also `data = kwargs["data"].deepcopy()` wouldn't either. This is because a new dictionary is created for `data` which has references to the existing existing `kwargs["data"]` fields, so **updating** a field will update the field `kwargs["data"]`. Although as `data["start_date_modifier"] = "exact"` would be **adding** a new field this wouldn't reference the original fields within `kwargs["data"]` and thus not update the original object.

measures/filters.py
a-gleeson: I have run by Steve with it and he says it's fine to return all Certificates. I think `.current()` is most suitable here you can check the definition at `TrackedModelQuerySet.current()` and see what you think

measures/filters.py
a-gleeson: Ah good to know that it doesn't select any of the values by default! 
The documentation wasn't the clearest when I was trying to check and I was being lazy not pulling the branch down locally to check myself.

measures/filters.py
CPrich905: Happy to take your steer & I've updated to `.current()`

measures/filters.py
CPrich905: Yep! And response from Steve is to hold off on being able to filter by inactive measures for now & consider it as a future feature

measures/filters.py
a-gleeson: Just to update on the why, `.all()` would bring back all certificate versions so for a single certificate it would return previous versions as well. We wouldn't want that, so `.current()` returns the most recent versions

measures/filters.py
CPrich905: OK, I've updated this to make a copy of `data`, added in the values and then re-adding the new data back in to `**kwargs`. This seems to work for me, playing around with `.pop()` had some issues for me. If you think it would be better approach, would you be up for a quick pair through to see if our combined brains can get it going?

measures/filters.py
CPrich905: Resolved through correcting test set up

measures/filters.py
a-gleeson: I am happy with that!

a-gleeson (approved):
LGTM!

PR #996, opened 2023-08-07
TP2000-984 - Delete workbasket and handle broken session

paulpepper-trade:
# TP2000-984 & TP2000-992 - Delete workbasket and handle broken session

## Why

Tariff managers must request that DDaT delete workbaskets as there's currently no provision to do this within the main UI - superusers may access the Django admin site to delete workbaskets.

If a user has a workbasket assigned as current and that workbasket is deleted, then access to the application becomes unavailable, displaying a 500-series response. In these circumstances a user must clear their cookies in order to clear out the "corrupt" current workbasket assignment.

## What

This PR:

* Provides UI capability to delete workbaskets, restricting to users with the `workbaskets.delete_workbasket` permission.
* Makes access to the workbasket delete views routable via the workbasket's PK, rather than the "current" (session-based) workbasket. This is to minimise the possibility for mishap for such a destructive action - for instance deleting a workbasket when, say, current workbasket has been changed in another tab.
* Implements middleware to validate a user's current workbasket assignment within their session. If the current workbasket is invalid (for instance, has been deleted or is no longer in EDITING status), then the workbasket is cleared from the session.

TODO: unit tests

"Delete workbasket" button, only visible to a user with the `workbaskets.delete_workbasket` permission and when a workbasket is empty (no tracked models items).
image

"Confirm workbasket deletion" page:
image

Workbasket validation failure (items may have been added to the workbasket):
image

"Done workbasket deletion" page:
image

## Checklist
- Requires migrations? No


workbaskets/jinja2/workbaskets/delete_workbasket.jinja
```diff
@@ -0,0 +1,79 @@
...
+<div class="govuk-breadcrumbs">
+  <ol class="govuk-breadcrumbs__list">
+    <li class="govuk-breadcrumbs__list-item">
+      <a class="govuk-breadcrumbs__link" href="{{ url('home') }}">Home</a>
+    </li>
+    <li class="govuk-breadcrumbs__list-item">
+      {{ page_title }}
+    </li>
+  </ol>
+</div>
```
a-gleeson, line 20: Should we use the breadcrumb macro here?

workbaskets/jinja2/workbaskets/delete_workbasket_done.jinja
```diff
@@ -0,0 +1,47 @@
...
+
+    <h2 class="govuk-heading-m">Next steps</h2>
+    <p class="govuk-body">
+      To manage another workbasket select one from the workbasket list.
+    </p>
+
+    {{ govukButton({
+      "text": "Go to workbasket list",
+      "href": url("workbaskets:workbasket-ui-list"),
+      "classes": "govuk-button--secondary"
```
a-gleeson, line 39: AS this is the main action (only) should we be using the primary button?

workbaskets/jinja2/workbaskets/delete_workbasket_done.jinja
```diff
@@ -0,0 +1,47 @@
...
+<div class="govuk-breadcrumbs">
+  <ol class="govuk-breadcrumbs__list">
+    <li class="govuk-breadcrumbs__list-item">
+      <a class="govuk-breadcrumbs__link" href="{{ url('home') }}">Home</a>
+    </li>
+    <li class="govuk-breadcrumbs__list-item">
+      {{ page_title }}
+    </li>
+  </ol>
+</div>
```
a-gleeson, line 19: dreadcrumb marco?

a-gleeson (approved):
LGTM! Left a few minor comments

workbaskets/jinja2/workbaskets/delete_workbasket.jinja
paulpepper-trade: The breadcrumb macro puts the "current" workbasket in the trail, but the view associated with this template isn't relative to current - this is for safety reasons, but also because this view is likely to become accessible from other locations that are themselves not related to current.

workbaskets/jinja2/workbaskets/delete_workbasket_done.jinja
paulpepper-trade: See [previous comment](https://github.com/uktrade/tamato/pull/996#discussion_r1296040384).

workbaskets/jinja2/workbaskets/delete_workbasket_done.jinja
paulpepper-trade: Actually, our other confirmation views use a secondary button! There's a UI consistency review going on just now so I'll flag this as a question.

a-gleeson: Thanks, yeah understanding the consitent approach will be good! The General feedback for myself has been the main action for a user to take on a screen should be the primary. So understanding the scenario's should be good!

PR #999, opened 2023-08-09
TP2000-994 - Preserve user's goods-only import choice

paulpepper-trade:
# TP2000-994 - Preserve user's goods-only import choice

## Why

No distinction between goods-only and general import is captured on the `ImportBatch` model. This leads to general, non-goods imports, normally performed by data engineers, appearing on the goods-only import list view.

## What

This PR:
* Introduces a new `BooleanField` model attribute, `ImportBatch.goods_import`.
* Set the new `ImportBatch.goods_import` according to which type of import is being performed - `False` when performed from the (admin only) master importer, `True` when performed from the goods importer.
* Update unit tests to validate `ImportBatch.goods_import` attribute correctness upon `Form.save()`.
* Remove the "Commodities only" checkbox from the master importer's create view to more clearly convey intent - i.e. it is not intended for goods import use, as that should be performed via the goods importer.
* Add help text to the master importer fields.

## Checklist
- Requires migrations? Yes


importer/forms.py
```diff
@@ -109,21 +109,47 @@ class Meta:
...
 
-    status = forms.ChoiceField(choices=WorkflowStatus.choices, required=True)
-    taric_file = forms.FileField(required=True)
-    commodities = forms.BooleanField(
-        label="Commodities Only",
-        required=False,
-        initial=False,
+    status = forms.ChoiceField(
+        choices=WorkflowStatus.choices,
+        # choices=WorkflowStatus.unchecked_statuses(),
```
a-gleeson, line 114: leftover comment?

a-gleeson (approved):
LGTM

importer/forms.py
paulpepper-trade: Updated to correctly filter the choices.

a-gleeson (approved):
LGTM!

a-gleeson (approved):

a-gleeson (approved):

PR #1000, opened 2023-08-09
TP2000-412  Edit draft regulations

dalecannon:
# TP2000-412  Edit draft regulations

## Why
Users are unable to edit certain fields on draft regulations once created. Also, a regulation's status can be set to approved without providing its public identifier, absolute url of the legislation and title.

## What
- Makes `regulation_usage`, `published_at` and `sequence_number` editable for draft regulations TP2000-412
- Creates a new regulation along with updated associated measures and deletes old one if the `regulation_id` has been changed TP2000-981
- Validates whether the `approved` status corresponds to the regulation's `regulation_usage`
- Makes `public_identifier`, `url` and `information_text` fields mandatory if regulation status is set to `approved` TP2000-209
- Adds and updates form and view tests

##
Example workbasket summary of transaction order of a twice-updated regulation ID:
Screenshot 2023-08-11 at 09 38 31


regulations/views.py
```diff
@@ -98,30 +121,46 @@ def get_queryset(self):
...
 ):
     """UI to create regulation UPDATE instances."""
 
+    def get_result_object(self, form):
+        if form.instance.regulation_id == form.cleaned_data["regulation_id"]:
+            return super().get_result_object(form)
+
+        model_fields = [f.name for f in self.model._meta.get_fields()]
+        form_changed_data = [f for f in form.changed_data if f in model_fields]
+        # Regulation_id is not a field on the form and so needs to be added to form.changed_data
```
eadpearce, line 136: regulation_id could be added to the form as a pre-populated hidden field

eadpearce: ah nvm. I just saw the _get_regulation_id method

eadpearce (approved):

PR #1004, opened 2023-08-14
fix: correct ON10 business rule to use effective_valid_between instead of valid_between

nboyse:
# TOPS-1108 - Correct ON10 business rule to use effective_valid_between instead of valid_between

## Why
The business rule for ON10 wasn't quite implemented correctly, it was not using the effective_valid_between method which checks the date range against the related models.

This PR also moves the current ON10 unit tests into a separate file from the other business rules' unit tests for clarity.

## What
Checks measures against the effective_valid_between property instead of valid_between. This prevents users from receiving a business rule violation when editing their workbasket.

We are moving ON10 unit tests into a separate file to make it easier to update and track where they are.


measures/business_rules.py
```diff
@@ -528,7 +528,7 @@ def validate(self, measure):
                 valid_between = origin.valid_between
                 if validity_range_contains_range(
                     valid_between,
-                    contained_measure.valid_between,
+                    contained_measure.effective_valid_between,
```
paulpepper-trade, line 531: Oh wow.

PR #1005, opened 2023-08-14
TP2000-982 Add new filters to find and edit regulation view

dalecannon:
# TP2000-982 Add new filters to find and edit regulation view

## Why
When using the find and edit regulation view to search for regulations, users are unable to filter search results by regulation usage, regulation group or approved status. 

## What
- Enables filtering by usage, group and approved status
- Adds unit test for new filters

##
Screenshot 2023-08-14 at 15 15 00

Screenshot 2023-08-14 at 15 15 09

Screenshot 2023-08-14 at 15 15 17


regulations/filters.py
```diff
@@ -25,8 +31,53 @@ class RegulationFilter(
...
+
+    clear_url = reverse_lazy("regulation-ui-list")
+
+    APPROVED_STATUS_CHOICES = (
+        (True, "Approved"),
+        (False, "Not approved"),
+    )
+
+    regulation_usage = LazyMultipleChoiceFilter(
+        choices=RegulationUsage.choices[1:],
```
paulpepper-trade, line 53: It would probably be clearer and less prone to error if the `__empty__` was prepended to the choices where necessary, rather than slicing it off here.

paulpepper-trade: ... alternatively create a base class for `RegulationUsage` that has all except for `__empty__` as its available choices.

paulpepper-trade (approved):

regulations/filters.py
dalecannon: I couldn't see a way to create and extend a base class given `RegulationUsage` is an enum behind it all. I opted to remove`__empty__` and prepend it to the choices where it's used by the field's dropdown in the regulation create form.

PR #1006, opened 2023-08-15
TP2000-990: Footnote associations delete

eadpearce:
# TP2000-990: Footnote associations delete

## Why

* Users need a way to delete commodity code footnote associations

## What

* Adds a delete link to the footnotes on the commodity detail page 

Screenshot 2023-08-15 at 13 20 51
Screenshot 2023-08-15 at 13 20 38
Screenshot 2023-08-15 at 13 24 59


commodities/urls.py
```diff
@@ -94,4 +94,14 @@
...
     ),
+    path(
+        f"{footnote_association_pattern}/delete/",
+        views.FootnoteAssociationGoodsNomenclatureDelete.as_view(),
+        name="footnote_association_goods_nomenclature-ui-delete",
+    ),
+    path(
+        f"commodities/<sid>/footnote-associations/confirm-delete/",
+        views.FootnoteAssociationGoodsNomenclatureConfirmDelete.as_view(),
+        name="footnote_association_goods_nomenclature-ui-confirm-delete",
```
a-gleeson, line 105: do you think there would be any need in passing in the deleted id, just to display the Id of the association deleted? May be overkill

a-gleeson (approved):
LGTM!

commodities/urls.py
eadpearce: So we can't do that in the url because then it would return a 404 (if we're using the TrackedModelDetailView). Instead I used django messages to pass a message to this view saying which object has been deleted 
https://github.com/uktrade/tamato/pull/1006/files#diff-03f665fbf23f873831b625edeee6b677f2a2ef9f0a65aa00af4cebda8cd28ba4R318

a-gleeson: Ah right it's because you are using TrackedModelDetailView, sounds fair to me :)

PR #1008, opened 2023-08-16
TP2000-991: Commodity footnotes search filter

eadpearce:
# TP2000-991: Commodity footnotes search filter

## Why

* Users need to be able to easily find commodity codes that have 1 or more footnotes

## What

* Adds a checkbox to the commodity search filters to only include commodity codes with 1 or more footnotes
* Also simplifies the templates for the commodity search page

Screenshot 2023-08-16 at 13 50 22


commodities/jinja2/includes/commodities/table.jinja
```diff
@@ -3,13 +3,19 @@
 {% set table_rows = [] %}
 {% for commodity in object_list|sort(attribute='item_id') %}
     {% set commodity_link -%}
-        <a class="govuk-link govuk-!-font-weight-bold" href="{{ url("commodity-ui-detail", kwargs={"sid": commodity.sid}) }}">{{ commodity.code }}</a>
+        <a class="govuk-link govuk-!-font-weight-bold" href="{{ url('commodity-ui-detail', kwargs={'sid': commodity.sid}) }}">{{ commodity.code }}</a>
+    {%- endset %}
+    {% set commodity_footnotes -%}
+    {% for association in commodity.footnote_associations.current() %}
+        <a class="govuk-link" href="{{ association.associated_footnote.get_url() }}">{{ association.associated_footnote }}</a>{% if not loop.last %}, {% endif %}
```
a-gleeson, line 10: How does column width affect how the associations list, are we happy with them on one line or would we want a break between each link?

commodities/tests/test_filters.py
```diff
@@ -0,0 +1,27 @@
+from commodities.filters import CommodityFilter
+from commodities.models.orm import GoodsNomenclature
+from common.tests import factories
+
+
```
a-gleeson, line 5: missing `pytestmark = pytest.mark.django_db` causing test to fail

commodities/jinja2/includes/commodities/table.jinja
eadpearce: Most of the time the description column forces the footnotes column to be quite narrow so they would naturally display on multiple lines even if we don't put line breaks in here

commodities/tests/test_filters.py
eadpearce: Oh good spot thanks

a-gleeson (approved):
LGTM!

PR #1009, opened 2023-08-16
TP2000 627 workbasket summary business rules text

mattjamc:
# TP2000-627 workbasket summary business rules text

## Why

* This ticket makes the live status text clearer for workbasket business rule checks following Patrick's recommendations. 

## What

* Updated the text for when a business rule check passes, fails or has not been run
* Added in a test that checks that success banner appears following a successful check or does not appear otherwise

### DIFF
--- a/workbaskets/tests/test_views.py
+++ b/workbaskets/tests/test_views.py
@@ -609,7 +609,7 @@ def test_workbasket_business_rule_status(valid_user_client):
 
     url = reverse("workbaskets:current-workbasket")
     response = valid_user_client.get(url)
-    page = BeautifulSoup(str(response.content), "html.parser")
+    page = BeautifulSoup(response.content.decode(response.charset))
     success_banner = page.find(
         "div",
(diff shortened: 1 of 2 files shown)


workbaskets/tests/test_views.py
```diff
@@ -589,6 +589,41 @@ def test_run_business_rules(check_workbasket, valid_user_client, session_workbas
...
+        TrackedModelCheckFactory.create(
+            transaction_check__transaction=transaction,
+            model=footnote,
+            successful=True,
+        )
+    workbasket.save_to_session(valid_user_client.session)
+
+    url = reverse("workbaskets:current-workbasket")
+    response = valid_user_client.get(url)
+    page = BeautifulSoup(str(response.content), "html.parser")
```
eadpearce, line 612: Better to use `response.content.decode(response.charset)` here instead of `str(response.content)`

eadpearce (approved):

PR #1011, opened 2023-08-18
TP2000-1008 Fix geo membership creation

dalecannon:
# TP2000-1008 Fix geo membership creation

## Why
Creating a new geographical membership via the ‘Create a new membership' link on the 'Memberships’ tab of a geographical area detail view does not show the membership association creation in the workbasket summary if the current workbasket has no previous transactions. It erroneously uses the workbasket's last transaction or the most recent global transaction.

## What
- Uses a new transaction in the current workbasket for each membership being created

### DIFF
--- a/geo_areas/views.py
+++ b/geo_areas/views.py
@@ -322,10 +322,9 @@ class GeographicalMembershipsCreate(
         members = [d["member"] for d in form.cleaned_data if "member" in d]
         geo_groups = [d["geo_group"] for d in form.cleaned_data if "geo_group" in d]
 
-        tx = self.get_transaction()
-
         if members:
             for member, valid_between in zip(members, validity_periods):
+                tx = self.get_transaction()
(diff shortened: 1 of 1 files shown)


geo_areas/views.py
```diff
@@ -166,7 +166,7 @@ def create_membership(self, form):
         else:
             return
 
-        tx = WorkBasket.get_current_transaction(self.request)
+        tx = self.workbasket.new_transaction()
```
a-gleeson, line 169: Should we have a comment to explain why?

a-gleeson (approved):
LGTM!

geo_areas/views.py
CPrich905: More a clarification than a question, but I'm assuming that creating a new transaction on the workbasket, if one does already exist, doesn't have a meaningful impact on this process?

PR #1015, opened 2023-08-22
Tp2000 819 importer refactor

dougmills-DIT:
# TP2000-667 - Importer refactor / rewrite

## Why
Ultimately this work is required to fully hand over the Comm code automation piece to tariff managers without DE massaging every 5th import, plus the current importer is buggy, inflexible and has a cache that leaks.  
* The caching implementation has been problematic, polluting future imports and failing repeat imports. 
* Codebase is hard to manage, extend and debug, or even read. 
* Child objects could be handled better
* Test coverage is poor, and hard to improve.
* No facility to report on import failures in the UI.
* Some objects are silently discarded (GoodsNomenclatureSuccessor) 

## What
The importer v2 addresses issues with the importer v1. Initially importer v2 will be added as an additional option within the footer menu, under "New TARIC parser" allowing the existing process to proceed as it did before, but allow the importer v2 to be employed. This will be for a time until confidence grows and its agreed that the importer v2 should replace importer v2 for comm code import capability.
* Removed the use of caching
** The implementation of caching was the cause if multiple issues.
** Cached data was not purged, and future imports could include previously cached data in certain circumstances. 
** The need for caching was questionable to begin with, since a TARIC file can and should be processed in a linear way, each successive change should be built on the previous changes. The need for caching may have been due to the merging of certain data types (description and period for example) and the caching was needed when a change was presented to a child object that does not have a table to attach to until the parent object is created. This problem is addressed with the  importer v2 by pre-processing child / parent objects and appending properties to the parent before commiting to the database. The only exception is when a child is explicitly deleted, without the subsequent deletion of the parent. This will raise a warning from the importer, but the whole import can proceed regardless, and the net effect is that the child record is not removed, unless the parent is removed. This is an unfortunate side effect of the choice to join these objects together in the database and unless we split these out to separate database tables, this possible edge case will exist. 
* Coded in a more linear way, leading to better developer understanding and easier improvements to existing behaviour.
** The ability to extend behaviour is far easier and more accessible to understanding. 
** Imports are not recursively handled like the importer v1, where a transaction object calls to a message object, that calls to a transmission object that calls to a record object, that is then identified as the data type - this process of parsing made debugging and altering the process very difficult and hard to see object properties and types.
** Its very easy to debug and see what's happening at any point.  
** Stack traces make a lot more sense. 
** Extending and changing the behaviour of an imported taric object, or a set of taric objects is far easier and simpler. 
** The process is split into a validation stage, then a commit stage, depending on the validation passing.
* Better child objects handling (where a TARIC record is appended to a parent table), Where before changes would be lost or could not be processed correctly, now a more robust process around handling this type of update.
** There were circumstances where the importer v1 would simply discard a change, the importer v2 will not discard anything and either import or add a report item to the failed import with details about the failure.
** child parent relationships are validated and reported on if anything data can't be processed for a number of reasons. 
** No data gets left behind!
* Very high test coverage for all TARIC types, and update types
** Unit Testing was a huge undertaking, to test every object type and every update type was a large task and very time consuming. It did however provide roughly 97% coverage on the new code. 
** Integration testing at the time of development allowed us to evidence multiple examples of previous EU imports that failed, and required Data Engineering to massage the data when imported with importer v1, worked seamlessly with importer v2. This will have a huge impact on the comm code automation piece and truly allow us to hand over EU imports to the TIPS team. 
* Reporting on import issues in the UI
** When an invalid scenario occurs, and data cant import correctly for one reason or another, there will be an error created against the import, available within the UI for a user to read and take action on. 
** If an issue is detected that can be negated, the importer v2 will create a warning but will allow the import to complete. An example of this would be to delete a description period, since this is not a first class object in the TAP database, it can't be deleted without deleting the associated description. A warning will be recorded against the import, but will still allow the import to complete successfully, but provide the warning details in the importer UI. 
* Importing all objects
** The importer v1 can miss objects and report success, we have seen this with both goods nomenclature origins and successors. the new importer imports EVERYTHING! 

## Checklist
- Requires migrations? YES
- Requires dependency updates? YES

Links to relevant material
See: [JIRA Ticket](https://uktrade.atlassian.net/browse/TP2000-819)


README.rst
```diff
@@ -396,6 +396,15 @@ We use a shared service accross the department for virus scanning to run locally
 3. add CLAM_AV_DOMAIN without http(s):// 
 4. set CLAM_AV_USERNAME,CLAM_AV_PASSWORD as the username and password found in the config.py in the dit-clamav-rest project
 
+Virus Scan and running locally
```
a-gleeson, line 399: Is this duplicated docs from above?

commodities/tests/test_migrations.py
```diff
@@ -83,16 +84,18 @@ def test_main_migration_works(migrator):
         6,
     )
 
-    migrator.reset()
+    print("last chance")
```
a-gleeson, line 87: leftover print?

common/tests/test_forms.py
```diff
@@ -315,87 +314,3 @@ def test_unprefix_formset_data(data, exp):
...
-CakesFormSet = formset_factory(
-    CakesForm,
-    prefix="cakes-formset",
-    formset=FormSet,
-    min_num=1,
-    max_num=10,
-    extra=0,
-    validate_min=True,
-    validate_max=True,
-)
```
a-gleeson, line 344: Was there a reason to remove the formset_factory?

importer/forms.py
```diff
@@ -19,6 +19,8 @@
 from common.util import parse_xml
 from importer.chunker import chunk_taric
 from importer.management.commands.run_import_batch import run_batch
+
+# from importer.management.commands.run_new_import_batch import new_run_batch
```
a-gleeson, line 23: left over comment?

importer/models.py
```diff
@@ -10,6 +10,7 @@
 
 from common.models import TimestampedMixin
 from importer.storages import CommodityImporterStorage
+from taric_parsers.importer_issue import *
```
a-gleeson, line 13: Should we use absolute imports?

importer/tasks.py
```diff
@@ -9,8 +9,19 @@
 from importer import models
 from importer.taric import process_taric_xml_stream
 from importer.utils import build_dependency_tree
+from taric_parsers.importer import *
```
a-gleeson, line 12: should we use absolute imports?

importer/tasks.py
```diff
@@ -9,8 +9,19 @@
...
 from workbaskets.models import get_partition_scheme
 
+# noinspection PyUnresolvedReferences
+# noinspection PyUnresolvedReferences
+# noinspection PyUnresolvedReferences
+# noinspection PyUnresolvedReferences
+# noinspection PyUnresolvedReferences
+# noinspection PyUnresolvedReferences
+# noinspection PyUnresolvedReferences
+# noinspection PyUnresolvedReferences
```
a-gleeson, line 22: excess comments?

publishing/models/envelope.py
```diff
@@ -52,7 +52,7 @@ class MultipleEnvelopesGenerated(Exception):
 
 class EnvelopeManager(Manager):
     @atomic
-    def create(self, packaged_work_basket: PackagedWorkBasket, **kwargs) -> "Envelope":
+    def create(self, packaged_work_basket, **kwargs):
```
a-gleeson, line 55: Was there a reason to remove Type signatures?

taric_parsers/chunker.py
```diff
@@ -0,0 +1,378 @@
...
+from django.template.loader import render_to_string
+
+from common.util import xml_fromstring
+from importer import models
+from importer.namespaces import make_schema_dataclass
+from importer.namespaces import nsmap
+from importer.namespaces import xsd_schema_paths
+from importer.utils import build_dependency_tree
+
+MAX_FILE_SIZE = 1024 * 1024 * 50  # Will keep chunks roughly close to 50MB
```
a-gleeson, line 19: Should this be a configuration variable in settings?

taric_parsers/chunker.py
```diff
@@ -0,0 +1,378 @@
...
+Tags = make_schema_dataclass(xsd_schema_paths)
+
+logger = getLogger(__name__)
+
+
+def get_chunk(
```
a-gleeson, line 25: as it can create a chunk too should the name of the method reflect that?

taric_parsers/chunker.py
```diff
@@ -0,0 +1,378 @@
...
+    are commodities with the same item ID and indent. In this case the commodities must also be sorted
+    on suffix, as 80 means an endline product and therefore must come last always - all other suffixes
+    are lesser than 80.
+
+    Lastly if there is still a conflict commodities are ordered on transaction order.
+
+    These four factors in that order produce an accurate hierarchical representation.
+    """
+
+    def comm_code_key(transaction):
```
a-gleeson, line 119: Might be my ignorance onthe taric import structure but do we need to be concerned about multiple different com codes within a single transaction?

taric_parsers/chunker.py
```diff
@@ -0,0 +1,378 @@
...
+    N.B. This happens entirely within memory.
+    """
+    chunks = batch.chunks.filter(record_code=record_code)
+
+    transactions = []
+
+    for chunk in chunks:
+        transactions.extend(xml_fromstring(chunk.chunk_text))
+
+    logger.info("%d to sort, deleting old ones", len(transactions))
```
a-gleeson, line 183: should we include an Id for this log as it may not be clear if we are investigating logs

taric_parsers/chunker.py
```diff
@@ -0,0 +1,378 @@
...
+
+    chunks.delete()
```
a-gleeson, line 185: what could be the implications of TAP crashing after the delete but before chunks are recreated? 
Is it just this import would be abandoned or is it picked up again?

taric_parsers/chunker.py
```diff
@@ -0,0 +1,378 @@
...
+    for message, transmission, record, identifier in pending_removals:
+        sequence_number = Tags.OUB_RECORD_SEQUENCE_NUMBER.first(record).text
+
+        transmission.remove(record)
+
+        msg = "Transaction %s: Removed record with code %s and sequence number %s."
+        logger.info(msg, transaction_id, identifier, sequence_number)
+
+        if not transmission:
+            message.remove(transmission)
```
a-gleeson, line 315: Am I miss understanding this that we remove transmission if theres no value?

publishing/models/envelope.py
dougmills-DIT: Must have been a wonky github thing - i have not changes that file. I have pulled latest and this has been updated.

commodities/tests/test_migrations.py
dougmills-DIT: removed, good spot

common/tests/test_forms.py
dougmills-DIT: no idea how this got removed - but restored!

importer/forms.py
dougmills-DIT: yep, removed!

taric_parsers/importer.py
```diff
@@ -0,0 +1,770 @@
...
+
+from bs4 import BeautifulSoup
+from django.contrib.auth.models import User
+from django.db import IntegrityError
+from django.db import transaction
+
+from importer.models import BatchImportError
+from importer.models import ImportBatch
+from taric.models import Envelope
+from taric_parsers.parsers.additional_code_parsers import *  # noqa
```
a-gleeson, line 11: should we use absolute importers?

README.rst
dougmills-DIT: good spot! removed duplication

importer/forms.py
dougmills-DIT: removed!

importer/models.py
dougmills-DIT: yep - updated!

importer/tasks.py
dougmills-DIT: yep - updated!

taric_parsers/importer.py
```diff
@@ -0,0 +1,770 @@
...
+        workbasket: WorkBasket = None,
+    ):
+        """
+        TaricImpoter initialiser. This class imports TARIC data into the TAP
+        database, or reports on data issues encountered.
+
+        Args:
+            import_batch: (required) ImportBatch, This object is used to link instances of ImportIssueReportItem
+            taric3_file_path: (required if taric3_xml_string not provided) str, Path to a local xml file that should be imported.
+            taric3_xml_string: (required if taric3_file_path not provided) str, string containing the Taric3 XML.
```
a-gleeson, line 55: Doesn't seem complex enough where dependency injection is required but instead of taric3_file_path and taric3_xml_string we could initialise a "taric3_config" which could be either a file or string which has a function for returning the raw_xml. That way this init doesn't need to worry about which one we require

importer/tasks.py
dougmills-DIT: removed! black removed imports since noinspect is ignored, and should have been noqa

taric_parsers/chunker.py
dougmills-DIT: not even sure if this is required, will investigate - was a straight copy from importer

taric_parsers/chunker.py
dougmills-DIT: renamed to find_or_create_chunk

taric_parsers/chunker.py
dougmills-DIT: removed as not used un new importer

taric_parsers/chunker.py
dougmills-DIT: removed containing function

taric_parsers/chunker.py
dougmills-DIT: removed containing function

taric_parsers/importer.py
dougmills-DIT: yep updated in most places

taric_parsers/chunker.py
dougmills-DIT: yes, looks like not transmission checks that its empty and is safe to remove

taric_parsers/importer.py
dougmills-DIT: good idea, will look at that.

dougmills-DIT: ended up creating a xml source base class, and created two adapters to address the different source types

a-gleeson: Docstring requires updaing for the new base calss, looks good!

taric_parsers/importer.py
```diff
@@ -0,0 +1,754 @@
...
+                            False,
+                        )
+                        for attribute_key in attributes.keys():
+                            setattr(parent, attribute_key, attributes[attribute_key])
+
+    @transaction.atomic
+    def commit_data(self):
+        """Commit the import to the database, iterating through each parsed
+        transaction and the parsed objects contained within."""
+        envelope = Envelope.new_envelope()
```
a-gleeson, line 178: This is an "old" model for envelope (seperate from the packaging envelope). Just wondering the usage of this as it seems to be just for the id in a composite key?
Thinking about it we couldn't really use the new model for this as the ids require to be sequential and if we create an envelope id for an import and then try use that id later it would be out of sequence. If this would be the goal maybe there's future work where you can create an Envelope at any point for an object but you defer the id creation until you generate the envelope for CDS.

taric_parsers/importer.py
```diff
@@ -0,0 +1,754 @@
...
+        Returns:
+            Parent parser or None
+        """
+        parent = None
+
+        matched_transaction = None
+        for transaction in self.parsed_transactions:
+            # skip once matched transaction is processed
+            if matched_transaction:
+                continue
```
a-gleeson, line 379: should this be a break?

taric_parsers/importer.py
```diff
@@ -0,0 +1,754 @@
...
+    ):
+        """
+        Creates an import issue that will be recorded against the database at
+        the end of the import process regardless of the success or failure.
+
+        Args:
+            parsed_message: (required) MessageParser, Parsed message from the XML import
+            related_tag: (optional) str, the XML tag for the parsed message
+            related_identity_keys: (optional) dict, a dictionary of keys defining the identity of the object.
+            message: (optional) str, A string describing the encountered issue / warning.
```
a-gleeson, line 413: missing issue type in docs

taric_parsers/importer.py
```diff
@@ -0,0 +1,754 @@
...
+                child_matches = {}
+                for child_parser_class in child_parser_classes:
+                    object_matches = self.find_child_objects_in_import(
+                        child_parser_class,
+                        parsed_message.taric_object,
+                        parsed_transaction,
+                    )
+                    for object_match in object_matches:
+                        if object_match.__class__.__name__ in child_matches.keys():
+                            child_matches[object_match.__class__.__name__] += 1
```
a-gleeson, line 470: I might have my python wrong but is it possible to rewrite this as `child_matches.get(object_match.__class__.__name__,0) += 1`

taric_parsers/views.py
```diff
@@ -0,0 +1,65 @@
+from django.db.models import Count
+from django.db.models import Q
+from django.urls import reverse_lazy
+from django.views.generic import DetailView
+from django.views.generic import FormView
+
+from common.views import RequiresSuperuserMixin
+from common.views import WithPaginationListView
+from importer.filters import ImportBatchFilter
+from importer.models import *
```
a-gleeson, line 10: should we use absolute imports here?

taric_parsers/tests/test_transaction_parser.py
```diff
@@ -0,0 +1,35 @@
+import os
+
+import pytest
+from bs4 import BeautifulSoup
+
+from taric_parsers.parsers.additional_code_parsers import *
+from taric_parsers.parsers.taric_parser import *
```
a-gleeson, line 7: a couple star imports

taric_parsers/parsers/taric_parser.py
```diff
@@ -0,0 +1,814 @@
...
+import bs4
+from bs4 import NavigableString
+
+from common import validators
+from common.models import Transaction
+from common.util import TaricDateRange
+from common.validators import UpdateType
+from quotas.models import QuotaEvent
+from taric_parsers.importer_issue import ImportIssueReportItem
+from taric_parsers.parser_model_link import *
```
a-gleeson, line 17: star import

taric_parsers/parsers/taric_parser.py
```diff
@@ -0,0 +1,814 @@
...
+            elif isinstance(sibling, NavigableString):
+                sibling = sibling.next_sibling
+                continue
+            elif isinstance(sibling, bs4.Tag):
+                self.object_type = sibling.name
+                break
+
+        for update_type in UpdateType:
+            if update_type.value == self.update_type:
+                self.update_type_name = update_type.name.lower()
```
a-gleeson, line 118: once matched would we want to break the loop?

taric_parsers/importer.py
dougmills-DIT: This is the same as importer/taric.py:119 

I think I need a better understanding of what you are suggesting here, it's not very clear.

dougmills-DIT: I think I need to understand the issue better. 

see importer/taric.py:119 

This is the same as the existing importer. Can you be a little more explicit on your proposition here?

taric_parsers/importer.py
dougmills-DIT: yes, minor refactor done

taric_parsers/importer.py
dougmills-DIT: done, also refactored issue type to a TextChoices parameter

taric_parsers/importer.py
dougmills-DIT: its possible, but I would question if that would be an improvement :) 

It makes the code less readable.

taric_parsers/parsers/regulation_parser.py
```diff
@@ -0,0 +1,315 @@
+from regulations.models import *
+from taric_parsers.parser_model_link import *
+from taric_parsers.parsers.mixins import *
+from taric_parsers.parsers.taric_parser import *
```
a-gleeson, line 4: should we use absolute imports?

taric_parsers/parsers/regulation_parser.py
```diff
@@ -0,0 +1,315 @@
...
+
+    xml_object_tag = "regulation.group.description"
+
+    identity_fields = [
+        "group_id",
+    ]
+
+    deletes_allowed = False
+    group_id: str = None
+    # language_id: str = None
```
a-gleeson, line 63: left over comment?

taric_parsers/parsers/regulation_parser.py
```diff
@@ -0,0 +1,315 @@
...
+        "officialjournal_page": "enacting_regulation__official_journal_page",
+        "validity_start_date": "enacting_regulation__valid_between_lower",
+        "validity_end_date": "enacting_regulation__valid_between_upper",
+        "effective_end_date": "enacting_regulation__effective_end_date",
+        "base_regulation_role": "target_regulation__role_type",
+        "base_regulation_id": "target_regulation__regulation_id",
+        # "complete_abrogation_regulation_role": "3",
+        # "complete_abrogation_regulation_id": "EF123123",
+        # "explicit_abrogation_regulation_role": "3",
+        # "explicit_abrogation_regulation_id": "GH123123",
```
a-gleeson, line 141: left over comment?

taric_parsers/parsers/regulation_parser.py
```diff
@@ -0,0 +1,315 @@
...
+    value_mapping = {
+        "full_temporary_stop_regulation_role": "enacting_regulation__role_type",
+        "full_temporary_stop_regulation_id": "enacting_regulation__regulation_id",
+        "published_date": "enacting_regulation__published_at",
+        "officialjournal_number": "enacting_regulation__official_journal_number",
+        "officialjournal_page": "enacting_regulation__official_journal_page",
+        "validity_start_date": "enacting_regulation__valid_between_lower",
+        "validity_end_date": "enacting_regulation__valid_between_upper",
+        "effective_enddate": "effective_end_date",
+        # "complete_abrogation_regulation_role": "3",
```
a-gleeson, line 188: left over comment?

taric_parsers/parsers/taric_parser.py
```diff
@@ -0,0 +1,814 @@
...
+                match = False
+
+        return match
+
+    def get_identity_fields_and_values_for_parent(self):
+        """Return a dict of values to be used to query the parent, using the
+        child values."""
+        result = {}
+
+        for field in self.identity_fields_for_parent():
```
a-gleeson, line 362: we could do `self.identity_fields_for_parent().values()` if we only need the values for this loop or `.items()` if we need `key, value` instead of calling the function twice

taric_parsers/parsers/taric_parser.py
```diff
@@ -0,0 +1,814 @@
...
+                elif mapped_data_item_key == "update_type_name":
+                    field_data_type = str
+                else:
+                    field_data_type = get_type_hints(self)[mapped_data_item_key]
+
+                field_data_typed = None
+
+                # convert string objects to the correct types, based on parser class annotations
+
+                if field_data_type == str:
```
a-gleeson, line 470: there's a few elifs, should we use a switch statement?

taric_parsers/parsers/taric_parser.py
```diff
@@ -0,0 +1,814 @@
...
+                if field in model_attributes.keys() and field in self.identity_fields:
+                    del model_attributes[field]
+
+        return model_attributes
+
+    def can_save_to_model(self) -> bool:
+        """
+        Can the model be saved to the database.
+
+        This method checks that the parser that represents a TARIC object can be saved to a TAP model by checking associated issues recorded during validation
```
a-gleeson, line 687: Does this check against issues arising in validation?

taric_parsers/tests/test_parser_helper.py
```diff
@@ -0,0 +1,143 @@
+import pytest
+
+from common.util import TaricDateRange
+from taric_parsers.parsers.additional_code_parsers import *
+from taric_parsers.parsers.certificate_parser import *
+from taric_parsers.parsers.footnote_parser import *
+from taric_parsers.parsers.geo_area_parser import *
+from taric_parsers.parsers.measure_parser import *
+from taric_parsers.parsers.quota_parser import *
+from taric_parsers.parsers.regulation_parser import *
```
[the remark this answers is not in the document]
dougmills-DIT: not so fussed about tests but done ;)

taric_parsers/parsers/quota_parser.py
```diff
@@ -0,0 +1,569 @@
+import json
+
+from geo_areas.models import *
+from measures.models import *
+from quotas.models import *
+from taric_parsers.parser_model_link import *
+from taric_parsers.parsers.taric_parser import *
```
a-gleeson, line 7: absolute imports over star imports?

taric_parsers/parsers/quota_parser.py
```diff
@@ -0,0 +1,569 @@
...
+    identity_fields = [
+        "quota_definition__sid",
+        "occurrence_timestamp",
+    ]
+
+    quota_definition__sid: str = None
+    occurrence_timestamp: datetime = None
+
+    @property
+    def data(self):
```
a-gleeson, line 355: whats the purpose of this parser having this property?

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
+from additional_codes.models import *
```
a-gleeson, line 1: absolute imports over star imports?

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+
+    xml_object_tag = "measure.type.series.description"
+
+    identity_fields = [
+        "sid",
+    ]
+
+    deletes_allowed = False
+    sid: str = None
+    # language_id: str = None
```
a-gleeson, line 68: leftover comment?

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+    description: str = None
+
+
+class NewMeasurementUnitParser(ValidityMixin, Writable, BaseTaricParser):
+    # handler = MeasurementUnitHandler
```
a-gleeson, line 73: leftover comment?

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+    subrecord_code = "05"
+
+    xml_object_tag = "measurement.unit.description"
+
+    identity_fields = [
+        "code",
+    ]
+
+    code: str = None
+    # language_id: str = None
```
a-gleeson, line 125: leftover comment?

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+
+    xml_object_tag = "measurement.unit.qualifier.description"
+
+    identity_fields = [
+        "code",
+    ]
+
+    deletes_allowed = False
+    code: str = None
+    # language_id: str = None
```
a-gleeson, line 187: leftover comment?

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+    xml_object_tag = "monetary.unit.description"
+
+    identity_fields = [
+        "code",
+    ]
+
+    deletes_allowed = False
+
+    code: str = None
+    # language_id: str = None
```
a-gleeson, line 290: leftover comment?

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+    ]
+
+    deletes_allowed = False
+    sid: int = None
+    # language_id: str = None
+    description: str = None
+
+
+class NewMeasureTypeParser(ValidityMixin, Writable, BaseTaricParser):
+    # handler = MeasureTypeHandler
```
a-gleeson, line 360: leftover comment?

taric_parsers/tests/test_transaction_parser.py
dougmills-DIT: done

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+
+    xml_object_tag = "measure.type.description"
+
+    identity_fields = [
+        "sid",
+    ]
+
+    deletes_allowed = False
+    sid: str = None
+    # language_id: str = None
```
a-gleeson, line 433: comment

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+    description: str = None
+
+
+class NewAdditionalCodeTypeMeasureTypeParser(
+    ValidityMixin,
+    Writable,
+    BaseTaricParser,
+):
+    # handler = AdditionalCodeTypeMeasureTypeHandler
```
a-gleeson, line 442: comment

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+    ]
+
+    measure_type__sid: str = None
+    additional_code_type__sid: str = None
+    valid_between_lower: date = None
+    valid_between_upper: date = None
+
+
+class NewMeasureConditionCodeParser(ValidityMixin, Writable, BaseTaricParser):
+    # handler = MeasureConditionCodeHandler
```
a-gleeson, line 485: comment

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+
+    xml_object_tag = "measure.condition.code.description"
+
+    identity_fields = [
+        "code",
+    ]
+
+    deletes_allowed = False
+    code: str = None
+    # language_id: str = None
```
a-gleeson, line 538: comment

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+    description: str = None
+
+
+class NewMeasureActionParser(ValidityMixin, Writable, BaseTaricParser):
+    # handler = MeasureActionHandler
```
a-gleeson, line 543: comment

taric_parsers/parsers/geo_area_parser.py
```diff
@@ -0,0 +1,177 @@
+from datetime import date
+
+from geo_areas.models import *
```
a-gleeson, line 3: Star imports used

taric_parsers/parsers/footnote_parser.py
```diff
@@ -0,0 +1,190 @@
...
+    subrecord_code = "05"
+
+    xml_object_tag = "footnote.type.description"
+
+    identity_fields = [
+        "footnote_type_id",
+    ]
+
+    footnote_type_id: str = None
+    # language_id: str = None
```
a-gleeson, line 64: leftover comment?

taric_parsers/parsers/footnote_parser.py
```diff
@@ -0,0 +1,190 @@
...
+
+    identity_fields = [
+        "described_footnote__footnote_type__footnote_type_id",
+        "described_footnote__footnote_id",
+    ]
+
+    allow_update_without_children = True
+
+    sid: int = None
+    # language_id: str = None
```
a-gleeson, line 138: leftover comment?

README.rst
```diff
@@ -296,6 +296,25 @@ This command is broken into two stages:
 
       $ python manage.py run_import_batch
 
+Using the new TARIC parser (new importer)
+-----------------------------------------
+
+The new importer addresses issues with the existing importer.
```
paulpepper-trade, line 302: Terms like "new importer" will be out of date as soon as this work is introduced as production code.

taric_parsers/parsers/commodity_parser.py
```diff
@@ -0,0 +1,338 @@
+from datetime import date
+
+from commodities.models import *
```
a-gleeson, line 3: Star imports

taric_parsers/parsers/commodity_parser.py
```diff
@@ -0,0 +1,338 @@
...
+    sid: int = None
+    item_id: str = None
+    suffix: int = None
+    valid_between_lower: date = None
+    valid_between_upper: date = None
+    statistical: int = None
+
+
+class NewGoodsNomenclatureOriginParser(Writable, BaseTaricParser):
+    # handler = GoodsNomenclatureOriginHandler
```
a-gleeson, line 39: leftover comment?

taric_parsers/parsers/commodity_parser.py
```diff
@@ -0,0 +1,338 @@
...
+    xml_object_tag = "goods.nomenclature.description"
+
+    identity_fields = [
+        "described_goods_nomenclature__sid",
+        "described_goods_nomenclature__item_id",
+        "described_goods_nomenclature__suffix",
+    ]
+
+    sid: int = None
+    # language_id: str = None
```
a-gleeson, line 181: leftover comment?

taric_parsers/parsers/certificate_parser.py
```diff
@@ -0,0 +1,184 @@
+from certificates.models import *
```
a-gleeson, line 1: star import

taric_parsers/parsers/additional_code_parsers.py
```diff
@@ -0,0 +1,264 @@
...
+    xml_object_tag = "additional.code.description"
+
+    identity_fields = [
+        "described_additionalcode__sid",
+        "described_additionalcode__type__sid",
+        "described_additionalcode__code",
+    ]
+
+    sid: int = None
+    # language_id: str = None
```
a-gleeson, line 139: leftover comment?

taric_parsers/parsers/additional_code_parsers.py
```diff
@@ -0,0 +1,264 @@
...
+from taric_parsers.parsers.mixins import *
+from taric_parsers.parsers.taric_parser import *
+
+
+class NewAdditionalCodeTypeParser(ValidityMixin, Writable, BaseTaricParser):
+    model = AdditionalCodeType
+    model_links = []
+
+    value_mapping = {
+        # "certificate_type_code": "certificate_type_sid",
```
a-gleeson, line 15: leftover comment?

taric_parsers/parsers/additional_code_parsers.py
```diff
@@ -0,0 +1,264 @@
+from datetime import date
+
+from additional_codes.models import *
```
a-gleeson, line 3: star imports

README.rst
```diff
@@ -296,6 +296,25 @@ This command is broken into two stages:
...
+under "New TARIC parser" allowing the existing process to proceed as it did before, but
+allow the new importer to be employed. This will be for a time until confidence grows
+and its agreed that the new importer should replace the existing one for comm code import
+capability.
+
+* Removed the use of caching, which has been a cause if multiple issues where changes leaked into future imports, and at times imports would stop all together.
+* Writen in a more linear way, leading to better developer understanding and easier improvements to existing behaviour.
+* Better child objects handling (where a TARIC record is appended to a parent table), Where before changes would be lost or could not be processed correctly, now a more robust process around handling this type of update.
+* Very high test coverage for all TARIC types, and update types
+* Reporting on import issues in the UI, when an invalid scenario occurs, and data cant import correctly for one reason or another, there will be an error created against the import. If an issue is detected that can be negated, the importer will create a warning but will allow the import to complete.
```
paulpepper-trade, line 314: This looks like the kind of content that could go in the PR description rather than in the README. I don't think offers enduring value as readme content.

commodities/models/dc.py
```diff
@@ -491,32 +491,32 @@ def get_dependent_measures(
         *commodities: Commodity,
         as_at: Optional[date] = NOT_PROVIDED,
     ) -> MeasuresQuerySet:
-        goods_sid_query = Q()
+        filter = Q()
```
paulpepper-trade, line 494: Why abbreviate a descriptive / self-documenting variable name?

commodities/models/dc.py
```diff
@@ -491,32 +491,32 @@ def get_dependent_measures(
...
         for commodity in commodities:
             if commodity not in self.commodities:
                 logger.warning(f"Commodity {commodity} not found in this snapshot.")
                 continue
 
-            goods_sid_query |= Q(goods_nomenclature__sid=commodity.obj.sid)
+            filter |= Q(goods_nomenclature__sid=commodity.obj.sid)
 
-        measure_qs = Measure.objects.filter(goods_sid_query)
+        qs = Measure.objects.filter(filter)
```
paulpepper-trade, line 502: Ditto re abbreviation.

commodities/models/dc.py
```diff
@@ -1604,7 +1604,7 @@ def _apply_filters(qs: TrackedModelQuerySet):
             .with_end_date()
             .filter(indented_goods_nomenclature__sid__in=goods_sids)
             .annotate(goods_sid=F("indented_goods_nomenclature__sid"))
-            .order_by("transaction", "validity_start")
+            .order_by("transaction")
```
paulpepper-trade, line 1607: It looks as though this is undoing a fix and naming improvement / refactor in a PR that was merged six months ago: https://github.com/uktrade/tamato/commit/211bd3ee00be9442f8e1315cbcbe6effc8f8d621. Is that intentional or a merge problem in this PR?

taric_parsers/namespaces.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @property
+    def qualified_name(self) -> str:
+        """Returns a fully qualified element tag."""
+        ns = self.namespace
+
+        if ns is None:
+            return self.name
+
+        return f"{{{ns}}}{self.name}"
```
a-gleeson, line 92: extra curly bracket?

taric_parsers/namespaces.py
```diff
@@ -0,0 +1,183 @@
...
+
+from common.xml.namespaces import SEED_MESSAGE
+from common.xml.namespaces import nsmap
+from settings.common import PATH_XSD_ENVELOPE
+from settings.common import PATH_XSD_TARIC
+
+TTag = TypeVar("TTag", bound="Tag")
+TTags = TypeVar("TTags", bound="SchemaTagsBase")
+
+RE_PATTERN_TEST = re.compile(r"[^A-Za-z\.\_]")
```
a-gleeson, line 23: describe the regex?

taric_parsers/parsers/taric_parser.py
dougmills-DIT: done

taric_parsers/parsers/taric_parser.py
dougmills-DIT: done

taric_parsers/models.py
```diff
@@ -0,0 +1 @@
+# Create your models here.
```
a-gleeson, line 1: file needed?

taric_parsers/namespaces.py
```diff
@@ -0,0 +1,183 @@
...
+Define additional groups in the below dictionary for use as a `record_group`
+argument to importer.chunker.chunk_taric.
+
+Check importer.forms.UploadTaricForm.save for example usage when users check the
+'Commodities Only' box in /importers/create.
+
+The only group defined at the moment is commodities, which is easily extensible
+to additional record groups.
+"""
+TARIC_RECORD_GROUPS: Dict[str, Sequence[str]] = dict(
```
a-gleeson, line 39: Is this a repeated constant used in `importer/namespaces.py`?

taric_parsers/parsers/measure_parser.py
dougmills-DIT: removed on all parser classes

taric_parsers/parsers/measure_parser.py
dougmills-DIT: removed

taric_parsers/parsers/measure_parser.py
dougmills-DIT: removed

taric_parsers/parsers/measure_parser.py
```diff
@@ -0,0 +1,1037 @@
...
+
+    xml_object_tag = "duty.expression.description"
+
+    identity_fields = [
+        "sid",
+    ]
+
+    deletes_allowed = False
+    sid: int = None
+    # language_id: str = None
```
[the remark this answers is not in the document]
dougmills-DIT: removed

taric_parsers/parsers/measure_parser.py
dougmills-DIT: removed

taric_parsers/parsers/measure_parser.py
dougmills-DIT: removed

taric_parsers/parsers/measure_parser.py
dougmills-DIT: removed

taric_parsers/models.py
dougmills-DIT: removed!

commodities/models/dc.py
dougmills-DIT: looks like an update that was not pulled across when updating from master - pulled manually! should be fixed

commodities/models/dc.py
dougmills-DIT: looks like an update that was not pulled across when updating from master - pulled manually! should be fixed

taric_parsers/parsers/measure_parser.py
dougmills-DIT: removed!

taric_parsers/parsers/measure_parser.py
dougmills-DIT: removed!

taric_parsers/parsers/measure_parser.py
dougmills-DIT: removed!

taric_parsers/parsers/geo_area_parser.py
dougmills-DIT: Updated!

taric_parsers/parsers/footnote_parser.py
dougmills-DIT: removed!

taric_parsers/parsers/footnote_parser.py
dougmills-DIT: removed!

README.rst
dougmills-DIT: yep - i thought that when writing it, will take a look

taric_parsers/parsers/commodity_parser.py
dougmills-DIT: updated!

taric_parsers/parsers/commodity_parser.py
dougmills-DIT: removed!

taric_parsers/parsers/commodity_parser.py
dougmills-DIT: removed!

taric_parsers/parsers/certificate_parser.py
dougmills-DIT: Updated!

taric_parsers/parsers/additional_code_parsers.py
dougmills-DIT: removed!

taric_parsers/parsers/additional_code_parsers.py
dougmills-DIT: removed!

taric_parsers/parsers/additional_code_parsers.py
dougmills-DIT: Updated!

README.rst
dougmills-DIT: fair, will have a look

dougmills-DIT: tidied thais up.

README.rst
dougmills-DIT: updated

taric_parsers/namespaces.py
dougmills-DIT: yes, copying all required files and referencing within the taric_parser app (namespaces & chunker with a mind to remove importer namespaces.py in the future, at the same time as removing the old importer.

taric_parsers/namespaces.py
dougmills-DIT: done

taric_parsers/namespaces.py
a-gleeson: should we add a comment to highlight this?

taric_parsers/importer.py
a-gleeson: it may be easier to cover this on a call, I can contact you about this and I can update this after.

taric_parsers/importer.py
a-gleeson: Readability is imprortant, I am happy with it if you are,

taric_parsers/importer.py
dougmills-DIT: agreed not an issue, will investigate the use of composit_key in another ticket

taric_parsers/parsers/taric_parser.py
dougmills-DIT: update comments and add together

taric_parsers/parsers/taric_parser.py
dougmills-DIT: updated using values

taric_parsers/parsers/quota_parser.py
dougmills-DIT: the object QuotaEvent has a data attribute thats is a joining of multiple properties, TAP mildly cares about. This is the way we join those properties together as a json string

taric_parsers/parsers/measure_parser.py
dougmills-DIT: updated!

taric_parsers/namespaces.py
dougmills-DIT: discussed on call, will skip this one. the duplication will be removed when the current importer is decommissioned

a-gleeson (commented):
Yield changes look good to me!

common/jinja2/layouts/layout.jinja
```diff
@@ -89,6 +89,10 @@
           "href": url("import_batch-ui-list"),
           "text": "Master importer"
         },
+        {
+          "href": url("taric_parser_import_ui_list"),
+          "text": "New TARIC parser"
```
paulpepper-trade, line 94: Could we give this parser a name other than "new" as that won't age well. Version 2 / V2 may be prefereable.

taric_parsers/chunker.py
```diff
@@ -0,0 +1,258 @@
...
+from settings import MAX_IMPORT_FILE_SIZE
+from taric_parsers.namespaces import make_schema_dataclass
+from taric_parsers.namespaces import xsd_schema_paths
+
+Tags = make_schema_dataclass(xsd_schema_paths)
+
+logger = getLogger(__name__)
+
+
+def find_or_create_chunk(
```
paulpepper-trade, line 23: What's the reasoning for keeping the chunker behaviour? If it's really required, then a module-level docstring would be useful to explain why.

taric_parsers/chunker.py
```diff
@@ -0,0 +1,258 @@
...
+    chapter_heading=None,
+) -> TemporaryFile:
+    """
+    Find or create the chunk currently being written to. If the chunk has to be
+    created write the initial envelope header for it.
+
+    Chunks are eventually written to the database as Django models, however
+    handling the large strings and storing it in memory before updating is (I
+    believe) O(n!). Whereas storing tempfiles and simply appending to the files
+    before writing to the database is O(n).
```
paulpepper-trade, line 36: I think this docstring is taken from the original importer? Are there better ways of avoiding that factoral time complexity growth (assuming that statement is about time and not space complexity) than using an object that is potentially saved on the file system!? I suspect strings aren't used because they're immutable and so you get lots of creation and garbage collection going on. If that's the reasoning, then perhaps a mutable data type such as [bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray) or [StringIO](https://docs.python.org/3/library/io.html#io.StringIO), which I think are functionally equivalent for the purpose of chunking, would be preferable?

taric_parsers/tasks.py
```diff
@@ -0,0 +1,156 @@
...
+):
+    """
+    Task for importing an XML chunk into the database.
+
+    This task must ensure the chunks workbasket_status reflects the import
+    process, whether it is currently running, errored or done. Once complete it
+    is also responsible for finding and setting up the next chunk tasks.
+    """
+
+    get_partition_scheme(partition_scheme_setting)
```
a-gleeson, line 34: Is this call used? as the return value isn't used

a-gleeson, line 34: Is this call used? as the return value isn't used

taric_parsers/importer.py
```diff
@@ -0,0 +1,764 @@
...
+
+    def can_save(self):
+        """
+        Indicates if the current import can be saved.
+        Note: If warnings have been detected, the import can still be completed. The warnings will be available to review.
+
+        Returns:
+            boolean, indicating the state of the import, and its ability to save due to no issues being recorded.
+        """
+        if self.status != "FAILED":
```
a-gleeson, line 325: could this be simplified to `return self.status != "FAILED"` also if status are constants you could then refer to `"FAILED"` as `constant.FAILED`

a-gleeson (approved):
Left a couple more comments but I have finished reviewing the importer and approve. While I will have missed stuff I am sure others with a fresh pair of eyes will catch them!

taric_parsers/namespaces.py
```diff
@@ -0,0 +1,185 @@
...
+    A dataclass for xml element tags.
+
+    :py:attr:`name` corresponds to the name attribute of the Element element in the XML
+    Schema.
+
+    :py:attr:`prefix` reflects namespace prefixes defined in the taric3 and envelope
+    xsd-s.
+
+    :py:attr:`nsmap` this is a prefix-namespace mapping in the format required by
+    xml.etree.ElementTree
```
paulpepper-trade, line 74: We'd normally add these as docstrings directly beneath the attributes that they're documenting. For instance like this: https://github.com/uktrade/tamato/blob/d3d0421d147e6d8c410231a7dba41e1947453ca1/common/models/trackedmodel.py#L66-L79

taric_parsers/namespaces.py
```diff
@@ -0,0 +1,185 @@
...
+
+        return f"{self.prefix}:{self.name}"
+
+    @property
+    def is_pattern(self) -> bool:
+        """Returns true if the tag name is a regex pattern."""
+        return RE_PATTERN_TEST.search(self.name) is not None
+
+    @property
+    def pattern(self):
```
paulpepper-trade, line 110: Return type annotation perhaps?

requirements.txt
```diff
@@ -60,6 +60,7 @@ pytest-celery==0.0.0
 pytest-cov==3.0.0
 pytest-django==4.5.2
 pytest-forked==1.4.0
+pytest-mock==3.11.1
```
paulpepper-trade, line 63: Is pytest-mock being used? I don't see any instances of `PyMock` in the diff - I may have missed something. Although it looks interesting and may add some value, its repo (https://github.com/diddi-/pymock) doesn't get much attention - recent commits, stars or watchers

taric_parsers/chunker.py
paulpepper-trade: The main concern is that it retains some of the complexity that we're hoping to get rid of. If it offer some value in exchange for that complexity, then it would be good to explain what that is in the module docstring.

taric_parsers/chunker.py
paulpepper-trade, line 1: Has this file been pulled into this taric_parsers app verbatim or are there changes that we should note?

taric_parsers/forms.py
```diff
@@ -0,0 +1,149 @@
...
+from importer.models import BatchImportError
+from importer.models import ImportBatch
+from importer.namespaces import TARIC_RECORD_GROUPS
+from taric_parsers.chunker import chunk_taric
+from taric_parsers.importer import run_batch
+from workbaskets.models import WorkBasket
+from workbaskets.validators import WorkflowStatus
+
+
+class NewImportFormMixin:
```
paulpepper-trade, line 22: The "New" prefix won't age well, but I don't think any prefix is required anyway since it's namespaced by the taric_parsers app - it won't clash with a class of the same name in the importer module.
```suggestion
class ImportFormMixin:
```

taric_parsers/forms.py
```diff
@@ -0,0 +1,149 @@
...
+                "submit",
+                "Upload",
+                data_module="govuk-button",
+                data_prevent_double_click="true",
+            ),
+        )
+
+    @transaction.atomic
+    def save(self, user: User):
+        workbasket = WorkBasket.objects.create(
```
paulpepper-trade, line 109: Is there an opportunity to avoid creating a workbasket at this point with the new parser, creating it only when the parser has successfully parsed its TARIC input? I think the current process only require the workbasket because of a combination of difficulties with the current importer: capturing workbasket title and parser workbasket dependencies.

paulpepper-trade (changes requested):

taric_parsers/importer.py
```diff
@@ -0,0 +1,764 @@
...
+from taric.models import Envelope
+from taric_parsers.parsers.additional_code_parsers import *  # noqa
+from taric_parsers.parsers.certificate_parser import *  # noqa
+from taric_parsers.parsers.commodity_parser import *  # noqa
+from taric_parsers.parsers.footnote_parser import *  # noqa
+from taric_parsers.parsers.geo_area_parser import *  # noqa
+from taric_parsers.parsers.measure_parser import *  # noqa
+from taric_parsers.parsers.quota_parser import *  # noqa
+from taric_parsers.parsers.regulation_parser import *  # noqa
+from taric_parsers.parsers.taric_parser import *  # noqa
```
paulpepper-trade, line 20: We shouldn't be using star imports - they make dependency navigation difficult, they hide what's being imported, etc.

paulpepper-trade (changes requested):

taric_parsers/importer.py
```diff
@@ -0,0 +1,764 @@
...
+            author_username: (required) str, Username used to associate an import with a user.
+            workbasket: (optional) Workbasket, If importing to an existing workbasket this variable will be used, else a new workbasket will be created.
+        """
+        # Guard Clauses
+        if not workbasket:
+            if not workbasket_title:
+                raise Exception(
+                    "Import title is required when no workbasket is provided",
+                )
+            elif not author_username:
```
paulpepper-trade, line 67: This could just be an `if` statement since there's no dependency relationship between this conditional statement and the previous one.

taric_parsers/importer.py
```diff
@@ -0,0 +1,764 @@
...
+        self.raw_xml = taric_xml_source.get_xml_string()
+
+        # load the taric3 file into memory, via beautiful soup
+        self.bs_taric3_file = BeautifulSoup(self.raw_xml, "xml")
+
+        # if all good, commit to workbasket
+        if workbasket is None:
+            author = User.objects.get(username=author_username)
+            self.workbasket = WorkBasket(title=workbasket_title, author=author)
+            self.workbasket.save()
```
paulpepper-trade, line 82: I think this would be unnecessary if the caller were required to pass a `WorkBasket` instance into this `__init__()`, wouldn't it? You could then ditch the `workbasket_title` and `author_username` params and associated checks (which should be done by the WorkBasket's own validation anyway, making the validation in this function a redundant duplicate of that).

taric_parsers/importer.py
```diff
@@ -0,0 +1,764 @@
...
+
+        if self.can_save():
+            self.populate_parent_attributes()
+            self.commit_data()
+
+        # Store issues against import
+        for issue in self.issues():
+            BatchImportError.create_from_import_issue_report_item(issue, import_batch)
+
+        return
```
paulpepper-trade, line 102: Redundant return.

taric_parsers/importer.py
```diff
@@ -0,0 +1,764 @@
...
+
+        # reorder models within a transaction to process the children
+
+        if self.can_save():
+            self.populate_parent_attributes()
+            self.commit_data()
+
+        # Store issues against import
+        for issue in self.issues():
+            BatchImportError.create_from_import_issue_report_item(issue, import_batch)
```
paulpepper-trade, line 100: It's unusual for an `__init__()` method to do so much than initialisation. I think conventionally this kind of functionality would be placed in a member function that's called after initialisation. It would be easier to extend this class, make clear there's more to it than simple initialisation, self-document the function's behaviour through its name, perhaps return a useful value, etc.

taric_parsers/importer.py
```diff
@@ -0,0 +1,764 @@
...
+
+        return
+
+    def ordered_transactions_and_messages(self, up_to_transaction=None):
```
paulpepper-trade, line 104: A docstring and return type annotation would be helpful I think, especially since it's yielding a tuple.

paulpepper-trade (changes requested):

taric_parsers/importer.py
```diff
@@ -0,0 +1,764 @@
...
+        encountered.
+
+        Returns:
+            str, string representation of the outcome of the import.
+            - COMPLETED : Successful import, no issues or warnings detected
+            - COMPLETED_WITH_WARNINGS : Successful import, but with warnings
+            - FAILED : failed import, no data committed to the database. Issues recorded against the import.
+        """
+        if len(self.issues(BatchImportErrorIssueType.ERROR)) > 0:
+            return "FAILED"
```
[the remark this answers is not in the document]
paulpepper-trade: Or constant values in a choices data type.

importer/models.py
```diff
@@ -179,6 +181,45 @@ def __repr__(self) -> str:
         )
 
 
+class BatchImportError(TimestampedMixin):
+    object_type = models.CharField(max_length=250)
```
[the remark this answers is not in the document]
eadpearce: Seems to be the same as ImportIssueReportItem.object_type which is documented here https://github.com/uktrade/tamato/pull/1015/files#diff-15211b62debbd8195d1d0a6ea7621b1f010d5e4a3ea3d00a3632a8ab21e34fd9R14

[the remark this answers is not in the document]
paulpepper-trade: It needs a docstring to that effect.

importer/models.py
```diff
@@ -179,6 +181,45 @@ def __repr__(self) -> str:
...
+    related_object_type = models.CharField(max_length=250)
+
+    related_object_identity_keys = models.CharField(max_length=1000)
```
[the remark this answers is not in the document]
dougmills-DIT: no, some taric object use a compound key to be identified, since not all have a SID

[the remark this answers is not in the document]
paulpepper-trade: It probably needs a docstring to explain that. Edie similarly responded to a question in another comment (thanks @eadpearce). I guess if it's necessary to ask for an explanation in the PR, then it might also suggest a need for a docstring or some other way of clarifying the code.

common/jinja2/layouts/layout.jinja
dougmills-DIT: Updated link to be importer v2

importer/migrations/0015_alter_batchimporterror_issue_type.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.2.23 on 2023-11-13 08:44
+
+from django.db import migrations
+from django.db import models
+
+
+class Migration(migrations.Migration):
```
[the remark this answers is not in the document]
dougmills-DIT: done

importer/models.py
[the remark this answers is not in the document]
dougmills-DIT: done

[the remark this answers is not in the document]
dougmills-DIT: I have been through all properties and added a description for the BatchImportError class

common/jinja2/layouts/layout.jinja
dougmills-DIT: updated to importer V2

importer/models.py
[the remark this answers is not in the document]
dougmills-DIT: ok, updated - on a broader note, and worth wider discussion - we could use flake8 to wrap pycodestyle and programatically enforce this type of check (on a file changed / created basis) to automate this sort of feedback. It would speed up the review process and improve standardisation across the code base

importer/models.py
```diff
@@ -179,6 +181,45 @@ def __repr__(self) -> str:
...
+    description = models.CharField(max_length=2000)
+    issue_type = models.CharField(
+        max_length=2000,
```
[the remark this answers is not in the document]
dougmills-DIT: done

importer/models.py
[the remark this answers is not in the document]
dougmills-DIT: I have updated the class comments to include all attributes, and methods. This will show up when calling help(BatchImportError) which I believe is an appropriate and standardised way of presenting this information, thoughts?

importer/models.py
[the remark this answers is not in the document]
dougmills-DIT: no - its a string of a dict - meta information for the relationship. It displays the actual values the importer was using to find a related object. Very useful for reviewing failed imports and resolving. 

This is an import error so the info is from the XML not the database, since in most cases it can't commit because if the error.

importer/models.py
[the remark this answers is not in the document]
paulpepper-trade: I'm unsure what your suggestion would do, Doug. Are you suggesting using flake8 to enfore docstrings against all class attributes?

importer/models.py
```diff
@@ -179,6 +181,45 @@ def __repr__(self) -> str:
...
+        choices=BatchImportErrorIssueType.choices,
+    )
+
+    batch = models.ForeignKey(
+        ImportBatch,
+        on_delete=models.PROTECT,
+        related_name="issues",
+    )
+
+    object_details = models.TextField(default=None)
```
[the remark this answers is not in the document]
dougmills-DIT: its a string representation of the objects TARIC data (dict). Typically this would be used by DE to pull out information and find / correct the XML so the file imported.

importer/models.py
```diff
@@ -179,6 +181,45 @@ def __repr__(self) -> str:
...
+    taric_change_type = models.CharField(max_length=20, default=None)
+    transaction_id = models.CharField(max_length=50, default=None)
+
+    @staticmethod
```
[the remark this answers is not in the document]
dougmills-DIT: done! resolving

importer/models.py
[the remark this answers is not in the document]
paulpepper-trade: I'm still not completely clear about what it is. Since it contains a compound value that would need unpacking, then it probably needs a docstring to explain that and perhaps accompanying example.

pyproject.toml
```diff
@@ -96,4 +96,7 @@ norecursedirs = [
     "run",
     "venv",
 ]
-addopts = "-n=auto"
+markers = [
+    "new_importer"
```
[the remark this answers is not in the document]
dougmills-DIT: done! resolving

importer/models.py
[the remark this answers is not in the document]
paulpepper-trade: The closer the docstring is to the thing it's documenting the better:

- It's more likely to be updated if there's a change (for instance, a naming change).
- I expect most readers of the code would be doing so directly rather than using `help()`, so an adjacent docstring is easier to see than something in the class docstring.
- Although it's a standard approach, it's not the only standard :-)

pyproject.toml
```diff
@@ -96,4 +96,7 @@ norecursedirs = [
...
+]
+#addopts = "-n=auto"
```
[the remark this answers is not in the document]
dougmills-DIT: removed

importer/models.py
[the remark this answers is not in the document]
paulpepper-trade: If it's a dict of the TARIC data, then would a JSONField be preferable here - similar to [a previous comment](https://github.com/uktrade/tamato/pull/1015#discussion_r1405940430).

importer/validators.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class BatchImportErrorIssueType(models.TextChoices):
```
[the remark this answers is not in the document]
dougmills-DIT: I think IssueType is too generic and does not appropriately describe the use, but ImportIssueType sounds fine! will update.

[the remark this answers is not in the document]
dougmills-DIT: Updated

requirements.txt
dougmills-DIT: yep - good spot - I had it in a test but was unused. removed!

taric_parsers/importer.py
dougmills-DIT: yeah, tried to avoid doing that one because of quantity - but fine - if it improves dev experience its a fair point.

taric_parsers/importer.py
dougmills-DIT: removed!

taric_parsers/importer.py
dougmills-DIT: added

taric_parsers/importer.py
dougmills-DIT: done

taric_parsers/importer.py
dougmills-DIT: agreed, removed!

taric_parsers/forms.py
dougmills-DIT: yep ok - was mirroring the existing behaviour in that case, happy to update this as I didnt like it either.

importer/forms.py
```diff
@@ -201,6 +201,7 @@ class CommodityImportForm(ImportFormMixin, forms.Form):
             "number. "
         ),
     )
+
```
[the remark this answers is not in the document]
dougmills-DIT: removed

importer/models.py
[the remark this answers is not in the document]
dougmills-DIT: yep, ok - updated as preferred

importer/models.py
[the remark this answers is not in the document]
dougmills-DIT: Primarily this will be displayed in the UI for a DE / Dev to review and identify the record in XML for manual change - I can update to json field no trouble.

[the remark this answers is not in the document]
dougmills-DIT: Done

Note: this attribute has value when resolving issues with imports, and would currently need a human to review why it's an issue in most cases, as long as the value is in the UI a DE / dev can use this info to resolve, it does not matter if its a string or a jsonfield to that end, however as json it does provide additional validation and utility future  undetermined value if we wanted to do some programmatic work around it in the future, so makes sense, to save a potential messy conversion later.

importer/models.py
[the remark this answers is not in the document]
dougmills-DIT: yep, updated this same time as i did related_object_identity_keys

importer/tasks.py
```diff
@@ -116,7 +118,7 @@ def setup_chunk_task(
     get_partition_scheme(partition_scheme_setting)
 
     if batch.ready_chunks.filter(
-        record_code=record_code, status=models.ImporterChunkStatus.RUNNING, **kwargs
+        record_code=record_code, status=ImporterChunkStatus.RUNNING, **kwargs
```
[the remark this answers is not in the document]
dougmills-DIT: will check for updates, however there is no change to the stacking, just removal of "models."

taric_parsers/forms.py
dougmills-DIT: yep, thats done

importer/tasks.py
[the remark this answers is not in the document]
dougmills-DIT: I have updated pre-commit hooks - I was under the impression it will update as it runs / as needed. I added a section to readme to cover pre-commit hooks also, since there was nothing.

taric_parsers/importer.py
dougmills-DIT: Ended up adding import Status rather than const, but implemented as above.

taric_parsers/importer.py
```diff
@@ -0,0 +1,988 @@
...
+)
+from taric_parsers.parsers.regulation_parser import NewRegulationGroupParser  # noqa
+from taric_parsers.parsers.regulation_parser import (  # noqa
+    NewRegulationReplacementParser,
+)
+from taric_parsers.parsers.taric_parser import BaseTaricParser  # noqa
+from taric_parsers.parsers.taric_parser import ImportIssueReportItem  # noqa
+from taric_parsers.parsers.taric_parser import MessageParser  # noqa
+from taric_parsers.parsers.taric_parser import ParserHelper  # noqa
+from taric_parsers.parsers.taric_parser import TransactionParser  # noqa
```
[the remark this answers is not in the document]
dougmills-DIT: no 0- its needed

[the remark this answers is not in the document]
paulpepper-trade: Why is it needed? It isn't obvious (to me at least).

taric_parsers/parsers/additional_code_parsers.py
```diff
@@ -0,0 +1,266 @@
...
+from footnotes.models import Footnote
+from taric_parsers.parser_model_link import ModelLink
+from taric_parsers.parser_model_link import ModelLinkField
+from taric_parsers.parsers.mixins import ChildPeriod
+from taric_parsers.parsers.mixins import ValidityMixin
+from taric_parsers.parsers.mixins import Writable
+from taric_parsers.parsers.taric_parser import BaseTaricParser
+
+
+class NewAdditionalCodeTypeParser(ValidityMixin, Writable, BaseTaricParser):
```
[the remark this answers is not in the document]
dougmills-DIT: yep - just finished renaming all the parsers now.

paulpepper-trade (approved):

PR #1016, opened 2023-08-23
TP2000-912: Commodity VAT & excise measures

eadpearce:
# TP2000-912: VAT & excise measures

## Why

* Users want to be able to see VAT & excise measures related to a commodity code that aren't held in TAP

## What

* Adds a VAT & excise measures tab to the commodity detail page
* Adds the commodity endpoint to the tariffs api module

Screenshot 2023-08-23 at 16 36 56

### DIFF
--- a/commodities/jinja2/includes/commodities/tabs/measures-vat-excise.jinja
+++ b/commodities/jinja2/includes/commodities/tabs/measures-vat-excise.jinja
@@ -33,7 +33,7 @@
     <div class="app-related-items" role="complementary">
         <h2 class="govuk-heading-s" id="subsection-title">Actions</h2>
         <ul class="govuk-list govuk-!-font-size-16">
-            <li><a class="govuk-link" href="https://www.trade-tariff.service.gov.uk/commodities/{{ object.item_id }}#vat_excise" target="_blank" noopener noreferrer>View this data on the UK Integrated Online Tariff </a></li>
+            <li><a class="govuk-link" href="{{ uk_tariff_url }}" target="_blank" noopener noreferrer>View this data on the UK Integrated Online Tariff </a></li>
         </ul>
     </div>
   </div>
--- a/commodities/jinja2/includes/commodities/tabs/measures-vat-excise.jinja
+++ b/commodities/jinja2/includes/commodities/tabs/measures-vat-excise.jinja
@@ -1,0 +1,84 @@
+{% extends "commodities/detail.jinja" %}
+{% from "components/sort_by.jinja" import sort_by %}
+
+{% set extra_tabs_links = [
+    {
+      "text": "Measures as defined",
+      "href": url('commodity-ui-detail-measures-as-defined', args=[object.sid]),
+      "selected": False,
+    },
+    {
+      "text": "Measures on declarable commodities",
+      "href": url('commodity-ui-detail-measures-declarable', args=[object.sid]),
+      "selected": False,
+    },
+    {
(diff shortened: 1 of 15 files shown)


common/tariffs_api.py
```diff
@@ -1,22 +1,53 @@
...
 import requests
 
 BASE_URL = "https://www.trade-tariff.service.gov.uk/api/v2/"
-QUOTAS = f"{BASE_URL}quotas/search"
 
 
-def get_quota_data(order_number):
-    params = urlencode({"order_number": order_number})
-    response = requests.get(f"{QUOTAS}?{params}")
+class Endpoints(Enum):
```
a-gleeson, line 11: Maybe beyond the scope of this story but do you think making this api more into a client which can be easily switched out for a stub would be useful testing/local dev if you do not want to call out to the service. I understand we can also mock endpoints in tests.
Dale made a good example for our "tariff_api" used by channel islands. https://github.com/uktrade/tamato/tree/master/publishing/tariff_api

commodities/jinja2/includes/commodities/tabs/measures-vat-excise.jinja
```diff
@@ -0,0 +1,84 @@
...
+{% block tab_content %}
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-three-quarters">
+    <h2 class="govuk-heading-l">VAT & excise measures</h2>
+  </div>
+  <div class="govuk-grid-column-one-quarter">
+    <div class="app-related-items" role="complementary">
+        <h2 class="govuk-heading-s" id="subsection-title">Actions</h2>
+        <ul class="govuk-list govuk-!-font-size-16">
+            <li><a class="govuk-link" href="https://www.trade-tariff.service.gov.uk/commodities/{{ object.item_id }}#vat_excise" target="_blank" noopener noreferrer>View this data on the UK Integrated Online Tariff </a></li>
```
a-gleeson, line 36: I noticed this URL used within the code too, is there use in extracting this out? Maybe to vault or just setting it within settings so it is set in one place and it's then pulled down through the context into the template.

commodities/views.py
```diff
@@ -227,6 +228,54 @@ def get_context_data(self, *args, **kwargs):
...
+    template_name = "includes/commodities/tabs/measures-vat-excise.jinja"
+
+    @cached_property
+    def commodity_data(self):
+        data = get_commodity_data(self.object.item_id)
+        if not data:
+            return None
+        return data
+
+    def get_related(self, measure, key):
```
a-gleeson, line 241: Comment to explain whata related data we are retrieving?

common/tariffs_api.py
eadpearce: Since there's no auth and only GET endpoints I don't really think it's worth making a whole client class for it. And mocking the responses in tests is pretty straightfoward (if longwinded due to the amount of and structure of the data)

commodities/jinja2/includes/commodities/tabs/measures-vat-excise.jinja
eadpearce: I might put it in tariffs_api.py with the other url variable just to keep them all in one place. Then build the full url in the view

common/tariffs_api.py
a-gleeson: That sounds fair to me, happy with that :)

a-gleeson (approved):
LGTM!

PR #1017, opened 2023-08-25
TP2000-1009

dougmills-DIT:
# TP2000-1009
Update ME67 and add a test to verify it handles deleted measures correctly

## Why
 * In production deleting a measure still triggers ME67 which fails because it cant find the measure .... becuase its deleted

## What
* Altered the query in ME67 to find the measure, if it cant find it - its fine, dont check. 
* Added test to verify (which also failerd before update)
* All rules passing ion workbasket that originally seeded the ticket 

## Checklist
- Requires migrations? No 

Links to relevant material
See: [Description](https://uktrade.atlassian.net/browse/TP2000-1009)

### DIFF
--- a/measures/business_rules.py
+++ b/measures/business_rules.py
@@ -1253,7 +1253,8 @@ class ME67(BusinessRule):
         # if there is a result, it should be only 1 result, but incase there are multiple it selects last, which is
         # the most recent record
         if measures.exists():
-            geo_area = measures.last().geographical_area
+            measure = measures.last()
+            geo_area = measure.geographical_area
             members = geo_area.members.approved_up_to_transaction(
                 self.transaction,
             )
@@ -1261,10 +1262,10 @@ class ME67(BusinessRule):
             matching_members_to_exclusion_period = members.filter(
                 Q(
                     member__area_id=exclusion.excluded_geographical_area.area_id,
-                    valid_between__startswith__lte=measures.last().valid_between.lower,
+                    valid_between__startswith__lte=measure.valid_between.lower,
                 ),
             )
-            if measures.last().valid_between.upper is None:
+            if measure.valid_between.upper is None:
                 matching_members_to_exclusion_period = (
                     matching_members_to_exclusion_period.filter(
                         valid_between__endswith__isnull=True,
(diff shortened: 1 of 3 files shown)


measures/business_rules.py
```diff
@@ -1246,46 +1245,48 @@ def validate(self, exclusion):
...
+                # e.g. if a date range is 1/1/2020 to 31/1/2020, in the database the upper will be stored as 1/2/2020
+                # which means we must use gt rather than gte here. See the tests for ME67 for clarity - they all work
+                # correctly and the queried dates are all exactly within the ranges, no days space so we can be
+                # confident this rule is behaving as expected.
+                matching_members_to_exclusion_period = matching_members_to_exclusion_period.filter(
                     Q(valid_between__endswith__isnull=True)
                     | Q(
                         valid_between__endswith__isnull=False,
-                        valid_between__endswith__gt=measure.valid_between.upper,
+                        valid_between__endswith__gt=measures.last().valid_between.upper,
```
a-gleeson, line 1284: as used twice should we set latest measure above for maintainability?

measures/business_rules.py
```diff
@@ -1334,8 +1335,11 @@ def validate(self, association):
 
 
 class ME71(FootnoteApplicability):
-    """Footnotes with a footnote type for which the application type = "CN footnotes"
-    cannot be associated with TARIC codes (codes with pos. 9-10 different from 00).
+    """
+    Footnotes with a footnote type for which the application type = "CN
+    footnotes" cannot be associated with TARIC codes (codes with pos.
+
```
a-gleeson, line 1341: extra line?

a-gleeson (approved):
LTGM!

measures/business_rules.py
```diff
@@ -1246,46 +1245,48 @@ def validate(self, exclusion):
...
-        measure = Measure.objects.approved_up_to_transaction(self.transaction).get(
+        measures = Measure.objects.approved_up_to_transaction(self.transaction).filter(
             sid=exclusion.modified_measure.sid,
         )
 
-        geo_area = measure.geographical_area
-        members = geo_area.members.approved_up_to_transaction(
-            self.transaction,
-        )
+        # If no measures are found, this mean the measure is deleted and we can skip the check
```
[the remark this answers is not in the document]
dougmills-DIT: That's correct, or it never existed I suppose, however no point checking exclusions against a measure that does not exist either way

measures/business_rules.py
dougmills-DIT: updated

measures/business_rules.py
dougmills-DIT: updated

dougmills-DIT: ok, this is black being an idiot, removed full stop which makes it appear less broken

a-gleeson (approved):

PR #1018, opened 2023-08-29
TP-2000-875 Adding Filter by Current Work Basket to all search sections

CPrich905:
TP-2000-875 Adding Filter by Current Work Basket to all search sections

## Why
As part of the improvements to the Find and edit measures, we’ve implemented the capability to filter the search to show only measures that are present within the users current workbasket.

We should include this filter for all other Find and edit journey’s to help with making review a lot easier.

## What
This PR adds a Mixin to filter an object based on the current workbasket.
This mixin has been added to the following search and filter pages:

- [ ] Additional Codes
- [ ] Certificates
- [ ] Footnotes
- [ ] GeoAreas
- [ ] Commodities
- [ ] Quotas
- [ ] Regulations

Tests have been added, but please confirm it is working as expected on each search/filter page.

## Checklist
No migrations required
The mixin has been corrected to use a boolean filter and checkbox widget.
Screenshot 2023-09-01 at 10 46 30

The bespoke Measures filter has been refactored to match the style of the mixin.
Screenshot 2023-09-01 at 10 46 40


common/filters.py
```diff
@@ -313,3 +316,38 @@ class AutoCompleteFilter(CharFilter):
     value."""
 
     field_class = AutoCompleteField
+
+
+class CurrentWorkBasketMixin(FilterSet):
+    """Generic filter mixin to filter an object by association with current Work
+    Basket."""
+
+    current_work_basket = ChoiceFilter(
```
eadpearce, line 325: This should be a boolean filter that uses a checkbox widget.  Currently this renders a single radio button that once checked, can't be unchecked

common/filters.py
```diff
@@ -313,3 +316,38 @@ class AutoCompleteFilter(CharFilter):
...
+        widget=forms.RadioSelect,
+        method="filter_work_basket",
+        empty_label=None,
+        label="Work Basket",
+        required=False,
+    )
+
+    def filter_work_basket(self, queryset, name, value):
+        """
+        Filter the given queryset to those which have the specified active
```
eadpearce, line 336: Docstring incorrect?

CPrich905: Corrected

CPrich905:
> Filtering seems to work fine but filter form field needs tweaking

Done! Also removed some bits I'd forgotten to remove previously & fixed some tests that were breaking due to this addition.

common/filters.py
```diff
@@ -34,8 +35,10 @@
 from common.util import StartDate
 from common.util import TaricDateRange
 from common.validators import SymbolValidator
+from workbaskets.models import WorkBasket
 
 ACTIVE_STATE_CHOICES = [Choice("active", "Active"), Choice("terminated", "Terminated")]
+CURRENT_WORKBASKET_CHOICES = (("current", "Current Workbasket"),)
```
eadpearce, line 41: Looks like this is unused now and can be removed

common/filters.py
```diff
@@ -313,3 +316,36 @@ class AutoCompleteFilter(CharFilter):
...
+        Filter the given queryset to those which are part of the current
+        workbasket.
+
+        :param queryset: The queryset to filterby
+        :param name: The name of the field
+        :param value: The value of the field
+        """
+        current_workbasket = WorkBasket.current(self.request)
+        target_model = self._meta.model
+        if value == "current":
```
eadpearce, line 343: Since the field is now a boolean this can just be `if value:`

common/tests/test_filters.py
```diff
@@ -42,3 +44,41 @@ def test_search_queryset_returns_case_insensitive():
...
+        self=self,
+        queryset=qs,
+        name="current_work_basket",
+        value="current",
+    )
+
+    for model in res:
+        assert type(model) == self._meta.model
+
+    assert commodity_not_in_workbasket_1, commodity_not_in_workbasket_2 not in res
```
eadpearce, line 84: probably add this line too
```
assert commodity_in_workbasket_1, commodity_in_workbasket_2 in res
```
because if we're just checking if things aren't in the filtered queryset it wouldn't catch if the filter broke and consistently returned nothing

eadpearce (approved):

PR #1021, opened 2023-08-30
Tp2000 959 measure search results page revamp

LaurenQurashi:
# TP-??? Your PR title here


common/jinja2/components/sort_by.jinja
```diff
@@ -1,14 +1,28 @@
 
-{% macro sort_by(request, sort_by, title, base_url, anchor="") %}
+{% macro sort_by(request, sort_by, title, base_url, query_params, anchor="") %}
 
-  {% if request.GET.sort_by == sort_by and request.GET.order == "desc" %}
-    <a class="govuk-link govuk-link--no-visited-state sort-icon--down" href="{{ base_url }}?sort_by={{ sort_by }}&order=asc{{ anchor }}">
+  {% if request.GET.sort_by == sort_by and request.GET.ordered == "desc" %}
+
+    {% if query_params %}
```
LaurenQurashi, line 6: I've added this, as I need to add to existing filter params.. I'll look to refactor this to be neater once I've sorted the other stuff

common/jinja2/components/sort_by.jinja
```diff
@@ -1,14 +1,28 @@
...
+
+  {% else %}
+    {% if query_params %}
+      <a class="govuk-link govuk-link--no-visited-state sort-icon--up" href="{{ base_url }}&sort_by={{ sort_by }}&ordered=desc{{ anchor }}">
       {{ title }} <img src="{{ static('common/images/sort_icon.svg') }}" alt="">
     </a>
-  {% else %}
-    <a class="govuk-link govuk-link--no-visited-state sort-icon--up" href="{{ base_url }}?sort_by={{ sort_by }}&order=desc{{ anchor }}">
+    {% else %}
+      <a class="govuk-link govuk-link--no-visited-state sort-icon--up" href="{{ base_url }}?sort_by={{ sort_by }}&ordered=desc{{ anchor }}">
```
LaurenQurashi, line 22: I changed the name of the order to ordered as it clashed with the order number in the measures filter

measures/jinja2/measures/list.jinja
```diff
@@ -1,4 +1,127 @@
+{% extends "layouts/layout.jinja" %}
```
LaurenQurashi, line 1: I changed this template as the design wants some changes to the wording of the pagination, so thought it would make sense just to have it all in one template as it's a one off different design

measures/views.py
```diff
@@ -125,21 +127,46 @@ class MeasureSearch(FilterView):
     filterset_class = MeasureFilter
 
     def form_valid(self, form):
+        print(form)
```
LaurenQurashi, line 130: I need to take this out oops

measures/views.py
```diff
@@ -125,21 +127,46 @@ class MeasureSearch(FilterView):
...
         if not self.request.GET:
             return HttpResponseRedirect(reverse("measure-ui-search"))
         return super().dispatch(*args, **kwargs)
 
+    def get_queryset(self):
+        queryset = super().get_queryset()
+
+        ordering = self.get_ordering()
+
+        if ordering:
```
LaurenQurashi, line 163: Here I am checking for the ordering that comes through on the params.. I can't do this for end date as it's not a db field so will need a custom way to sort

measures/views.py
```diff
@@ -153,14 +180,42 @@ def paginator(self):
         return MeasurePaginator(self.filterset.qs, per_page=20)
 
     def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
+        context = {}
```
LaurenQurashi, line 183: The get_context_data call ruined everything I had set in the paginator. It goes up the calling tree and somehow undoes all of the max counts, limits breached etc. I've taken the approach to manually do it in this get context.. and this seems to work and apply on the page.

measures/views.py
```diff
@@ -153,14 +180,42 @@ def paginator(self):
...
             SelectableObjectsForm.object_id_from_field_name(name)
             for name in self.measure_selections
         ]
         context["measure_selections"] = Measure.objects.filter(
             pk__in=measure_selections,
         )
+        context["query_params"] = True
+
+        # Remove sort by and order here, as the queryset will have already been ordered
+        if "sort_by" and "ordered" in self.filterset.data:
```
LaurenQurashi, line 207: Here I am taking out the sort by and ordered once I've sorted and ordered as otherwise it works it's way into the base url and ruins the next sort..

settings/common.py
```diff
@@ -736,7 +736,7 @@
 LIMITED_PAGINATOR_MAX_COUNT = 200
 # Default max number of objects that will be accurately counted by MeasurePaginator.
 MEASURES_PAGINATOR_MAX_COUNT = int(
-    os.environ.get("MEASURES_PAGINATOR_MAX_COUNT", "1000"),
+    os.environ.get("MEASURES_PAGINATOR_MAX_COUNT", "30000"),
```
LaurenQurashi, line 739: Whacking the limit up from the previously unapplied 1000.. could be a mistake but 1000 measures doesn't seem sensible for what TM's will want to do at a time

PR #1023, opened 2023-08-30
TP2000-1012: Add extra fields to quota definition edit form

eadpearce:
# TP2000-1012: Add extra fields to quota definition edit form

## Why

* Allows users to edit more info on quota definition periods

## What

* Allows editing of: description, volume, initial volume, critical state, critical threshold, measurement unit, measurement unit qualifier
* Maximum precision is always set to 3 so this is not editable

FireShot Capture 062 - Edit quota definition details - Manage Trade Tariffs - localhost


SteveCN7:
Screenshot looks good, would it be possible to swap Measurements and Volumes around in the list so that volume is above the unit of measurements and measurement unit qualifier options?

quotas/forms.py
```diff
@@ -273,21 +273,90 @@ class Meta:
...
+    quota_critical_threshold = forms.DecimalField(
+        label="Threshold",
+        help_text="The point at which this quota definition period becomes critical, as a percentage of the total volume.",
+        widget=forms.TextInput(),
+        error_messages={"invalid": "Critical threshold must be a number"},
+    )
+    quota_critical = forms.TypedChoiceField(
+        label="Is the quota definition period in a critical state?",
+        help_text="This determines if a trader needs to pay securities when utilising the quota.",
+        coerce=lambda value: value == "True",
```
CPrich905, line 304: What is the reason for using a coerced value here rather than a boolean filter?

CPrich905 (changes requested):
One clarification for my own interest, then just the change Steve has requested to the layout

quotas/forms.py
eadpearce: Annoying a BooleanField in Django doesn't let you render it as radios with Yes/No options (which is what we wanted instead of a checkbox) so we have to use a TypedChoiceField instead

eadpearce:
> Screenshot looks good, would it be possible to swap Measurements and Volumes around in the list so that volume is above the unit of measurements and measurement unit qualifier options?

This was changed after feedback from Patrick. We thought it made more sense to state the units before the volume but tariff managers may feel differently

SteveCN7:
> > Screenshot looks good, would it be possible to swap Measurements and Volumes around in the list so that volume is above the unit of measurements and measurement unit qualifier options?
> 
> This was changed after feedback from Patrick. We thought it made more sense to state the units before the volume but tariff managers may feel differently

Good to go then :-)

quotas/forms.py
CPrich905: Thanks! I might experiment with this as well for some of the search/filter page widgets as well as the different styles there are frustrating some of the TMs

CPrich905 (approved):

PR #1026, opened 2023-09-08
TP2000-959 measure list pagination fix

LaurenQurashi:
# TP-2000-959

## Why
 Previously pagination on the Measures List page was slow and broken. It wouldn't limit how many measures came back, or how many pages you could see. It was really slow, and while still is slow, these tweaks hopefully makes it a little quicker. 

## What
When running locally, I've gotten results on master around the 80 second mark for responses. After some investigation (thanks Paul), it seems that the bottleneck is how we reference `page` and `paginator` in the templates. Each time we reference them, SQL queries are duplicated to calculate these objects' values. To get round this, if we assign variables in the context so that the fetching is done in one bout and not over and over again in the template aswell. 

I've also fixed the page and results limiters on the pagination. I did this by simply not calling `super().get_context_data()` in the MeasureList's get context data function, and providing all the data manually instead. This is because a mixin up the inheritance chain was calling `super().get_context_date()` which referenced to seemingly itself? but by doing that it reset all of the limiter settings that had been provided on the MeasureList's paginator property. 

Ultimately by doing all of this, but also upping the items per page number to 40, we're sitting at around 51 seconds (filtering for just erga omnes) which is a little better.


measures/jinja2/measures/list.jinja
```diff
@@ -1,4 +1,119 @@
+{% extends "layouts/layout.jinja" %}
+
+{% from "components/button/macro.njk" import govukButton %}
+{% from "components/input/macro.njk" import govukInput %}
+{% from "components/table/macro.njk" import govukTable %}
+
 {% set form_url = "measure-ui-list" %}
 {% set list_include = "includes/measures/list.jinja" %}
 
-{%- include "layouts/list_vertical.jinja" -%}
```
a-gleeson, line 4: is this unused now and can be deleted?

a-gleeson (approved):
Minor comments, otherwise LGTM!

PR #1028, opened 2023-09-11
TP2000-1044  Include geo areas with future start dates in dropdown lists

dalecannon:
# TP2000-1044  Include geo areas with future start dates in dropdown lists

## Why
Tariff changes that involve geographical areas with a future start date are unable to be carried out in advance because such geographical areas are unavailable for selection in dropdown lists on geographical area forms (e.g in the create a new measure journey).

## What
-  Replaces `as_at_today()`  with `as_at_today_and_beyond()` for all geographical area forms (create measures journey and quota origins edit journey)
- Adds test for `as_at_today_and_beyond()`


common/tests/test_models.py
```diff
@@ -81,6 +81,23 @@ def test_as_at_today(model1_with_history):
...
+    """Ensure only records active at the current date and future records are
+    fetched."""
+    outdated_record = {validity_factory.create(valid_between=date_ranges.earlier).pk}
+
+    active_and_future_records = {
+        validity_factory.create(valid_between=date_ranges.normal).pk,
+        validity_factory.create(valid_between=date_ranges.later).pk,
+    }
+
+    Model = validity_factory._meta.get_model_class()
```
a-gleeson, line 94: should we maybe still use a variable name here just incase an accidental class is introduced?

geo_areas/forms.py
```diff
@@ -85,7 +85,7 @@ def __init__(self, *args, **kwargs):
             GeographicalArea.objects.filter(area_code=AreaCode.COUNTRY)
             .current()
             .with_latest_description()
-            .as_at_today()
+            .as_at_today_and_beyond()
             .order_by("description")
```
a-gleeson, line 89: not including querysets with `.exclude` but should we maybe extracte this as a common queryset on `GeographicalArea` where we pass in the `area_code`?

geo_areas/forms.py
```diff
@@ -803,7 +803,7 @@ def __init__(self, *args, **kwargs):
             GeographicalArea.objects.current()
             .with_latest_description()
             .filter(area_code=AreaCode.GROUP)
-            .as_at_today()
+            .as_at_today_and_beyond()
             .order_by("description")
```
a-gleeson, line 807: not apart of your story but whats the effect of the order of these querysets being different from the ones in the other form.py?

dalecannon: From what I can test, any difference in orders of `current()`, `with_latest_description` and `filter()` doesn't really matter in this case (or at least it doesn't appear to return different results).

common/tests/test_models.py
dalecannon: Yeah, I've renamed the variable for good measure.

geo_areas/forms.py
a-gleeson: That's fair, I guess there's maybe a performance (slight) hit in different orders. I guess you would want to narrow down the list of results generally. Although probably mineut here

a-gleeson (approved):
LGTM!

PR #1029, opened 2023-09-11
feat: TP2000-1018 - expose report endpoints to tariff managers

nboyse:
# TP2000-1018 - Expose report endpoints to tariff managers

## Why

TAP reports link isn't accessible to Tariff Managers

## What

Creating a minimal model for reports so that custom permissions can be added for tariff managers to access reports. I was tempted to remove the custom permissions entirely but due to plans to further expand on reports feature, I instead built the functionality for them to work as already put in place to save future migration pains.


reports/views.py
```diff
@@ -16,7 +16,7 @@ def index(request):
     return render(request, "reports/index.jinja", context)
 
 
-@permission_required("app.view_report")
+@permission_required("reports.view_report")
```
a-gleeson, line 19: There's a conftest called policy_group that it may be useful to add to for testing pemissions. It's in the main conftest.py and  around line 247

reports/models.py
```diff
@@ -1 +1,7 @@
-# Create your models here.
+from django.db import models
+
+
+class Report(models.Model):
+    class Meta:
+        # Define the name for the database table (optional)
```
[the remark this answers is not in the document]
nboyse: Maybe in future, but I just had to build this to register the custom permissions for now as we're not actually interacting with the DB

reports/views.py
nboyse: Perfect, have added it to the fixture, good idea.

reports/migrations/0002_create_custom_permissions.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+    edit_permission, _ = Permission.objects.get_or_create(
+        codename="view_report",
+        name="Can view reports",
+        content_type=content_type,
+    )
+
+
+class Migration(migrations.Migration):
```
a-gleeson, line 23: I can't view the error but a lot of tests are failing I imagine it's around the migration?

a-gleeson (approved):

PR #1030, opened 2023-09-11
feat: hide Import EU Taric files

nboyse:
# TP2000-1028 - Goods importer journey selection is visible to all users

## Why

Goods importer journey selection is visible to all users

The “Import EU Taric files” journey is visible as a selection from the TAP home page to all users. The journey should be resricted to those users with the relevant permissions only - similar to other journeys that have a creational, edit or deletion action.

## What

Following the same logic of the view and only exposing the link to users with permission to add or change trackedmodel


a-gleeson (approved):
LGTM :)

PR #1031, opened 2023-09-11
TP2000-1029 - feat: Measure list view requires optimising

nboyse:
# TP2000-1029 - feat: Measure list view requires optimising

## Why

Loading measures list views takes a long time to load. On a typical developement development workstation (normally having more computing resource available than the TAP production environment), the response time is in excess of 60 seconds.

The response time should be reduced to an acceptable amount of time

## What

A variety of approaches have been taken, primarily using select related rather than fetching all from certain models, also, setting defaults for things like start dates, prefixes and units to prevent data being fetched when it does not exist.

This has reduced large queries from 40s to 20s.

### DIFF
--- a/measures/filters.py
+++ b/measures/filters.py
@@ -81,7 +81,7 @@ class MeasureFilter(TamatoFilter):
     goods_nomenclature = AutoCompleteFilter(
         label="Specific commodity code",
         field_name="goods_nomenclature__item_id",
-        queryset=GoodsNomenclature.objects.all().select_related("goods_nomenclature"),
+        queryset=GoodsNomenclature.objects.all().select_related("trackedmodel_ptr"),
         attrs={"display_class": GOV_UK_TWO_THIRDS, "min_length": 4},
     )
 
--- a/measures/filters.py
+++ b/measures/filters.py
@@ -74,7 +74,7 @@ class MeasureFilter(TamatoFilter):
     measure_type = AutoCompleteFilter(
         label="Type",
         field_name="measure_type__sid",
-        queryset=MeasureType.objects.all().select_related("measure_type"),
+        queryset=MeasureType.objects.all().select_related("trackedmodel_ptr"),
         attrs={"display_class": GOV_UK_TWO_THIRDS, "min_length": 3},
     )
 
--- a/measures/filters.py
+++ b/measures/filters.py
@@ -51,41 +51,38 @@ class MeasureTypeFilterBackend(TamatoFilterBackend):
 
 class MeasureFilter(TamatoFilter):
     def __init__(self, *args, **kwargs):
-        if kwargs["data"]:
-            data = kwargs["data"].copy()
-            if "start_date_modifier" not in data:
-                data["start_date_modifier"] = "exact"
-            if "end_date_modifier" not in data:
-                data["end_date_modifier"] = "exact"
-            kwargs["data"] = data
-        super(MeasureFilter, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
+        self.optimize_query()
+
+    def optimize_query(self):
+        data_copy = self.data.copy()  # Create a copy of the QueryDict
+
+        data = self.data
+        if not data:
+            return
+
+        # Check for missing date modifiers and set defaults
+        data_copy["start_date_modifier"] = data_copy.get("start_date_modifier", "exact")
+        data_copy["end_date_modifier"] = data_copy.get("end_date_modifier", "exact")
 
     sid = CharFilter(
         label="ID",
         widget=forms.TextInput(),
-        validators=[
-            NumericValidator,
-        ],
+        validators=[NumericValidator],
     )
 
     measure_type = AutoCompleteFilter(
         label="Type",
         field_name="measure_type__sid",
-        queryset=MeasureType.objects.all(),
-        attrs={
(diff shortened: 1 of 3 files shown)


measures/querysets.py
```diff
@@ -26,160 +19,81 @@
...
+                    default=F("duty_expression__prefix"),
+                    output_field=CharField(),
+                ),
+                monetary_unit_value=Case(
+                    When(
+                        Q(duty_amount__isnull=False) & Q(monetary_unit=None),
+                        then=Value("%"),
+                    ),
+                    When(duty_amount__isnull=True, then=Value("")),
+                    default=Value(" "),
```
a-gleeson, line 68: This no longer includes anything in regards to  `F("monetary_unit__code"),` is that okay?

measures/querysets.py
```diff
@@ -26,160 +19,81 @@
...
+
+        # Filter components by the latest transaction_id.
         component_qs = component_qs.filter(transaction_id=latest_transaction_id)
 
-        # Aggregate all the current Components for component_parent to form its
-        # duty sentence.
-        duty_sentence = component_qs.aggregate(
-            duty_sentence=StringAgg(
-                expression=Trim(
+        # Construct the duty sentence.
```
a-gleeson, line 51: Nice query, simplfied a lot. I think a comment to describe the basics of it could help future maintenance!

measures/querysets.py
```diff
@@ -26,160 +19,81 @@
...
-                                duty_amount__isnull=True,
-                                then=Value(""),
-                            ),
-                            default=Concat(
-                                Value(" "),
-                                F("monetary_unit__code"),
-                            ),
+                        "prefix_value",
+                        Value(" "),
+                        F("duty_expression__monetary_unit_applicability_code"),
```
a-gleeson, line 77: would this be the same value `F("monetary_unit__code")` code then?

measures/querysets.py
```diff
@@ -271,82 +185,54 @@ def with_effective_valid_between(self):
 
 class MeasureConditionQuerySet(TrackedModelQuerySet):
     def with_reference_price_string(self):
-        """
```
a-gleeson, line 274: I think having a comment to explain this is generally useful, what do you think?

measures/querysets.py
```diff
@@ -271,82 +185,54 @@ def with_effective_valid_between(self):
...
                             then=Value("%"),
                         ),
-                        default=Concat(
-                            Value(" "),
-                            F("monetary_unit__code"),
-                        ),
+                        default=Value(" "),  # Space separator
+                        output_field=CharField(),
                     ),
+                    F("monetary_unit__code"),
```
a-gleeson, line 209: I may be reading this wrong but I think when we set the value to % we do not want to include `F("monetary_unit__code")` in the concatenation. I'd have to check too but if `monetary_unit` is null and we try to access `code` would that error?

measures/querysets.py
```diff
@@ -271,82 +185,54 @@ def with_effective_valid_between(self):
...
-                            Value(" / "),
-                            F(
-                                "condition_measurement__measurement_unit_qualifier__abbreviation",
-                            ),
-                        ),
+                        default=Value(" / "),
+                        output_field=CharField(),
+                    ),
+                    F(
+                        "condition_measurement__measurement_unit_qualifier__abbreviation"
```
a-gleeson, line 232: Just sense checking this to, when it is null/none it doesn't concat Null onto the string? This third case looks good to me if it doesn't do that

measures/querysets.py
```diff
@@ -271,82 +185,54 @@ def with_effective_valid_between(self):
...
-                        default=Concat(
-                            Value(" / "),
-                            F(
-                                "condition_measurement__measurement_unit__abbreviation",
-                            ),
-                        ),
+                        default=Value(" / "),
+                        output_field=CharField(),
                     ),
+                    F("condition_measurement__measurement_unit__abbreviation"),
```
a-gleeson, line 222: From the previous code it looks like we wouldn't Concat `" / "` if `monetary_unit__isnull=True,`, would the default value always concat this now?
Just sense checking this to, when `F("condition_measurement__measurement_unit__abbreviation")` is null/none it doesn't concat Null/make the whole value Null? I think if this is the case Coalesce may be an option to use in these cases?

paulpepper-trade:
It may be useful to coordinate with @LaurenQurashi as she's doing work with the measures list view, including optimisations.

measures/querysets.py
a-gleeson: Ah did you just undo the changes to MeasureConditionQuerySet?

measures/filters.py
```diff
@@ -51,41 +51,38 @@ class MeasureTypeFilterBackend(TamatoFilterBackend):
...
+    def optimize_query(self):
+        data_copy = self.data.copy()  # Create a copy of the QueryDict
+
+        data = self.data
+        if not data:
+            return
+
+        # Check for missing date modifiers and set defaults
+        data_copy["start_date_modifier"] = data_copy.get("start_date_modifier", "exact")
+        data_copy["end_date_modifier"] = data_copy.get("end_date_modifier", "exact")
```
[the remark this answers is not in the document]
CPrich905: I think adding this line back in should resolve that?
`kwargs["data"] = data_copy`

PR #1032, opened 2023-09-12
TP2000-885  Use quota origins and exclusions to populate geo data in create measure journey

dalecannon:
# TP2000-885  Use quota origins and exclusions to populate geo data in create measure journey

## Why
There exists a requirement for quotas that geographical data (origins and exclusions) held on a quota order number must match geographical data associated with a measure. To save the user from having to manually select the geographical areas and exclusions, we can take the quota's origins and exclusions to populate automatically the geographical data for measures whenever a quota order number is selected in the create measure journey. 

## What
- Moves away from `cleaned_data["geo_areas_list"]` and `cleaned_data["geo_area_exclusions"]` to combine `cleaned_data["geo_areas_and_exclusions"]` (list of dictionaries) to accommodate multiple pairs of geo areas and exclusions that can sometimes result from the geo data held on a quota order number

- Sets `cleaned_data["geo_areas_and_exclusions"]` either with geo data held on order number if one has been selected or with geo area form data in `MeasureGeographicalAreaForm.clean()`

- Conditionally shows geo areas form or a summary of geo data that will be used to create measures depending if an order number has been selected

- Adds and updates form and view unit tests

- Amends help text on quota order number form page

- Add help text to geo area form/summary page

##

Example of the summary page of geo areas/exclusions automatically populated from the selected quota order number in place of the geo areas form:

Screenshot 2023-09-15 at 10 48 39


a-gleeson:
Is this an either or scenario? Where if you add an order number and the geo area's are populated you cannot add more geo area's?

dalecannon:
> Is this an either or scenario? Where if you add an order number and the geo area's are populated you cannot add more geo area's?

That's right. Using an order number replaces selecting geo areas/exclusions. 

I've just realised it would be good to handle

measures/forms.py
```diff
@@ -1121,29 +1148,35 @@ def clean(self):
...
+                    exclusions = [
                         exclusion[constants.FIELD_NAME_MAPPING[geo_area_choice]]
-                        for exclusion in cleaned_data[
-                            constants.EXCLUSIONS_FORMSET_PREFIX_MAPPING[geo_area_choice]
-                        ]
+                        for exclusion in geo_area_exclusions
                     ]
+                    cleaned_data["geo_areas_and_exclusions"][0][
+                        "exclusions"
+                    ] = exclusions
```
a-gleeson, line 1179: is the index [0] just because there's only ever one set of exclusions to add?

measures/tests/conftest.py
```diff
@@ -311,6 +311,11 @@ def commodity2():
     return factories.GoodsNomenclatureFactory.create()
 
 
+@pytest.fixture()
+def order_number():
+    return factories.QuotaOrderNumberFactory.create()
```
a-gleeson, line 316: Does this not exist on line 1458?

measures/forms.py
dalecannon: Yeah. Since only one geo area group can be selected, there will only be one set of exclusions that can apply.

measures/views.py
```diff
@@ -597,11 +597,11 @@ def create_measures(self, data):
...
         for commodity_data in data.get("formset-commodities", []):
             if not commodity_data.get("DELETE"):
-                for geo_area in data["geo_area_list"]:
+                for geo_data in data["geo_areas_and_exclusions"]:
                     measure_data = {
                         "measure_type": data["measure_type"],
-                        "geographical_area": geo_area,
-                        "exclusions": data.get("geo_area_exclusions", None) or [],
+                        "geographical_area": geo_data["geo_area"],
+                        "exclusions": geo_data.get("exclusions", None) or [],
```
a-gleeson, line 604: could just replace `None` with `[]` here without the `or`?

a-gleeson (approved):
LGTM! just a couple minor comments

measures/views.py
dalecannon: Done!

a-gleeson (approved):
Good catch on the origins! LGTM!

PR #1033, opened 2023-09-12
TP2000-1021 Measures tab for additional codes

mattjamc:
# TP2000-1021 Measures tab for additional codes

## Why

Adding further detail to the additional code details view so users can see active measure information.

## What

It allows users to see all associated active measures and updates the action menu so they can view the additional code in find and edit measures.

### Previously
image

### Update
Includes:

- measures tab displaying currently active measures linked to that additional code
- updated action menu to include 'View in find and edit measures'

image


additional_codes/jinja2/additional_codes/detail.jinja
mattjamc, line 1: Ended up deleting version control tab in next commit and reverting back to inheriting from common as this panel doesn't have the action menu anyway

additional_codes/jinja2/includes/additional_codes/tabs/version_control.jinja
mattjamc, line 1: Ended up deleting version control tab in next commit and reverting back to inheriting from common as this panel doesn't have the action menu anyway

additional_codes/jinja2/includes/additional_codes/actions.jinja
```diff
@@ -0,0 +1,16 @@
+<div class="govuk-grid-column-one-quarter">
+  <div class="app-related-items" role="complementary">
+      <h2 class="govuk-heading-s" id="subsection-title">Actions</h2>
+      <ul class="govuk-list govuk-!-font-size-16">
+            <li><a class="govuk-link" href="{{ measures_url }}">View in find and edit measures</a></li>
```
a-gleeson, line 5: If you do `{% if measure_url %}` here could you include this in the common action menu and keep using `includes/common/tabs/descriptions.jinja` above? 
There may be another reason that I am missing to why you created a new additional codes tab that's specific?

additional_codes/tests/test_views.py
```diff
@@ -186,3 +188,30 @@ def test_additional_code_type_api_list_view(valid_user_client):
         expected_results,
         valid_user_client,
     )
+
+
+def test_additional_code_details_list_current_measures(
```
a-gleeson, line 193: Maybe a test to test what's displayed for no active measures?

additional_codes/views.py
```diff
@@ -134,6 +139,29 @@ class AdditionalCodeConfirmCreate(AdditionalCodeMixin, TrackedModelDetailView):
 class AdditionalCodeDetail(AdditionalCodeMixin, TrackedModelDetailView):
     template_name = "additional_codes/detail.jinja"
 
+    def get_context_data(self, *args, **kwargs):
+        measures = (
```
a-gleeson, line 143: Do we have access to the selected tab? If we do we could only exectued this query when the tab is selected, which means we don't executed the query for other tabs. 
Although I may be miss understanding how our tabs work and they don;t have page loads inbetween

mattjamc: @a-gleeson I think all tabs load at once with the page - there aren't separate views for each tab. Maybe I am misunderstanding though

a-gleeson: In this implementation I think you are correct, I think some other tab implementations have changed though. You will see them referenced as `FakeTab` or something.
This isn't an issue here then happy to resolve

a-gleeson (approved):
LTGM!

PR #1034, opened 2023-09-12
TP2000-959 Add sorting to the MeasuresList filter results table

LaurenQurashi:
# TP2000-959-measure-list-table-sorting

## Why
In order for tariff managers to get to the right results quicker and easier, Table sorting has been added to the designs. Initially this was for all headers but we whittled it down to the most important which are: 
- ID
- Type
- Start and End dates
- Geographical areas

When tariff managers click on the table headers, it will reload the page with only that column sorted.

## What
Sorting options have been added to the measures list results table

## Checklist
- Requires migrations? - no


commodities/tests/test_views.py
```diff
@@ -245,7 +245,7 @@ def test_commodity_measures_sorting_geo_area(
         url_name,
         kwargs={"sid": commodity.sid},
     )
-    response = valid_user_client.get(f"{url}?sort_by=geo_area&order=desc")
+    response = valid_user_client.get(f"{url}?sort_by=geo_area&ordered=desc")
```
LaurenQurashi, line 248: A lot of these changes are changing order to ordered.. down the page a bit I have some code to remove sort_by and ordered from the url params which clashed a bit with the order_number filter.

measures/tests/test_views.py
```diff
@@ -2478,3 +2479,87 @@ def test_measure_list_redirects_to_search_with_no_params(valid_user_client):
 def test_measure_search_200(valid_user_client):
     response = valid_user_client.get(reverse("measure-ui-search"))
     assert response.status_code == 200
+
```
LaurenQurashi, line 2482: If anyone knows how to make this test a bit smaller that would be great. I couldn't parameterise is as I needed the measure id's that didn't exist yet in the params so I thought list iteration would be the next best work around.

measures/views.py
```diff
@@ -133,24 +134,62 @@ def form_valid(self, form):
...
+
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
         page = self.paginator.get_page(self.request.GET.get("page", 1))
         kwargs["objects"] = page.object_list
         return kwargs
 
+    def cleaned_query_params(self):
+        # Remove the sort_by and ordered params in order to stop them being duplicated in the base url
+        if "sort_by" and "ordered" in self.filterset.data:
```
LaurenQurashi, line 185: The base url is used for the sort_by in the template.. I needed the query params from the filters in there, but if I didn't remove the last sort_by and Ordered from it, it would get bundled in with the rest of them so I have to get rid before resetting the base url.

common/jinja2/components/sort_by.jinja
```diff
@@ -1,14 +1,21 @@
 
-{% macro sort_by(request, sort_by, title, base_url, anchor="") %}
+{% macro sort_by(request, sort_by, title, base_url, query_params="False", anchor="") %}
```
paulpepper-trade, line 2: I know it's not your naming, Lauren, but maybe this macro would be more helpfully named something like `create_anchor` or `create_sortable_anchor`, since `sort_by` is kind of incidental to its main purpose of creating an anchor element.

Similarly with the `anchor` param naming. It isn't clear what that is or what values it should take. I don't think it's an anchor in the HTML sense! I think it's order "direction" isn't it?

paulpepper-trade, line 2: Should the `query_params` parameter take a default value having type `bool` rather than a string - `query_params=False`?

common/jinja2/components/sort_by.jinja
```diff
@@ -1,14 +1,21 @@
...
 
-  {% if request.GET.sort_by == sort_by and request.GET.order == "desc" %}
-    <a class="govuk-link govuk-link--no-visited-state sort-icon--down" href="{{ base_url }}?sort_by={{ sort_by }}&order=asc{{ anchor }}">
-      {{ title }} <img src="{{ static('common/images/sort_icon.svg') }}" alt="">
-    </a>
+  {% if query_params %}
```
dalecannon, line 4: I think the default value of `query_params` ought to be a boolean rather than a string otherwise the conditional will always evaluate as `True`.

measures/tests/test_views.py
```diff
@@ -2478,3 +2479,87 @@ def test_measure_list_redirects_to_search_with_no_params(valid_user_client):
...
+        geographical_area=area_3,
+        valid_between=date_ranges.earlier,
+    )
+    measure3 = factories.MeasureFactory.create(
+        measure_type=type3,
+        geographical_area=area_1,
+        valid_between=date_ranges.later,
+    )
+
+    sort_by_list = [
```
paulpepper-trade, line 2524: You could perhaps pull this list out as a set of parameters using [pytest's parameterize](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html).

measures/views.py
```diff
@@ -133,24 +134,62 @@ def form_valid(self, form):
...
             return HttpResponseRedirect(reverse("measure-ui-search"))
         return super().dispatch(*args, **kwargs)
 
+    def get_queryset(self):
+        queryset = super().get_queryset()
+
+        ordering = self.get_ordering()
+
+        if ordering:
+            if isinstance(ordering, str):
```
paulpepper-trade, line 168: I think you could slightly simplify this by `and`ing these `if`s:
```suggestion
        if ordering and isinstance(ordering, str):
```

paulpepper-trade: Although, would `ordering` ever be anything other than type `str` if it's truthy value is True?

measures/tests/test_views.py
paulpepper-trade: On reading through this test again, perhaps there isn't actually much to be gained from parameterising it.

paulpepper-trade (approved):

PR #1035, opened 2023-09-13
TP2000-976: Fix flaky test

eadpearce:
# TP2000-976: Fix flaky test

## Why

* Rerunning tests is annoying

## What

* Ensures validity dates and transactions are explicitly set on test fixtures so that ordering will be the same every time


a-gleeson (approved):
LGTM, make sense!

PR #1036, opened 2023-09-14
Notifications model and goods report notification

a-gleeson:
# TP2000-975 Notifications model and goods report notification

## Why
The purpose of the story was to implement a notification to send to channel islands onces a goods report has been reviewed. This is a synchronously sent notification that should only be sent once, which means we need some sort of way to track if a notification has been triggered. The current implmenetation of notifications made this difficult. 
We will also be expanding notificaitons as we bring in other users and implement a larger workflow.

## What
- Notify module for gov notify specific functions
- Introduced a generic Notfiication model which implements 
  - Has field notified_object_pk (ok of object which notification relates to)
  - Has field notification_type which is a test choices fields for the different type of notificaitons
  - send_notification_emails() sends the email via the notify module and creates notification logs tied to this notification object
  - return_subclass_instance() returns the subclass instance of the notfication object
  - synchronous_send_emails() triggers celery task directly
  - schedule_send_emails() triggers celery task asyncly
- Proxy Notification Sub classes implemented for each type of notifiaction
  - has it's own init to set notification type
  - has it's own method to retrieve linked object
  - has it's own email personalistation which it gets details from linked object
  - maintains it's own gov notify template id 
  - has it's own notfied_user() function to retrieve it's email list
  - GoodsSuccessfulImportNotification get_personalisation() calls notify.prepare_link_to_file() to prepare attachment on email
- Tests to cover this model and the celery task trigger
- A new conformation screen to confirm notification for good report sent
- A view and button on the goods report tab to send the notification
- If an import is a commodity import disable the send to packaging button until a goods report notification is sent
- Tests to cover the views

NotificationProxyModel(1)
image
image
image
Button disappears after notfiication sent
image
image
image
 

## Checklist
- Requires migrations? - Ues

Links to relevant material
See: [Options for notification model](https://uktrade.atlassian.net/wiki/spaces/T/pages/3586555958/Notification+Technical+Design)


workbaskets/tests/test_views.py
```diff
@@ -1285,3 +1393,93 @@ def test_workbasket_compare_found_measures(
...
+        #     mock_xlsx_open,
+        # ):
+        response = valid_user_client.get(
+            reverse("workbaskets:workbasket-ui-review-goods"),
+        )
+
+    assert response.status_code == 200
+    soup = BeautifulSoup(str(response.content), "html.parser")
+
+    # notify_button = soup.find("a", href=f"/notify-goods-report/{import_batch.id}/")
```
CPrich905, line 1478: left over comment

workbaskets/tests/test_views.py
```diff
@@ -1285,3 +1393,93 @@ def test_workbasket_compare_found_measures(
...
+    def mock_xlsx_open(filename, mode):
+        if os.path.basename(filename) == "goods.xlsx":
+            return mock_open().return_value
+        return open(filename, mode)
+
+    with patch(
+        "importer.goods_report.GoodsReport.xlsx_file",
+        return_value="",
+    ) as mocked_xlsx_file:
+        # with patch(
```
CPrich905, line 1467: left over comment

notifications/models.py
```diff
@@ -10,17 +20,447 @@ class NotifiedUser(models.Model):
     email = models.EmailField()
     enrol_packaging = models.BooleanField(default=True)
     enrol_api_publishing = models.BooleanField(default=False)
+    enrol_goods_report = models.BooleanField(default=False)
 
-    def __str__(self):
-        return self.email
+    def __str__(self) -> str:
```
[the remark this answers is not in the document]
a-gleeson: Probably, though it may just be good to give most things a signature but this is a different type of method

workbaskets/tests/test_views.py
a-gleeson: This is a skipped test that I wasn't able to get working, I should currently leave these comments as they might be used when the test is working

PR #1014, opened 2023-08-22
TP2000 958 implement the improved find and edit measures design search page

CPrich905:
# TP2000-958 implement the improved find and edit measures design search page

nb: Geographical areas should not have been omitted from the design; The `Clear` option should be a button, not a link

## Why

This is a repeat of [PR  995](https://github.com/uktrade/tamato/pull/995) without the active/live measures search & associated tests. This has been descoped and will be picked up in a later ticket.

## What

1. Reorganised the search and filter options, implemented some small wording changes to match the figma design on the ticket. Adding a new option to search by `Certificates`
2. Adding a new filter section to contain `Include inherited measures for specific commodity code` and a new filter on a radio button:  `Only show measures in this current workbasket` (the `Show live measures only` filter will be added here in a future ticket)
3. Disable the Include inherited measures for a specific commodity code if there is no commodity code.
4. Styling changes

Additionally: refactored `kwargs["data"]._mutable = True` and modifying it directly in favour of making a copy of `kwargs["data"]` modifying the copy and re-adding to `kwargs` - see [this conversation](https://github.com/uktrade/tamato/pull/995#discussion_r1299781069) for further context.

## Checklist

- [ ] Do the new filters work as intended?
- [ ] Does the styling match the figma design?
- [ ] Does the conditional rendering on `Include inherited measures` work as intended, including when a commodity code pre-populates (eg. when reaching this page from `/commodities/{commodity_code}/measures-as-defined/` and choosing `View in find and edit measures`)
- [ ] Confirm the other search/edit pages have not been impacted by the changes to this design.
- [ ] Confirm tests pass

Screenshot 2023-08-24 at 12 57 41

Screenshot 2023-08-24 at 12 57 53

PR #1038, opened 2023-09-15
TP2000-1043 Final implementation changes

CPrich905:
# TP2000-1043 Final implementation changes

## Why
Following the roll out of [TP2000-958](https://github.com/uktrade/tamato/pull/1014) the Tariff Managers requested some additional changes to the search and filter page.

## What
Implementing the changes requested on the ticket.

Accordion:
- Hard lock it on “open” when a user first selects “Find and edit measures”
- Remove the black “Search and filter” header
- Change the blue font on the “accordion” to black and change the text to read “Search and filter”
- Add text underneath the accordion title but above the filter boxes which will read “Select one or more options to search”
- When a search is returned, the default behaviour of the search box is to be closed.

Other:
- Bring back the footnotes filter on a new fourth row, the filter should sit below regulations. 
- Help text should be added to the filter for “Include inherited measures” that explains this option only works when the filter “Specific commodity code” is used
- Remove the grey background from the Time periods area

Note: the `initOpenCloseAccordionSection` will require a little refactor for it to be used with the other search/filter pages.
Currently it checks that (1) there is an AccordionSection and (2) the pathname includes `measures/search` to ensure there is no contamination with other AccordionSections throughout the site.
Suggestion is to add a classname to the Accordion section & ensure the url path for all search pages includes `/search/`. The AccordionSections can then be closed by default and opened on search pages by checking the pathname.

Before filtering
Screenshot 2023-09-15 at 16 14 51

After filtering
Screenshot 2023-09-15 at 16 14 36

[TP2000-958]: https://uktrade.atlassian.net/browse/TP2000-958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


a-gleeson:
Accessing from Find Measures.
image
opens in collapsed 
image

common/static/common/js/openCloseAccordion.js
```diff
@@ -0,0 +1,18 @@
...
+const initOpenCloseAccordionSection = () => {
+    document.addEventListener('DOMContentLoaded', function() {
+
+        let expandedSection = document.getElementsByClassName('govuk-accordion__section')
+        let pathname = window.location.pathname
+        if(expandedSection.length > 0){
+            if (pathname.includes('measures/search')){
+                expandedSection[0].classList.add('govuk-accordion__section--expanded');
+            } else if (pathname.includes('measures') && !pathname.includes('search')){
+                expandedSection[0].classList.remove('govuk-accordion__section--expanded');
```
a-gleeson, line 11: still morning so struggling to think if it is possible but it may be possible to add this within the crsipy form class depending on the page loaded. Thinking about it that may have to be done depending on the view and having to access the Form and gets complicated fast. Just trying to consider JS alternatives

common/static/common/js/openCloseAccordion.js
```diff
@@ -0,0 +1,18 @@
+// TODO: adjust the pathname filter to account for all objects search/filter pages
+const initOpenCloseAccordionSection = () => {
+    document.addEventListener('DOMContentLoaded', function() {
+
+        let expandedSection = document.getElementsByClassName('govuk-accordion__section')
```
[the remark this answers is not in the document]
CPrich905: Good call, I've actioned this.

common/static/common/js/openCloseAccordion.js
CPrich905: I think possibly? I'm still building familiarity around django/crispy forms so I'd have to have a play about. This approach made sense to me as the search form is still available in the list view, we just want it to default to close & the most obvious way to do this to me at the time was to check the pathname. 
My intention when the changes are rolled out to the rest of the search/filter pages, is to see if we can update the url to be `search/` and results page will be `list/`, then rather than check for the absence of search, check for the presence of list and close the accordion based on that, which I think would be less repetitive/vulnerable than adding it directly to the crispy form (happy to hear any arguments against though!

a-gleeson: I agree the approach makes sense to me, if I used JS I would go with this approach for sure.
The suggestion of a non-js approach is general common for Government web pages (I know our app is internal), I know from experience using an app with JS with a spotty connect can be furstrating or just not work. 
I am not sure what you mean by vulnerable here?
I think as this is setting the state on load and not throughout activity on the page JS may not be needed.

common/static/common/js/openCloseAccordion.js
```diff
@@ -0,0 +1,18 @@
...
+const initOpenCloseAccordionSection = () => {
+    document.addEventListener('DOMContentLoaded', function() {
+
+        let expandedSection = document.getElementById('accordion-open-close-section')
+        let pathname = window.location.pathname
+        if(expandedSection){
+            if (pathname.includes('search')){
+                expandedSection.classList.add('govuk-accordion__section--expanded');
+            } else if (!pathname.includes('search')){
+                expandedSection.classList.remove('govuk-accordion__section--expanded');
```
a-gleeson, line 11: This is a little pseudo cody but you could add something like this to the views for search and list
```
def get_form(self):
    form - super().get_form()
    
    nestest_div = form.helper.layout.find_by_id("accordion-open-close-section")
    current_class = nested_div.attrs.get("class", "")
    new_class = f"{current_class} govuk-accordion__section--expanded".strip()
    nested_dic.attrs["class"] = new_class
    return form
```
```suggestion
                expandedSection.classList.remove('govuk-accordion__section--expanded');
```

a-gleeson (approved):
LGTM!

PR #1039, opened 2023-09-18
TP2000-1048  Add geo area exclusions step to multiple measure edit journey

dalecannon:
# TP2000-1048  Add geo area exclusions step to multiple measure edit journey

## Why
Users require the ability to edit the geographical area exclusions of multiple measures. 

## What
- Adds `MeasuresGeographicalAreaExclusionsForm`
- Adds geographical area exclusions step to `MeasureEditWizard`
- Adds view, form unit tests

##
Screenshot 2023-09-20 at 16 39 37

Screenshot 2023-09-20 at 16 39 57


measures/constants.py
```diff
@@ -1,6 +1,7 @@
 from django.db import models
 
 START = "start"
+GEOGRAPHICAL_AREA_EXCLUSIONS = "geographical_area_exclusions"
```
a-gleeson, line 4: Could we not use the `label` or `value`within the `MeasureEditSteps` instead of a seperate contsant?

measures/views.py
```diff
@@ -328,6 +345,111 @@ def get_form_kwargs(self, step):
...
+        measure: Measure,
+        exclusions: List[GeographicalArea],
+        workbasket: WorkBasket,
+    ):
+        """Updates the excluded geographical areas associated to the measure."""
+        existing_exclusions = measure.exclusions.current()
+
+        if not edited:
+            for exclusion in existing_exclusions:
+                exclusion.new_version(
```
a-gleeson, line 389: why do we create a new version if it's not edited?

a-gleeson (approved):
A couple queries but other than that LGTM!

measures/views.py
dalecannon: My understanding is that measures are associated to exclusions via foreign key (pk) which changes when the measure is updated (a new version is created).

a-gleeson: ah so the exclusion needs to reference the new version of the measure not the old? My understanding is a little spotty but that sounds fair to me. I guess if we have updated measures in the past we always update exclusions to reference the new version?
 Maybe a little comment like "update exclusion to new measure version".

measures/views.py
```diff
@@ -328,6 +345,111 @@ def get_form_kwargs(self, step):
...
+                    modified_measure=measure,
+                    workbasket=workbasket,
+                )
+            return
+
+        if not exclusions:
```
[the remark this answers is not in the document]
dalecannon: You're right. I've removed this if block.

PR #1040, opened 2023-09-18
fix: add apsw package back in due to latest update causing failure

nboyse:
# TP N/A

## Why

http://sentry.ci.uktrade.digital/organizations/dit/issues/83955/?project=200
Caused by latest apsw release: https://pypi.org/project/apsw/#history

As we removed the direct relation in this PR:
https://github.com/uktrade/tamato/commit/7f4df66963001241be2a590775d637479faf3428

And the dependency from [sqlite-s3vfs](https://github.com/uktrade/sqlite-s3vfs/commit/98da943566704277fafa9590be076a7a2d2686b8) does not specify a version. So we need to pin it and perhaps resolve this in sqlite-s3vfs at a later date (I think we are the only team actively using this package so will ask DI)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 aiohttp==3.8.5
+apsw==3.43.0.0
 aioresponses==0.7.4
 allure-pytest-bdd==2.8.40
 api-client==1.3.0


a-gleeson (approved):
LGTM!

PR #1043, opened 2023-09-19
footnote description fix for MultipleObjectsReturned error

a-gleeson:
# TP2000-1054 footnote description fix for MultipleObjectsReturned error

## Why
When trying to create a footnote description for a footnot with multiple versions the .get() returns more than one result

## What
added `.current()` to only return the latest version of the footnote. So the footnotes/xxxxx/description-create/ page does not error when trying to access it.


a-gleeson:
Duplicate work

PR #1044, opened 2023-09-19
TP2000-1054  Fix DescriptionCreate MultipleObjectsReturned error

dalecannon:
# TP2000-1054 Fix DescriptionCreate MultipleObjectsReturned error

## Why
`FootnoteDescriptionCreate` view raises a `MultipleObjectsReturned` error when it encounters a footnote with more than one version.

## What
- Uses current() to return the current version of the object only
- Prevents the same issue from occurring in `AdditionalCodeDescriptionCreate`
- Adds view unit tests


a-gleeson (approved):
LGTM!

PR #1047, opened 2023-09-21
fixed get packaged workbasket bug

a-gleeson:
# TP-??? Your PR title here

## Why
In the notification model when accepting an envelope theres a bug that cause the notification to fail

## What
- Added unit tests for those models
- Fixed the `.get` syntax

### DIFF
--- a/notifications/models.py
+++ b/notifications/models.py
@@ -393,9 +393,9 @@ class CrownDependenciesEnvelopeFailedNotification(Notification):
         )
 
     def get_personalisation(self) -> dict:
-        self.notified_object()
+        crown_dependicies_envelope = self.notified_object()
         personalisation = {
-            "envelope_id": self.packagedworkbaskets.last().envelope.envelope_id,
+            "envelope_id": crown_dependicies_envelope.packagedworkbaskets.last().envelope.envelope_id,
         }
         return personalisation
 
--- a/notifications/models.py
+++ b/notifications/models.py
@@ -283,7 +283,7 @@ class EnvelopeAcceptedNotification(Notification):
         from publishing.models import PackagedWorkBasket
 
         return (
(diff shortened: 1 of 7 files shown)


notifications/models.py
```diff
@@ -283,7 +283,7 @@ def notified_object(self) -> models.Model:
         from publishing.models import PackagedWorkBasket
 
         return (
-            PackagedWorkBasket.objects.get(self.notified_object_pk)
+            PackagedWorkBasket.objects.get(pk=self.notified_object_pk)
```
paulpepper-trade, line 286: I've done this so many times!

a-gleeson: I did it for the other notifications and fixed it, this one missed the mark :D

notifications/tests/test_models.py
```diff
@@ -73,6 +73,62 @@ def test_create_packaging_notification(ready_for_packaging_notification):
...
+        == "Loading report(s): REPORT_DBT23000.html"
+    )
+    assert "comments" in content
+
+
+def test_create_rejected_envelope(rejected_packaging_notification):
+    """Test that the creating a notification correctly assigns users."""
+
+    expected_present_email = f"packaging@email.co.uk"  # /PS-IGNORE
+    expected_not_present_email = f"no_packaging@email.co.uk"  # /PS-IGNORE
```
paulpepper-trade, line 108: Is it worth factoring out these email addresses in some way to allow them to be accessible from conftest / fixture info? It would minimise the horrible `PS-IGNORE` directive and lessen the special knowledge relationship between the test and the fixture's data. Similarly for other uses below.

a-gleeson: Yeah they are there, I guess I could return a tuple of data from the fixtures

PR #1050, opened 2023-09-25
TP2000-1067  Fix measure editing footnote save

dalecannon:
# TP2000-1067  Fix measure editing footnote save

## Why
Editing a measure to add a new footnote requires pressing "Add another footnote" before "Save" otherwise the footnote is not added to the measure.

## What
- Checks submitted form data for footnotes other than those already found in session data (i.e those that have been added via "Add another footnote" button)
- Adds test
- Moves formset prefixes to constants file
- Tidy up footnote formset template

##
Screenshot 2023-09-25 at 15 21 46


measures/forms.py
```diff
@@ -692,10 +689,22 @@ def save(self, commit=True):
...
+        # Footnotes added via "Add another footnote" button
         footnote_pks = [
-            dct["footnote"]
-            for dct in self.request.session.get(f"formset_initial_{sid}", [])
+            form["footnote"]
+            for form in self.request.session.get(f"formset_initial_{sid}", [])
         ]
+
+        # Footnote submitted directly via "Save" button
+        if hasattr(self.request, "POST"):
```
paulpepper-trade, line 699: Are there circumstances when this `save()` is called from something other than servicing a POST request?

paulpepper-trade: If not, then it may be preferable to let things blow up on invalid access to `self.request.POST`, avoiding silent failures.

paulpepper-trade (approved):

measures/forms.py
dalecannon: There were a couple form tests that called the form's `save()` method without a POST request, but I've updated them and removed this conditional.

PR #1052, opened 2023-09-25
TP2000-1069 - Packaging queue management command

paulpepper-trade:
# TP2000-1069 - Packaging queue management command

## Why

There is no application capability to manage and override PackagedWorkBasket instances in the packaging queue. This is useful in the interm (until Celery issues are resolved) to fix queue problems associated with failing Envelope generating Celery tasks.

## What

This PR adds a Django management command to view and manipulate `PackagedWorkBasket` instances in the packaging queue. Only those instances with a processing_state of `AWAITING_PROCESSING` are be candidates for repositioning. Instances with a `processing_state` of `CURRENTLY_PROCESSING` should remain unchanged to avoid surprising queue consumers.

## Checklist
- Requires migrations? No


publishing/management/commands/packaging_queue.py
```diff
@@ -0,0 +1,117 @@
...
+from publishing.models.packaged_workbasket import ProcessingState
+
+
+class Command(BaseCommand):
+    help = (
+        "View and manage the packaging queue. The bare command prints the "
+        "state of queued PackageWorkBasket instances, including those in both "
+        "that has state AWAITING_PROCESSING. Instances in state "
+        "CURRENTLY_PROCESSING will be unchanged to avoid surprises for "
+    )
```
dalecannon, line 15: *for queue consumers (or equivalent)?

dalecannon (approved):

publishing/management/commands/packaging_queue.py
paulpepper-trade: Done!

PR #1054, opened 2023-09-26
adding debounce to submit new measure

CPrich905:
# TP-2000-468 Double click error on measure create complete

## Why
The preventDoubleClick functionality has a time out on it. Occasionally, we exceed this timeout and the user is then able to double click the button, creating 2x the number of intended measures.

## What
We have duplicated the data-prevent-double-click functionality, extending the timer to 5 seconds. Should we discover that some events are taking longer than 5 seconds, we can extend this timer again.

### DIFF
--- a/common/static/common/js/buttonDebounce.js
+++ b/common/static/common/js/buttonDebounce.js
@@ -1,12 +1,12 @@
 let button = document.getElementById('add-debounce')
+let infoText = document.getElementById('add-debounce-inset-text')
 
 const debounceButton = () => {
     if (button) {
-        console.log('button', button)
         addEventListener('submit', function() {
             button.disabled = true,
-            alert("Your action is being processed and the submit button is now disabled. You will be taken to a confirmation screen once the action has been processed.")}
-        )
+            infoText.classList.remove('js-hidden')
+        })
     }
 }
 
--- a/common/static/common/js/buttonDebounce.js
+++ b/common/static/common/js/buttonDebounce.js
@@ -1,13 +1,24 @@
-let button = document.getElementById('add-debounce')
-let infoText = document.getElementById('add-debounce-inset-text')
+const tapDebounce = function(event) {
+    console.log('b fires')
+    let DEBOUNCE_TIMEOUT_IN_SECONDS = 5;
 
-const debounceButton = () => {
-    if (button) {
-        addEventListener('submit', function() {
-            button.disabled = true,
-            infoText.classList.remove('js-hidden')
-        })
+    if (this.tapDebounceFormSubmitTimer) {
+        console.log('c fires')
+        event.preventDefault();
+        button.disabled = true
+        return false
     }
-}
 
-export default debounceButton;
+
+    this.tapDebounceFormSubmitTimer = setTimeout(function () {
+        this.tapDebounceFormSubmitTimer = null;
+    }.bind(this), DEBOUNCE_TIMEOUT_IN_SECONDS * 1000);
+};
+
+document.querySelectorAll("[data-prevent-double-click").forEach(
+    (element) => {
+        element.addEventListener('click', tapDebounce);
+    }
+)
+
+export default tapDebounce;
--- a/common/static/common/js/buttonDebounce.js
+++ b/common/static/common/js/buttonDebounce.js
@@ -26,7 +26,7 @@ const tapDebounce = (event) => {
 
 
 const initTapDebounce = () => {
-  document.querySelectorAll("[data-prevent-double-click").forEach(
+  document.querySelectorAll("[data-prevent-double-click]").forEach(
     (element) => {
         
       element.addEventListener('click', tapDebounce);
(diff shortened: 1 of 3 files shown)


measures/jinja2/measures/create-review.jinja
```diff
@@ -130,5 +130,5 @@
     {% endif %}
   {% endcall %}
 
-  {{ govukButton({"text": "Create", "preventDoubleClick": true,}) }}
+  <span class="add-debounce">{{ govukButton({"text": "Create", "preventDoubleClick": true, css_class:"test"}) }}</span>
```
paulpepper-trade, line 133: If we're replacing the `preventDoubleClick` capability (assuming we are), then would it be cleaner to apply the `add-debounce` to the `govukButton` element on its `classes` attr?

CPrich905: Yes it would - I went a little code blind/foolish while working on this and couldn't assign a class to the `govukButton` - this was entirely a PICNIC problem (Problem In Chair...)
I've actually used the ID as using a class resulted in some funky behaviour (I'm not sure if this is due to the way the FormWizard works or if this is another PICNIC) however this also resolves the issue of using `<span>` tags and indexing the result of getElementsByClassName.

common/static/common/js/buttonDebounce.js
```diff
@@ -0,0 +1,14 @@
+let button = document.querySelectorAll('span.add-debounce > button')[0]
+
+const debounceButton = () => {
+    if (!button) {
+        return
+    } else if (button) {
```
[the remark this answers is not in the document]
CPrich905: removed

common/static/common/js/buttonDebounce.js
```diff
@@ -0,0 +1,14 @@
+let button = document.querySelectorAll('span.add-debounce > button')[0]
```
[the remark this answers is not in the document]
CPrich905: Yes - resolved (see later comment)

common/static/common/js/buttonDebounce.js
```diff
@@ -0,0 +1,14 @@
...
+
+const debounceButton = () => {
+    if (!button) {
+        return
+    } else if (button) {
+        addEventListener('submit', function() {
+            button.disabled = true,
+            alert("Your action is being processed and the submit button is now disabled. You will be taken to a confirmation screen once the action has been processed.")}
```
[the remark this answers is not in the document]
CPrich905: Hmm, good question. I'd originally thought to implement a loading spinner, which Ash & Wren had agreed to, however I'd missed the conversation on the ticket about adding the debounce & disabling the button. I had felt that disabling the button without explanation would raise problems, so added the alert but forgot to check with them. I'll reach out to them now and see what their thoughts are.

common/static/common/js/buttonDebounce.js
```diff
@@ -0,0 +1,13 @@
+let button = document.getElementById('add-debounce')
```
paulpepper-trade, line 1: I think using the `id` attribute restricts this functionality to one button per page, because `id` attributes should be unique within the context of a page:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id
Because we have some pages with multiple submit buttons, I don't think we'd want to do that.

CPrich905: Understood and yes, that makes sens. I'll have another play with this then to see if I can resolve the funkiness with the class attribute

common/static/common/js/buttonDebounce.js
```diff
@@ -1,22 +1,31 @@
...
+        target.style.cursor = targetDefaultCursor;
+      }.bind(target),
+      DEBOUNCE_TIMEOUT_IN_SECONDS * 1000
+    );
+    target.style.cursor = "wait";
 };
 
-document.querySelectorAll("[data-prevent-double-click").forEach(
+
+const initTapDebounce = () => {
```
paulpepper-trade, line 22: It's probably worth adding a docstring either at the top of the file or to the exported function. Maybe something about why there's a TAP-specific debounce and what it does?

paulpepper-trade (approved):

common/static/common/js/buttonDebounce.js
```diff
@@ -0,0 +1,31 @@
...
+        target.style.cursor = targetDefaultCursor;
+      }.bind(target),
+      DEBOUNCE_TIMEOUT_IN_SECONDS * 1000
+    );
+    target.style.cursor = "wait";
+};
+
+
+const initTapDebounce = () => {
+  document.querySelectorAll("[data-prevent-double-click").forEach(
```
paulpepper-trade, line 23: Oh, just noticed that there's a closing brace missing in the selector string!
```suggestion
  document.querySelectorAll("[data-prevent-double-click]").forEach(
```

common/static/common/js/buttonDebounce.js
CPrich905: Added :)

PR #1056, opened 2023-09-29
TP2000-1053  Implement workbasket review tab redesign

dalecannon:
# TP2000-1053  Implement workbasket review tab redesign

## Why
A redesign of the current workbasket review tabs was proposed following user research to improve peer review and quality assurance within TAP. 

## What
- Consolidates review goods and review measures tabs into one review tab
- Adds other nested tabs under review tab for additional codes, certificates, footnotes, geographical areas, quotas, regulations
- Filters objects in tab by update type
- Adds and updates view unit tests
- Renames commodity column on goods report

##

## Before:
Screenshot 2023-09-29 at 10 40 54

## After:
Screenshot 2023-09-29 at 10 20 49

Screenshot 2023-09-29 at 10 17 50

Screenshot 2023-09-29 at 10 20 55

Screenshot 2023-09-29 at 10 21 06


SteveCN7:
Small content tweak, instead of "Filter results" - can we say 
"Filter by action"

publishing/static/publishing/scss/_publishing.scss
```diff
@@ -19,10 +19,6 @@ table.queued-envelopes .icon {
   display: inline-block;
   width: 1.4rem;
 }
-table.queued-envelopes .down-arrow,
```
paulpepper-trade, line 22: Sanity check: does the envelope download link still render okay?

dalecannon: Yeah, the link renders as before. Only the arrow span class styling has been moved to `common/static/common/scss/_text.scss`.

workbaskets/tests/test_views.py
```diff
@@ -494,20 +580,55 @@ def test_workbasket_measures_review(valid_user_client):
     assert not measure_end_dates.difference(table_end_dates)
 
 
-def test_workbasket_measures_review_pagination(
+@pytest.mark.parametrize(
+    ("update_type", "expected_result"),
```
paulpepper-trade, line 584: Maybe something like `expected_measure_count` would self document this param?

workbaskets/views/ui.py
```diff
@@ -777,3 +704,180 @@ def get_context_data(self, *args, **kwargs):
...
+        # Get actual values from the ImportBatch instance if one is associated
+        # with the workbasket.
+        try:
+            import_batch = self.workbasket.importbatch
+        except ObjectDoesNotExist:
+            import_batch = None
+
+        if import_batch and import_batch.taric_file:
+            reporter = GoodsReporter(import_batch.taric_file)
+            goods_report = reporter.create_report()
```
paulpepper-trade, line 783: I have recall that this has blown up (AKA 500-series response) in test / staging / dev environments, when using a DB from another environement. The DB representation of `import_batch.taric_file` saves the path value, which gives `import_batch.taric_file` a Truthy value, but if it was created in another environment there'll be no file stored against it in the current S3 environment. It's probably worth managing that situation more gracefully.

paulpepper-trade (approved):

SteveCN7:
One final observation Dale, can we make the font size in the tabs one size smaller - request from Wren?
Otherwise this is looking good to go

PR #1059, opened 2023-10-03
Tp2000-955  environment label

a-gleeson:
# TP2000-955 Add environment label to enviornments

## Why
To ensure that work is not done on the incorrect environment we should have a label for the given environment. As the URL is not enough to always spot which environment we are in.

## What
Using the env value stored in the context we display the environment banner for non production environments

image
image
image


common/static/common/scss/_environment.scss
```diff
@@ -0,0 +1,22 @@
...
+    width: 0;
+    height: 0;
+    border-left: 150px solid transparent;
+    border-top: 150px solid green;
+    color: white;
+    text-align: center;
+    font-size: 18px;
+}
+
+.text-wrapper {
```
CPrich905, line 14: I can't see any other examples of this classname so I don't think it's much of a problem, but could/should this have a more specific name, ie. `env-banner-text-wrapper` or could it be a nested class?

CPrich905 (approved):

common/static/common/scss/_environment.scss
a-gleeson: makes sense to me updated with nested css

common/jinja2/layouts/layout.jinja
```diff
@@ -53,6 +53,12 @@
     {{ breadcrumbs(request, [{"text": page_title}]) }}
   {% endif %}
   {% endblock %}
+
+  {% if env != "production"%}
+    <div class="environment-banner"> 
+      <div class="text-wrapper">{{ env|upper}}</div>
```
eadpearce, line 59: You can leave out the `|upper` here and use `text-transform: uppercase` in the style instead. I think it's slightly nicer for screenreaders.

eadpearce: Also! For the font, you can get it to match the rest of the page using the `govuk-body` class though you might need to tweak the size

PR #1064, opened 2023-10-05
fixing certificates filter

CPrich905:
# Clean up of Certificates filter on Measures Search page

## Why
Some clean up of the Certificate filter implemented while investigating a bug with TAP data.

## What
The value returned by the filter is a Certificate object meaning we don't need to retrieve the certificate object with 
`certificates = Certificate.objects.filter(sid=value.sid)`
Filtering by `trackedmodel_ptr_id` saves passing the whole object.


mattjamc (approved):
Looks good to me :)

measures/filters.py
```diff
@@ -257,24 +257,12 @@ def measures_filter(self, queryset, name, value):
...
-            measure_ids = set()
-            certificates = Certificate.objects.filter(sid=value.sid)
-            for certificate in certificates:
-                measure_conditions = MeasureCondition.objects.filter(
-                    required_certificate=certificate,
-                )
-
-                for condition in measure_conditions:
-                    measure_ids.add(condition.dependent_measure_id)
+            measure_ids = MeasureCondition.objects.filter(
```
paulpepper-trade, line 263: I think this will get al versions, which is probably not desirable. It's unusual to reference the low-level `trackedmodel_ptr_id` too.

paulpepper-trade (changes requested):

measures/filters.py
CPrich905: I might be mistaken, but in the test I ran it returned the same number of results as the previous filter. I may have caught an edge case however - happy to jump on a call and we can discuss?

measures/filters.py
```diff
@@ -257,24 +257,12 @@ def measures_filter(self, queryset, name, value):
...
+                required_certificate=value.trackedmodel_ptr_id,
```
paulpepper-trade, line 264: Using `trackedmodel_ptr_id` will reference a specific version of this Measure. It's necessary to use either the `TrackedModel.identifying_fields` (`sid` in the case of `Measure`) or `TrackedModel.version_group` - they're generally, functionally equivalent when it comes to getting the versions of any `TrackedModel`. You can then use `TrackedModelQuerySet.current()` and related methods to get the correct version.

measures/filters.py
paulpepper-trade: It looks as though `measures.tests.test_filters.test_filter_by_certificates` doesn't test for the case of measures with multiple versions - only multiple measures each with a single version.

measures/filters.py
paulpepper-trade, line 264: I think it would be enough to replace this with

```required_certificate=value
```

PR #1066, opened 2023-10-09
TP2000-1071  Implement workbasket detail view redesign

dalecannon:
# TP2000-1071  Implement workbasket detail view redesign

## Why
In preparation for the Workbasket Summary design improvements, the Workbasket Detail view (accessible via Find and View Workbasket) must undergo a redesign, bringing across the component list from the summary view from which users will be able to view and remove workbasket items.

## What
- Adds `WorkBasketDetailView` to display details of a workbasket under 'Details' tab
- Adds `WorkBasketChangesView` to display workbasket items and allow deletion under 'Changes' tab
- Enables sorting of workbasket items by component type, action or activity date
- Prevents users without `change_workbasket` permission from removing items in a workbasket or when the workbasket is not in EDITING state
- Adds two new URL routes (`<pk>/changes/delete/` and `<pk>/changes/confirm-delete/`) for deleting workbasket items from new Workbasket Changes view
- Modifies the existing `WorkBasketDeleteChanges` view to maintain deletion capability with Workbasket Summary view until the full redesign work is implemented
- Styles pagination as a 'Load more' button, returning the previous pages' plus the new page's results and displaying a progress indicator
- Adds view tests

## Before
Screenshot 2023-10-09 at 15 35 42

### Detail view
Screenshot 2023-10-09 at 15 22 01

### Changes view
Screenshot 2023-10-09 at 15 22 26

### Changes view pagination
Screenshot 2023-10-09 at 15 22 53


workbaskets/jinja2/workbaskets/changes.jinja
```diff
@@ -0,0 +1,183 @@
...
+
+{% block breadcrumb %}
+  {{ breadcrumbs(request, [
+      {"text": "Find and view workbaskets", "href": url("workbaskets:workbasket-ui-list-all")},
+      {"text": page_title}
+    ])
+  }}
+{% endblock %}
+
+{% macro checkbox_item(field) %}
```
eadpearce, line 18: Looks like we already have this macro defined in `measures/jinja2/includes/measures/list.jinja` and `workbaskets/jinja2/includes/workbaskets/workbasket-selectable-items.jinja` so this could probably be consolidated into the common folder

workbaskets/views/ui.py
```diff
@@ -189,7 +188,11 @@ class WorkBasketDeleteChanges(PermissionRequiredMixin, ListView):
 
     @property
     def workbasket(self) -> WorkBasket:
-        return WorkBasket.current(self.request)
+        workbasket_pk = self.kwargs.get("pk", None)
```
paulpepper-trade, line 191: I think it would be cleaner to use two concrete views to support the different URL routes for for deleting workbasket changes (one for current workbasket, the other to support an explicit workbasket primary key as a URL param). It's more (OO) idiomatic, removes the, albeit relatively small, amount of complexity that conditions scattered around the class bring. The common behaviour could be factored out into a base class, which may well be suitable as one of the two concrete views.

paulpepper-trade: Or... actuallly.. if it's possible... would it be preferable to use just the single route that takes the workbasket primary key as a captured URL parameter?

workbaskets/views/ui.py
```diff
@@ -476,40 +499,145 @@ def get_queryset(self):
...
-        context["page_obj"] = page_obj
-        context["is_paginated"] = True
-        context["object_list"] = items
-        context["page_links"] = build_pagination_list(
-            page_number,
-            page_obj.paginator.num_pages,
+    @property
+    def paginator(self):
+        return Paginator(
+            self.workbasket.tracked_models.select_related(
```
paulpepper-trade, line 537: Would it be worth factoring out this nice efficiency gain into a new method, something like `TrackedModelQuerySet.with_transactions()`?

paulpepper-trade: ... and perhaps also something for the model-specific `polymorphic_ctype`?

workbaskets/views/ui.py
dalecannon: Having just the single route would be better, yeah. I've now made that change.

workbaskets/views/ui.py
```diff
@@ -241,14 +246,23 @@ def post(self, request, *args, **kwargs):
...
 
-class WorkBasketDeleteChangesDone(TemplateView):
+class WorkBasketChangesConfirmDelete(TemplateView):
     template_name = "workbaskets/delete_changes_confirm.jinja"
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        session_workbasket = WorkBasket.current(self.request)
+        workbasket = WorkBasket.objects.get(pk=self.kwargs["pk"])
+        if workbasket.pk != session_workbasket.pk:
```
paulpepper-trade, line 262: You should be able to just compare the objects - https://docs.djangoproject.com/en/dev/topics/db/queries/#comparing-objects - which may avoid this blowing up if there's no session workbasket.

workbaskets/views/ui.py
```diff
@@ -241,14 +246,23 @@ def post(self, request, *args, **kwargs):
...
+            context["workbasket"] = workbasket
```
paulpepper-trade, line 263: Although, if the session workbasket (if any) were added to the context, then the conditional tests in the template (testing whether the view's workbasket is working with a session workbasket or not) may make intent a little clearer there?

paulpepper-trade (approved):

PR #1067, opened 2023-10-11
feat: Add correct path to allow unpublished quota order numbers to be edited

nboyse:
# TP2000-1079 Edit Quota Order Number link missing on newly created quota order numbers

## Why

When creating new quota order numbers, we can't edit them as the url pattern wasn't in place
## What

Adding two new URL patterns to expose on the front end the ability to edit new quota order numbers. One is autogenerated, one is explicitly added. This is because the quota edit page cascades down to edit the quota origin pages, so we needed to create a new URL pattern to handle edge cases.


quotas/urls.py
```diff
@@ -59,6 +59,11 @@
         views.QuotaOrderNumberOriginUpdate.as_view(),
         name="quota_order_number_origin-ui-edit",
     ),
+    path(
+        f"quotas/<sid>/edit-create/",
+        views.QuotaEditUpdate.as_view(),
+        name="quota-ui-edit-create",
```
dalecannon, line 65: It may possible to add a `QuotaEditCreate` view (like `QuotaEditUpdate`) that `get_ui_paths()` can use to generate the route automatically. Would it also be worth writing a test for this new view?

paulpepper-trade (approved):
It may help make the PR's intent clearer if the "what" part of the description explains that there are two new URL routes (one auto-generated and one explicitly added) and why they're both needed.

PR #1068, opened 2023-10-11
TP2000-1023 Measures tab for footnotes

mattjamc:
# TP2000-1023 Measures tab for footnotes

## Why

In the footnote detail view, users cannot currently view all measures associated with a footnote. This adds a measures tab to the detail view and changes the current set-up to fake tabs to allow for pagination of measures.
## What

- Changes existing detail tabs to fake tabs and adds a new tab for measures
- Allows sorting of measures by commodity code and start date
- Adds action menu to the measures tab with a link to view the footnote in find and edit measures
- Adds unit tests and updates existing bdd tests to reflect new fake tabs layout

### Currently
image

### Update
image


footnotes/views.py
```diff
@@ -159,6 +167,84 @@ class FootnoteConfirmCreate(FootnoteMixin, TrackedModelDetailView):
...
+    template_name = "includes/footnotes/tabs/measures.jinja"
+    paginate_by = 20
+    sort_by_fields = ["goods_nomenclature", "start_date"]
+    custom_sorting = {
+        "start_date": "valid_between",
+    }
+
+    @property
+    def footnote(self):
+        return Footnote.objects.get(
```
eadpearce, line 214: since you've already got `from footnotes import models` you can do `models.Footnote` here and remove the `from footnotes.models import Footnote` import

footnotes/views.py
```diff
@@ -159,6 +167,84 @@ class FootnoteConfirmCreate(FootnoteMixin, TrackedModelDetailView):
...
+            footnote_type__footnote_type_id=self.kwargs[
+                "footnote_type__footnote_type_id"
+            ],
+            footnote_id=self.kwargs["footnote_id"],
+        )
+
+    def get_queryset(self):
+        footnote_associations = FootnoteAssociationMeasure.objects.filter(
+            associated_footnote_id=self.footnote.trackedmodel_ptr_id,
```
eadpearce, line 223: you can just do `associated_footnote=self.footnote` here I think

footnotes/views.py
```diff
@@ -159,6 +167,84 @@ class FootnoteConfirmCreate(FootnoteMixin, TrackedModelDetailView):
...
+        ).values_list("footnoted_measure", flat=True)
+        queryset = (
+            Measure.objects.all()
+            .current()
```
eadpearce, line 227: if you're doing .current() and .filter() you don't need .all() as well

eadpearce: So I had a play around in the shell and you can simplify this a lot here if you set the view's model to Footnote and do `self.object.measure_set.all()`. It should get the same results

eadpearce (approved):

PR #1069, opened 2023-10-12
TP-2000-1073 Add a description tab to the Commodity code details view

CPrich905:
# TP-2000-1073-Add a description tab to the Commodity code details view

## Why
To provide users with a history of descriptions, allowing them to visually inspect the description data coming from EU TARIC imports.
Visually aligns with Add Codes/Footnotes/Certificates.

## What
Adds a descriptions tab to the Commodity Code details view. 
On click, the user is taken to a view for the associated descriptions.

NOTE: This does not include the edit/add functionality discussed in the ticket. Discussed with Steve 12/10 that this will be broken into a separate ticket pending further discussion.

Screenshot 2023-10-12 at 12 00 20


commodities/jinja2/includes/commodities/tabs/descriptions.jinja
```diff
@@ -0,0 +1,36 @@
+{% extends "commodities/detail.jinja" %}
+
+{% from 'macros/create_link.jinja' import create_link %}
+
+{% block tab_content %}
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+        {% set description_rows = [] %}
+        {% set descriptions = object.get_descriptions() %}
+        {% set is_only_description = descriptions|length == 1 %}
```
eadpearce, line 10: Seems like this isn't used anywhere

eadpearce (approved):

commodities/jinja2/includes/commodities/tabs/descriptions.jinja
CPrich905: Thanks, have removed :)

commodities/jinja2/includes/commodities/tabs/descriptions.jinja
```diff
@@ -0,0 +1,35 @@
...
+        {"text": "Start date", "classes":  "govuk-!-width-one-eighth"},
+        {"text": "Description"},
+        ]%}
+        {% if not omit_actions and description_rows.0|length >= 3 %}
+            {{ head.append({"text": "Actions"}) or "" }}
+        {% endif %}
+
+        <h2 class="govuk-heading-l">Descriptions</h2>
+        
+        {{ govukTable({
```
paulpepper-trade, line 29: I think I just missed the boat here... I think it's preferable (explicit better than implicit) to import a macro at the top of the template file in which it's used, rather than relying upon its parent to do that.

CPrich905: I'd based this on the various other tabs, which rely on inheritance. There will be some more work on this tab though, so I'm happy to update it then unless you'd prefer it to be done sooner than later? 
I can also update the other files to include the import for consistency.

paulpepper-trade: Yes, unfortunately there's a mixture of styles. I'd leave this as is for now.

PR #1072, opened 2023-10-13
Tp2000 959 measure list design changes

LaurenQurashi:
# TP-2000-959-measure-list-design-changes

## Why
This PR adds Regulations to the measures list search results table, as well as a list of filters that the user has selected. The list sits at the top of the search results under the result count. It should be 2 columns of 7 list items (one column if less are selected) written up in a human friendly way. A simple test to check for the lists and columns have been added. 

## What
Screenshot 2023-10-23 at 14 02 05

## Checklist
- Requires migrations? - no


measures/views.py
```diff
@@ -189,6 +195,112 @@ def cleaned_query_params(self):
...
+            footnote = Footnote.objects.current().get(id=selected_filters["footnote"])
+            selected_filters_strings.append(f"Footnote {footnote.structure_code}")
+
+        if "start_date_0" and "start_date_1" and "start_date_2" in selected_filters:
+            if selected_filters["start_date_modifier"] == "exact":
+                modifier = ""
+            else:
+                modifier = selected_filters["start_date_modifier"]
+            selected_filters_strings.append(
+                f"start date: {modifier} {selected_filters['start_date_0']}/{selected_filters['start_date_1']}/{selected_filters['start_date_2']}",
```
eadpearce, line 278: "start" here should be capitalised

measures/views.py
```diff
@@ -189,6 +195,112 @@ def cleaned_query_params(self):
...
+            )
+
+        if "end_date_0" and "end_date_1" and "end_date_2" in selected_filters:
+            if selected_filters["end_date_modifier"] == "exact":
+                modifier = ""
+            else:
+                modifier = selected_filters["end_date_modifier"]
+            selected_filters_strings.append(
+                f"end date: {modifier} {selected_filters['end_date_0']}/{selected_filters['end_date_1']}/{selected_filters['end_date_2']}",
```
eadpearce, line 287: "end" should be capitalised

measures/views.py
```diff
@@ -189,6 +195,112 @@ def cleaned_query_params(self):
...
+            )
+
+        if "modc" in selected_filters:
+            selected_filters_strings.append("include inherited measures")
```
eadpearce, line 291: These other filters should also probably begin with a capital letter

eadpearce (approved):

PR #1073, opened 2023-10-16
TP2000-980: Quota definition create

eadpearce:
# TP2000-980: Quota definition create

## Why

* Users need the ability to create new definition periods for an existing quota order number

## What

* Link to create a new definition on the quota order number detail page under the "Actions" list
* Single page form to create a new definition
* Success page

FireShot Capture 064 - Create a new quota definition period - Manage Trade Tariffs - localhost
Screenshot 2023-10-16 at 15 44 22

### DIFF
--- a/quotas/forms.py
+++ b/quotas/forms.py
@@ -391,7 +391,6 @@ class QuotaDefinitionCreateForm(
     class Meta:
         model = models.QuotaDefinition
         fields = [
-            "order_number",
             "valid_between",
             "description",
             "volume",
@@ -444,13 +443,8 @@ class QuotaDefinitionCreateForm(
     maximum_precision = forms.IntegerField(
         widget=forms.HiddenInput(),
     )
-    order_number = forms.ModelChoiceField(
-        queryset=models.QuotaOrderNumber.objects.all(),
-        widget=forms.HiddenInput(),
-    )
 
     def __init__(self, *args, **kwargs):
-        self.order_number = kwargs.pop("order_number")
         super().__init__(*args, **kwargs)
         self.init_layout()
         self.init_fields()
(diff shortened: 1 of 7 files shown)


SteveCN7:
Couple of things I think we might need to consider/update for both this design and the edit:

1. End dates should be enforced - Definition periods always run for specific periods of time
2. Some help text on what the two volume fields do

quotas/forms.py
```diff
@@ -287,24 +288,38 @@ class Meta:
...
     initial_volume = forms.DecimalField(
         widget=forms.TextInput(),
-        error_messages={"invalid": "Initial volume must be a number"},
+        error_messages={
+            "invalid": "Initial volume must be a number",
+            "required": "Enter the initial volume",
+        },
+    )
+    measurement_unit = forms.ModelChoiceField(
+        queryset=MeasurementUnit.objects.all(),
```
a-gleeson, line 304: As MeasurementUnit is a trcaked model would we want to use current() so the drop down only returns the latest version?

quotas/forms.py
```diff
@@ -369,6 +384,137 @@ def init_layout(self):
...
+    )
+    initial_volume = forms.DecimalField(
+        widget=forms.TextInput(),
+        error_messages={
+            "invalid": "Initial volume must be a number",
+            "required": "Enter the initial volume",
+        },
+    )
+    measurement_unit = forms.ModelChoiceField(
+        queryset=MeasurementUnit.objects.all(),
```
a-gleeson, line 422: Ditto comment above

quotas/jinja2/includes/quotas/actions.jinja
```diff
@@ -3,7 +3,10 @@
       <h2 class="govuk-heading-s" id="subsection-title">Actions</h2>
       <ul class="govuk-list govuk-!-font-size-16">
             <li><a class="govuk-link" href="{{ url('quota-definitions', args=[object.sid]) }}">View definition periods</a></li>
```
a-gleeson, line 5: should there be a `<br>` after this list element too?

eadpearce: No. The top two items are grouped together because they're both to do with definition periods

a-gleeson (approved):
LGTM!

PR #1075, opened 2023-10-20
TP2000-1078  Implement transaction order tab

dalecannon:
# TP2000-1078  Implement transaction order tab

## Why
A feature improvement to workbaskets is to introduce the capability to reorder workbasket transactions, allowing users to resolve business rule violations related to transaction ordering.

## What
**This PR is based on the experimental branch created by Paul**: [EXPERIMENTAL--move-tranx-within-workbasket](https://github.com/uktrade/tamato/compare/master...EXPERIMENTAL--move-tranx-within-workbasket)

- Creates a new tab on workbasket detail view to allow users with the requisite permissions to view transactions and reorder transactions in workbaskets in an editing state

- Adds view unit tests

- Stops an empty transaction from being left behind when creating a new certificate

## 
Screenshot 2023-10-20 at 14 01 34

### DIFF
--- a/workbaskets/views/ui.py
+++ b/workbaskets/views/ui.py
@@ -645,7 +645,7 @@ class WorkBasketChangesView(SortingMixin, WorkBasketChangesMixin):
 
 
 class WorkBasketTransactionOrderView(WorkBasketChangesMixin):
-    """UI endpoint for re-ordering transactions in a workbasket."""
+    """UI endpoint for reordering transactions in a workbasket."""
 
     template_name = "workbaskets/transaction_order.jinja"
 
@@ -754,7 +754,7 @@ class WorkBasketTransactionOrderView(WorkBasketChangesMixin):
     @atomic
     def promote_transaction(self, form_action):
         """Swap the transaction order of the promoted transaction with the
-        transaction above it (now demoted_transaction)."""
+        (demoted) transaction above it."""
         promoted_transaction = self._get_transaction_pk_from_form_action(form_action)
         demoted_transaction = (
             self.workbasket_transactions()
@@ -780,7 +780,7 @@ class WorkBasketTransactionOrderView(WorkBasketChangesMixin):
     @atomic
     def demote_transaction(self, form_action):
         """Swap the transaction order of the demoted transaction with the
-        transaction below it (now promoted transaction)."""
+        (promoted) transaction below it."""
         demoted_transaction = self._get_transaction_pk_from_form_action(form_action)
         promoted_transaction = (
             self.workbasket_transactions()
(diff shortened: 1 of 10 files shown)


workbaskets/views/ui.py
```diff
@@ -635,6 +614,216 @@ def get_success_url(self):
...
+            [demoted_transaction, promoted_transaction],
+            ["order"],
+        )
+
+        return HttpResponseRedirect(self.get_success_url())
+
+    def is_transaction_first_in_workbasket(self, transaction):
+        """Returns `True` if the transaction is the first in the workbasket,
+        `False` otherwise."""
+        return self.workbasket_transactions().first() == transaction
```
a-gleeson, line 809: would `workbasket_transactions()` be called for each transaction/ item in transaction? If this is the case would it have a negative impact on performance and page load time?
Could we cache the `workbasket_transactions()` as a property or something to avoid it querying the data base for each check within the db?

dalecannon: Ah yes, caching the first and last transaction would reduce database hits.

workbaskets/views/ui.py
```diff
@@ -635,6 +615,214 @@ def get_success_url(self):
...
+
+    @cached_property
+    def first_transaction_in_workbasket(self):
+        return self.workbasket_transactions().first()
+
+    @cached_property
+    def last_transaction_in_workbasket(self):
+        return self.workbasket_transactions().last()
+
+    def is_obj_first_in_transaction(self, tracked_model):
```
a-gleeson, line 815: would doing the same for is_obj_first_in_transaction and  is_obj_last_in_transaction help or is that only called once for each transaction?

workbaskets/views/ui.py
a-gleeson: changes look good

workbaskets/views/ui.py
dalecannon: `is_obj_first_in_transaction()` and `is_obj_last_in_transaction()` are called for each tracked model (obj) in the template, but I'm unsure if it's feasible to cache the results because the tracked models can belong to different transactions.

a-gleeson: that makes sense, there may be a more ideal way to create these results but you are correct in caching this particular value.
Do you think creating a different property that would be a list of tracked models for the first in a transaction from a single query would help performance? In the template you could then just check if the id is in the list and it would pass the check

paulpepper-trade: Is there a performance issue that needs fixing?

a-gleeson: not an issue and maybe premature optimisation at this point. It was just an observation that each tracked model on the list would execute a query to check if it is the first or last tracked model in the transaction. I was just wondering if there's a better way to reduce the query count as maybe for large workbaskets we could experience slow down

a-gleeson (approved):
LGTM!

PR #1078, opened 2023-10-24
TP-2000-1085 Replace View Violations and Worksheet checker with a blended tab

CPrich905:
# TP-2000-1085 Replace View Violations and Worksheet checker with a blended tab

## Why
To keep the workbasket design refreshed and up to date, we are combining `View violations` and the `Worksheet checker`.

## What
The View violations button and the Worksheet checker is now hosted on a single tab called `Check workbasket`
The `Run business rules` is triggered from this new page.
The `Worksheet checker` is accessed from this new page via the `Compare data` button.
The behaviour of both remains the same, with the addition of a formatting method on the duties sentence of the Worksheet checker. This ensures there are 3 decimal places after the value passed, e.g. `3.2%` becomes `3.200%`. Note that this will not show in the first table (Formatted worksheet data) but ensures a match against the measures returned.

image

Screenshot 2023-11-06 at 16 36 48


workbaskets/forms.py
```diff
@@ -156,7 +156,9 @@ def cleaned_data_no_prefix(self):
 class WorkbasketCompareForm(forms.Form):
     data = forms.CharField(
         label="Compare worksheet data against the measures in this workbasket",
-        widget=forms.Textarea(),
+        widget=forms.Textarea(
+            attrs={"placeholder": "Add your worksheet data here"},
+        ),
```
dalecannon, line 161: Would this be better as help text perhaps?

workbaskets/jinja2/includes/workbaskets/workbasket-business-rules.jinja
```diff
@@ -0,0 +1,142 @@
...
+  {% set rule_check_button = run_business_rules_form %}
+  {% set rule_check_block %}{% endset %}
+
+{% else %}
+  {% set live_rule_check_status = "No objects in workbasket" %}
+  {% set rule_check_button %}{% endset %}
+  {% set rule_check_block %}{% endset %}
+{% endif %}
+{% block content %}
+<div>
```
dalecannon, line 126: The `<div>` element and `block` tag are probably not required.

workbaskets/jinja2/workbaskets/checks.jinja
```diff
@@ -0,0 +1,36 @@
+{% extends "layouts/layout.jinja" %}
+{% from "components/button/macro.njk" import govukButton %}
+{% from "includes/workbaskets/navigation.jinja" import navigation %}
+{% from "macros/fake_tabs.jinja" import fake_tabs %}
```
dalecannon, line 4: This can be removed if it's unused.

workbaskets/jinja2/workbaskets/compare.jinja
```diff
@@ -19,7 +19,7 @@
       "items": [
       {"text": "Home", "href": url("home")},
       {"text": "Edit an existing workbasket", "href": url("workbaskets:workbasket-ui-list")},
-      {"text": "Workbasket " ~ request.session.workbasket.id ~ " - Summary" }
+      {"text": "Workbasket " ~ request.session.workbasket.id ~ " - Summary", "href": url("workbaskets:current-workbasket") }
```
dalecannon, line 22: Could the user's position be more accurately reflected, say, with breadcrumbs akin to `Home > Current workbasket > Check workbasket > Compare`.

workbaskets/jinja2/workbaskets/compare.jinja
```diff
@@ -29,15 +29,23 @@
     {{ page_title }}
   </h1>
 
-  {{ navigation(request, "compare") }}
+  {{ navigation(request) }}
```
dalecannon, line 32: Would it be useful to pass in `active_link=check` to highlight 'Check workbasket' as the active tab?

workbaskets/views/ui.py
```diff
@@ -1059,6 +1036,24 @@ def data_upload(self):
         except DataUpload.DoesNotExist:
             return None
 
+    @staticmethod
+    def format_duty_sentence(duty_sentence):
```
dalecannon, line 1040: Perhaps a docstring commenting on its use?  And would it be worth having an accompanying unit test?

workbaskets/views/ui.py
```diff
@@ -1059,6 +1036,24 @@ def data_upload(self):
...
+        if duty_sentence:
+            duty_rate = re.search("[0-9]+(.[0-9]+)?", duty_sentence).group(0)
```
dalecannon, line 1042: Does this only match the first non-decimal number in a duty sentence? Would it also need to handle compound duties (for example `1% + 10 GBP / 100 kg` -> `1.000% + 10.000 GBP / 100 kg`).

workbaskets/views/ui.py
```diff
@@ -1059,6 +1036,24 @@ def data_upload(self):
...
+        if "." in duty_rate:
+            before_decimal = duty_rate.split(".")[0]
+            after_decimal = duty_rate.split(".")[1].ljust(3, "0")
+            new_ds = duty_sentence.replace(
+                duty_rate,
+                before_decimal + "." + after_decimal,
+            )
+            return new_ds
+        else:
+            new_ds = duty_sentence.replace(duty_rate, duty_rate + ".000")
```
dalecannon, line 1054: Would this replace all occurrences of `duty_rate` ? As in if `duty_sentence = 1% + 10 GBP / 100 kg"` and `duty_rate` regex above returns `1` then `new_ds = 1.000% + 1.0000 GBP / 1.00000kg`.

workbaskets/views/ui.py
```diff
@@ -1113,6 +1108,163 @@ def get_context_data(self, *args, **kwargs):
...
+    def _append_url_page_param(self, url, form_action):
+        """Based upon 'form_action', append a 'page' URL parameter to the given
+        url param and return the result."""
+        page = self.paginator.get_page(self.request.GET.get("page", 1))
+        page_number = 1
+        if form_action == "page-prev":
+            page_number = page.previous_page_number()
+        elif form_action == "page-next":
+            page_number = page.next_page_number()
+        return f"{url}?page={page_number}"
```
dalecannon, line 1162: I think all of this is unused on this view and can be removed.

workbaskets/views/ui.py
```diff
@@ -1113,6 +1108,163 @@ def get_context_data(self, *args, **kwargs):
...
+                form_action,
+            )
+        except KeyError:
+            return reverse("home")
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        page = self.paginator.get_page(self.request.GET.get("page", 1))
+        kwargs["objects"] = page.object_list
+        return kwargs
```
dalecannon, line 1201: Ditto.

workbaskets/views/ui.py
```diff
@@ -1113,6 +1108,163 @@ def get_context_data(self, *args, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        page = self.paginator.get_page(self.request.GET.get("page", 1))
+        user_can_delete_workbasket = (
+            self.request.user.is_superuser
+            or self.request.user.has_perm("workbaskets.delete_workbasket")
+        )
```
dalecannon, line 1209: Ditto.

workbaskets/views/ui.py
```diff
@@ -1113,6 +1108,163 @@ def get_context_data(self, *args, **kwargs):
...
+        store.remove_items(form.cleaned_data)
+        if form_action == "remove-all":
+            object_list = {
+                self.form_class.field_name_for_object(obj): True
+                for obj in self.workbasket.tracked_models
+            }
+            store.add_items(object_list)
+        else:
+            to_add = {key: value for key, value in form.cleaned_data.items() if value}
+            store.add_items(to_add)
```
dalecannon, line 1264: Ditto.

CPrich905: Ditto

workbaskets/views/ui.py
CPrich905: Ditto

workbaskets/views/ui.py
CPrich905: Ditto

workbaskets/views/ui.py
CPrich905: You are correct, removed

workbaskets/jinja2/workbaskets/compare.jinja
CPrich905: That's a good point. I've added this in & we'll see how it fares in user testing.

workbaskets/jinja2/workbaskets/compare.jinja
CPrich905: Added this in - I think it looks a little long now, but we'll see how it does in user testing :)

workbaskets/jinja2/workbaskets/checks.jinja
CPrich905: Removed

workbaskets/jinja2/includes/workbaskets/workbasket-business-rules.jinja
CPrich905: Removed

workbaskets/forms.py
CPrich905: This was asked for in a meeting with Steve/Ash/Wren to be as placeholder text, the page itself has some instructions.

workbaskets/views/ui.py
CPrich905: My understanding of the task (and this work was passed to me by @LaurenQurashi so hopefully she'll correct me if I'm wrong) is that this should only impact the first number, ie. the duty rate, so `1% + 10 GBP / 100 kg` -> `1.000% + 10GBP / 100 kg`

workbaskets/forms.py
dalecannon: Ah fair enough. Thought I'd point out just in case as GDS generally recommend against it for accessibility.

workbaskets/jinja2/workbaskets/compare.jinja
dalecannon: Maybe the `Edit and exisiting workbasket` breadcrumb could be taken out to shorten it?

workbaskets/views/ui.py
CPrich905: Yes, correct - this is to account for the various ways the duty rate can be expressed vs what the compare tool is expecting/will match, e.g. 4% wouldn't match a measure with a duty sentence of 4.000% in the db, so to use the examples from the help text within the measure create journey:
`4%` --> `4.000%`
`33 GBP/100kg` --> `33.000 GBP/100kg`
`3.5% + 11 GBP / 100kg` --> `3.500% + 11 GBP / 100kg`

workbaskets/views/ui.py
CPrich905: I'll add these in

CPrich905: Discussed this with @eadpearce, we're going to remove this work from this PR as we both feel it should be factored in to a larger piece of work in the form of refactoring/improving the [DutySentenceParser](https://github.com/uktrade/tamato/blob/fbbe5a870d106e94eb828be3f7fab473cf183094/measures/parsers.py#L121)

dalecannon (approved):

PR #1081, opened 2023-10-31
TP2000-1083: Quota order number create

eadpearce:
# TP2000-1083: Quota order number create

## What

* Adds a link to the create quota form from the workbasket page and the quota search page
* Adds the quota create form and confirmation page
* Confirmation page links to the quota origin create form

Create quota confirmation page
Create quota form validation errors
Create quota form


quotas/forms.py
```diff
@@ -198,6 +200,119 @@ def init_layout(self):
...
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.init_fields()
+        self.set_initial_data(*args, **kwargs)
+        self.init_layout()
+
+    def set_initial_data(self, *args, **kwargs):
+        self.fields["category"].initial = self.instance.category
```
dalecannon, line 247: Would `self.instance` always be `None` in the case of this create form?

quotas/urls.py
```diff
@@ -44,6 +44,16 @@
...
+    path(
+        f"quotas/create/",
+        views.QuotaCreate.as_view(),
+        name="quota-ui-create",
+    ),
+    path(
+        f"quotas/<sid>/confirm-create/",
+        views.QuotaConfirmCreate.as_view(),
+        name="quota-ui-confirm-create",
+    ),
```
dalecannon, line 56: I think `get_ui_paths()` can automatically generate these routes so long as the corresponding views exist.

workbaskets/jinja2/workbaskets/edit-workbasket.jinja
```diff
@@ -27,123 +27,63 @@
...
+      {"text": "Find and edit measures", "url": url('measure-ui-search')},
+      ])
+    }}
+    {{ workbasket_column("Quotas", [
+      {"text": "Create new quota", "url": url('quota-ui-create')},
+      {"text": "Find and edit quotas", "url": url('quota-ui-list')},
+      ])
+    }}
+    {{ workbasket_column("Regulations", [
+      {"text": "Create new regulation", "url": url('regulation-ui-create')},
```
dalecannon, line 84:
```suggestion
      {"text": "Create a new regulation", "url": url('regulation-ui-create')},
```
to match the other links?

quotas/views.py
```diff
@@ -105,6 +105,20 @@ def get_queryset(self):
         return models.QuotaOrderNumber.objects.approved_up_to_transaction(tx)
 
 
+class QuotaCreate(QuotaOrderNumberMixin, CreateTaricCreateView):
```
dalecannon, line 108: Would it be worth adding validation for business rules ON1 and ON2?

quotas/urls.py
eadpearce: I've been avoiding using that because it obscures how the urls are defined. I'd much rather have them declared explicitly.

workbaskets/jinja2/workbaskets/edit-workbasket.jinja
eadpearce: Oh I didn't notice this. Yeah I had just copy pasted and changed the model name. This probably reads better

quotas/forms.py
eadpearce: Yeah not sure why this didn't blow up. It can be removed

dalecannon (approved):

PR #1082, opened 2023-11-01
TP2000-1052 - UI support for envelope validation failure

paulpepper-trade:
# TP2000-1052 - UI support for envelope validation failure

## Why

TARIC XML files are validated immediately after they have been generated from a workbasket's contents. The XML file creation process is performed in a Celery background task. Envelope creation is bound up with its associated XML file creation and validation: if an `Envelope`'s XML file fails validation, then the `Envelope` instance process fails in its entirety. Information about file validation is therefore lost without means of establishing what error may have occurred. There's therefore no way of presenting failure infomation in TAP's packaging and envelope queues.

## What

This PR introduces a `ValidationState` Choices class and new attribute, `Envelope.validation_state`, that takes a `ValidationState` value. `Envelope` instance creation no longer fails upon XML validation failure. When `Envelope` instances are created, TARIC XML file validation state is preserved, aiding failure diagnosis. However, because invalid XML files have no use, they are not stored to S3 and their associated `Envelope`'s `xml_file` takes a value of `None`.

While implementing this UI-related feature, issues were found in the validation process. These have been fixed as part of this PR.

A unit test against empty transaction validation has been removed because it was redundant. The same validation feature is a part of the TARIC XML schema validation check, which has associated unit tests.

## Checklist
- Requires migrations? Yes


publishing/models/envelope.py
```diff
@@ -42,12 +44,37 @@ class EnvelopeInvalidQueuePosition(Exception):
...
 
 
-class MultipleEnvelopesGenerated(Exception):
-    pass
+class ValidationState(IntegerChoices):
+    """
+    Envelope validation states.
+
+    Failure states directly map to the exceptions that may be raised by
+    validating an envelope using `publishing.util.validate_envelop()`.
```
a-gleeson, line 56: small spelling mistake `validate_envelop()`

publishing/migrations/0009_envelope_validation_state.py
```diff
@@ -0,0 +1,44 @@
+# Generated by Django 3.2.19 on 2023-10-27 13:45
+
+from django.db import migrations
+from django.db import models
+
+
+def envelope_validation_state_data_setup(apps, schema_editor):
```
a-gleeson, line 7: Should we be concerned about updated failed envelopes to anything? I think we have a deleted flag but by the sounds of it that's not applicable anymore as we want to keep failed generations for reporting purposes

paulpepper-trade: The scope of this change means that failed Envelope instances only need to hang around long enough for the validation failure to be reported in the UI and dealt with by a user, so I think the deleted flag is still applicable.

publishing/models/envelope.py
paulpepper-trade: Done.

publishing/tests/test_util.py
```diff
@@ -301,3 +304,31 @@ def test_validate_envelope_records_out_of_order(queued_workbasket):
         envelope_file.seek(0, os.SEEK_SET)
         validate_envelope(envelope_file, workbaskets=workbaskets)
         assert "Elements out of order in XML:" in e
+
+
+def test_validate_envelope_no_declaration(caplog):
```
a-gleeson, line 309: Nice test, should we check that `skip_declaration=True` passes as expected?

publishing/migrations/0009_envelope_validation_state.py
a-gleeson: Sounds fair to me, and I guess we will still play the tech debt story to clean up envelopes mark for delete too

a-gleeson (approved):
LGTM!

publishing/tests/test_util.py
paulpepper-trade: Thorough :) Done!

PR #1083, opened 2023-11-02
Increment message id & record sequence number correctly

a-gleeson:
# TP2000-827 Increment message id & record sequence number correctly

## Why
When the envelope is generated a couple Id's are not generated correctly. The message id starts from 1 on a fresh start of TAP but continues to increment between envelopes, this should reset between envelopes and start at 1. The record sequence number does not increment between records in an envelope when it should. 

## What
- move counter_generator from default param into the init in the envelope serialiser for message_id
- move counter_generator into the init for envelope serialiser from within the transaction jinja
- test to validate message id and record sequence number


common/tests/test_serializers.py
```diff
@@ -165,3 +169,105 @@ def create_output_constructor():
...
+            transaction=approved_transaction,
+        ),
+    )
+    factories.FootnoteFactory(
+        transaction=approved_transaction,
+        description=factories.FootnoteDescriptionFactory(
+            transaction=approved_transaction,
+        ),
+        footnote_type=factories.FootnoteTypeFactory(transaction=approved_transaction),
+    )
```
paulpepper-trade, line 204: There may be value in pulling this out as a populated workbasket fixture.

paulpepper-trade (approved):

PR #1084, opened 2023-11-02
TP2000-1100  Reset rule check status after transaction reordering

dalecannon:
# TP2000-1100  Reset rule check status after transaction reordering

## Why
When transactions are reordered in a workbasket that has already passed business rules, the successful rule check result remains despite the transaction order in the workbasket having changed. This opens the potential for a workbasket with an invalid transaction sequence to proceed through the packaging process.

## What
- Deletes the workbasket's business rule checks after transactions have been reordered to ensure a new business rule check is required to be run before it can be packaged.
- Updates unit tests


paulpepper-trade (approved):
`test_workbasket_transaction_order_view_promote_transaction()` and `test_workbasket_transaction_order_view_demote_transaction()` are near identical. Could they be collapsed down into a single [parametrize](https://docs.pytest.org/en/7.3.x/reference/reference.html#pytest-mark-parametrize)d unit test, say, `test_workbasket_transaction_order_view_move_transaction()`?

But otherwise LGTM!

PR #1086, opened 2023-11-06
Tp2000 1055 update import status labels

LaurenQurashi:
# TP2000-1055 Update import Status Labels

In order to align TAP to GDS standards, a few tweaks to the importer pages have been added: 
- Change the status labels to rainbow colours (very cute) 
- Change status label text from 'Completed' to 'Ready'
- Change status label text from 'Errored' to 'Failed'
- Change top filters to match the new status name
- Rename Importer Status column to Status and add "Action" to the last column which was previously nameless
- Swap around the form fields on the upload page
- Update the copy on the success page

## Checklist
- Requires migrations? - no

### DIFF
--- a/common/static/common/scss/_text.scss
+++ b/common/static/common/scss/_text.scss
@@ -1,34 +1,3 @@
-.status-badge {
-  font-weight: bold;
-  color: govuk-colour("white");
-  background-color: govuk-colour("blue");
-  padding: 0px 10px;
-}
-
-.status-badge-purple {
-  @extend .status-badge;
-  color: #3d2375;
-  background-color: #dbd5e9;
-}
-
-.status-badge-grey {
-  @extend .status-badge;
-  color: #383f43;
-  background-color: #eeefef;
-}
-
-.status-badge-red {
-  @extend .status-badge;
-  color: #942514;
-  background-color: #f6d7d2;
-}
-
-.status-badge-green {
-  @extend .status-badge;
-  color: #005a30;
-  background-color: #cce2d8;
-}
-
 // makes a button or input element look like a hyperlink
 .button-link {
   background: none;
--- a/common/static/common/scss/_text.scss
+++ b/common/static/common/scss/_text.scss
@@ -1,9 +1,32 @@
-
 .status-badge {
   font-weight: bold;
   color: govuk-colour("white");
   background-color: govuk-colour("blue");
   padding: 0px 10px;
+}
+
+.status-badge-purple {
+  @extend .status-badge;
+  color: #3d2375;
+  background-color: #dbd5e9;
+}
+
+.status-badge-grey {
+  @extend .status-badge;
+  color: #383f43;
+  background-color: #eeefef;
+}
+
+.status-badge-red {
+  @extend .status-badge;
+  color: #942514;
+  background-color: #f6d7d2;
+}
+
+.status-badge-green {
+  @extend .status-badge;
+  color: #005a30;
+  background-color: #cce2d8;
 }
 
 // makes a button or input element look like a hyperlink
--- a/importer/forms.py
+++ b/importer/forms.py
@@ -205,7 +205,7 @@ class CommodityImportForm(ImportFormMixin, forms.Form):
         label="Upload a TARIC file",
         help_text=(
             "Valid TARIC files contain XML and usually have a .xml file name "
-            "extension. They contain goods nomenclure items and related "
+            "extension. They contain goods nomenclature items and related "
             "items."
         ),
     )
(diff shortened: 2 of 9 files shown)


importer/forms.py
```diff
@@ -188,14 +188,6 @@ class CommodityImportForm(ImportFormMixin, forms.Form):
     """Form used to create new instances of ImportBatch via upload of a
     commodity code file."""
 
-    taric_file = forms.FileField(
```
LaurenQurashi, line 191: Confusing but this is just where I'm swapping the fields round on the upload page

importer/jinja2/components/create-status-tag.jinja
```diff
@@ -0,0 +1,8 @@
+
```
LaurenQurashi, line 1: Made a cute little macro for the status label as they all needed different text and CSS

importer/views.py
```diff
@@ -102,12 +102,13 @@ def get_context_data(self, **kwargs):
             and workbasket_status == WorkflowStatus.ARCHIVED
         ):
             context["selected_link"] = "empty"
-        elif import_status == ImportBatchStatus.IMPORTING and workbasket_status == None:
+        elif import_status == ImportBatchStatus.IMPORTING:
```
LaurenQurashi, line 105: Okay so I'm not sure if this was the right thing to do.. but basically when an import_batch  is created it now comes with a workbasket with a status of editing. Historically it didn't, but now that it does, it leads me to think that these links were broken cause they'll never hit the workbasket status = none. 

We also don't rely on knowing the workbasket status for importing and failed as it doesn't change the outcome like the ones above do - so I took it completely out but I'm not sure if it should be changed to add a conditional for if workbasket_status is editing.. but it feels a bit redundant..

common/static/common/scss/_text.scss
```diff
@@ -1,11 +1,34 @@
...
   color: govuk-colour("white");
   background-color: govuk-colour("blue");
   padding: 0px 10px;
 }
 
+.status-badge-purple {
+  @extend .status-badge;
+  color: #3d2375;
+  background-color: #dbd5e9;
+}
```
dalecannon, line 12: Would it be preferable to use [`govuk-tag--[colour]` classes](https://design-system.service.gov.uk/components/tag/all-colours/index.html) ?

importer/forms.py
```diff
@@ -209,6 +201,14 @@ class CommodityImportForm(ImportFormMixin, forms.Form):
             "number. "
         ),
     )
+    taric_file = forms.FileField(
+        label="Upload a TARIC file",
+        help_text=(
+            "Valid TARIC files contain XML and usually have a .xml file name "
+            "extension. They contain goods nomenclure items and related "
```
dalecannon, line 208:
```suggestion
            "extension. They contain goods nomenclature items and related "
```

importer/jinja2/includes/taric_importer_list.jinja
```diff
@@ -1,20 +1,23 @@
...
         <input type="hidden" name="workbasket" value="{{ object.workbasket.pk }}">
         <input type="hidden" name="workbasket-tab" value="review-goods">
-        <input type="submit" class="button-link" name="submit" value="Edit/View workbasket">
+        <input type="submit" class="button-link" name="submit" value="View Workbasket {{object.workbasket.pk}}">
       </form>
     {% elif goods_status(object) == "no_goods" %}
-      <span>0 goods items</span>
+      <span>N/A</span>
     {%- else -%}
       <span></span>
```
dalecannon, line 22: Would it be good to indicate that no action is available for this case too?

importer/views.py
```diff
@@ -136,6 +137,39 @@ def goods_status(cls, import_batch: ImportBatchFilter) -> str:
...
+    @classmethod
+    def status_tag_generator(cls, import_batch: ImportBatchFilter) -> dict:
+        """Returns a dict with text and a css class for a ui friendly label for
+        an import batch."""
+        workbasket = import_batch.workbasket
+
+        if (
+            import_batch.status == ImportBatchStatus.SUCCEEDED
+            and workbasket.status == WorkflowStatus.EDITING
+        ):
```
dalecannon, line 149: Could there be historical imports without associated workbaskets on the list view? If so, would it need to check first if the workbasket exists to avoid an `AttributeError`  being raised on `workbasket.status`?

common/static/common/scss/_text.scss
LaurenQurashi: It would! however only half or so of the colours come from GDS so I'd have to also use codes too so went with uniform codes rather than hald GDS half codes? But I am easy with this one!

dalecannon (approved):

PR #1087, opened 2023-11-06
TP2000-1099 Add 'Move to top' and 'Move to bottom' transaction reordering actions

dalecannon:
# TP2000-1099 Add 'Move to top' and 'Move to bottom' transaction reordering actions

## Why
To move a transaction to the top or bottom of a workbasket currently requires pressing 'Move up' or 'Move down'  buttons many times. The addition of 'Move to top' and 'Move to bottom' actions would make this easier for users to achieve.

## What
- Adds 'Move to top' and 'Move to bottom' actions to transactions listed in Transaction Order tab
- Handles in the view setting the transaction order of the moved transaction to be either first or last in the workbasket, updating other transactions accordingly
- Refactores transaction reordering form action unit tests into a single parametrised test

##
Screenshot 2023-11-06 at 15 50 14

### DIFF
--- a/workbaskets/jinja2/workbaskets/macros/transaction_details.jinja
+++ b/workbaskets/jinja2/workbaskets/macros/transaction_details.jinja
@@ -3,8 +3,15 @@
 
   <div id="transaction-order-actions">
     {% if transaction == first_transaction_in_workbasket or not user_can_delete_items %}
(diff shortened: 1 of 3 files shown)


CPrich905 (approved):
All makes sense to me :)

PR #1091, opened 2023-11-14
TP2000-1114: React enhanced forms proof of concept

eadpearce:
# TP2000-1114: React enhanced forms proof of concept

## Why

* FormSet code in Django is getting to be very complicated and due to the need to submit and refresh the page to add or delete items, makes the UX clunky
* We are already using FormSets in a way that is not idiomatic for Django. React is better equipped to handle this kind of interactivity on the frontend and can make the UX smoother

## What

* Adds the required npm packages for React support
* Adds a QuotaOrigins React component that renders a QuotaOriginForm that handles adding/deleting multiple quota origins before data is submitted to the backend for database changes
* Adds govuk-react to allow the use of govuk styled React form components
* Adds a custom add_extra_error method to the Django form that allows us to add errors to be passed to the React form which allows us to keep business rules checks and other validation on the backend
* Data is passed from Django to React with a script tag in the form template

Screenshot 2023-11-14 at 16 35 53
Screenshot 2023-11-14 at 16 36 31


common/static/common/js/QuotaOriginForm.js
```diff
@@ -0,0 +1,85 @@
...
+                        day: `origins-${index}-start_date_0`,
+                        month: `origins-${index}-start_date_1`,
+                        year: `origins-${index}-start_date_2`,
+                    }}
+                    defaultValues={{
+                        day: data.start_date_0,
+                        month: data.start_date_1,
+                        year: data.start_date_2,
+                    }}
+                    errorText={originsErrors[`origins-${index}-start_date`]}
```
a-gleeson, line 29: I was never good at react and haven't done any in a while. Where does `originsErrors` come from?

a-gleeson (commented):
Instead of just in the `common/js` shouldeer we put this in a specific folder for quota's or use a more "react" folder layout

common/static/common/js/QuotaOriginForm.js
eadpearce: Probably should add a comment here explaining. This comes from django and is set in a script tag in the template

eadpearce:
@a-gleeson I was trying to find an example folder structure in the React docs but couldn't find anything and ended up just winging it. If you've got an example somewhere I can look at that would help a ton

common/static/common/js/QuotaOriginForm.js
a-gleeson: ah that makes sense I thought that but hadn't got that far in the review! I guess having some sort of commenting strucutre for components could be good.

a-gleeson:
> @a-gleeson I was trying to find an example folder structure in the React docs but couldn't find anything and ended up just winging it. If you've got an example somewhere I can look at that would help a ton

I remember using this structure before and it was used in datadub too
https://github.com/uktrade/data-hub-frontend/tree/main/src/client/components

quotas/jinja2/includes/quotas/quota-edit-origins.jinja
```diff
@@ -10,6 +11,23 @@
...
 </div>
+<script nonce="{{ request.csp_nonce }}">
+  var originsErrors = {
+    {% for k, errorlist in errors.as_data().items() %}
+      "{{ k }}": "{% for errors in errorlist %}{% for error in errors %}{{ error|safe }}{% if not loop.last %}\n{% endif %}{% endfor %}{% endfor %}",
+    {% endfor %}
+  };
+  var originsData = {{origins_initial|safe}};
+  var geoAreasOptions = [
+    {"name": "", "value": ""},
```
a-gleeson, line 22: in the `options.map` in QuotaOriginForm will this empty value always display?

a-gleeson: nevermind I think this is just the default value within the options select!

quotas/forms.py
```diff
@@ -156,7 +163,86 @@ def init_fields(self):
...
+            )
+            if not form.is_valid():
+                for field, e in form.errors.as_data().items():
+                    self.add_extra_error(
+                        f"{QUOTA_ORIGINS_FORMSET_PREFIX}-{i}-{field}",
+                        e,
+                    )
+            else:
+                self.cleaned_data["origins"].append(form.cleaned_data)
+        return super().clean()
```
paulpepper-trade, line 243: We'd normally call `super().clean()` [at the beginning of the overridden funtion to allow the parent behaviour to do its thing](https://docs.djangoproject.com/en/dev/ref/forms/validation/#validating-fields-with-clean). Probably worth using that convention or otherwise documenting why this is different.

quotas/jinja2/includes/quotas/quota-edit-origins.jinja
```diff
@@ -10,6 +11,23 @@
   </div>
   {% endfor %}
 </div>
+<script nonce="{{ request.csp_nonce }}">
+  var originsErrors = {
```
paulpepper-trade, line 15: We should use `let` and `const` in place of `var` these days to avoid the weird hoisting and scoping behaviours that `var` applies.

paulpepper-trade (commented):
This is likely to be the beginning of a lot more JavaScript in Tamato, so probably a good point at which to introduce a JavaScript linter. It'd also be worth popping [a note in the docs about any linting capability that we start using](https://github.com/uktrade/tamato/blob/react-poc/docs/source/CONTRIBUTING.rst#formatting).

quotas/jinja2/includes/quotas/quota-edit-origins.jinja
eadpearce: Haha I actually did want this variable to be hoisted here but I think that was while I was still figuring out the script load order. If it gets hoisted it'll at least come back as undefined instead of blowing up the whole thing if it's not defined by the time the react is initialised. I might be able to change it back now though

eadpearce: Switched all these out for `const` instead and it works!

quotas/views.py
```diff
@@ -295,27 +297,135 @@ class QuotaUpdateMixin(
...
+
+                else:
+                    models.QuotaOrderNumberOrigin.objects.create(
+                        order_number=instance,
+                        valid_between=origin["valid_between"],
+                        geographical_area=origin["geographical_area"],
+                        update_type=UpdateType.CREATE,
+                        transaction=instance.transaction,
+                    )
+        else:
```
dalecannon, line 362: Would any exclusions for the new origin need to be created here too?

quotas/forms.py
```diff
@@ -93,6 +96,7 @@ class QuotaOriginExclusionsForm(forms.Form):
         help_text="Select a country to be excluded:",
         required=False,
     )
+    pk = forms.IntegerField(required=False)
```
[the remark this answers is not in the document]
eadpearce: Removed since it wasn't actually being used!

quotas/views.py
```diff
@@ -295,27 +297,137 @@ class QuotaUpdateMixin(
...
 
-        existing_origins = (
-            models.QuotaOrderNumberOrigin.objects.approved_up_to_transaction(
-                object.transaction,
-            ).filter(
-                order_number__sid=object.sid,
+    def update_origins(self, instance, form_origins):
+        existing_origin_pks = {
+            o.pk for o in instance.quotaordernumberorigin_set.current()
+        }
```
dalecannon, line 320: I think that the first time an update is created for the quota, the reverse lookup for existing origins will always return as empty because `instance` will have a new pk.

quotas/views.py
```diff
@@ -295,27 +297,137 @@ class QuotaUpdateMixin(
...
+                        transaction=instance.transaction,
+                    )
+
+                # whether it's edited or new we need to add/update exclusions
+                if origin.get("exclusions"):
+                    self.update_exclusions(
+                        instance,
+                        updated_origin,
+                        origin.get("exclusions"),
+                    )
```
dalecannon, line 363: I'm not sure this will get called if the origin has only one exclusion and that exclusion is then removed from the submitted form.

eadpearce: Oh wait nvm. Exclusions. Not origins

eadpearce: Should be fixed now

dalecannon (approved):

PR #1094, opened 2023-11-15
Remove workbasket items and add items to table

LaurenQurashi:
# TP-2000-1117-workbasket-view-summary-changes

## Why
Due to a reshuffle in the workbasket journeys, we're deleting some stuff as it's no longer needed once the new changes come in

## What
On the workbasket summary tab, the selectable objects table with the workbaskets items on has been deleted as it can now be seen on the workbaskets changes page. 

We've also added two new rows to the table at the top of the page with links to the workbaskets changes page and to the transaction reordering page

## Checklist
- Requires migrations? - No


workbaskets/views/ui.py
```diff
@@ -321,13 +321,11 @@ class EditWorkbasketView(PermissionRequiredMixin, TemplateView):
 
 
 @method_decorator(require_current_workbasket, name="dispatch")
-class CurrentWorkBasket(FormView):
```
LaurenQurashi, line 324: A lot of these deletes are just out of date code. As I deleted the workbasket items list.. which was a selectable form, a lot of this stuff could go

workbaskets/tests/test_views.py
```diff
@@ -879,7 +787,13 @@ def test_submit_for_packaging_disabled(
         import_batch.workbasket_id = session_workbasket.id
         if isinstance(import_batch, ImportBatch):
             import_batch.save()
-
+    url = reverse(
+        "workbaskets:workbasket-checks",
+    )
+    response = valid_user_client.post(
```
eadpearce, line 793: This response will get overwritten by the one below it so it's still the old GET request being tested here

workbaskets/jinja2/workbaskets/summary-workbasket.jinja
```diff
@@ -54,7 +57,31 @@
         </a>
       </dd>
     </div>
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">
```
eadpearce, line 61: These `dt` and `dd` elements should be inside a `dl`

eadpearce (approved):

PR #1095, opened 2023-11-15
TP2000-1124: Add validation for quota definition volume

eadpearce:
# TP2000-1124: Add validation for quota definition volume

Screenshot 2023-11-15 at 12 29 51

### DIFF
--- a/quotas/forms.py
+++ b/quotas/forms.py
@@ -438,6 +438,11 @@ class QuotaDefinitionUpdateForm(
         self.init_layout()
         self.init_fields()
 
+    def clean(self):
+        if self.cleaned_data["volume"] > self.cleaned_data["initial_volume"]:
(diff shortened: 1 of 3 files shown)


CPrich905 (approved):
LGTM, one non-blocking nit-picky question: if there is an intention to add more validation tests, would this one be better named something like `test_quota_definition_validation_for_higher_than_initial_volume`?

eadpearce:
@CPrich905 Yeah I think if we did add more validation we could modify that test or parametrize it

eadpearce:
@CPrich905 ...that is to say. I'll update it to test more of the validation because I probably should have done that anyway

CPrich905 (approved):
Nice one! Also like calling the single function twice rather than duplicating it :)

PR #1096, opened 2023-11-16
docker db managing improvements

a-gleeson:
# TP-??? Docker Db managing improvement

## Why
When working with TAP DB's you may not always want to work with just a single db or run a db dump everytime you need to clear down the database. This is especially relevant to data engineers 

## What
- added two new make commands `docker-import-new-db` which imports a new db into a template name & `docker-restore-db` which loads a template into the db (deafult name tamato)
- updated readme
- corrected sample.env around minio

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,11 @@ DEV=true
 #default run can be replaced with exec
 DOCKER_RUN?=run --rm
 #db import file name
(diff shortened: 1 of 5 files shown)


paulpepper-trade (approved):
Nice

PR #1097, opened 2023-11-21
TP2000-1119 : add processing_time field to TrackedModelCheck

dougmills-DIT:
# TP2000-1119 (part 1) add performance metric to rules checks
Right now, we have some qualitative information that ME32 and ME16 are taking a lot longer than the other rules. 
This PR adds a processing_time to the TrackedModelCheck model and updates the creation to include processing time 
so that we can report / query rule running duration in a rules check, and better isolate outliers.  

## Why
We need better information on what time a rule takes to process and a way to monitor improvements in production. 
This change gives us data to analyze and report on, giving us quantitative information to work with.  

## What
Really simple, add a field to store the time a rule check took to run and update the code to add the appropriate field to the database when recording the rule check results

## Checklist
- Requires migrations? yes

Links to relevant material
See: [Description](https://uktrade.atlassian.net/browse/TP2000-1126)


a-gleeson (approved):
LGTM!

PR #1099, opened 2023-11-21
TP2000-1131 Increase pagination value for transaction order view

dalecannon:
# TP2000-1131 Increase pagination value for transaction order view

## Why
The transaction order view only loads 20 transactions at a time. This number is too low for larger workbaskets. 

## What
- Sets the view's paginate_by value to 100


a-gleeson (approved):
LGTM!

PR #1101, opened 2023-11-22
TP2000-1133  Fix editing geographical memberships to allow members to be re-added

dalecannon:
# TP2000-1133  Fix editing geographical memberships to allow members to be re-added

## Why
Geo areas that were once members of an area group but have been end dated are unable to be re-added with new validity periods.

## What
- Uses `current().as_at_and_beyond()` when checking if a geo area is already a member to allow for members to be re-added only with non-overlapping validity periods.
- Adds validity period to dropdown labels to help users distinguish between old and new when selecting to edit an existing membership

##
Screenshot 2023-11-22 at 17 14 48

Screenshot 2023-11-22 at 17 10 50

### DIFF
--- a/geo_areas/forms.py
+++ b/geo_areas/forms.py
@@ -305,17 +305,26 @@ class GeographicalMembershipEditForm(
 
         self.fields["membership"].queryset = self.instance.get_current_memberships()
 
-        self.fields["membership"].label_from_instance = (
-            lambda obj: f"{obj.member.area_id} - {obj.member.structure_description} ({obj.valid_between.lower} - {obj.valid_between.upper})"
-            if self.instance.is_group()
-            else f"{obj.geo_group.area_id} - {obj.geo_group.structure_description} ({obj.valid_between.lower} - {obj.valid_between.upper})"
-        )
+        self.fields["membership"].label_from_instance = self.label_from_instance
 
         self.fields["membership"].help_text = (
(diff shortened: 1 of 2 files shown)


geo_areas/forms.py
```diff
@@ -297,9 +305,9 @@ def __init__(self, *args, **kwargs):
         self.fields["membership"].queryset = self.instance.get_current_memberships()
 
         self.fields["membership"].label_from_instance = (
-            lambda obj: f"{obj.member.area_id} - {obj.member.structure_description}"
+            lambda obj: f"{obj.member.area_id} - {obj.member.structure_description} ({obj.valid_between.lower} - {obj.valid_between.upper})"
             if self.instance.is_group()
-            else f"{obj.geo_group.area_id} - {obj.geo_group.structure_description}"
+            else f"{obj.geo_group.area_id} - {obj.geo_group.structure_description} ({obj.valid_between.lower} - {obj.valid_between.upper})"
```
paulpepper-trade, line 310: Does this now have the whiff of being in need of a refactor, say, by pulling this out into a utility function on the form or module?

paulpepper-trade (approved):
It's maybe worth adding a unit test to cover the situation where the bug occurred.

geo_areas/forms.py
dalecannon: I've refactored it into a small utility function on the form since it's only really applicable here.

PR #1107, opened 2023-11-30
TP2000-1084 Archive empty, previously queued workbaskets on attempt to delete

dalecannon:
# TP2000-1084  Archive empty, previously queued workbaskets on attempt to delete

## Why
A workbasket that has previously been packaged, and then removed from the packaging queue, will have a non-nullable PackagedWorkBasket instance associated with it, preventing deletion owing to a ProtectedError exception. Users should however be given the appearance of a successful workbasket deletion.

## What
- Adds a Django FSM condition to prevent non-empty workbaskets from being archived
- Checks whether a PackagedWorkBasket exists for a workbasket, archiving the workbasket instead of attempting to delete it if so.


paulpepper-trade (approved):
LGTM

PR #1109, opened 2023-12-06
TP2000-1144 Change pagination on workbasket changes and transaction views to avoid timeout issues

dalecannon:
# TP2000-1144 Change pagination on workbasket changes and transaction views to avoid timeout issues

## Why
The workbasket changes view and workbasket transaction order view paginate by returning the previous pages' results plus the new page's result (per paginate_by) when the load more button is pressed. For pages with high object counts, this can eventually lead to timeout errors, preventing users from proceeding further.

## What
- Removes the load more button in favour of traditional pagination (prev and next buttons) limiting the results of any one page to 100
-  Removes sortable column functionality (will be re-added in a follow-up ticket together with object filtering capability)
- Updates tests to reflect changes


CPrich905 (approved):
LGTM!

PR #1113, opened 2023-12-15
TP2000-1152-handling-invalid-workbaskets

mattjamc:
# TP2000-1152 Handling invalid workbaskets

## Why

Currently if a user accesses a view covered by the _require_current_workbasket_ decorator and they don't have an active workbasket, they will be assigned a different workbasket without being notified.

This is particularly problematic if a user's current workbasket gets edited/queued by someone else and it is no longer in the EDITING state. In this scenario, when the user submits their next http request their workbasket will change and they might start making updates in the wrong workbasket.

## What (part 1)

This PR:

- updates the _require_current_workbasket_ decorator. If a workbasket is required and the user doesn't have an active workbasket then it redirects them to select a new workbasket instead of assigning them a different one
- adds template, view and url for the new page
- adds tests to confirm redirect when a user's session workbasket is deleted or queued

As various tests relied on being randomly assigned a workbasket without creating one, this PR also makes the following test changes:

- adds two new fixtures similar to valid_user_client / valid_user_api_client but which also create a workbasket and add it to the session
- updates tests to use the new fixtures where a workbasket wasn't previously created in the test but was needed
- updates _workbasket_ fixture to return existing workbasket if one was already created by a different fixture rather than making a second 

image

## What (part 2)
In order to merge with TP2000-1130--extend-user-model this PR updates current WorkBasket methods and pytest fixtures to account for the workbasket now being stored in the user model rather than session data.

### New model methods

- assign_to_user - assigns workbasket to the user of request
- remove_users_current_workbasket - removes users assigned current workbasket 

### Removed model methods

- save_to_session - previously saved workbasket to the session of request
- load_from_session - previously queried session workbasket from database
- remove_current_from_session - previously deleted the workbasket from the session data

### New fixtures

- client_with_current_workbasket - returns a valid user client with a workbasket associated to the user (accounts for tests which did not previously have a workbasket created since now one doesn't get created automatically)
- api_client_with_current_workbasket - returns a valid user client with a workbasket associated to the user
- session_request_with_workbasket - returns a request object with a valid user, session and workbasket associated
- user_workbasket - returns a workbasket which has been assigned to a valid logged-in user with 2 footnotes in it
- user_empty_workbasket - returns an empty workbasket which bas been assigned to a valid logged in user

### Removed fixtures

- session_workbasket - previously returned a workbasket in the session
- session_empty_workbasket - previously returned an empty workbasket in the session


common/models/utils.py
mattjamc, line 1: Can ignore this as I ended up doing it in the require_current_workbasket decorator. This middleware stuff is undone in a later commit.

measures/tests/test_views.py
```diff
@@ -1083,16 +1081,14 @@ def test_measure_update_group_exclusion(client, valid_user, erga_omnes):
     assert area_2.sid in area_sids
 
 
-def test_measure_edit_update_view(valid_user_client, erga_omnes):
+def test_measure_edit_update_view(valid_user_client_workbasket, erga_omnes):
     """Test that a measure UPDATE instance can be edited."""
     measure = factories.MeasureFactory.create(
         update_type=UpdateType.UPDATE,
-        transaction=factories.UnapprovedTransactionFactory(),
```
mattjamc, line 1090: Had to remove the unapproved transaction for it to work - could that cause any problem I'm not seeing?

measures/tests/test_views.py
```diff
@@ -1117,16 +1113,19 @@ def test_measure_edit_update_view(valid_user_client, erga_omnes):
...
-def test_measure_edit_create_view(valid_user_client, duty_sentence_parser, erga_omnes):
+def test_measure_edit_create_view(
+    valid_user_client_workbasket,
+    duty_sentence_parser,
+    erga_omnes,
+):
     """Test that a measure CREATE instance can be edited."""
     measure = factories.MeasureFactory.create(
         update_type=UpdateType.CREATE,
-        transaction=factories.UnapprovedTransactionFactory(),
```
mattjamc, line 1124: Had to remove the unapproved transaction for it to work - could that cause any problem I'm not seeing?

geo_areas/tests/test_views.py
```diff
@@ -434,17 +437,16 @@ def test_geo_area_create_view(valid_user_client, session_workbasket, date_ranges
 
 
 def test_geo_area_edit_create_view(
-    valid_user_client,
-    session_workbasket,
-    date_ranges,
     use_edit_view,
+    workbasket,
+    date_ranges,
```
mattjamc, line 442: use_edit_view already calls valid_user_api_client so not sure valid_user_client is needed

workbaskets/views/ui.py
```diff
@@ -1489,3 +1490,10 @@ def get_context_data(self, *args, **kwargs):
         context["selected_tab"] = "regulations"
         context["tab_template"] = "includes/regulations/list.jinja"
         return context
+
+
+def workbasket_not_active(request, *args, **kwargs):
```
paulpepper-trade, line 1495: We normally use class-based views rather than function-based view. [TemplateView](https://docs.djangoproject.com/en/dev/ref/class-based-views/base/#django.views.generic.base.TemplateView) would probably offer a straight swap here.

conftest.py
```diff
@@ -156,7 +156,7 @@ def tap_migrator_factory(migrator_factory):
     its migration unit testing, continues to cause problems in migration unit
     tests. A couple of examples of the reported issue:
     https://code.djangoproject.com/ticket/10827
-    https://github.com/wemake-services/django-test-migrations/blob/93db540c00a830767eeab5f90e2eef1747c940d4/django_test_migrations/migrator.py#L73
```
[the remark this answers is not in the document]
mattjamc: Yeah I did originally do that but a different pre-commit hook kept putting it on the line below. I could do at the end without a space but then it would look like part of the link

PR #1114, opened 2023-12-18
TP2000-1142 Synchronous only operation errors

paulpepper-trade:
# TP2000-1142 Synchronous only operation errors

## Why

Use of Python's async features to concurrently retrieve data from 3rd party APIs across the network is causing Django to raise `SynchronousOnlyOperation` exceptions.

## What

[Non-async Django does not permit syncronous code execution within the same context (i.e. on the same thread) as async code](https://docs.djangoproject.com/en/dev/topics/async/#async-views). It guards against this by [trying to get an event loop on the current thread and raises an SynchronousOnlyOperation exception if a loop / message pump is found](https://github.com/django/django/blob/08306bad57761b5eb176894649ac7d4e735c52dd/django/utils/asyncio.py#L8-L28).

This PR:

- Replaces the use of async with the equivalent Python threading behaviour to concurrently retrieve data from 3rd party APIs.
- Adds unit tests for the API querying function used to retrieve quota data.
- Removes async behaviour and related test library dependency.

## Checklist
- Requires migrations? No


common/tariffs_api.py
```diff
@@ -77,23 +75,76 @@ def build_quota_definition_urls(order_number, object_list):
...
+
+def threaded_get_request_session() -> requests.Session:
+    """Return a requests.Session instance scoped to the current thread."""
 
+    if not hasattr(_thread_locals, "requests_session"):
+        _thread_locals.requests_session = requests.Session()
+    return _thread_locals.requests_session
+
+
+def threaded_get_from_endpoint(url: str) -> str:
```
a-gleeson, line 105: as None is an option maybe the return type should be a union with the none type

common/tariffs_api.py
```diff
@@ -77,23 +75,76 @@ def build_quota_definition_urls(order_number, object_list):
...
+    """
+
+    requests_session = threaded_get_request_session()
+    try:
+        with requests_session.get(url) as response:
+            response.raise_for_status()
+            return response.json()
+    except requests.exceptions.ConnectionError:
+        logger.error(f"Unable to establish connection during HTTP GET {url}")
+        return None
```
a-gleeson, line 122: do we need repeated returns or can just have a single return at the end?

a-gleeson (approved):
LGTM!

common/tariffs_api.py
paulpepper-trade: Done what I think is the preferred alternative of using `Optional`.

common/tariffs_api.py
paulpepper-trade: Yes! Done.

PR #1115, opened 2023-12-19
Impl'n: New Celery task infra to process CreateMeasures instances

CPrich905:
# TP-2000-1147 Impl'n: New Celery task infra to process CreateMeasures instances

## Why
Currently, bulk editing occurs within the context of a web worker. Because an edit can take many seconds and even minutes to complete, the confirmation web page presented to a user is bounded at the lower end by the amount of time an edit takes. It’s not the best approach to performing this type of computation for a number of reasons, including:

- Generally, users are accustomed to web pages loading in the order of milliseconds or a few seconds in the worst case, so a long delay is frustrating.
- Errors may be accumulate through spurious UI interaction with an in-progress creation process.
- Proxy server time-outs may occur during edit measure processing - the TAP team and SRE do not have the necessary access permissions to extend the proxy server’s timeout period.

## What
This PR is concerned with:

1. Implementing the Celery infrastructure for processing the task asynchronously.
2. Prepare the form data in serializable format. It has been established that the best way to do this with Django Form Wizard is to add a serializer and deserializer onto each Form. The serializer returns the user input to the form, the deserializer provides the completed form object.

## Charlie's Notes
- Migration might be irrelevant
- We're duplicating a lot of the serialization code. TODO: Investigate extracting the serialization code into a helper function that can be imported and reused.


paulpepper-trade (commented):
I think it would be worth separating this PR into smaller ones:

1. Celery-related changes.
2. CreateMeasures model command class.
3. Form serialization changes and tests.

The form serialization could be broken down into smaller PRs, which would help with divvying up those changes too - ticketing them as needed

PR #1116, opened 2024-01-03
Updated NIG5 to correctly handle only goods create and update and NIG5_origin to only handle origin deletions. Also added NIG11 skip delete

dougmills-DIT:
# TP2000-1158 : Updated NIG5, create NIG5_origin, added skip delete to NIG11

 * NIG5 should not be triggered when an origin is added, but validate on the goods transaction not the origins transaction
 * NIG11 should not trigger if a goods is being deleted
 * Full details on ticket 

## Why
Why is this change happening, e.g. goals, use cases, stories, etc.?
 * Both NIG5 and NIG11 recently triggered false positives for an EU update. 
 * Upon investigation it was observed that NIG5 was associated with Goods origins as an indirect business rule, but the validate method used the goods transaction not the origins transaction so would fail if both goods and origin were not present in the same transaction. 
 * NIG11 triggered and failed on a goods deletion

## What
 * Updated the query for the validation method for NIG5 to use the current transaction 
 * Added an NIG5_origin rule, to run against the origin on deletion to prevent updates that leave goods with no origin.
 * Added tests to cover the scenarios identified as issues for both NIG5 and NIG11 
 * Updated tests to work in 2024, some didn't age well

## Checklist
- Requires migrations? No

Links to relevant material
See: [Description](https://uktrade.atlassian.net/browse/TP2000-1158)

### DIFF
--- a/notifications/tests/test_models.py
+++ b/notifications/tests/test_models.py
@@ -1,3 +1,4 @@
+from datetime import date
 from unittest.mock import patch
 
 import factory
@@ -69,7 +70,7 @@ def test_create_packaging_notification(ready_for_packaging_notification):
 
     content = notification.get_personalisation()
     assert content == {
-        "envelope_id": "230001",
+        "envelope_id": f"{str(date.today().year)[2:]}0001",
         "description": "",
         "download_url": "http://localhost/publishing/envelope-queue/",
         "theme": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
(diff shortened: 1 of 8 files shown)


notifications/tests/test_models.py
```diff
@@ -69,7 +70,7 @@ def test_create_packaging_notification(ready_for_packaging_notification):
 
     content = notification.get_personalisation()
     assert content == {
-        "envelope_id": "230001",
+        "envelope_id": f"{str(date.today().year)[2:]}0001",
```
eadpearce, line 73: Would the `date_ranges` pytest fixture work for this? Or are these tests set up a bit differently?

dougmills-DIT: im sure its possible to get the current year out of date ranges, I expect that it would be better / simpler / more readable to just use current year as its such a simple value.

eadpearce (approved):

PR #1117, opened 2024-01-04
Quota origins and exclusions geo area description query optimisation

eadpearce:
# Quota origins and exclusions geo area description query optimisation

## Why

* Getting geo area descriptions was very slow. Using a custom query with annotations speeds this up a lot

## What

* Adds annotated queries for quota order number origins and exclusions that adds the latest geo area description to a custom field

### DIFF
--- a/quotas/querysets.py
+++ b/quotas/querysets.py
@@ -1,5 +1,7 @@
 from django.db.models import BooleanField
 from django.db.models import Case
+from django.db.models import OuterRef
+from django.db.models import Subquery
 from django.db.models import When
 
 from common.models.tracked_qs import TrackedModelQuerySet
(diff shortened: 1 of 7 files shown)


quotas/querysets.py
```diff
@@ -16,3 +18,65 @@ def with_is_origin_quota(self) -> TrackedModelQuerySet:
                 output_field=BooleanField(),
             ),
         )
+
+
+class QuotaOrderNumberOriginQuerySet(TrackedModelQuerySet):
+    def with_latest_geo_area_description(qs):
```
CPrich905, line 24: Given that these two queries are identical, could we not pull them out and then call them from within the class?

eadpearce: Annoyingly they're not exactly identical. Origins have the field `geographical_area` and exclusions have `excluded_geographical_area` which is honestly baffling to me. We could add another annotation to each model to standardise the field name but then I feel like since that would then deviate from the data spec it could just make it more confusing

CPrich905: 🤦‍♂️ Ahh yes. Despite re-reading this 3 times (including once backwards) I still didn't see this

CPrich905 (approved):

PR #1123, opened 2024-01-11
TP2000-1130 Move current workbasket from Session to custom User model

mattjamc:
# TP2000-1130 Move current workbasket from Session to custom User model

## Why
TMs have reported losing their current workbasket and having it switch to an entirely different one after a short period of inactivity, which has led on occasion to tariff changes being applied to the wrong workbasket. Saving the users' current workbasket in the database rather than in the session would prevent this from happening following session expiration.

## What
- Updates User model referencing to use `get_user_model()`

- Substitutes Django's default User model for a custom user model (see [#Note](#Note))

- Adds `current_workbasket` field to `User` model to store the user's current workbasket in the DB

- Add User model to admin

- Updates `require_current_workbasket` view decorator to redirect users without an active current workbasket to select a new workbasket

- Removes now unused `ValidateSessionWorkBasketMiddleware`

- Adds a new view, template and route, `NoActiveWorkBasket`, to be used as a redirect endpoint for users without an active current workbasket

- Updates WorkBasket model methods and test fixtures to account for the workbasket now being stored in the user model rather than session data

    **New model methods**:
    - `assign_to_user` sets the workbasket to be the user's current workbasket
    - `remove_users_current_workbasket` removes user's current workbasket
  
    **Removed model methods**:
    - `save_to_session` previously saved workbasket to the session of request
    - `load_from_session` previously queried session workbasket from database
    - `remove_current_from_session` previously deleted the workbasket from the session data

    **New fixtures**:
    - `client_with_current_workbasket` returns a valid user client with a workbasket associated to the user
    - `api_client_with_current_workbasket` returns a valid user api client with a workbasket associated to the user
    - `session_request_with_workbasket` returns a request object with a valid user with a current workbasket and session
    - `user_workbasket` returns a workbasket which has been assigned to a valid logged-in user
    -  `user_empty_workbasket` - returns an empty workbasket which bas been assigned to a valid logged-in user
    -  `client_with_current_workbasket_no_permissions` - returns a client with a logged in user who has a current workbasket but no permissions

    **Removed fixtures**:
    - `session_workbasket` previously returned a workbasket in the session
    - `session_empty_workbasket` previously returned an empty workbasket in the session

- Adds new unit tests for the new changes

## Note <a href="#note" id="note"/>
This PR followed the guide set out in https://code.djangoproject.com/ticket/25313#comment:24 to change to a custom user model mid project.

It's possible to place the custom User model in its own app, say users, however that requires either truncating the `django_migrations` table or manually inserting the initial migration for the user model (using raw SQL) into the table making it as if it had already been applied. This would need doing only once for production but need doing every time a new DB snapshot was imported for local dev environments if the snapshot preceded the manual intervention to the migrations table on prod.

## Screenshot
Screenshot 2024-01-12 at 09 45 30

Screenshot 2024-01-24 at 15 50 20

Screenshot 2024-01-24 at 15 50 33

## Checklist


workbaskets/models.py
```diff
@@ -512,25 +509,27 @@ def measures(self) -> MeasuresQuerySet:
...
-        if "workbasket" not in session:
-            raise KeyError("WorkBasket not in session")
-        return WorkBasket.objects.get(pk=session["workbasket"]["id"])
-
-    @classmethod
-    def remove_current_from_session(cls, session):
-        """Remove the current workbasket from the user's session."""
-        if "workbasket" in session:
-            del session["workbasket"]
+    def remove_users_current_workbasket(cls, request):
```
eadpearce, line 512: Because of the lack of apostrophe this method name may be confusing to read (mistaking singular possessive for plural). Use 'from' instead like the old method eg `remove_current_from_user`

workbaskets/migrations/0001_initial.py
```diff
@@ -6,14 +6,149 @@
 from django.db import models
 
 
+def change_user_content_type(apps, schema_editor):
```
eadpearce, line 9: Shouldn't this be in a new migration rather than initial? I think envs that have already run the old version of this file won't rerun it with the new changes

dalecannon: Ah yes. This was meant to go in  `0009_user_current_workbasket.py` migration file so that it will be applied and performed.

workbaskets/models.py
```diff
@@ -512,25 +509,27 @@ def measures(self) -> MeasuresQuerySet:
...
+            pass
 
     @classmethod
     def current(cls, request):
-        """Get the current workbasket in the session."""
-        if "workbasket" in request.session:
-            workbasket = cls.load_from_session(request.session)
+        """Get the user's current workbasket."""
+        try:
+            current_user = User.objects.get(id=request.user.id)
```
paulpepper-trade, line 525: Redundant DB access to re-fetch `request.user` - see previous comment.

workbaskets/admin.py
```diff
@@ -168,4 +173,29 @@ def save_model(self, request, instance, form, change):
...
+    add_form = UserCreationForm
+    readonly_fields = ("current_workbasket",)
+    fieldsets = UserAdmin.fieldsets + (
+        ("Workbasket", {"fields": ("current_workbasket",)}),
+    )
+    add_fieldsets = UserAdmin.add_fieldsets
+
+
+admin.site.register(User, UserAdmin)
+
```
paulpepper-trade, line 200: Perhaps worth including some screenshots of the User admin pages on the cover of this PR?

workbaskets/models.py
```diff
@@ -665,3 +664,17 @@ class DataRow(ValidityMixin, models.Model):
         null=True,
         blank=True,
     )
+
+
+class User(AbstractUser):
```
paulpepper-trade, line 669: I can see why an existing app has to be used, but I'm somewhat curious to understand why the `User` model has been placed in the workbaskets app, rather than, say, `common`. Were there some difficulties putting there?

paulpepper-trade (changes requested):

workbaskets/models.py
```diff
@@ -512,25 +509,27 @@ def measures(self) -> MeasuresQuerySet:
...
-        if "workbasket" not in session:
-            raise KeyError("WorkBasket not in session")
-        return WorkBasket.objects.get(pk=session["workbasket"]["id"])
-
-    @classmethod
-    def remove_current_from_session(cls, session):
-        """Remove the current workbasket from the user's session."""
-        if "workbasket" in session:
-            del session["workbasket"]
+    def remove_current_from_user(cls, request):
```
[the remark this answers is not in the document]
mattjamc: I think it was just a case that the previous remove current workbasket from session function was here and it only gets used by the current() function directly below - but happy to move if that makes more sense :)

workbaskets/models.py
dalecannon: There's a circular dependency issue where `common.0001_initial` has `workbaskets.0001_initial` as a dependency and `workbaskets.0001_initial` requires the `User` model for foreign key fields on `WorkBasket` model.

Saying that, I wonder if the migration operation that adds the User model foreign key fields can be moved to the next migration file (workbaskets.0002...), resolving the dependency issue.

commodities/tests/test_migrations.py
```diff
@@ -26,12 +27,13 @@ def test_main_migration_works(migrator, setup_content_types):
     Transaction = old_state.apps.get_model("common", "Transaction")
     Workbasket = old_state.apps.get_model("workbaskets", "WorkBasket")
     VersionGroup = old_state.apps.get_model("common", "VersionGroup")
+    User = old_state.apps.get_model("workbaskets", "User")
 
-    QueuedWorkBasketFactory.create().save()
-    workbasket = Workbasket.objects.create(id=238, author_id=1)
+    user = User.objects.create()
```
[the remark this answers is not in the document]
dalecannon: Using factories seemed to cause errors in existing migration tests following the change to a custom user model. 

From what I could gather, the factories use current model definitions to create objects but the migration test sets the state of the database according to the specified migration, making it so only model definitions as they were defined at that point in migration history (`old_state.apps.get_model("app", "model")`) can be used.

This poses a problem for factories that directly or indirectly require a `UserFactory`, as the test database will complain that `current_workbasket` column doesn’t exist in the `User` table because of a mismatch between the factory model definition and the state of the test database preceding the addition of the new field on the User model.

common/models/utils.py
```diff
@@ -93,48 +93,6 @@ def override_current_transaction(transaction=None):
...
-
-        # A current, editable workbasket is necessary in order to set the
-        # current transaction (below), so abandon this middleware's action and
-        # return early if there is no current editable workbasket in the
-        # session.
-        if not is_session_workbasket_valid(request):
-            WorkBasket.remove_current_from_session(request.session)
-        return self.get_response(request)
-
-
```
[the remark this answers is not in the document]
mattjamc: Yes I think this achieves more or less the same thing but I can bring the middleware back for extra security - it was more difficult to keep the middleware when workbasket was still in session but now it should be fine

commodities/tests/test_migrations.py
```diff
@@ -18,6 +17,8 @@ def test_main_migration_works(migrator, setup_content_types):
 
     setup_content_types(old_state.apps)
```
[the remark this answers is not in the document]
paulpepper-trade: Was there a problem with removing the `setup_content_types` fixture?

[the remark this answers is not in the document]
paulpepper-trade: Oh, I can see it's gone now.

paulpepper-trade (approved):

PR #1130, opened 2024-01-16
Tp2000 652  force rule check after real edit

a-gleeson:
# TP2000-652 force rule check after real edit

## Why
Successful business rule checks on a workbasket are currently valid for as long as the head transaction of the workbasket is equal to (actually less than or equal in Tamato’s code) the transaction check. As soon as a new transaction is created in a workbasket, it is ahead of the last transaction check and so becomes invalid.

However, real edits do not create new transactions, but instead update them. This could allow hidden invalid workbaskets should an edit cause a change that would otherwise cause a business rule to fail.

## What
- add TimestampedMixin to TransactionCheck
- add TimestampedMixin to TrackedModel
- data migration for TransactionCheck which takes the TrackedModelCheck timestamps to set TransactionCheck timestamps
- data migration for TrackedModel which takes the Transaction timestamp to set TrackedModel timestamps
- updated unchecked_or_errored_transactions to check that the TransactionCheck is created after the last TrackedModel in the workbasket was updated. (A real edit)
- unit tests for a real edit
- unit tests for data migrations

## Checklist
- Requires migrations? - Yes
- Requires dependency updates?


checks/tests/test_migrations.py
```diff
@@ -0,0 +1,60 @@
+# test transaction checks
+
+# create tracked model, transaction, workbasket, tracked model check and transaction check
+# create a workbasket in published, editing, queued
+# apply 0006
+# check transaction check datetime
+# apply 0007
+# check transaction check incompletful, unsuccessful and equal to tracked model checl
```
paulpepper-trade, line 8: An old todo list?

common/migrations/0009_set_tracked_model_datetime.py
```diff
@@ -0,0 +1,32 @@
...
+from django.db import migrations
+from django.db.transaction import atomic
+
+
+@atomic
+def generate_timestamps(apps, schema_editor):
+    TrackedModel = apps.get_model("common", "trackedmodel")
+    all_models = TrackedModel.objects.select_related("transaction").all()
+    paginator = Paginator(all_models, settings.DATA_MIGRATION_BATCH_SIZE)
+    print(f"Total: {paginator.num_pages}")
```
paulpepper-trade, line 14: Could the prints be replaced by log output to allow some control over it?

checks/tests/test_migrations.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+    from common.models import TrackedModel
+
+    tracked_models = TrackedModel.objects.all()
+    for tracked_model in tracked_models:
+        print(tracked_model.name)
+
+    transaction_check = TransactionCheck.objects.get(
+        pk=tracked_model_check_1.transaction_check.transaction.pk,
+    )
```
paulpepper-trade, line 45: Hangover debug code?

workbaskets/models.py
```diff
@@ -619,15 +620,22 @@ def delete_checks(self):
 
     @property
     def unchecked_or_errored_transactions(self):
```
paulpepper-trade, line 622: It may be worth adding a docstring to this function now that logic is a bit more involved. A high-level account of how `Transaction` and now `TrackedModel` play their parts in invalidating business rule checks?

common/migrations/0009_set_tracked_model_datetime.py
```diff
@@ -0,0 +1,32 @@
...
+    for page_num in range(1, paginator.num_pages + 1):
+        print(f"number {page_num}")
```
paulpepper-trade, line 16: Log for print?

paulpepper-trade (approved):

workbaskets/models.py
a-gleeson: good shout

checks/tests/test_migrations.py
a-gleeson: good catch, debugging from this morning :)

PR #1134, opened 2024-01-17
feat: invoke UI changes to reports and create new URL path to handle reports with multiple tabs

nboyse:
# TP2000-169 - Invoke UI changes to various reports

## Why

After review by our UI designer, there are a few changes that need to be made to the way we are rendering our reports, i.e. description updates, remove the "type" of report and make it explicit through the description and change spacing/layout of buttons.

In par with this PR, I also realised that the export_to_csv functionality will not work as expected when tabs have been invoked in specific reports. 
## What

Description of reports have been updated, CSV button has been moved to float right, new endpoint has been created for the CSV button to handle multiple tabs and in relation to this had to write some JS as we can't access the fragment with anything django request-y related (as its hardcoded backend from load and not working with present data).

### DIFF
--- a/reports/jinja2/generics/table.jinja
+++ b/reports/jinja2/generics/table.jinja
@@ -27,7 +27,7 @@
     </ul>
     <div class="govuk-tabs__panel" id="{{ report.tab_name | replace(' ', '_') | lower }}">
         <h2 class="govuk-heading-l">{{ report.tab_name }}</h2>
-            <a id="exportToCSVLink" class="govuk-button" style="float:right; margin-top:-70px" href="{{ url("reports:export_report_to_csv", kwargs={"report_slug": report.slug() }) }}">Export to CSV</a>
+            <a id="exportToCSVLink{{report.tab_name | replace(' ', '_') | lower}}" class="govuk-button report-button-inline" href="{{ url("reports:export_report_to_csv", kwargs={"report_slug": report.slug() }) }}">Export to CSV</a>
         {{ govukTable({
         "head": report.headers(),
         "rows": report.rows()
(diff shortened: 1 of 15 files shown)


reports/jinja2/reports/report_table.jinja
```diff
@@ -9,15 +9,42 @@
...
+window.onhashchange = initialisePage;
+</script>
+
     <h1 class="govuk-heading-xl govuk-!-margin-bottom-3">
         Report: {{ report.name }}        
     </h1>
-                    <a class="govuk-button" href="{{ url("reports:export_report_to_csv", kwargs={"report_slug": report.slug()}) }}">Export to CSV</a>
-    <span class="govuk-caption-xl govuk-!-margin-bottom-9">Table</span>
+
+    <a id="exportToCSVLink" class="govuk-button govuk-!-margin-bottom-0" style="float:right" href="{{ url("reports:export_report_to_csv", kwargs={"report_slug": report.slug()}) }}">Export to CSV</a>
```
eadpearce, line 42: Wasn't sure what was going on with this so I had a play around in my local env. You could avoid needing to use JS to change the href by moving the button inside the tab content. As it is now, this button feels to me like it should export the data from all the tabs below it since it's at the very top of the page. Might be a thing to go over with user research/design though

reports/jinja2/generics/table.jinja
```diff
@@ -27,27 +27,33 @@
     </ul>
     <div class="govuk-tabs__panel" id="{{ report.tab_name | replace(' ', '_') | lower }}">
         <h2 class="govuk-heading-l">{{ report.tab_name }}</h2>
+            <a id="exportToCSVLink" class="govuk-button" style="float:right; margin-top:-70px" href="{{ url("reports:export_report_to_csv", kwargs={"report_slug": report.slug() }) }}">Export to CSV</a>
```
eadpearce, line 30: Since all these buttons have the same styling it's probably better to make a CSS class instead of using inline styles. Keeps all our styling in one place too

eadpearce: Also these links should have different IDs. Could namespace them by tab name

eadpearce (changes requested):
couple small things

eadpearce (approved):

PR #1137, opened 2024-01-22
TP2000-1185  Add maintenance mode

mattjamc:
# TP2000-1185 Add maintenance mode

## Why

Some types of application update require that TAP is out of action for a short period of time - or at least, no data should be changed during that time. A means of preventing active user interfaction with the service should provided, while also showing an informative message.

## What

This PR:
- adds ENV variable for MAINTENANCE_MODE to be set to True whenever we need to take TAP down temporarily
- adds a url/template and view for a page letting users know TAP is down for maintenance (following [GDS guidelines](https://design-system.service.gov.uk/patterns/service-unavailable-pages/))
- adds middleware to check the maintenance mode variable and redirect accordingly
- conditionally removes middleware that might write to the database during maintenance mode as well as the admin app
- adds tests for the maintenance mode redirect and page content
- updates incorrect contact us links on the 403 and 500 error pages

image


common/urls.py
```diff
@@ -26,4 +26,5 @@
     path("login", views.LoginView.as_view(), name="login"),
     path("logout", views.LogoutView.as_view(), name="logout"),
     path("api-auth/", include("rest_framework.urls")),
+    path("503/", views.maintenance, name="maintenance"),
```
paulpepper-trade, line 29: I don't think the maintenance page should be displayed for 503 HTTP responses - 500-series responses being error response. Although the service will be unavailable, it would be intentionally so and not because of an error.

Would it be preferable to have the maintenance page located at a URL route of "/maintenance/" or simiar, so that we can distinguish between a real 503 and maintenance mode.

common/views.py
```diff
@@ -464,5 +464,9 @@ def handler500(request, *args, **kwargs):
     return TemplateResponse(request=request, template="common/500.jinja", status=500)
 
 
+def maintenance(request, *args, **kwargs):
+    return TemplateResponse(request=request, template="common/503.jinja", status=503)
```
paulpepper-trade, line 468: Ditto comment above about 503 being reserved for error rather than intended maintenance activity - the service is still available, you just can't do a great deal with it!

paulpepper-trade (changes requested):

paulpepper-trade (approved):

PR #1143, opened 2024-01-30
TP2000-1190  Prevent unassigned workbaskets from being packaged

mattjamc:
# TP2000-1190 Prevent unassigned workbaskets from being packaged

## Why

As part of the wider goal to move away from Jira, tasks have been created and users will soon be able to assign workbaskets to others.

A requirement is there must be a worker and reviewer assigned before a workbasket can be packaged. If no users are assigned, packaging must be prevented until those fields are completed.
## What

- Adds the transition requirement _fully_assigned_ in order for a workbasket to be queued for packaging
- Disables the 'Send to packaging queue' button for workbaskets which have not been assigned
- Updates old tests which involved packaging to assign the workbaskets
- Adds a view and model test to confirm unassigned workbaskets cannot be queued but otherwise can

image


workbaskets/models.py
```diff
@@ -402,6 +402,24 @@ def archive_workbasket_condition_is_empty(self) -> bool:
         """Django FSM condition: workbasket must be empty (no tracked models and no transactions) to transition to ARCHIVED status."""
         return not self.tracked_models.exists() and not self.transactions.exists()
 
+    def fully_assigned(self) -> bool:
```
dalecannon, line 405: Since this is a boolean-returning method, would it perhaps make it a little clearer if it were renamed as `is_fully_assigned()`?

conftest.py
```diff
@@ -1210,8 +1231,16 @@ def check(
...
+        workbasket = factories.WorkBasketFactory.create()
+
+        task = factories.TaskFactory.create(workbasket=workbasket)
+        factories.UserAssignmentFactory.create(
+            assignment_type=UserAssignment.AssignmentType.WORKBASKET_WORKER,
+            task=task,
+        )
+        factories.UserAssignmentFactory.create(
+            assignment_type=UserAssignment.AssignmentType.WORKBASKET_REVIEWER,
+            task=task,
```
dalecannon, line 1243: Would it be worth refactoring as an `AssignedWorkBasketFactory` that can create and return a fully assigned workbasket?

For example:
```
class AssignedWorkBasketFactory(WorkBasketFactory):
    @factory.post_generation
    def user_assignments(self, create, extracted, **kwargs):
        if not create:
            return
        
        task = TaskFactory.create(workbasket=self)
        UserAssignmentFactory.create(
            assignment_type=UserAssignment.AssignmentType.WORKBASKET_WORKER,
            task=task,
        )
        UserAssignmentFactory.create(
            assignment_type=UserAssignment.AssignmentType.WORKBASKET_REVIEWER,
            task=task,
        )
```

PR #1144, opened 2024-01-30
TP2000-1179 Serialize commodities and duties formset

paulpepper-trade:
# TP2000-1179 Serialize commodities and duties formset

## Why

Complete support for serialization / deserialization of `MeasureConditionsWizardStepFormSet` required.
Support for serialization / deserialization of `MeasureCommodityAndDutiesFormSet` required.
Supporting capability for serializing dates and associated tests required.
Further refinement of form serialization is desirable.
Removal of debug / temporary test capability is desirable.

## What

This PR:

- More flexible handling of form data key serialization. `SerializableFormMixin.ignored_data_key_regexs` is now used to represent the default list of keys that should be skipped.
- Adds `SerializableFormMixin.get_serializable_data_keys()` which is used to get the non-ignored data keys of a Form instance.
- Provides unit test for `SerializableFormMixin.get_serializable_data_keys()`.
- Adds supporting capability for serializing and deserializing dates.
- Provides unit tests for date serialization / deserialization.
- Adds support for `serializable_init_kwargs()` and `deserialize_init_kwargs()` on `MeasureConditionsWizardStepFormSet`.
- Full support for serialization / deserialization of `MeasureCommodityAndDutiesFormSet`.
- Provides a temporary unit test `test_measure_commodities_and_duties_form_set_serialization()` to be superceded by a fuller suite of tests.
- Introduces a destinction between forms at wizard steps that collect user data and those that don't collect data. This allows serialization of only those data-providing forms. `MeasureCreateWizard.data_form_list` is used to hold the data forms and has an accompanying `MeasureCreateWizard.get_data_form_list()` function which is analogous to the base class's `WizardView.get_form_list()` - they get their associated form lists while applying the conditions filter in `MeasureCreateWizard.condition_dict`.

## Checklist
- Requires migrations? No

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -111,7 +111,7 @@ class SerializableFormMixin:
         Return serializable form data that can be serialized / stored as, say,
         `django.db.models.JSONField` which can be used to recreate a valid form.
 
-        if `remove_key_prefix` is a non-empty string, then the keys in the
+        If `remove_key_prefix` is a non-empty string, then the keys in the
         returned dictionary will be stripped of that string where it appears as
         a key prefix in the origin `data` dictionary.
 
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -84,8 +84,8 @@ class SerializableFormMixin:
         "MAX_NUM_FORMS$",
     ]
     """
-    Regexs of keys that may appear in a Form's `data` attribute and which should
-    be ignored when creating a serializable version of `data`.
+    Regexs of keys that may appear in a Form's `data` dictionary attribute and
+    which should be ignored when creating a serializable version of `data`.
 
     Override this on a per form basis if there are other, redundant keys that
     should be ignored. See the default implementation of
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+import re
 from itertools import groupby
 from typing import Dict
 from typing import List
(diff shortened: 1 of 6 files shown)


measures/forms.py
```diff
@@ -70,36 +73,46 @@ class SerializableFormMixin:
...
+        Override this function if neither `ignored_data_key_regexs` or this default
         implementation is sufficient for identifying which of `Form.data`'s keys
         should be used during a call to this mixin's `serializable()` method.
         """
-        return [k for k in self.data if k not in self.data_keys_ignored]
+        combined_regexs = "(" + ")|(".join(self.ignored_data_key_regexs) + ")"
+        return [k for k in self.data.keys() if not re.search(combined_regexs, k)]
 
-    def serializable(self, with_prefix=True) -> Dict:
+    def serializable(self, remove_key_prefix: str = "") -> Dict:
```
CPrich905, line 108: Minor point, but should this be called `serializable` or `serialize` (or `serialize_data` or similar)? 
The name `serializable` implies to me that this is a check to see whether a data set is serializable, not necessarily a function to serialize data.

measures/forms.py
```diff
@@ -527,10 +546,40 @@ class MeasureConditionsWizardStepFormSet(
 ):
     form = MeasureConditionsWizardStepForm
 
-    def get_serializable_data_keys(self) -> List[str]:
-        keys = super().get_serializable_data_keys()
-        # Additionally filter out unused auto-complete data.
-        return [k for k in keys if not k.endswith("certificate_autocomplete")]
+    @classmethod
+    def serializable_init_kwargs(cls, kwargs: Dict) -> Dict:
```
CPrich905, line 550: Ditto naming comment re serialized/serializable.
Would a docstring/comment be helpful to here to explain when and how `serializable_init_kwargs` should be used (ie. will we need to add in other kwargs?) be useful?

CPrich905 (approved):
Very minor comments which you can feel free to ignore if you disagree as otherwise LGTM.
On the naming topic, would `FormSerializerMixin` be more appropriate?

measures/forms.py
paulpepper-trade: [My response to the previous comment](https://github.com/uktrade/tamato/pull/1144#discussion_r1474348198) mostly applies to this. It's maybe also worth saying, though, that the `deserialize_init_kwargs()` function is actually taking a serializable parameter (a Python dictionary) and returning something that is more Python-like. I've tweaked the docstring on `SerializableFormMixin.deserialize_init_kwargs()` to make it clearer.

measures/forms.py
paulpepper-trade: Naming refactor from `serializable()` to `serializable_data()` now done.

LaurenQurashi (approved):

PR #1148, opened 2024-02-07
Tp2000 1211

dougmills-DIT:
# TP2000-1211 persist old version of govuk jinja template code until it can be updated 

* https://github.com/LandRegistry/govuk-frontend-jinja got updated, and the commit we were pinned to for this project was removed. 

## Why
Why is this change happening, e.g. goals, use cases, stories, etc.?
 * The git commit we were pinned to was no longer available - the project won't build / deploy

## What
What is this PR doing, e.g. implementations, algorithms, etc.?
 * committed a copy of the code required into TAP and removed the git reference in requirements

## Checklist
- Requires migrations? No 

Links to relevant material
See: [Description](https://uktrade.atlassian.net/browse/TP2000-1211)

### DIFF
--- a/govuk_frontend_jinja/flask_ext.py
+++ b/govuk_frontend_jinja/flask_ext.py
@@ -1,0 +1,29 @@
+from flask.templating import Environment as FlaskEnvironment
+from jinja2 import select_autoescape
+
+from govuk_frontend_jinja.templates import Environment as NunjucksEnvironment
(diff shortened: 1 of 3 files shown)


paulpepper-trade (approved):
Perhaps worth suggesting `pip uninstall govuk-frontend-jinja` on existing vitual envs?

PR #1151, opened 2024-02-08
TP2000-1148 Create asynchronous confirmation page

CPrich905:
# TP-2000-1148 Create asynchronous confirmation page

## Why
The current create measures confirmation page has updates to align with the asynchronous processing of the task and latest designs.
The intention is to retain the option for processing the creation synchronously, therefore the original confirmation page has been left in place for the synchronous create journey.

## What
Created a new file, `confirm-create-multiple-async.jinja` containing the new changes.
Update the path for asynchronous task completion, using `redirect()` rather than `render()`.
The value for `expected_measures_count` has been hard coded until we have the creation process completed. 
A note has been added to explain how this count is calculated.

**Note 1:** the omission in the actual from the design of the `Review your measure` breadcrumb. It is neither possible or desired for the user to return to this page after clicking complete.

**Note 2:** we don't currently have a `View status` page, so the url for this button will need to be added later. The button has been disabled until this url is added in.

image

To run this locally, pull the branch and copy the return for the `async_done()` function into the return of `sync_done()`. You will need to hard code a value for `expected_measures_count`.


measures/views.py
```diff
@@ -1030,6 +1030,14 @@ def async_done(self, form_list, **kwargs):
         #   this point, or on the Celery side of measure creation. (The current
         #   user is assigned to the measures workbasket, but is that redundant?)
         # - Redirect from summary page to done page.
+        return redirect(
+            reverse(
```
paulpepper-trade, line 1034: The `reverse()` isn't needed. I think the same parameters can be passed directly to `redirect()`.

paulpepper-trade (approved):

PR #1153, opened 2024-02-13
TP2000-1222 Refactor measures based models

paulpepper-trade:
# TP2000-1222 Refactor measures based models

## Why

The Measures-related `models.py` file is already relatively large. It also currently contains only `TrackedModel` related models. Adding measure bulk creation / editing models and related classes to this confuses its purpose.

## What

This PR factors out models into two separate modules that live under a new `measures/models/` directory: one retaining the existing `TrackedModel`s-based models, and another containing the new bulk creation/editing-relate models.

## Checklist
- Requires migrations? No

### DIFF
--- a/measures/models/bulk_processing.py
+++ b/measures/models/bulk_processing.py
@@ -1,0 +1,132 @@
+import json
+import logging
+from typing import Iterable
+
+from django.db import models
+from django.db.transaction import atomic
+from django.forms.formsets import BaseFormSet
+
+from common.models.utils import set_current_transaction
+from measures.models.tracked_models import Measure
+
+logger = logging.getLogger(__name__)
+
+
+class MeasuresBulkCreator(models.Model):
+    """
+    Model class used to bulk create Measures instances from serialized form
+    data.
+
+    The stored form data is serialized and deserialized by Forms that subclass
+    SerializableFormMixin.
+    """
+
+    form_data = models.JSONField()
+    """Dictionary of all Form.data, used to reconstruct bound Form instances as
+    if the form data had been sumbitted by the user within the measure wizard
+    process."""
+
+    form_kwargs = models.JSONField()
+    """Dictionary of all form init data, excluding a form's `data` param (which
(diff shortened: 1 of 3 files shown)


measures/models/bulk_processing.py
```diff
@@ -0,0 +1,132 @@
...
+        "common.Transaction",
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="measures_bulk_creators",
+        editable=False,
+    )
+    """
+    The 'current' Transaction instance at the time `form_data` was constructed.
+
+    This is normally be set by
```
CPrich905, line 43:
```suggestion
    This is normally set by
```

measures/models/bulk_processing.py
```diff
@@ -0,0 +1,132 @@
...
+    # - Is it preferable to save the Workbasket rather than the current
+    #   transaction in the workbasket?
+    #   The current transaction can change if more transactions are created
+    #   before create_meassures() has chance to run. That could be good (we
+    #   want objects at the time the user performed the create measures action)
+    #   or bad (the current transaction may get deleted).
+    #   However if workbasket immutability is guarenteed until create_measures()
+    #   has completed, then this is moot. It'd need a 'protected' attribute, or
+    #   something like that, on the WorkBasket class that freezes it, say, in
+    #   the save() and update() methods.
```
CPrich905, line 59: Recording/registering a response to this now as we're looking at the models & it may require model changes:
As it's possible for the `current_transaction` to change (i.e. afaik there are no blocks in place to stop it from happening) and also that they could happen for legitimate reasons/might be desirable behaviour in the future (allowing TMs the option to work on multiple transactions at once), my opinion is that we should save the Workbasket and ensure it's immutability during the editing process.

paulpepper-trade: This is for the next PR. This PR is strictly a refactor that doesn't introduce functional change in order to reduce confusion and difficulties.

CPrich905 (approved):

PR #1156, opened 2024-02-15
TP2000-1207 Create measures from reconstructed forms

paulpepper-trade:
# TP2000-1207 Create measures from reconstructed forms

## Why

Measures bulk creation and supporting functionality is required in the Celery background task that is spawned from the measure creation wizard.

## What

This PR:

- Adds a Model Manager class for `MeasuresBulkCreator` object management.
- Adds `ProcessingState` Choices class used to represent the state of a bulk creation/editing process.
- Adds a `DateTimeStampedMixin` defining `created_at` and `updated_at`. An existing class name `TimeStampedMixin` exists, but that inherits from Django's `Model` class, making it a base class and a little more difficult to use.
- Adds a `BulkProcessorResult` model class used to save bulk processing result data.
- Adds a `BulkProcessorMixin` class for inheritance by model classes that are used to perform bulk processing tasks. The mixin defines attributes and behaviour that are common to all bulk processing models (current only `MeasuresBulkCreator`).
- Performs a refactor of `create_measures()` and `create_measures_conditions()` functions, moving them from the `MeasuresCreateWizard` to a new `MeasuresCreator` class that can be used by the wizard view or the `MeasuresBulkCreator` class.
- Adds unit tests for measures bulk creation - more coverage to follow in subsequent PRs.

## Checklist
- Requires migrations? Yes


measures/tasks.py
```diff
@@ -14,7 +14,15 @@ def bulk_create_measures(measures_bulk_creator_pk: int) -> None:
...
-    )
+    if len(measures) > 0:
+        transaction = measures[0].transaction
+        workbasket = transaction.workbasket
+        logger.info(
+            f"MeasuresBulkCreator({measures_bulk_creator.pk}) created "
+            f"{len(measures)} Measures in WorkBasket({workbasket.pk}) on "
+            f"Transaction({transaction.pk}).",
+        )
+    else:
```
CPrich905, line 25: Would it be useful/practical/possible to log errors at this point?

measures/tests/test_views.py
```diff
@@ -2689,3 +2691,27 @@ def test_measure_list_results_show_chosen_filters(valid_user_client, date_ranges
             f"Regulation {measure.generating_regulation.autocomplete_label}" in items[3]
         )
         assert f"Measure Type {measure.measure_type.autocomplete_label}" in items[4]
+
+
+def test_measures_wizard_create_confirm_view(valid_user_client):
```
CPrich905, line 2696: Should we have a test for failure her as well? Ie. 'force' a failure to ensure we're not getting false positive results?

CPrich905 (commented):
A couple of very minor, non-blocking clarification questions & one typo, otherwise LGTM!

measures/tasks.py
paulpepper-trade: I'm not sure what errors we'd catch at this level in order to log useful messages. I think any uncaught exceptions would appear in the application logs and there's log output down in the measures creation functions to give context to failures.
Are there particular errors / exceptions that you think we should catch and log additional details of at this level in the call stack?

measures/tests/test_views.py
paulpepper-trade: What kind of failures do you have in mind? This is testing confirmation page that's shown after queuing a bulk creation - i.e. after at the end of the measures create wizard journey, before the creation task has run and had a chance to succeed or fail.

LaurenQurashi (approved):

measures/tasks.py
CPrich905: Ahh yeah, this makes sense! I think I'd looked at this part in isolation (checked part last night, part this morning) so hadn't stitched them together properly :)

measures/tests/test_views.py
CPrich905: I think I'm resolving this after asking the question! Basically my thought was that we should have tests that confirm our error checking, but I think this is already managed by existing tests, so another non-issue :)

CPrich905 (approved):

PR #1157, opened 2024-02-19
TP2000 1175 Measure Create Wizard Forms Serialization and Deserialization tests

LaurenQurashi:
# TP2000-1195 to 1204 Add tests for Measure Create Wizard Forms' serialization capabilities

## Why
We are currently working on separating measure creation functionality to celery tasks in order to avoid long wait times for the process to complete. Celery however can't handle having a whole form wizard's amount of data passed to it.. so we're saving it in the database and dragging it back out once we're inside the task. This means serializing all the form data so that it can be saved in the db in a JSON field, but also deserializing it once in the celery task in order to submit the form data for measure creation.

## What
This PR adds testing to the form's newly added serialization and deserialization capabilities. The tests check that each form's serialization and deserialization capabilities work, by creating a form, serializing it, then deserializing it, and checking the form is still valid. This PR does not include the geo areas form test - as they're taking a bit of time for us to get our heads round. 

## Checklist
- Requires migrations? No


measures/forms.py
```diff
@@ -345,6 +345,7 @@ def conditions_clean(
 
         # Note this is a quick fix & hard coded for now
         # Action code's 1,2,3,4 are flexible and have edge cases that all neither Price or certificate to be present
+
```
paulpepper-trade, line 348: Probably remove this to avoid pulling this file into the PR.

CPrich905 (approved):
LGTM, but might be worth including a bit more info in the docstrings as to why there are separate tests/where different parameters are required, e.g. `test_selectableobjects_measure_forms_serialize_deserialize` and it's need for `quota_order_number_origin_form` isn't immediately obvious.

paulpepper-trade (approved):

PR #1159, opened 2024-02-20
TP2000-1226 Fix workbasket user assignment assignable users queryset

dalecannon:
# TP2000-1226 Fix workbasket user assignment assignable users queryset

## Why
The queryset used to populate selectable users for workbasket user assignment includes inactive and duplicate user profiles. 

## What
- Adds `.filter(is_active=True)` to the users queryset to filter out inactive users
- Adds `.distinct()` to prevent duplicate user objects being included in the queryset (users who belong to both TM and TM Lead user groups)

### DIFF
--- a/workbaskets/views/ui.py
+++ b/workbaskets/views/ui.py
@@ -429,10 +429,15 @@ class CurrentWorkBasket(TemplateView):
             )
         ]
 
-        users = User.objects.filter(
-            Q(groups__name__in=["Tariff Managers", "Tariff Lead Profile"])
-            | Q(is_superuser=True),
-        ).order_by("first_name", "last_name")
+        users = (
+            User.objects.filter(
+                Q(groups__name__in=["Tariff Managers", "Tariff Lead Profile"])
+                | Q(is_superuser=True),
+            )
+            .filter(is_active=True)
+            .distinct()
+            .order_by("first_name", "last_name")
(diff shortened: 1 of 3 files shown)


workbaskets/views/ui.py
```diff
@@ -429,10 +429,15 @@ def get_context_data(self, **kwargs):
...
-        ).order_by("first_name", "last_name")
+        users = (
+            User.objects.filter(
+                Q(groups__name__in=["Tariff Managers", "Tariff Lead Profile"])
+                | Q(is_superuser=True),
+            )
+            .filter(is_active=True)
+            .distinct()
+            .order_by("first_name", "last_name")
+        )
```
paulpepper-trade, line 440: This looks to be the same filtering performed at lines 216-223 in `workbaskets/forms.py`. Perhaps a refactor to share it - it could be lifted out into a custom filter, say, `UserQuerySet.tariff_managers()`.

paulpepper-trade (approved):

workbaskets/views/ui.py
dalecannon: Yeah, that's a good idea. I ran into a small issue trying to create the custom QuerySet. I'll follow up the refactoring in another ticket so it doesn't block the fix for now while I'm resolving the error.

PR #1160, opened 2024-02-21
TAP Review 1 - Standardisation

mattjamc:
# TAP Review 1 - Standardisation

## Why

User journeys across TAP vary in content and style and it's better for users to be consistent throughout.

## What

This PR addresses comments from Ash's [TAP Review 1 document](https://dbis-my.sharepoint.com/:w:/r/personal/ash_patel_businessandtrade_gov_uk/Documents/Review%20of%20TAP%20UI%20shared%20with%20Wren/TAP%20Review%201%20-%20create%20a%20workbasket_workbasket%20pages_%20additional%20codes.docx?d=w9df72e7c46ef4ddab5d010203402317f&csf=1&web=1&e=hZOz0j) (which is document 1 in a series of 10). It primarily looks at create and edit workbasket journeys but may also include updates to other sections where they are part of a specific user journey or use common shared templates.

It does not change much functionality on TAP and is more surface level changes. It also updates some view tests that failed as a result of the page content slightly changing.


common/filters.py
```diff
@@ -131,9 +131,9 @@ def __init__(self, *args, **kwargs):
 
         self.helper.layout = Layout(
             *field_layout,
-            Button("submit", "Search and Filter"),
+            Button("submit", "Search and filter"),
             HTML(
-                f'<a class="govuk-button govuk-button--secondary" href="{self.clear_url}"> Clear </a>',
+                f'<a class="govuk-button govuk-button--secondary" href="{self.clear_url}" style="margin-left:6px"> Clear </a>',
```
eadpearce, line 136: Could one of these classes be used instead of an inline style? https://design-system.service.gov.uk/styles/spacing/#overriding-spacing

mattjamc: This is cool didn't know they had override styles! Now updated

measures/tests/test_business_rules.py
```diff
@@ -1723,10 +1722,13 @@ def test_ME70():
     ),
 )
 def test_ME71_ME72(application_code, item_id, error_expected):
-    """Footnotes with a footnote type for which the application type = "CN
-    footnotes" cannot be associated with TARIC codes (codes with pos. 9-10
-    different from 00). Footnotes with a footnote type for which the application
-    type = "measure footnotes" can be associated at any level."""
+    """
+    Footnotes with a footnote type for which the application type = "CN
+    footnotes" cannot be associated with TARIC codes (codes with pos.
```
eadpearce, line 1727: I think the full stop here after 'pos' is making the autoformatter incorrectly add a line break. It should probably be removed and change 'pos' to (I'm assuming) 'position'

workbaskets/jinja2/includes/workbaskets/workbasket-business-rules.jinja
```diff
@@ -71,7 +74,7 @@
 
 {% if rule_check_in_progress %}
   {% set live_rule_check_status %}
-    Rule check in progress. Checking {{ workbasket.tracked_models.count() }} objects in basket.
+    Rule check in progress. Checking {{ workbasket.tracked_models.count() }} items in basket.
```
eadpearce, line 77: Can you add a pluralize here? `item{{ workbasket.tracked_models.count()|pluralize }}`

workbaskets/jinja2/workbaskets/delete_changes.jinja
```diff
@@ -27,7 +27,7 @@
   <div class="govuk-grid-row govuk-!-margin-bottom-5">
     <div class="govuk-grid-column-full">
       {{ govukWarningText({
-        "text": "Are you sure you want to permanently remove the " ~ object_list|length ~ " tariff changes below?",
+        "text": "Are you sure you want to permanently remove the " ~ object_list|length ~ " tariff changes below from workbasket " ~ request.user.current_workbasket.pk ~ "?",
```
eadpearce, line 30: We should pluralize here as well for "changes"

workbaskets/tests/test_views.py
```diff
@@ -242,7 +242,8 @@ def test_select_workbasket_page_200(valid_user_client):
     assert response.status_code == 200
     soup = BeautifulSoup(str(response.content), "html.parser")
     statuses = [
-        element.text for element in soup.select(".govuk-table__row .status-badge")
+        element.text.strip("\\n ")
```
eadpearce, line 245: If you instead do `soup = BeautifulSoup(response.content.decode(response.charset), "html.parser")` you can use `strip()` here without any args. (When you cast response.content to string instead of decoding it the newlines get escaped as `\\n` which doesn't get stripped whereas `\n` does)

eadpearce (approved):

PR #1161, opened 2024-02-22
TP2000-1224 Update crown logo

mattjamc:
# TP2000-1224 Update crown logo

## Why

The official GOV.UK logo has updated to feature the chosen crown of His Majesty King Charles III.

## What

This PR:
- bumps govuk-frontend to 3.15.0 which includes the new crown logo, automatically updating the favicon
- updates the govuk-header to show the new crown logo

### Before
Favicon
Screenshot 2024-02-22 at 17 21 41

Banner
image

### After
Favicon
image

Banner
image


common/jinja2/layouts/layout.jinja
```diff
@@ -28,7 +29,7 @@
   {% endif %}
 {% endset %}
 
-{{ govukHeader({
```
PanosParalakis, line 31: I believe that we need to replace the crown and not remove it completely.  Since you already bump the version to 3.15.0 I think that all you need to do is add  

{{ govukHeader({
  ...
  useTudorCrown: true
}) }}

or locate and add the image

@see https://github.com/alphagov/govuk-frontend/releases/tag/v3.15.0

SteveCN7: @mattjamc  - I think I'm inclined to leave the crown image in, the gov.uk design system indicates it must be used on any service.service.gov.uk - no mention of internal service.
I've heard of the internal directive but I think we can leave it place for now until we have a clearer steer.

mattjamc: Happy to leave it in for me - that is easier

PR #1162, opened 2024-02-22
Tp2000 1199 test measure geo area serialization

LaurenQurashi:
# TP2000-1199 Test CreateMeasure Geo Area serialization and deserialization

## Why
This PR contains the serialization and deserialization tests for the Geo Areas form - the last form in the bunch to have it's new serialization capabilities tested for the async path for creating measures
 
## What
Adds fixtures and tests to ensure that different options selected in the CreateMeasures Geo Areas form serialize and deserialize correctly. 

I was debating whether I merge these test cases into the other tests, but thought it could be cleaner to keep it specifically geo areas due to the different options you would need to test thanks to the nested forms and all that. 

## Checklist
- Requires migrations? - No


paulpepper-trade (approved):
LGTM

PR #1163, opened 2024-02-23
TP2000-1193  Upgrade to Django 4.2

dalecannon:
# TP2000-1193  Upgrade to Django 4.2

## Checklist
- Requires dependency updates? **Yes**


common/static/common/scss/_components.scss
```diff
@@ -42,6 +42,19 @@
...
+    text-decoration: none;
+  }
+
+  button.sign-out:hover {
+    text-decoration: underline;
+    text-decoration-thickness: 3px;
+    text-underline-offset: 0.1em;
+  }
+
 }
```
PanosParalakis, line 58: Using Sass this can also be typed as:

```
 button.sign-out {
    .....
    &:hover {
        .....
    }
  }
```

I believe that nesting in this case slightly improves readability and reduces repetition.  Further you can use # as a  selector since sign out button is a single specific element and add the id attribute to the element.

workbaskets/views/ui.py
```diff
@@ -1061,7 +1060,7 @@ def post(self, request, *args, **kwargs):
         return redirect("workbaskets:workbasket-ui-violations")
 
 
-class WorkBasketDelete(PermissionRequiredMixin, FormMixin, DeleteView):
+class WorkBasketDelete(PermissionRequiredMixin, DeleteView):
```
mattjamc, line 1063: Can we can now get rid of the post override method as the new DeleteView post now calls is_valid which it didn't before 4.0

quotas/forms.py
```diff
@@ -796,7 +799,7 @@ def get_geo_area_initial(self):
         field_name = "exclusion"
         initial = {}
         initial_exclusions = []
-        if hasattr(self, "instance"):
+        if self.instance.pk:
```
paulpepper-trade, line 802: Sanity check: Is this fixing a mistaken assumption about how to check whether instance was taken from the DB?

requirements.txt
```diff
@@ -8,34 +8,34 @@ black==23.3.0
 boto3==1.18.51
 celery[redis]>=5.2.2
 coverage[toml]==5.4
-crispy-forms-gds @ git+https://github.com/uktrade/crispy-forms-gds.git@b50168d0e23ffacbdd30e7819ec8f9a08e055c8e
+crispy-forms-gds @ git+https://github.com/uktrade/crispy-forms-gds.git@0ed314933ce601e208c2f263bde3943e63b1e554
```
paulpepper-trade, line 11: Sanity check: Is this pulling in our own upgrades to the crispy-forms-gds repo?

paulpepper-trade (approved):

common/forms.py
```diff
@@ -561,6 +561,7 @@ def __init__(
         error_class=ErrorList,
         form_kwargs=None,
         error_messages=None,
+        renderer=None,
```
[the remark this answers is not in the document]
dalecannon: Without the `renderer` parameter, formsets that sublass our `FormSet` raise an `AttributeError`. (Interestingly our `formset_factory()` already has it.) I'm not sure if it's because it would normally default to the renderer specified by `FORM_RENDERER` setting and we're using `TemplatesSetting`.

quotas/forms.py
dalecannon: Yeah, the `QuotaOrderNumberOriginUpdateForm` is a `ModelForm` which would always have an instance attribute, but this now checks whether the instance has been taken from the DB (i.e has a pk value) or is unsaved before attempting to call a related manager for the instance.

requirements.txt
dalecannon: Yes, it's pulling the latest current commit for that repo.

common/forms.py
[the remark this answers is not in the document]
paulpepper-trade: But I think it actually requires that the `formset_factory()` apply the correct renderer, not the `FormSet`. If we want to maintain compatibility and consistency with Django's `FormSetBase` then there shouldn't be an extra parameter `renderer` param on the `__init__()`. [Looking at how `formset_factory()` creates new FormSet classes](https://github.com/django/django/blob/79739511397367c56d20393b93c5edd35f5cca53/django/forms/formsets.py#L573), it appears that the `formset_factory()` attrs are applied to the attributes in the class's body - if I've read that correctly - and not passed in as constructor params.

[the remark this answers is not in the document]
dalecannon: Ah that's right. I've moved to be a class-wide attribute on `FormSet`.

PR #1170, opened 2024-03-04
fix all files using  pre-commit

PanosParalakis:
# TP-??? Your PR title here
Fix all files using  pre-commit

## Why
It looks like some files have not been checked using pre-commit hooks at the past. By running pre-commit against all files fixing some errors found.

## What
- Fix all errors to skipped files
- Remove whitespace from docker related files
- Remove whitespace from Makefile
- Add missing .nvmrc file for managing node version 

- Requires migrations? No


common/business_rules.py
```diff
@@ -132,8 +132,8 @@ def get_linked_models(
 
         :param model TrackedModel: Get models linked to this model instance
         :param transaction Transaction: Get latest approved versions of linked
-            models as of this transaction
-        :rtype Iterator[TrackedModel]: The linked models
+            models as of this transaction :rtype Iterator[TrackedModel]: The
```
paulpepper-trade, line 135: Is the `:rtype` meant to be on a new line, in keeping with the `:param` lines?

common/business_rules.py
```diff
@@ -226,8 +228,8 @@ def skip_when_update_type(cls: Type[BusinessRule], update_types: Iterable[Update
     """
     Skip business rule validation for given update types.
 
-    :param cls Type[BusinessRule]: The BusinessRule to decorate
-    :param update_types Iterable[int]: The UpdateTypes to skip
+    :param cls Type[BusinessRule]: The BusinessRule to decorate :param
```
paulpepper-trade, line 231: Should there be a line break before the second `:param`?

common/business_rules.py
PanosParalakis: Fix applied

common/business_rules.py
PanosParalakis: Fix applied

common/business_rules.py
```diff
@@ -226,8 +229,10 @@ def skip_when_update_type(cls: Type[BusinessRule], update_types: Iterable[Update
     """
     Skip business rule validation for given update types.
 
-    :param cls Type[BusinessRule]: The BusinessRule to decorate
-    :param update_types Iterable[int]: The UpdateTypes to skip
+    :param cls: The BusinessRule to decorate :param
```
paulpepper-trade, line 232: Is that second `:param` on this line a stray one?

PanosParalakis: Sorry no I have already pushed a fix for it :)

.nvmrc
```diff
@@ -0,0 +1 @@
+16.*
```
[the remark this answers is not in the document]
PanosParalakis: Nice one, I will update both. Thank you!

settings/dev.py
```diff
@@ -51,6 +51,6 @@
     )
 
 try:
-    from settings.dev_override import *
```
paulpepper-trade, line 54: That a very odd formatting change. I think the import needs reinstating!

PanosParalakis: This was occurred from autoflake as it was identified as an unused import. Fix applied :)

exporter/management/commands/upload_transactions.py
```diff
@@ -14,15 +14,12 @@ def run_task_or_exit(task, local=False, *args, **kwargs):
...
-    :return: Result of task.
-
-    Utility function for management commands that require the ability
-    to start a celery task, then wait for the result.
-
-    Tasks may optionally be run locally, which can be useful if the
-    user needs to run a task, but celery infrastructure is unavailable.
+    :param local: Run locally (bypass celery)
+    :param task: Celery task to run.
+    :return: Result of task. Utility function for management commands that
```
dalecannon, line 19: Should there be a line break after `:return: Result of task.` ?

common/business_rules.py
```diff
@@ -169,9 +170,11 @@ def violation(
         """
         Create a violation exception object.
 
-        :param model Optional[TrackedModel]: The model that violates this business rule
-        :param message Optional[str]: A message explaining the violation
-        :rtype BusinessRuleViolation: An exception indicating a business rule violation
+        :param model Optional[TrackedModel]: The model that violates this
+        business rule :param message Optional[str]: A message explaining the
```
dalecannon, line 174: Does `:param message` belong on a new line?

settings/dev.py
paulpepper-trade: It's a good example of why we shouldn't blindly rely upon automation tooling. In this case it would have introduced an error (in development environments that override settings)!

common/business_rules.py
PanosParalakis: All should be fine now :)

dalecannon (approved):

PR #1172, opened 2024-03-06
TP-2000-1074: New duty sentence parser + reference pages

eadpearce:
# TP-2000-1074: New duty sentence parser + reference pages

## Why

* The current duty sentence parser is difficult to understand and uses a library (parsec) that has very little documentation
* It also mixes together parsing and validation, increasing code complexity
* Duty sentence syntax itself is difficult to understand 

## What

* Implements a new parser using the Lark python library
* Separates parsing into a Lark Parser class and validation/data serialization into a Transformer class
* Adds reference pages for duty sentence syntax, DutyExpression, Measurement (MeasurementUnit and MeasurementUnitQualifier) and MonetaryUnit objects in TAP

## Checklist
- Requires migrations? **No**
- Requires dependency updates? **Yes. New package added (Lark)**

Screenshot 2024-03-06 at 16 11 57
Screenshot 2024-03-06 at 16 12 05

### DIFF
--- a/measures/duty_sentence_parser.py
+++ b/measures/duty_sentence_parser.py
@@ -2,7 +2,6 @@ from datetime import datetime
 
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.exceptions import ValidationError
-from django.template.loader import render_to_string
 from lark import Lark
 from lark import Transformer
 from lark import UnexpectedInput
@@ -59,6 +58,17 @@ class DutySentenceParser:
     """A dumb parser that does minimal validation and only matches a duty
     sentence according to a set pattern."""
 
+    @staticmethod
+    def create_rule(items, field_name):
+        output = []
+        for i, item in enumerate(items, start=1):
+            if getattr(item, field_name):
+                if i != len(items):
+                    output.append(f'"{getattr(item, field_name)}"i | ')
+                else:
+                    output.append(f'"{getattr(item, field_name)}"i')
+        return "".join(output)
+
     def __init__(self, date=datetime.now()):
         duty_expressions = models.DutyExpression.objects.as_at(date)
         monetary_units = models.MonetaryUnit.objects.as_at(date)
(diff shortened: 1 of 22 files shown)


measures/duty_sentence_parser.py
```diff
@@ -0,0 +1,437 @@
...
+    sentence according to a set pattern."""
+
+    @staticmethod
+    def create_rule(items, field_name, strip=False):
+        output = []
+        for item in items:
+            if getattr(item, field_name):
+                val = getattr(item, field_name)
+                if strip:
+                    # For measurement abbreviations we want to be able to match with and without spaces and commas (in the case of measurement units in the thousands) so add the spaceless and commaless values to the rule as well
```
PanosParalakis, line 72: Nothing super important but from readability point of view might be better if this long comment can be break to more lines

measures/tests/test_lark_parser.py
```diff
@@ -0,0 +1,305 @@
...
+        request,
+    )
+
+
+def test_only_permitted_measurements_allowed(lark_duty_sentence_parser):
+    with pytest.raises(ValidationError) as e:
+        lark_duty_sentence_parser.transform("1.0 EUR / kg / lactic.")
+        assert (
+            e.message
+            == "Measurement unit qualifier lactic. cannot be used with measurement unit kg."
```
PanosParalakis, line 247: Is the dot after lactic needed?

eadpearce: Yes because it matches the abbreviation of the measurement unit given in the duty sentence above

measures/duty_sentence_parser.py
eadpearce: Huh. I was expecting our autoformatter to do that since it does it for docstrings. There may be a setting to tweak somewhere. I'll fix it manually for now

measures/tests/test_lark_parser.py
eadpearce: We could change the fixture to remove the dot but it's probably useful to have one in this format since it matches one we currently use in production (% sacchar.)

PanosParalakis (approved):
LGTM 👍🏻

measures/duty_sentence_parser.py
```diff
@@ -0,0 +1,439 @@
...
+        duty_expressions = models.DutyExpression.objects.as_at(date) or duty_expressions
+        monetary_units = models.MonetaryUnit.objects.as_at(date) or monetary_units
+        measurement_units = (
+            models.MeasurementUnit.objects.as_at(date) or measurement_units
+        )
+        measurement_unit_qualifiers = (
+            models.MeasurementUnitQualifier.objects.as_at(date)
+            or measurement_unit_qualifiers
+        )
+        measurements = models.Measurement.objects.as_at(date) or measurements
```
dalecannon, line 114: Do the boolean expressions need swapping for it to load any custom lists?

PR #1173, opened 2024-03-08
TP2000-1241 Implement new homepage design

dalecannon:
# TP2000-1241 Implement new homepage design

## Why
The current homepage is becoming less in keeping with other services under the data portfolio.

## What
- Applies new header and footer across the application

- Updates homepage layout and content to match the design
    - Replaces `workbasket_action.jinja` with `homepage.jinja` template
    - Uses custom CSS styling
    - Displays relevant cards/links/header items according to user permissions
    - Removes now unused `HomeForm`

- Introduces a new search capability to homepage
    - Adds a new form, `HomeSearchForm`, using an alphanumeric validator and limiting input to 18 chars (longest possible search term for its purpose)
    - Redirects to the detail view of the tariff element if a match is found
    - Redirects to the search page if no match can be found

- Adds unit tests

## Before
Screenshot 2024-03-08 at 10 56 05

## After
Screenshot 2024-03-14 at 11 00 11


common/views.py
```diff
@@ -26,50 +26,189 @@
...
-        return kwargs
+        assignments = (
+            UserAssignment.objects.filter(user=self.request.user)
+            .assigned()
+            .select_related("task__workbasket")
+        )
+        assigned_workbaskets = []
+        for assignment in assignments:
+            workbasket = assignment.task.workbasket
+            if workbasket.status == WorkflowStatus.EDITING:
```
paulpepper-trade, line 74: Would this filtering be better done by the database? Could `assignments` be filtered so that only those `UserAssignments` that are related to workbaskets with `EDITING` status are selected?

paulpepper-trade, line 74: How do we deal with workbaskets that have a status of `ERRORED` to allow users to re-edit them?

dalecannon: I've now filtered them in following the suggestion above so they'll appear in the list too.

common/jinja2/layouts/layout.jinja
```diff
@@ -4,116 +4,150 @@
...
 {% block pageTitle %}
   {{ page_title }} | {{ service_name }}
 {% endblock %}
 
 {% set workbasket_svg %}
-<svg width="25" height="25" viewBox="0 0 362 360" fill="none" xmlns="http://www.w3.org/2000/svg"> {# /PS-IGNORE #}
-<path d="M316.75 135H295.332L239.343 37.5601L213.158 52.4401L260.595 135H101.405L148.858 52.4401L122.658 37.5601L66.6686 135H45.2502C42.9319 135.003 40.6452 135.536 38.5676 136.559C36.4899 137.582 34.6771 139.067 33.2696 140.899C31.862 142.731 30.8976 144.861 30.4511 147.123C30.0046 149.386 30.088 151.72 30.6948 153.945L72.9131 307.905C74.6681 314.247 78.4654 319.842 83.7233 323.833C88.9811 327.823 95.4096 329.99 102.024 330H259.992C273.536 330 285.513 320.925 289.102 307.875L331.321 153.93C331.935 151.704 332.023 149.367 331.578 147.102C331.133 144.836 330.168 142.704 328.757 140.871C327.346 139.038 325.528 137.554 323.446 136.536C321.363 135.517 319.071 134.992 316.75 135ZM259.977 300V315V300H102.024L65.0094 165H296.991L259.977 300Z" fill="white"/> {# /PS-IGNORE #}
-<path d="M135.75 195H165.917V270H135.75V195ZM196.083 195H226.25V270H196.083V195Z" fill="white"/> {# /PS-IGNORE #}
-</svg>
+  <svg
```
paulpepper-trade, line 15: We normally put SVGs in their own file and load them using ``. Worth doing here too?

common/views.py
```diff
@@ -19,63 +19,210 @@
...
+                    "workbaskets.add_workbasket",
+                ),
+                "can_view_reports": self.request.user.has_perm(
+                    "reports.view_report_index",
+                ),
+            },
+        )
+        return context
+
+    def get_search_result(self, search_term):
```
paulpepper-trade, line 107: A docstring would be useful to understanding how this method works. For instance what `search_term` is and values it can take (only keys phrases in the keys of `list_view_map` for instance), where / what `list_view_map` is referencing (is there a duplication of this map somewhere?).

PR #1174, opened 2024-03-11
Tp2000-1243 Cancel task journey

paulpepper-trade:
# TP2000-1243 Cancel task journey

## Why

Past experience of using Celery to perform asynchronous processing tells us that we should provide a means of cancelling tasks that "get stuck" or in some way fail, leaving tasks in an incomplete state.

## What

This PR adds a journey, accessible to superusers from the measures process queue, that allows measures bulk creation tasks in one of the queued states to be cancelled. Specifically, this PR:

- Adds a view asking the user to confirm whether they would like to cancel a task.
- Adds a done view, displayed after a user has confirmed cancelling a task.
- Adds unit tests for views, testing scenarios with various combinations of processing state and user capability.

**Cancel process confirmation form**
image

**Cancel process done view**
image

## Checklist
- Requires migrations? No

### DIFF
--- a/measures/jinja2/measures/cancel-bulk-processor-task-done.jinja
+++ b/measures/jinja2/measures/cancel-bulk-processor-task-done.jinja
@@ -1,0 +1,37 @@
+{% extends "layouts/layout.jinja" %}
+
+{% from "components/panel/macro.njk" import govukPanel %}
+{% from "components/button/macro.njk" import govukButton %}
+
+
+{% set page_title = "Cancel process done" %}
+
+
+{% block breadcrumb %}
+  {{ breadcrumbs(
+      request,
+      [
+        {"text": "Measures process queue", "href": url("measure-create-process-queue")},
+        {"text": page_title},
+      ],
+      with_workbasket=False,
+  ) }}
+{% endblock %}
+
+
+{% block content %}
+  <div class="govuk-grid-row">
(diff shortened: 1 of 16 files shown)


measures/jinja2/measures/cancel-bulk-processor-task-done.jinja
```diff
@@ -0,0 +1,37 @@
+{% extends "layouts/layout.jinja" %}
+
+{% from "components/panel/macro.njk" import govukPanel %}
+{% from "components/button/macro.njk" import govukButton %}
+
+
+{% set page_title = "Cancel process done" %}
```
CPrich905, line 7: Ash may be the best person to confirm this (and may already have done so, in which case disregard this comment) but given that the cancellation process is not guaranteed, would `"Cancel process requested"` be more appropriate?

paulpepper-trade: We'll hopefully review the content later today, Charles, so will wait to see what comes out of that. But I agree with your comment :-)

CPrich905 (approved):
LGTM, I have one other question but it's related to the behaviour after cancellation rather than the PR, so will take it off line.

CPrich905 (approved):
Approval for changes related to display name

PR #1175, opened 2024-03-13
TP2000-1249 Archive empty, previously assigned workbaskets upon deletion

dalecannon:
# TP2000-1249 Archive empty, previously assigned workbaskets upon deletion

## Why
Users with the requisite permission are able to delete a workbasket with `EDITING` status once its `Transactions` and `TrackedModels` have been removed. However, a workbasket that has previously been assigned to a user will have a non-nullable `Task` instance associated with it, preventing deletion.

## What
- Archives empty, previously assigned workbaskets, giving the appearance of workbasket deletion while preventing a `ProtectedError` from being raised

### DIFF
--- a/workbaskets/views/ui.py
+++ b/workbaskets/views/ui.py
@@ -1087,13 +1087,13 @@ class WorkBasketDelete(PermissionRequiredMixin, DeleteView):
 
     def form_valid(self, form):
         if (
-            not PackagedWorkBasket.objects.filter(workbasket=self.object).exists()
-            and not self.object.tasks.exists()
+            PackagedWorkBasket.objects.filter(workbasket=self.object).exists()
+            or self.object.tasks.exists()
         ):
-            self.object.delete()
-        else:
             self.object.archive()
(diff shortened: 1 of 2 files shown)


workbaskets/views/ui.py
```diff
@@ -1086,7 +1086,10 @@ def get_form_kwargs(self):
...
 
     def form_valid(self, form):
-        if not PackagedWorkBasket.objects.filter(workbasket=self.object).exists():
+        if (
+            not PackagedWorkBasket.objects.filter(workbasket=self.object).exists()
+            and not self.object.tasks.exists()
+        ):
             self.object.delete()
         else:
             self.object.archive()
```
paulpepper-trade, line 1095: Now that a compound bool expression is required, I think this could be simplified.
```suggestion
        if (
           PackagedWorkBasket.objects.filter(workbasket=self.object).exists()
           or self.object.tasks.exists()
        ):
            self.object.archive()
        else:
            self.object.delete()
```

PR #1176, opened 2024-03-13
TP2000-1119 List view filtering

CPrich905:
# TP2000-1119 Task Queue view filtering

## Why
The new Task Queue view page requires users to be able to filter tasks by `ProcessingState`

## What
- adds functionality to filter for tasks which are Processing, ie. have a `ProcessingState` of `CURRENTLY_PROCESSING` or `AWAITING_PROCESSING`
- uses TamatoFilter to filter for the other states using the TamatoFilter
- tests that applying the filters displays the relevant task

Note: This PR has a work around for tasks which are Processing. We should investigate a refactor of/addition to the TamatoFilter that will allow querying using multiple parameters on the same filed, eg. a query of `?processing_state=AWAITING_PROCESSING&processing_state=CURRENTLY_PROCESSING` would return relevant tasks.
image

### DIFF
--- a/measures/views.py
+++ b/measures/views.py
@@ -13,6 +13,7 @@ from crispy_forms.helper import FormHelper
 from django.conf import settings
 from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.contrib.auth.mixins import UserPassesTestMixin
+from django.db.models import Q
 from django.http import HttpRequest
 from django.http import HttpResponse
 from django.http import HttpResponseRedirect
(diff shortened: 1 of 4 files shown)


measures/views.py
```diff
@@ -1116,48 +1116,47 @@ class MeasuresCreateProcessQueue(
...
+    def get_processing_queryset(self):
+        """Returns a combined queryset of tasks either AWAITING_PROCESSING or
+        CURRENTLY_PROCESSING."""
+        awaiting_processing = self.queryset.filter(
+            processing_state="AWAITING_PROCESSING",
+        )
+        currently_processing = self.queryset.filter(
+            processing_state="CURRENTLY_PROCESSING",
+        )
+        return awaiting_processing.union(currently_processing)
```
paulpepper-trade, line 1158: I don't think a union is the right expression here. A pair or or'd Q objects are the right way to go. Also, accessing state values via `ProcessingState` is a safer and more maintainable approach:
```suggestion
        return self.queryset.filter(
            Q(processing_state=ProcessingState.AWAITING_PROCESSING)
            | Q(processing_state=ProcessingState.CURRENTLY_PROCESSING)
        )
```

paulpepper-trade (changes requested):

paulpepper-trade (approved):

PR #1182, opened 2024-03-21
TP2000-1255 Correct homepage card permission

dalecannon:
# TP2000-1255 Correct homepage card permission

## Why
The ‘Currently working on’ card, which details a user’s assigned workbaskets, is shown to all users with the ‘view workbaskets' permission. However, not all users who can view workbaskets can be assigned to them.

## What
- Changes permission from 'can_view_workbasket' to 'can_edit_workbasket' (i.e 'workbaskets.change_workbasket') so that only the users who can be assigned to work on and to review workbaskets can see card

### DIFF
--- a/common/jinja2/common/homepage.jinja
+++ b/common/jinja2/common/homepage.jinja
@@ -32,7 +32,9 @@
                 class="govuk-button homepage-workbasket-action"
               >Create a <br> workbasket</a>
             </div>
+          {% endif %}
 
+          {% if can_edit_workbasket %}
             <div class="govuk-grid-column-one-half">
               <a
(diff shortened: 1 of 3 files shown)


common/jinja2/common/homepage.jinja
```diff
@@ -58,7 +58,9 @@
           </div>
         </div>
       </article>
+    {% endif %}
 
+    {% if can_add_workbasket %}
```
paulpepper-trade, line 63: Would assignment to a workbasket imply edit permissions on a workbasket rather than the permission to add?

paulpepper-trade (approved):

PR #1187, opened 2024-03-22
TP2000-1236  Improve measure search performance

dalecannon:
# TP2000-1236  Improve measure search performance

## Why
Searching for measures using the end date filter can result in slow loading times and eventual timeouts: (1) calls to the view's paginator are duplicated many times over and (2) performing a `.count()` operation on a measure queryset chained with the filter `.with_effective_valid_between()` (added to enable filtering by end date) appear to be contributing to performance issues.

## What
- Caches `limit_breached` and `count` paginator methods
- Caches the view's `paginator` property
- Evaluates the paginator's queryset to determine its length as a faster alternative to the slow count() operation (and without affecting non-end-date-filtered searches)
- Chains a `.select_related()` filter on the queryset to fetch objects used in the template
- Removes `object_list` from the view's context as the view's form is used to handle the resulting measure objects
- Passes the sids of any selected measures to the template rather than measure objects
- Fixes a couple `KeyError`s resulting from missing filter values (for instance, searching for a measure SID that doesn't exist or not entering all date values in a date filter)
- Amends the button text and aria label of the filter accordion to reflect whether it is expanded or not

## Screenshots

**Before**:
Using the search criteria detailed in the ticket, loading times on a local dev env sat around 5 minutes
Screenshot 2024-03-22 at 14 40 23 Screenshot 2024-03-22 at 14 40 30
Screenshot 2024-03-22 at 11 29 16

**After**:
Using the search criteria detailed in the ticket, loading times on a local dev env now sit around 10 seconds
Screenshot 2024-03-22 at 14 43 19 Screenshot 2024-03-22 at 14 43 12

### DIFF
--- a/measures/views.py
+++ b/measures/views.py
@@ -303,7 +303,17 @@ class MeasureList(
     def paginator(self):
         filterset_class = self.get_filterset_class()
         self.filterset = self.get_filterset(filterset_class)
-        return MeasurePaginator(self.filterset.qs, per_page=40)
+        return MeasurePaginator(
+            self.filterset.qs.select_related(
+                "additional_code",
+                "generating_regulation",
+                "geographical_area",
+                "goods_nomenclature",
+                "measure_type",
+                "order_number",
+            ),
+            per_page=40,
+        )
 
     def get_context_data(self, **kwargs):
         # References to page or pagination in the template were heavily increasing load time. By setting everything we need in the context,
--- a/measures/views.py
+++ b/measures/views.py
@@ -334,7 +334,6 @@ class MeasureList(
                 "has_next_page": page.has_next(),
                 "page_number": page.number,
                 "list_items_count": self.paginator.per_page,
-                "object_list": page.object_list,
                 "page_links": build_pagination_list(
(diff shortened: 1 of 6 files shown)


measures/views.py
```diff
@@ -266,7 +262,10 @@ def selected_filter_formatter(self) -> List[List[str]]:
             footnote = Footnote.objects.current().get(id=selected_filters["footnote"])
             selected_filters_strings.append(f"Footnote {footnote.structure_code}")
 
-        if "start_date_0" and "start_date_1" and "start_date_2" in selected_filters:
+        if all(
```
CPrich905, line 265: One very minor point would be that there might be a better name rather than `sf`? 
Otherwise LGTM

CPrich905 (approved):

measures/views.py
paulpepper-trade: I think the `sf` has such narrow scope (within the context of the comprehension) that we could let it go. Feel free to change it though.

eadpearce (approved):

PR #1188, opened 2024-03-22
TP2000-1254 Set Celery concurrency to 1

paulpepper-trade:
# TP2000-1254 Set Celery concurrency to 1

## Why

If more than one Celery task executes measures bulk creation, then there'll be contention over IDs for identifying fields in the bulk instances. This is because TAP relies upon the default database isolation level - READ COMMITTED. The first of multiple concurrent transactions will succeed, while others fail (and roll-back) due to duplicate IDs.

## What

This PR:
- Ensures Celery tasks used to bulk create measures are executed in serial order by applying a concurrency setting of one to the Celery worker.
- Concurrency configuration has been created for CF infrastructure and for docker compose, causing a single worker instance only to be spawed by the controller Celery process.
- The changes as they apply to docker containers and CF infrastructure have been tested, ensuring only one task is executed at any one time.
- The same concurrency pattern has been applied for the rules check Celery worker.

## Checklist
- Requires migrations? No

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,6 +1,6 @@
 web: gunicorn wsgi --bind 0.0.0.0:$PORT --timeout 1000 --worker-class=gevent --worker-connections=1000 --workers 9
 worker: celery -A common.celery worker -O fair -l info -Q standard
 beat: celery -A common.celery beat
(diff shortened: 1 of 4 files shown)


CPrich905 (approved):
LGTM :)

PR #1190, opened 2024-03-25
Quota definition create link fix

eadpearce:
# Quota definition create link fix

## Why

* User research flagged that the create link on the confirm create page for quota definitions is broken

## What

* Updates the get_url method to make sure the correct url is generated
* Removes the try/except so it no longer fails silently
* Adds tests


CPrich905 (approved):

CPrich905:
Looks like I might have jumped the gun on approving - failing test 
`NoReverseMatch: Reverse for 'quota_definition-ui-detail' not found. 'quota_definition-ui-detail' is not a valid view function or pattern name.`

PR #1191, opened 2024-03-25
Content updates based on tap review 1 and 3 changes

mattjamc:
# Content updates based on tap review 1 and 3 changes

## Why

User journeys across TAP vary in content and style and it's better for users to be consistent throughout.

## What

This PR addresses comments from Ash's [TAP Review 1 - updates based on changes ](https://dbis-my.sharepoint.com/:w:/r/personal/ash_patel_businessandtrade_gov_uk/Documents/Review%20of%20TAP%20UI%20shared%20with%20Wren/TAP%20Review%201%20-%20updates%20based%20on%20changes_29_02_24.docx?d=w104f0f7836f34adabb20a0c000fc9a5d&csf=1&web=1&e=MLRt4S) and [TAP Review 3 - updates based on changes ](https://dbis-my.sharepoint.com/:w:/r/personal/ash_patel_businessandtrade_gov_uk/Documents/Review%20of%20TAP%20UI%20shared%20with%20Wren/TAP%20Review%203%20updates%20based%20on%20changes_14_03_24.docx?d=w0c4a8cb9e7f24136bbb4c6ba763e5aeb&csf=1&web=1&e=aKabDq). It primarily looks at create and edit workbasket journeys but may also include updates to other sections where they are part of a specific user journey or use common shared templates.

It does not change any functionality on TAP and is more surface level changes. It also updates some view tests that failed as a result of the page content slightly changing.


additional_codes/views.py
```diff
@@ -238,7 +238,9 @@ class AdditionalCodeDescriptionEditCreate(
     """UI endpoint for editing AdditionalCodeDescription CREATE instances."""
 
     form_class = AdditionalCodeDescriptionForm
-    template_name = "common/edit_description.jinja"
+
+    def get_template_names(self):
```
eadpearce, line 242: This change isn't necessary. Usually you'd only want to use the `get_template_names` method if you need to access some attributes on `self` such as `request` to generate the return value

eadpearce (approved):

additional_codes/views.py
mattjamc: It seemed to be using the _"common/edit.jinja"_ from the super class which was wrong despite the correct template_name being specified here

Overriding super's get_template_names() seemed to work

mattjamc: See here on UAT the template it's getting for description edit-create is defo not right - https://tamato-uat.london.cloudapps.digital/footnotes/TN049/description/201357/edit-create/

eadpearce: I wonder if we changed EditTaricView to use the template_name attribute instead of the method would that fix template overriding issues

mattjamc: That works! As opposed to my fix which doesn't seem to be working anymore anyway

PR #1202, opened 2024-04-08
feat: add origins to quotas expiring soon reports

nboyse:
# Add quota order number origins to the quotas expiring soon reports

## Why
Requested by product manager, small change to add a new column to show users the quota's associated origins for the expiring soon report

## What
Adding in a new element to the existing array in the python script that generates the expiring soon report


mattjamc (approved):
LGTM :)

PR #1204, opened 2024-04-11
TP2000-1295 Dockerise tamato unit tests

paulpepper-trade:
# TP2000-1295 Dockerise tamato unit tests

## Why

There’s currently no docker-based way of directly executing Tamato’s unit tests from a development environment. For instance, piggybacking on the tamato container service, defined in Tamato's docker-compose.yml, and opening a tty to execute pytest from a container shell instance, results in spurious / incorrect failures.

This makes it difficult to reliably test environment changes, such as Python upgrades, within a clean environment from a developer workstation, because the (unclean) host environment has to be used. 

## What

This PR:
- Fixes the issue causing the spurious unit test errors described above.
- Defines a new Dockerfile from which a testing image can be built (`Dockerfile.pytest`). This Dockerfile has been kept to a minimal working definition and could be effectively used to as the Dockerfile for other Tamato-based images (i.e. Celery workers and Tamato web worker).
- Create a docker-compose service to run all Tamato unit tests using pytest. An associated dockerised clean / empty test database instance should be started as part of the test process.

## Checklist
- Requires migrations? No


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
-version: '3'
+version: "3"
 services:
   db:
-    image: "postgres:12.8"
+    image: "postgres:16"
```
PanosParalakis, line 4: Will bumping postgres version create any deployment problems?

paulpepper-trade: No, this is docker, which isn't used in any of our staging or production environments. The PR is still in draft, so we'd normally hold off on comments as it's a work in progress.

CPrich905 (approved):

PR #1206, opened 2024-04-15
feat: link geo areas to origins on expiring quotas report

nboyse:
# Link a quota order number origin to the relevant geo area on the expiring quotas report

## Why

For easability of report usage, we are allowing users to have a follow-through journey to the relevant geo area of related origins. CSV export was also giving the whole link values for both quota order number and quota origins so have fixed that.
## What

Simple update of logic so picks up if the row is a link and only grabs the value using regex. Also adding a new function that renders a link for quota order number origins.


PanosParalakis (approved):
Looks OK  👍🏻

PR #1212, opened 2024-04-18
TP2000-1264: Add new duty sentence parser to forms

eadpearce:
# TP2000-1264: Add new duty sentence parser to forms

## Why

* The new duty sentence parser had not yet been hooked up to fields that required it in forms

## What

* Plugs in the duty sentence parser to forms so we can have improved error messages
* Tweaks the styling of the commodities and duties page of the measure create journey so the long error text doesn't break the layout
* Also fixes a flaky test caused by random values being assigned to MeasureAction.code in the MeasureActionFactory

Screenshot 2024-04-02 at 12 12 58


measures/forms.py
```diff
@@ -1352,18 +1360,16 @@ def clean(self):
...
+        duty_sentence_parser = LarkDutySentenceParser(
+            date=self.measure_start_date or datetime.datetime.now(),
         )
         for duty, form in duties:
             try:
-                duty_sentence_parser.parse(duty)
-            except ParseError as error:
-                error_index = int(error.loc().split(":", 1)[1])
+                duty_sentence_parser.transform(duty)
+            except (SyntaxError, ValidationError) as e:
```
dalecannon, line 1369: Does the new parser validation allow for empty values in cases where the selected measure type doesn't permit duties to be specified (for example measure type 278)?

eadpearce: I didn't come across that logic in the old parser. I assume that validation is at the form level (since it needs data from elsewhere in the form), not the parser level

dalecannon: Ah, it turns out the old parser considered an empty value as a valid match. But it looks like the new parser expects one or more phrases. Would it be preferable to add an empty rule (e.g `!sentence: phrase+ | `) or attempt to parse `duty` only if it has a value (to prevent possible conflicts in validation https://github.com/uktrade/tamato/blob/master/measures/forms.py#L1478)?

measures/forms.py
```diff
@@ -280,6 +280,7 @@ def conditions_clean(
                 )
 
         if price and not price_errored:
+            # TODO: add condition sentence parsing functionality to the new parser
```
paulpepper-trade, line 283: This might be better tracked as a Jira ticket. It'd allow more description and give it a better chance of being schedule in for implementation.

paulpepper-trade (approved):
LGTM but can see that @dalecannon has an interest in this PR too.

measures/forms.py
eadpearce: It's also going to be a jira ticket but thought I'd flag in the code that the new parser doesn't yet have this functionality to avoid confusion

measures/forms.py
eadpearce: Updated this. Didn't break tests. Going to assume it's fine 🤞

measures/forms.py
SteveCN7: I'm going to add this to the measures improvement epic, with conditions, i'd also like to prototype a templating solution, so it feels there are a couple of opportunities to explore.

The big difference between measure.duty and measure.condition.duty is that you don't do compound duties in the condition - so it'll be simpler maybe?

PR #1213, opened 2024-04-18
TP2000-1279 Upgrade to Python version 3.9

paulpepper-trade:
# TP2000-1279 Upgrade to Python version 3.9

## Why

Platform migration requires a minimum of Python version 3.9.

## What

This PR:
- Provides an initial, intermediate upgrade step unblocking other upgrade activities that require 3.9 or greater.
- Upgrades Python in Docker config files.
- Upgrades Python in Cloud Foundry config.
- Fixes module imports and requirements dependencies.
- Pins the set of Sphinxcontrib packages to the latest version requiring Sphinx versions less than version 5. This is because Sphinx does not pin its dependencies and Python version 3.9 or greater causes Sphinx's most recent dependencies to be used, which in turn pulls in Sphinx version 5. However, only one version of the dependent package govuk-tech-docs-sphinx-theme exists and it requires versions of Sphinx less than 5.

## Checklist
- Requires migrations? No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,10 +109,9 @@ celerybeat.pid
 .env.*
 .venv
 env/
-venv/
+venv*/
 ENV/
 env.bak/
-venv.bak/
 docker-compose.override.yml
 settings/envs/docker.override.env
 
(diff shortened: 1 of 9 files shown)


PanosParalakis (approved):
👍🏻

PR #1216, opened 2024-04-23
TP2000-1277 AWS X -Ray integration

PanosParalakis:
TP2000-1277 AWS X -Ray integration

## Why
This ticket resolves the need for integrating X-Ray as required
for migrating to DBT platform

## What
* Installed dbt-copilot-python package
* Introduced scripts folder used from Procfile in order
changes to be backwards compatible
* Integrate X-Ray at WSGI config

## Checklist
- Requires migrations? No


scripts/entry.sh
paulpepper-trade, line 1: Perhaps worth naming this file script something along the lines of, `web-worker-entrypoint.sh`, to make its purpose clear.

scripts/entry.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+
+if [ -n "${COPILOT_ENVIRONMENT_NAME}" ]; then
+    echo "Running in DBT Platform"
+    opentelemetry-instrument gunicorn wsgi --bind 0.0.0.0:$PORT --timeout 1000 --worker-class=gevent --worker-connections=1000 --workers 9
```
paulpepper-trade, line 5: I'm unsure how much value there is in logging which platform the script is executing on once this script has been tested. Although it would just be a shortcut to reading this script, it might still be more useful to log the command used to start the web workers. And perhaps it could be refactored to something like:

```
#!/bin/sh -e

WORKER_COMMAND="gunicorn wsgi --bind 0.0.0.0:$PORT --timeout 1000 --worker-class=gevent --worker-connections=1000 --workers 9"
    
if [ -n "${COPILOT_ENVIRONMENT_NAME}" ]; then
    WORKER_COMMAND="opentelemetry-instrument ${WORKER_COMMAND}

echo "Starting web workers with: '${WORKER_COMMAND}'"

eval $WORKER_COMMAND
```

common/tests/test_models.py
```diff
@@ -856,7 +856,7 @@ def test_user_get_displayname():
     )
     assert user.get_displayname() == str(user)
 
-    user.email = "tester@test.com"
+    user.email = "tester@test.com"  # /PS-IGNORE Ignore false positive secret check
```
paulpepper-trade, line 859: Might this be preferably ignored by adding to the `pii-ner-exclude.txt` file, reducing comment noise a little?

paulpepper-trade (approved):

scripts/entry.sh
PanosParalakis: No problem, I will apply that change

common/tests/test_models.py
PanosParalakis: According to the documentation this is the preferred way.

`In order to exclude files from the checks, add them to this file. HOWEVER, you should heavily favour excluding lines using /PS-IGNORE, rather than files.`

paulpepper-trade: Aaagh! That's a shame as it's encouraging clutter in source code for the sake of quietening a tool that is throwing up false positives!
But your call, anyway, Panos.

scripts/web-worker-entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+WORKER_COMMAND="gunicorn wsgi --bind 0.0.0.0:$PORT --timeout 1000 --worker-class=gevent --worker-connections=1000 --workers 9"
+
+if [ -n "${COPILOT_ENVIRONMENT_NAME}" ]; then
+    echo "Running in DBT Platform"
+    WORKER_COMMAND="opentelemetry-instrument ${WORKER_COMMAND}
```
paulpepper-trade, line 7: I think the string literal part of this line is missing a closing quote.

PR #1217, opened 2024-04-23
TP2000-1275 Output TAP logs in ASIM format

mattjamc:
# TP2000-1275 Output TAP logs in ASIM format

## Why

As part of the migration to DBT PaaS, TAP must start outputting logs in [ASIM format](https://learn.microsoft.com/en-us/azure/sentinel/normalization). This change will ensure that several teams will be able to analyse your application logs for indicators of compromise, error trends and debugging issues.

## What

This PR:

- Incorporates the [django-log-formatter-asim ](https://github.com/uktrade/django-log-formatter-asim) package
- Adds a new formatter and handler to TAP Django logging settings
- Makes TAP django, celery and root loggers output to the ASIM handler when TAP is running in an AWS Copilot provisioned environment

Next steps:
- Inspect logs once we are on DBT PaaS and see how they are getting formatted
- This works fully locally but when deployed it doesn't format HTTP 3XX or 2XX responses. Locally these are handled by the django.server logger which doesn't work in production - instead it looks like cloud foundry is creating the 2XX and 3XX logs which don't pass through the Django logging handler created.


paulpepper-trade (approved):
> ... it looks like cloud foundry is creating the 2XX and 3XX logs which don't pass through the Django logging handler created.

... or perhaps the WSGI server?

PR #1220, opened 2024-04-25
TP2000-1331  Fix certificates related measure queryset error

dalecannon:
# TP2000-1331  Fix certificates related measure queryset error

## Why
Users are encountering server errors because of an unhandled `Measure.DoesNotExist` exception when:
1) filtering by certain certificates to search for measures (for example certificate Y901)
2) navigating to the measures tab on the detail view of certain certificates (for example certificate C672)

## What
- Catches the exception in `MeasureFilter.certificates_filter()` which is raised when calling .get() on the queryset that has returned empty because the latest version of the measure is excluded as it has since been deleted (update_type=DELETE).

- Updates the certificates filter unit test

- Refactors the same queryset in `CertificateDetailMeasures.get_queryset()`, preventing the exception from being raised but also improving response times for the view.


measures/filters.py
```diff
@@ -258,24 +258,27 @@ def measures_filter(self, queryset, name, value):
...
 
             measure_ids = []
             for condition in measure_conditions:
-                measure_ids.append(
-                    condition.dependent_measure.get_versions().current().get().pk,
-                )
+                try:
+                    # `dependent_measure` may not be the latest version of the measure.
+                    measure_ids.append(
+                        condition.dependent_measure.get_versions().current().get().pk,
```
paulpepper-trade, line 277: I wonder if there's any API or doc enhancement that we apply to make it clearer that a DELETE on the last version update will result in `current()` returning an empty set.

paulpepper-trade (approved):

PR #1222, opened 2024-04-26
Update Configuration Env Variables for DBT PaaS Migration

LaurenQurashi:
# TP2000-1274 Update Configuration env variables 

## Why
As we're migrating from Govuk PaaS to DBT PaaS, the way our env variables are obtained from Paas need to change. Previously we would access variables provided by Govuk paas through the VCAP_SERVICES variable. Once we've migrated, we need to get them through the CoPilot helper methods. 

## What
This PR makes some amendments to the common.py file to cater for all three environment cases we may have over the next coming months.. DBT PaaS hosting, Govuk PaaS hosting, and local development. We've also had to make some amendments to the existing AWS keys. Anything regarding an access key id or secret access key has had DBT PaaS if blocks wrapped around them to set them to None as default, as DBT PaaS will be setting these in a different way in future. 

## Checklist
- Requires migrations? - No

Links to relevant material
- [Codebase Prep Docs](https://platform.readme.trade.gov.uk/migrations/conducting-a-migration/02b-codebase-preparation/#application-settings)


settings/common.py
```diff
@@ -269,14 +274,22 @@
 # -- Security
 SECRET_KEY = os.environ.get("SECRET_KEY", "@@i$w*ct^hfihgh21@^8n+&ba@_l3x")
 
+
 # Whitelist values for the HTTP Host header, to prevent certain attacks
 # App runs inside GOV.UK PaaS, so we can allow all hosts
 ALLOWED_HOSTS = re.split(r"\s|,", os.environ.get("ALLOWED_HOSTS", ""))
+
+# DBT PaaS
+if is_copilot():
```
paulpepper-trade, line 283: I think it would improve readability to integrate this condition block with those for the other platform cases below - for instance, using `elif "VCAP_APPLICATION" in os.environ:` in the following condition.

paulpepper-trade (approved):
It'd be good to change the PR title from "Amend ALLOWED_HOSTS..." to something about config env vars to aid recognising what this PR does in the future.

Otherwise LGTM!

PR #1232, opened 2024-05-14
Celery log level always be DEBUG

mattjamc:
# Celery log level always be DEBUG

## Why

Since updating the Celery logging in TP2000-1275 there has been an error at 7:05pm every day when the sqlite export task is run.
* ValueError: Unable to configure logger 'celery'

This comes from the Celery logger being assigned a loglevel of '20'. Log levels can only be integers (10, 20 etc) or set string values ("INFO", "DEBUG" etc). A string of '20' does not work.

[TP2000-1275](https://github.com/uktrade/tamato/blob/3d1922c66bc066304a83fb1ef507c01bc07a1068/common/celery.py#L20) overrode the default Celery logging to allow us to do ASIM formatting. This seems to be causing the new error.

I've not yet been able to figure out this error and debugging is difficult since:

- I cannot reproduce this error locally - the task seems to run without any logger issues on my machine
- I understand Celery beat / sqlite export isn't used in non-production environments so I'm not able to experiment in the test environment

## What

I suspect the string value '20' is coming from Celery's logging configuration which converts the log level number to a string and then assigns this to the env variable CELERY_LOG_LEVEL here:
- https://github.com/celery/celery/blob/4cf6ba3858779a3b71c87001133506c286aec0cd/celery/app/log.py#L69

I'm unsure why this is happening but wasn't before / how to resolve it. In particular why it's only an issue for this one task.

A potential patch is to just ignore the CELERY_LOG_LEVEL env variable. It is not a variable we set or have used before.

This PR stops the celery logger getting the log level from the env variable and instead just sets it to "DEBUG" all the time.

[TP2000-1275]: https://uktrade.atlassian.net/browse/TP2000-1275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


paulpepper-trade (approved):
Let's test it then!

PR #1238, opened 2024-05-30
TP2000-1381 Celery logging config error pt2

paulpepper-trade:
# TP2000-1381 Celery logging config error pt2

## Why

The sqlite dump process executes Django commands in new processes. In doing so, environment variables are copied into the environment. One of the environment variables, CELERY_LOG_LEVEL, is incorrectly represented as a string integer "20" rather than its correct loglevel string value, "INFO".

## What

This PR:

- Adds a utility function, `Runner.normalise_loglevel()` to normalise (correct) loglevel values.
- Adds a call to the newly introduced`Runner.normalise_loglevel()` for the `CELERY_LOG_LEVEL` env var.
- Adds a new env var to docker-specific env file along with docstrings.

## Checklist
- Requires migrations? No

### DIFF
--- a/exporter/sqlite/runner.py
+++ b/exporter/sqlite/runner.py
@@ -25,6 +25,21 @@ class Runner:
         self.database = database
 
     @classmethod
+    def normalise_loglevel(cls, loglevel):
+        """
+        Attempt conversion of `loglevel` from a string integer value (e.g. "20")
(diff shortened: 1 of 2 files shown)


mattjamc (approved):
Looks good! Interesting solution :)

PR #1239, opened 2024-05-30
TP2000-1334 Geo area exclusions front-end validation

eadpearce:
# TP2000-1334 Geo area exclusions front-end validation

## Why

* The list of geo area exclusions was not updated to match the members of a group when a group was selected
* This ended up causing errors right at the end of the measure creation journey, wasting a lot of the user's time
* To avoid this we can do filtering on the form fields on the front-end with react components

## What

* Replaces the current form field with a react version that filters the exclusions by the selected group in real time
* Form still functions when JS is disabled

## Checklist
- Requires migrations? ❌ 
- Requires dependency updates? ✅ in package.json

Screenshot 2024-05-30 at 14 19 19
Screenshot 2024-05-30 at 14 19 27


common/static/common/js/components/GeoAreaField/tests/index.test.js
```diff
@@ -0,0 +1,195 @@
...
+    expect(
+      screen.getByTestId("erga_omnes_exclusions_select"),
+    ).toBeInTheDocument();
+    expect(screen.queryByTestId("group_select")).not.toBeInTheDocument();
+    expect(
+      screen.queryByTestId("country_region_select"),
+    ).not.toBeInTheDocument();
+  });
+
+  // it("should update the exclusions options list when a group is selected", () => {
```
paulpepper-trade, line 163: Maybe skip instead of commenting out?
https://jestjs.io/docs/api#describeskipname-fn

measures/forms.py
```diff
@@ -1318,14 +1382,51 @@ def init_layout(self):
...
+        field_name = f"{self.prefix}-{key}"
+        exclusion_pks = None
+        for k, value in self.data.lists():
+            if k == field_name:
+                exclusion_pks = value
+        # we assume the react filtering has prevented the user selecting an exclusion that is not valid for the group
+        validated_exclusion_pks = [
+            pk
+            for pk in exclusion_pks
+            if GeographicalArea.objects.filter(pk=pk).exists()
```
dalecannon, line 1427: Should `pk` be checked it has a value before using it to filter the queryset in case no exclusions were submitted in the form to prevent a ValueError from being raised?

common/static/common/js/components/GeoAreaField/CountryRegionForm.js
```diff
@@ -0,0 +1,46 @@
...
+        data-testid="country_region_select"
+      >
+        <div className="govuk-form-group">
+          <Select
+            className="react-select-container"
+            classNamePrefix="react-select"
+            options={countryRegionsOptions}
+            defaultValue={countryRegionsInitial}
+            isMulti={true}
+            max={2}
```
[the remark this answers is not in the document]
eadpearce: Right. That's super annoying. I'll have to add my own validation 💀

common/static/common/js/components/GeoAreaField/CountryRegionForm.js
```diff
@@ -0,0 +1,46 @@
...
+            name={`${fieldsPrefix}-countries`}
```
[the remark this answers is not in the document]
eadpearce: YEP. I just tested this now and I forgot to add this

common/static/common/js/components/SubmitButton/index.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+function SubmitButton({ buttonText }) {
+  return (
+    <button className="govuk-button" id="submit-id-submit">
```
dalecannon, line 6: Could `data-prevent-double-click` be added as an attribute here?

geo_areas/constants.py
```diff
@@ -38,6 +38,12 @@ class GeoAreaType(models.TextChoices):
     GeoAreaType.COUNTRY: None,
 }
 
+EXCLUSIONS_REACT_PREFIX_MAPPING = {
+    GeoAreaType.ERGA_OMNES: ERGA_OMNES_EXCLUSIONS_PREFIX,
+    GeoAreaType.GROUP: GROUP_EXCLUSIONS_PREFIX,
+    GeoAreaType.COUNTRY: COUNTRY_REGION_PREFIX,
```
dalecannon, line 44: Would `GeoAreaType.COUNTRY` having a value of `None` better reflect the fact it has no associated exclusions field?

common/static/common/js/components/GeoAreaForm/index.js
```diff
@@ -0,0 +1,171 @@
...
+        />
+        {data.ergaOmnesExclusions.map((exclusion, i) => (
+          <input
+            type="hidden"
+            name={`erga_omnes_exclusions_formset-${i}-erga_omnes_exclusion`}
+            value={exclusion.value}
+            key={exclusion.value}
+          />
+        ))}
+        {data.geoGroupExclusions.map((exclusion, i) => (
```
dalecannon, line 83: Does there also need to be a hidden field for the chosen geographical group (`data.geographicalAreaGroup`) for it to be submitted with the rest of the form data?

measures/forms.py
```diff
@@ -1357,29 +1378,49 @@ def clean(self):
...
-                if geo_area_exclusions:
-                    exclusions = [
-                        exclusion[constants.FIELD_NAME_MAPPING[geo_area_choice]]
-                        for exclusion in geo_area_exclusions
-                    ]
-                    cleaned_data["geo_areas_and_exclusions"][0][
-                        "exclusions"
-                    ] = exclusions
+                # format exclusions for all options
+                if self.data.get("react") == "true":
```
dalecannon, line 1382: Is this code block still required now that the new method of using a hidden form fields to submit the React form data is being used?

geo_areas/constants.py
eadpearce: I can actually remove this as it's not used

common/static/common/js/components/GeoAreaForm/index.js
```diff
@@ -0,0 +1,176 @@
...
+  );
+}
+
+function init() {
+  const csrfToken = document.querySelector(
+    "input[name=csrfmiddlewaretoken]",
+  ).value;
+  const helpText = document.querySelector(
+    "#id_geographical_area-geo_area_hint",
+  ).innerHTML;
```
dalecannon, line 116: I wonder if we need to catch if this selector exists first to prevent any error impacting other JS component initialisation. Perhaps using the optional chaining operator would suffice? 
```javascript
const helpText = document.querySelector(
  "#id_geographical_area-geo_area_hint",
)?.innerHTML;
```

dalecannon (approved):

common/static/common/js/components/GeoAreaForm/index.js
eadpearce: Cool this still runs but the help text is blank. We just get a warning in the console that helpText is undefined since it's a required prop

PR #1240, opened 2024-05-31
TP2000-1384 Allow end dates to be removed when bulk editing measures

dalecannon:
# TP2000-1384  Allow end dates to be removed when bulk editing measures

## Why
The measure end date form in the bulk measure editing journey requires an end date otherwise it raises a validation error, preventing end dates from being removed.

## What
- Refactors both `MeasureStartDateForm.clean()` and `MeasureEndDateForm.clean()` as the forms already normalise their respective field data to a datetime object

- Adds `required=False` to the `end_date` field on `MeasureEndDateForm` to allow empty values to be submitted

- Makes `MeasureEndDateForm` set its end date cleaned data to `None` if no end date was inputted

- Changes how `MeasureEditWizard.done()` gets the new start/end date for the edited measures
    - If end date key is in cleaned data, then use its value to set as the new end date (even if the value is `None` as that means the end date is intended to be removed)
    - If end date key is not in cleaned data, then the end date has not been updated and should default to the measure's current end date

- Updates unit tests

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1697,6 +1697,7 @@ class MeasureEndDateForm(forms.Form):
     end_date = DateInputFieldFixed(
         label="End date",
         help_text="For example, 27 3 2008",
+        required=False,
     )
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


mattjamc (commented):
This looks good :) I've been playing around and just wanted to check - when I remove end dates on the measures it doesn't seem to show the new empty end date on the _Review changes_ page (red) whereas if I'd updated the end date the new one would be showing there (yellow)
image

dalecannon:
> This looks good :) I've been playing around and just wanted to check - when I remove end dates on the measures it doesn't seem to show the new empty end date on the _Review changes_ page (red) whereas if I'd updated the end date the new one would be showing there (yellow)

I'd have to check but it could be that since those measures don't give an explicit end date (`valid_between.upper == None`),  the measures have been end-dated based on the generating regulation (`effective_end_date`).

mattjamc (approved):

PR #1243, opened 2024-06-06
TP2000-1389 Restrict sqlite dump to published only

paulpepper-trade:
# TP2000-1389 Restrict sqlite dump to published only

## Why

Approved transactions are used to determine whether the sqlite dump process should be run and also to create the data for those dumps. This is incorrect because approved transactions include queued, unpublished transactions. Only transactions that are approved **and** associated with a workbasket in PUBLISHED status should be used.

There is also poor visibility of the most recent sqlite dump files in S3. This makes it unnecessarily difficult to identify when problems occur with sqlite dump creation.

## What

This PR:

- Adds a custom queryset filter `TrackedModelQuerySet.published()` to retrieve `TrackedModel` instances associated with approved Transactions and workbaskets that are have PUBLISHED status.
- Adds unit tests for the new `TrackedModelQuerySet.published()` filter.
- Extends unit test for `TransactionQueryset.published()` filter.
- Adds a list of the most recent (last 30 days) sqlite dump files associated with the service instance on the app-info page. Visibility of the list is restricted to those with superuser status - the same as for deployment information that is also displayed on the app-info page.
- Factors out functionality to normalise value representations of `Transaction.order`, which is used when creating a database dump name and also when testing.
- Updates unit tests that create and upload database dumps using published rather than approved transactions.
- Updates unit test on /app-info view to include mocked sqlite dump data and test that the sqlite dumps section is present.

image

## Checklist
- Requires migrations? No

### DIFF
--- a/common/jinja2/common/app_info.jinja
+++ b/common/jinja2/common/app_info.jinja
@@ -61,6 +61,12 @@
           {"text": LAST_TRANSACTION_TIME},
           {"text": "Last transaction change time"},
         ],
+        [
+          {"text": "LAST_PUBLISHED_TRANSACTION_ORDER"},
+          {"text": LAST_PUBLISHED_TRANSACTION_ORDER},
+          {"text": "Last published transaction order value"},
+        ],
+
       ] %}
 
       {{ govukTable({
--- a/common/jinja2/common/app_info.jinja
+++ b/common/jinja2/common/app_info.jinja
@@ -162,7 +162,7 @@
               "colspan": 4,
             }
           ]) or "" }}
-      {% endif %}
+      {% endif -%}
 
       {{ govukTable({
           "head": table_head,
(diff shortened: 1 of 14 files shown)


common/jinja2/common/app_info.jinja
```diff
@@ -156,13 +162,54 @@
...
+          "head": table_head,
+          "rows": table_rows
+      }) }}
+    </div>
+  </div>
+
+
+  {%- if request.user.is_superuser and SQLITE_DUMP_LIST %}
+  <div class="govuk-grid-row govuk-!-margin-bottom-6 info-section">
+    <div class="govuk-grid-column-full">
```
eadpearce, line 177: If the grid only contains one full width column using the grid/column classes feel superfluous. You could remove this div and the govuk-grid-row class on the parent div to simplify it (unless there's some styling/layout need I'm missing here)

common/static/common/scss/_text.scss
```diff
@@ -19,3 +19,7 @@
 .right-arrow {
   font-size: 1rem;
 }
+
+.fixed-width {
```
eadpearce, line 23: Where is this class used?

common/jinja2/common/app_info.jinja
paulpepper-trade: All the other sections on the page are structured this way, so it would be odd to include one that breaks the convention. But I have a vague memory of left and right margin behaviour changes if the row and column classes aren't used, causing alignment problems - something to do with the horrible float system that the GDS kit uses for its responsive column and row layout.

We could strip out the whole lot on this page if you feel strongly enough about it?

common/static/common/scss/_text.scss
paulpepper-trade: In an old, superceded commit (⸝⸝⸝O﹏ O⸝⸝⸝). Now gone!

common/jinja2/common/app_info.jinja
eadpearce: Just tested it and it behaves exactly the same like this:
```
<div class="govuk-!-margin-bottom-6 info-section">
  [content]
</div>
```
So we can strip it out or not. Don't feel super strongly on it being done for this PR. Would be nice to trim down the amount of html we write

paulpepper-trade: Perhaps for another PR then, to optimise our page content across the whole application. We could collect some figures on page size and download time as part of that PR.

paulpepper-trade: https://uktrade.atlassian.net/browse/TP2000-1405

eadpearce (approved):

PR #1244, opened 2024-06-07
TP2000-1403  Fix missing quota origin associations

dalecannon:
# TP2000-1403  Fix missing quota origin associations

## Why
An origin that should be associated to quota order number 058801 doesn't appear on the quota's detail view nor does it appear as an existing origin when editing the quota.

## What
- Changes the queryset to find current/existing origins using the order number SID not PK as a way to catch associations that may have been lost when a quota's version has changed

### DIFF
--- a/quotas/models.py
+++ b/quotas/models.py
@@ -112,6 +112,11 @@ class QuotaOrderNumber(TrackedModel, ValidityMixin):
 
         return sorted(descriptions)
 
+    def get_current_origins(self):
+        return QuotaOrderNumberOrigin.objects.filter(
+            order_number__sid=self.sid,
+        ).current()
+
     class Meta:
         verbose_name = "quota"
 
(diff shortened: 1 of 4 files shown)


CPrich905 (approved):
Code LGTM and does what it says on the tin

quotas/models.py
```diff
@@ -112,6 +112,11 @@ def geographical_exclusion_descriptions(self):
 
         return sorted(descriptions)
 
+    def get_current_origins(self):
+        return QuotaOrderNumberOrigin.objects.filter(
+            order_number__sid=self.sid,
```
paulpepper-trade, line 117: For info really, but I think this aught to be `order_number__version_group=self.version_group`. They're functionally equivalent here, but the intent of `VersionGroup` is to provide a stable version grouping as I understand it - recycled SIDs was offered as an example justification.

PR #1245, opened 2024-06-12
TP2000-1378  Add option to new duty sentence parser to limit parsing to simple duties

dalecannon:
# TP2000-1378  Add option to new duty sentence parser to limit parsing to simple duties

## Why
The measure conditions form doesn't currently take full advantage of the new duty sentence parser because the 
the reference price field doesn't permit compound duties and the parser cannot be limited to simple duties.

## What
- Adds `compound_duties: bool` attribute/param to duty sentence parser, which determines the start symbol of the grammar used by the parser instance so that parsing can be limited to ad valorem and specific duties only (if required)
- Hooks up new duty sentence parser in `MeasureConditionsFormMixin.conditions_clean()` to parse and validate duties inputted in the reference price field of the measure conditions form
- Fixes the display of condition duties both on the measure review view and on the measure detail view conditions tab to measurement units
- Updates unit tests  

---

Before | After
------  | ------
Screenshot 2024-06-12 at 15 45 55|Screenshot 2024-06-12 at 16 50 26
&nbsp; | Screenshot 2024-06-12 at 15 39 54
Missing measurement unit (and measurement unit qualifier) | Screenshot 2024-06-12 at 15 48 17
Screenshot 2024-06-12 at 16 09 55 | Screenshot 2024-06-12 at 15 54 35

### DIFF
--- a/measures/duty_sentence_parser.py
+++ b/measures/duty_sentence_parser.py
@@ -94,7 +94,7 @@ class DutySentenceParser:
 
     def __init__(
         self,
-        full_grammar: bool = True,
+        compound_duties: bool = True,
         date: datetime = datetime.now(),
         duty_expressions: Sequence[models.DutyExpression] = None,
         monetary_units: Sequence[models.MonetaryUnit] = None,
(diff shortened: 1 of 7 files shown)


measures/duty_sentence_parser.py
```diff
@@ -89,6 +94,7 @@ def create_rule(items, field_name, strip=False):
 
     def __init__(
         self,
+        full_grammar: bool = True,
```
paulpepper-trade, line 97: Would `compound_duties` (or something similarly specific) more clearly express the intent and self-document this param / attr, do you think?

paulpepper-trade (approved):

measures/duty_sentence_parser.py
dalecannon: Renamed!

PR #1249, opened 2024-06-17
TP2000-1262 upgrade psycopg package

paulpepper-trade:
# TP2000-1262 upgrade psycopg package

## Why

A new major version (3) of the psycopg Python package was released some years ago. The previous version (2), currently used by Tamato, is now relatively inactive and should be replaced by the current, active version, allowing Tamato to take advantage of optimisations and fixes.

## What

In this PR:

- Bump psycopg version numbers in config, docs and settings.
- Swap to using equivalent date types introduced by psycopg 3.
- Correctly handle new `value` param behaviour in `Model.from_db_value()` as a result of new psycopg behaviour.
- Update test to use pscopg 3's changed exception message.

## Checklist
- Requires migrations? No

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -71,22 +71,6 @@ Installing
     $ npm install
     $ npm run build
 
-Those using Mac m1 laptops may have problems installing certain packages (e.g.
-psycopg2 and lxml) via requirements-dev.txt. In this scenario you should run the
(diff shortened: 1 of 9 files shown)


a-gleeson (approved):
LTGM!

PR #1253, opened 2024-06-24
TP2000-1268  Resolve XML external entity expansion alert

dalecannon:
# TP2000-1268  Resolve XML external entity expansion alert

## Why
External entity expansion hasn't been disabled when calling `etree.parse()` within `parse_xml()`.

## What
- Bumps `lxml` to 5.2.2 (latest version), which no longer expands external entities by default as of version 5.0.0
- Explicitly disables entity expansion in `parse_xml()` and `xml_fromstring()`
- Uses `werkzeug.utils.secure_filename` to validate the file name of uploaded taric files to guard against path traversal
- Updates unit tests

## Screenshot
Screenshot 2024-06-28 at 14 55 00

## Checklist
- Requires dependency updates? **Yes**

### DIFF
--- a/common/tests/test_util.py
+++ b/common/tests/test_util.py
@@ -306,7 +306,7 @@ def test_parse_xml_vulnerabilities(file_name):
         "test_files",
         file_name,
     )
-    with pytest.raises(XMLSyntaxError):
+    with pytest.raises((XMLSyntaxError, DTDForbidden)):
         util.parse_xml(file)
 
 
@@ -318,5 +318,5 @@ def test_xml_fromstring_vulnerabilities(file_name):
         file_name,
     )
     string = open(file).read()
-    with pytest.raises(XMLSyntaxError):
+    with pytest.raises((XMLSyntaxError, DTDForbidden)):
         util.xml_fromstring(string)
--- a/common/util.py
+++ b/common/util.py
@@ -554,14 +554,16 @@ def check_docinfo(elementtree, forbid_dtd=False):
 
 
 def parse_xml(source, forbid_dtd=True):
-    elementtree = etree.parse(source)
+    parser = etree.XMLParser(resolve_entities=False)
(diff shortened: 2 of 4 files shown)


common/validators.py
```diff
@@ -127,3 +129,18 @@ def get_help_text(self):
...
+    characters such as spaces, underscores and round brackets and that it has
+    the expected `extension` if provided."""
+
+    name_message = "File name must only include alphanumeric characters and special characters such as spaces, hyphens, underscores and round brackets. "
+    extension_message = f"The file must also have a {extension} file name extension."
+    error_message = (
+        "".join([name_message, extension_message]) if extension else name_message
+    )
+
+    if not re.match(rf"^[A-Za-z0-9_() -]+{extension}$", name):
```
paulpepper-trade, line 145: When `extension` is provided, does its contents also need validating to cater for this rule:

> Do not allow more than a single "." character.

common/validators.py
```diff
@@ -127,3 +129,18 @@ def get_help_text(self):
     "tbody",
     "td",
 ]
+
+
+def validate_filename(name: str, extension: str = ""):
```
paulpepper-trade, line 134: Could `werkzeug.utils.secure_filename()` be used in place of this `validate_filename()`? It would take a slightly different approach, but may (?) better satisfy code scanners and give us less code to manage and test ourselves.

common/validators.py
dalecannon: Ah, yes I believe it should. That way it guards against errors when providing the expected extension in the first place.

common/validators.py
dalecannon: That would be preferable, yeah. I did test it but it silently removes from file names round brackets (which have been included to re-upload TARIC files in the past).

paulpepper-trade (approved):
LGTM

PR #1254, opened 2024-06-25
TP2000 1407 Sqlite dump process failing

Luisella21:
TP2000 1407 daily sqlite dump process failing

Add logging to pin point the problem.

TP2000-1407-daily-sqlite-dump-process-failing

### DIFF
--- a/exporter/sqlite/tasks.py
+++ b/exporter/sqlite/tasks.py
@@ -40,10 +40,16 @@ def export_and_upload_sqlite() -> bool:
     Returns a boolean that is ``True`` if a file was uploaded and ``False`` if
     not.
     """
+    from django.conf import settings
+
     storage = SQLiteStorage()
     db_name = get_output_filename()
 
     export_filename = storage.generate_filename(db_name)
+    logger.debug(
+        f" In export_and_upload_sqlite "
+        f"SQLITE_S3_ENDPOINT_URL {settings.SQLITE_S3_ENDPOINT_URL}",
+    )
 
     logger.debug("Checking for need to upload tariff database %s", export_filename)
     if storage.exists(export_filename):
--- a/exporter/storages.py
+++ b/exporter/storages.py
@@ -1,9 +1,12 @@
+import logging
 from functools import cached_property
 from os import path
 
 import apsw
 from sqlite_s3vfs import S3VFS
 from storages.backends.s3boto3 import S3Boto3Storage
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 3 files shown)


exporter/storages.py
```diff
@@ -47,8 +50,19 @@ def generate_filename(self, filename: str) -> str:
         return super().generate_filename(filename)
 
     def exists(self, filename: str) -> bool:
+        from django.conf import settings
+
+        logger.debug(f"Check that filename {filename} exists")
+        logger.debug(
+            f" ****** SQLITE_S3_ENDPOINT_URL {settings.SQLITE_S3_ENDPOINT_URL}",
```
paulpepper-trade, line 57: For debugging this particular problem, we could look at the env vars on the Celery instance to validate correct config - otherwise I think this adds noise to the logs.

exporter/storages.py
```diff
@@ -47,8 +50,19 @@ def generate_filename(self, filename: str) -> str:
...
+        )
         return any(self.listdir(filename))
 
+        # except Exception as e:
+        #     logger.error(f"Error encountered while checking {filename}")
+        #     logger.error(f"Exception: {e}")
+        #     return False
+
```
paulpepper-trade, line 65: Hangover comment requiring deletion.

exporter/sqlite/tasks.py
```diff
@@ -40,10 +40,16 @@ def export_and_upload_sqlite() -> bool:
...
+    from django.conf import settings
+
     storage = SQLiteStorage()
     db_name = get_output_filename()
 
     export_filename = storage.generate_filename(db_name)
+    logger.debug(
+        f" In export_and_upload_sqlite "
+        f"SQLITE_S3_ENDPOINT_URL {settings.SQLITE_S3_ENDPOINT_URL}",
+    )
```
paulpepper-trade, line 52: For debugging this particular problem, we should look at the env vars on the Celery instance to validate correct config - otherwise I think this adds noise to the logs.

settings/common.py
```diff
@@ -713,7 +713,7 @@
         },
         "exporter": {
             "handlers": ["console"],
-            "level": os.environ.get("LOG_LEVEL", "DEBUG"),
+            "level": os.environ.get("LOG_LEVEL_EXPORTER", "DEBUG"),
```
paulpepper-trade, line 716: Checking our secrets store and env vars, I don't think we actually need to adapt the log level to get the debug info that we'd like to gather.

PR #1257, opened 2024-07-02
TP2000-1269 Resolve uncontrolled data used in path expression alert

dalecannon:
# TP2000-1269 Resolve uncontrolled data used in path expression alert

## Why
The filenames of TARIC files uploaded in the importer form could use some validation and sanitisation.

## What
- Adds two new functions, `validate_filename` and `validate_filepath`, that can be used to guard against malicious filenames and path traversal attacks
- Validates the filename of uploaded TARIC files
- Sanitises the name used to create an `ImportBatch` based on the uploaded TARIC file

### DIFF
--- a/importer/forms.py
+++ b/importer/forms.py
@@ -15,6 +15,7 @@ from django.core.files.base import ContentFile
 from django.core.files.uploadedfile import InMemoryUploadedFile
 from django.db import transaction
 from sentry_sdk import capture_exception
+from werkzeug.utils import secure_filename
 
 from common.util import get_mime_type
 from common.util import parse_xml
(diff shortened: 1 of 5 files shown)


importer/forms.py
```diff
@@ -154,13 +151,8 @@ def clean_taric_file(self):
...
-            uploaded_taric_file.name == secure_file_name
-        else:
-            raise ValidationError(
-                "File name must only include alphanumeric characters and special characters such as spaces, hyphens and underscores.",
-            )
+        validate_filename(uploaded_taric_file.name)
+        validate_filepath(uploaded_taric_file)
 
         try:
             xml_file = parse_xml(uploaded_taric_file)
```
dalecannon, line 158: Despite the current validation (and after various other validation attempts), the CodeQL alert warning that `uploaded_taric_file` is an uncontrolled user input still remains. It may be a false positive, where the scanner overlooks how we're having to access the `name` attribute on the `InMemoryUploadedFile` object to validate its filename, and not a `str` or other such object that can be directly acted upon to satisfy the scanner's expectations.

paulpepper-trade: If the code scan alert is a false positive, then I think it should be okay to mark it as such, perhaps referencing this comment in the accompanying message.

paulpepper-trade (approved):

PR #1259, opened 2024-07-05
TP2000-1270  Resolve URL redirection from remote source alerts

dalecannon:
# TP2000-1270  Resolve URL redirection from remote source alerts

## Why
Some redirect URLs may incorporate unvalidated user input.

## What
- Uses string concatenation instead of f-string formatting to form the redirect URL in `MeasureList.form_valid()` to compensate for false positive alert (see comment in [UrlRedirectCustomizations.qll](https://github.com/github/codeql/blob/main/python/ql/lib/semmle/python/security/dataflow/UrlRedirectCustomizations.qll#L103))

- Replaces `request.build_absolute_uri()` with functionally equivalent calls to `reverse(viewname)` in both `EnvelopeQueueView` and `PackagedWorkbasketQueueView`

### DIFF
--- a/measures/views.py
+++ b/measures/views.py
@@ -388,7 +388,7 @@ class MeasureList(
             self.session_store.add_items(selected_objects)
 
             params = urlencode(self.request.GET)
-            url = f"{reverse('measure-ui-list')}?{params}"
+            url = reverse("measure-ui-list") + "?" + params
         else:
             url = reverse("measure-ui-list")
 
--- a/publishing/views.py
+++ b/publishing/views.py
@@ -174,20 +174,19 @@ class EnvelopeQueueView(
         return data
 
     def post(self, request, *args, **kwargs):
-        """Manage POST requests, including download, accept and reject
-        envelopes."""
+        """
+        Manage POST requests that signal the start of envelope processing.
 
-        post = request.POST
+        Valid and invalid POST requests alike redisplay the envelope queue view.
+        """
 
-        if post.get("process_envelope"):
-            url = self._process_envelope(request, post.get("process_envelope"))
-        else:
-            # Handle invalid post content by redisplaying the page.
-            url = request.build_absolute_uri()
+        envelope_pk = request.POST.get("process_envelope")
+        if envelope_pk:
+            self._process_envelope(envelope_pk)
 
-        return redirect(url)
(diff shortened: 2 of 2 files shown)


measures/views.py
```diff
@@ -388,7 +388,7 @@ def form_valid(self, form):
             self.session_store.add_items(selected_objects)
 
             params = urlencode(self.request.GET)
-            url = f"{reverse('measure-ui-list')}?{params}"
+            url = reverse("measure-ui-list") + "?" + params
```
paulpepper-trade, line 391: Curious. Did this make a difference to code scanning results, or?

publishing/views.py
```diff
@@ -41,6 +41,7 @@ class PackagedWorkbasketQueueView(
 
     model = PackagedWorkBasket
     permission_required = "publishing.manage_packaging_queue"
+    redirect_url = reverse_lazy("publishing:packaged-workbasket-queue-ui-list")
```
paulpepper-trade, line 44: Would something like `view_url` help make it clear that this doesn't change and actually is this view's URL (which happens to be used for redirects)?

publishing/views.py
```diff
@@ -174,28 +172,26 @@ def get_context_data(self, *, object_list=None, **kwargs):
...
-    def _process_envelope(self, request, pk):
+    def _process_envelope(self, pk):
         if not OperationalStatus.is_queue_paused():
             packaged_work_basket = PackagedWorkBasket.objects.get(pk=pk)
             try:
                 packaged_work_basket.begin_processing()
             except TransitionNotAllowed:
                 # No error page right now, just reshow the list view.
                 pass
-        return request.build_absolute_uri()
```
paulpepper-trade, line 198: Has Django's `HttpRequest.build_absolute_url()` been the cause of scanner issues!?

paulpepper-trade (approved):

publishing/views.py
dalecannon: I suspect it's because the scanner's query isn't taking it into account when determining the sanitisation state of the `url`, causing it to warn about it as an unvalidated value still.

measures/views.py
paulpepper-trade: Oh my goodness!

publishing/views.py
paulpepper-trade: (－‸ლ)

PR #1263, opened 2024-07-12
Fix measure bulk create error

eadpearce:
# Fix measure bulk create error

## Why

* Bug was missed on last code merge due to missing test coverage

## What

* Changes MeasureGeographicalAreaForm.deserialize_init_kwargs back to a class method
* Adds missing test coverage for MeasuresBulkCreator.get_forms_cleaned_data

### DIFF
--- a/measures/forms.py
+++ b/measures/forms.py
@@ -1379,10 +1379,11 @@ class MeasureGeographicalAreaForm(
         # Perculiarly, serializable data in this form keeps its prefix.
         return super().serializable_data()
 
-    def deserialize_init_kwargs(self, form_kwargs: Dict) -> Dict:
+    @classmethod
+    def deserialize_init_kwargs(cls, form_kwargs: Dict) -> Dict:
         # Perculiarly, this Form requires a prefix of "geographical_area".
         return {
-            "prefix": self.prefix,
+            "prefix": cls.prefix,
         }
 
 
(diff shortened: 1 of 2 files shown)


measures/forms.py
```diff
@@ -1379,10 +1379,11 @@ def serializable_data(self, remove_key_prefix: str = "") -> Dict:
         # Perculiarly, serializable data in this form keeps its prefix.
         return super().serializable_data()
 
-    def deserialize_init_kwargs(self, form_kwargs: Dict) -> Dict:
+    @classmethod
+    def deserialize_init_kwargs(cls, form_kwargs: Dict) -> Dict:
         # Perculiarly, this Form requires a prefix of "geographical_area".
         return {
-            "prefix": self.prefix,
+            "prefix": cls.prefix,
```
dalecannon, line 1386: Does the prefix need to be specified as `"geographical_area"` like before or added as new class attribute on the form before it can be used?

eadpearce: Since I (clearly wrongly!!) assumed that tests would fail if it WAS empty I just assumed it was fine. It's actually empty! I have another regression test to write! 🤦‍♀️

eadpearce: Discussed with Steve. Writing tests for this code is going to take a bit longer (currently working on it) so he's given the go-ahead to deploy this fix without tests for now

dalecannon (approved):

CPrich905 (approved):

PR #1264, opened 2024-07-16
AWS Paas Migration Changes

PaulWheatcroft:
# Changes to enable AWS migration

## Why
To facilitate the application(s) in the repo to be deployed on to the new AWS platform

### DIFF
--- a/.copilot/phases/build.sh
+++ b/.copilot/phases/build.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -ex
+
+#python -m pip install --upgrade pip
+#pip install -r requirements.txt
+#
+#npm install
+#npm run build
--- a/.copilot/phases/build.sh
+++ b/.copilot/phases/build.sh
@@ -1,9 +1,3 @@
 #!/usr/bin/env bash
 
-set -ex
-
-#python -m pip install --upgrade pip
-#pip install -r requirements.txt
-#
-#npm install
-#npm run build
+set -e
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,5 @@
 web: scripts/web-worker-entrypoint.sh
+app: scripts/web-worker-entrypoint.sh
 worker: celery -A common.celery worker -O fair -l info -Q standard
 beat: celery -A common.celery beat -l info
 rule-check-worker: celery -A common.celery worker -O fair -l info -Q rule-check --concurrency 1
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,4 @@
 web: scripts/web-worker-entrypoint.sh
-app: scripts/web-worker-entrypoint.sh
 worker: celery -A common.celery worker -O fair -l info -Q standard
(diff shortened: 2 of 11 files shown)


.copilot/phases/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -ex
+
+#python -m pip install --upgrade pip
+#pip install -r requirements.txt
+#
+#npm install
+#npm run build
```
paulpepper-trade, line 9: Old, defunct comments?

Procfile
```diff
@@ -1,4 +1,5 @@
 web: scripts/web-worker-entrypoint.sh
+app: scripts/web-worker-entrypoint.sh
```
paulpepper-trade, line 2: Will a change to `app` be okay on Cloud Foundry deployments? Looking at the docs (https://docs.cloudfoundry.org/buildpacks/prod-server.html#procfiles):

> PROCESS-TYPE: COMMAND
> 
> Where:
> 
>     PROCESS-TYPE is web. A web process handles HTTP traffic.

I'm unlear if use of `web` is expected as a convention, or if that's just an example name. Do you know which?

package.json
```diff
@@ -44,6 +44,7 @@
     "clean": "rm -f ./run/static/webpack_bundles/*",
     "heroku-prebuild": "",
     "heroku-postbuild": "npm run build",
+    "start": "echo starting",
```
paulpepper-trade, line 47: Is `start` a placeholder for a future script? (We've tended not to introduce placeholders into production code.)

Procfile
PaulWheatcroft: This is a work-around for paketo to install the node buildpack. We will still continue to rely on the web entry point.

package.json
PaulWheatcroft: Again this is a work-around for node.js buildpack

Procfile
paulpepper-trade: Just discussed with @emileswarts and we believe there's some uncertainty about whether the `app` entry will start additional, unwanted web worker processes, possibly leading to problems (e.g. port conflicts with the processes started by `web`). Emile said he'll discuss with you @PaulWheatcroft and probably respond here.

package.json
paulpepper-trade: Agreed with @emileswarts to replace `"echo starting"` with something that documents the dummy `"start"` script entry - e.g. `"start": "dummy entry for node.js buildpack"`.

Procfile
emileswarts: Confirmed that this will spin up a new process on GOVPaas, was able to isolate this workaround to the package.json file

package.json
emileswarts: Pointing this at `bash web-worker-entrypoint.sh` will be sufficient for this workaround, but means that we can't add the comment to indicate that it's a workaround.

paulpepper-trade (approved):
LGTM

PR #1265, opened 2024-07-16
TP2000-1407 Sqlite dump failure fix

paulpepper-trade:
# TP2000-1407 Sqlite dump failure fix

## Why

Until recently, the SQLite export to S3 capability was taking over four hours to complete. With a continued increase in database size, export performance had sharply degraded and was taking in the region of three days to complete. Performance improvements were required.

## What

This PR:

- Reduces the time to create a SQLite export and upload to S3 down to approximately six minutes.
- Refactors methods on storage classes that are used to save SQLite export files to allow multiple strategies to be applied for different export situations (e.g. exporting to local file system from a command line, or exporting to S3 via a Celery Beat initiated task).
- Add new storage classes `SQLiteLocalStorage` and `SQLiteS3Storage` to apply the different storage strategies.
- Adds a decorator that is used to log the start time, end time and total elapsed time taken for a function to execute. This was useful for comparing the performance of different storage strategies and discovering the issue that this PR fixes, but has been placed under the `common` application for wider use.
- Added command-line support for generating exports to the local file system.
- Refactored some (but not all) of the export code in order to simplify the process - more could be done, but the prospect of a more radical replacement means a fuller refactor (and simplification) is probably not worthwhile.
- Added docstrings to export classes and functions in place of the fuller refactor that could otherwise have been carried out.
- Added config support to enable / disable scheduled (via Celery Beat) SQLite exports.
- Added unit tests for the newly added storage classes and updated existing tests.
- Added unit test for `dump_sqlite` Django management command.
- Applied incidental minor fixes and corrections.

## Checklist
- Requires migrations? No


exporter/storages.py
```diff
@@ -46,17 +68,73 @@ def generate_filename(self, filename: str) -> str:
...
+    """
+    Storage class used for remotely storing SQLite database files to an AWS
+    S3-like backing store.
+
+    This class applies a strategy that first creates a temporary instance of the
+    SQLite file on the local file system before transfering its contents to S3.
+    """
+
+    @log_timing(logger_function=logger.info)
+    def export_database(self, filename: str):
```
eadpearce, line 110: Is it possible to test this function?

exporter/management/commands/dump_sqlite.py
```diff
@@ -11,20 +11,39 @@
 
 
 class Command(BaseCommand):
+    help = (
+        "Create a snapshot of the application database to a file in SQLite "
```
eadpearce, line 15: Is this command tested? This whole file is not showing up in the coverage report. Not sure if that's a bug with codecov or not

exporter/storages.py
paulpepper-trade: It's implicitly tested as part of [test_s3_export_task_uploads() ](https://github.com/uktrade/tamato/blob/d1046d7fe97bcd5093e7236dafcf4d575bf4ad75/exporter/tests/test_sqlite.py#L179-L198) - called [within the export_and_upload_sqlite()](https://github.com/uktrade/tamato/blob/d1046d7fe97bcd5093e7236dafcf4d575bf4ad75/exporter/sqlite/tasks.py#L67) function.

exporter/management/commands/dump_sqlite.py
paulpepper-trade: It looks as though there's never been a test for this command. I can only guess that's because it's fairly simple and the function it executes is tested by [/exporter/tests/test_sqlite.py](https://github.com/uktrade/tamato/blob/TP2000-1407--sqlite-dump-failure/exporter/tests/test_sqlite.py).

[Test now added](https://github.com/uktrade/tamato/blob/6d27f64a68d8f1b8f951c2dc4a97d9953a3bfdb9/exporter/tests/test_exporter_commands.py#L18-L45).

exporter/storages.py
eadpearce: It's showing up for me as red in codecov. Is that right? Can you check it on your end as well and confirm?

eadpearce (approved):

PR #1266, opened 2024-07-17
TP2000-1404: Reorganise measures and common code

eadpearce:
# TP2000-1404: Reorganise measures and common code

## Why

* measures/forms.py and measures/views.py were getting very long and hard to navigate (1000+ lines)

## What

* Splits up measures forms and views and common views code into separate files inside directories
* Adds imports to `__init__` files so we can still use the same import paths - measures.forms, measures.views, common.views


common/views/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .base import *
```
eadpearce, line 1: I considered importing every function explicitly but decided against it since it would just be a burden on devs making changes to those files in future having to remember to explicitly import new functions in here (I know I would definitely forget and then wonder why I'm getting an error)

measures/forms/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .base import *
```
eadpearce, line 1: Same again here. `import *` in this particular context is more future-proof than importing everything explicitly

paulpepper-trade (commented):
Seems sensible!
Is this purely a reorganisation without any changes that could introduce side effects (there're a lot of changes making it difficult to see whether there are any)?

eadpearce:
@paulpepper-trade Nope. No code changes. Just moving things around and adding the init imports

paulpepper-trade (approved):

PR #1267, opened 2024-07-17
TP2000-1434 Prevent packaging queue race conditions

dalecannon:
# TP2000-1434 Prevent packaging queue race conditions

## Why
A race condition can occur during concurrent requests to create/promote/demote/process packaged workbaskets, resulting in duplicate or non-contiguous queue positions that stop the queue from functioning.

## What
- Uses [select_for_update()](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#select-for-update) queryset operation to lock affected packaged workbasket rows while queue position are being updated

    - Makes queue position update queries non-blocking (`nowait=True`), catching the raised exception if a conflicting lock is already acquired

- Returns early without updating queue positions if the selected packaged workbasket is no longer a queue member (position=0)
- Prevents accidental double-clicking of buttons in the packaging queue

## Examples
<details><summary>Shared queue positions after creating packaged workbaskets</summary>
<p>

https://github.com/user-attachments/assets/c495ba82-b076-43b7-aaa3-3aea3744fc85

</p>
</details> 

<details><summary>Shared queue positions after repositioning packaged workbaskets</summary>
<p>

https://github.com/user-attachments/assets/81e4f827-a165-454b-aa2f-791f1bd6f03c

</p>
</details> 

<details><summary>Invalid and inconsistent positions and processing states after processing a packaged workbasket while repositioning others</summary>
<p>

https://github.com/user-attachments/assets/a053895b-d3d7-4425-bfca-8554fc227609

</p>
</details>

<details><summary>Invalid and inconsistent positions and processing states after repositioning a since-removed packaged workbasket</summary>
<p>

https://github.com/user-attachments/assets/dcbd9e11-b2d5-4656-b2de-03bf6bc87c4b

</p>
</details>

### DIFF
--- a/publishing/models/packaged_workbasket.py
+++ b/publishing/models/packaged_workbasket.py
@@ -439,6 +439,7 @@ class PackagedWorkBasket(TimestampedMixin):
 
         return not PackagedWorkBasket.objects.currently_processing()
 
+    @atomic
     @pop_top_after
     @save_after
     @transition(
@@ -468,6 +469,7 @@ class PackagedWorkBasket(TimestampedMixin):
         multiple instances it's necessary for this method to perform a save()
         operation upon successful transitions.
         """
+        PackagedWorkBasket.objects.select_for_update(nowait=True).get(pk=self.pk)
         self.processing_started_at = datetime.now()
         self.save()
 
@@ -642,6 +644,10 @@ class PackagedWorkBasket(TimestampedMixin):
         Management of the queued instance's `processing_state` is not altered by
         this function and should be managed separately by the caller.
         """
+
+        PackagedWorkBasket.objects.select_for_update(nowait=True).get(pk=self.pk)
+        self.refresh_from_db()
+
         if self.position == 0:
             raise PackagedWorkBasketInvalidQueueOperation(
                 "Unable to remove instance with a position value of 0 from "
@@ -667,6 +673,9 @@ class PackagedWorkBasket(TimestampedMixin):
         """Promote the instance to the top position of the package processing
         queue so that it occupies position 1."""
 
+        PackagedWorkBasket.objects.select_for_update(nowait=True).get(pk=self.pk)
+        self.refresh_from_db()
+
         if self.position <= 1:
             return self
 
@@ -687,6 +696,9 @@ class PackagedWorkBasket(TimestampedMixin):
     def promote_position(self) -> "PackagedWorkBasket":
         """Promote the instance by one position up the package processing
         queue."""
+
+        PackagedWorkBasket.objects.select_for_update(nowait=True).get(pk=self.pk)
+        self.refresh_from_db()
 
         if self.position <= 1:
             return self
@@ -709,6 +721,9 @@ class PackagedWorkBasket(TimestampedMixin):
     def demote_position(self) -> "PackagedWorkBasket":
         """Demote the instance by one position down the package processing
         queue."""
+
+        PackagedWorkBasket.objects.select_for_update(nowait=True).get(pk=self.pk)
+        self.refresh_from_db()
 
         if self.position in {0, PackagedWorkBasket.objects.max_position()}:
             return self
--- a/publishing/models/packaged_workbasket.py
+++ b/publishing/models/packaged_workbasket.py
@@ -125,7 +125,7 @@ class PackagedWorkBasketInvalidQueueOperation(Exception):
 
 class PackagedWorkBasketManager(Manager):
     @atomic
-    @TableLock.acquire_lock("publishing.PackagedWorkBasket", lock="EXCLUSIVE")
+    @TableLock.acquire_lock("publishing.PackagedWorkBasket", lock=TableLock.EXCLUSIVE)
     def create(self, workbasket: WorkBasket, **kwargs):
         """Create a new instance, associating with workbasket."""
         if workbasket.status in WorkflowStatus.unchecked_statuses():
--- a/publishing/models/packaged_workbasket.py
+++ b/publishing/models/packaged_workbasket.py
@@ -669,7 +669,7 @@ class PackagedWorkBasket(TimestampedMixin):
         """Promote the instance to the top position of the package processing
         queue so that it occupies position 1."""
 
-        if self.position == 1:
+        if self.position <= 1:
             return self
 
         position = self.position
@@ -680,6 +680,7 @@ class PackagedWorkBasket(TimestampedMixin):
 
         self.position = 1
         self.save()
+        self.refresh_from_db()
 
         return self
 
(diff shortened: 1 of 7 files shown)


publishing/models/packaged_workbasket.py
```diff
@@ -142,15 +142,19 @@ def create(self, workbasket: WorkBasket, **kwargs):
...
+                packaged_workbaskets.aggregate(
+                    out=Coalesce(
+                        Max("position"),
+                        Value(0),
+                    ),
+                )["out"]
+                + 1
+            )
+        else:
+            position = 1
```
mattjamc, line 157: how come this if/else has been added in? If there is nothing in the queue already then wouldn't the position default to 1 anyway

publishing/views.py
```diff
@@ -94,45 +95,60 @@ def _unpause_queue(self, request):
...
         try:
-            packaged_work_basket = PackagedWorkBasket.objects.get(pk=pk)
+            packaged_work_basket = PackagedWorkBasket.objects.select_for_update(
+                nowait=True,
+            ).get(pk=pk)
             packaged_work_basket.promote_position()
         except (
             PackagedWorkBasket.DoesNotExist,
             PackagedWorkBasketInvalidQueueOperation,
+            OperationalError,
```
mattjamc, line 108: how come it is expecting an OperationalError? In the docs it seems like DatabaseError might be raised when nowait=True https://docs.djangoproject.com/en/5.0/ref/models/querysets/#select-for-update

dalecannon: I too was expecting to see `DatabaseError` but `OperationalError` ([which is a subclass](https://docs.djangoproject.com/en/4.2/_modules/django/db/utils/#OperationalError)) was the exception in particular that was being raised during testing when a conflicting lock was already acquired.

publishing/models/packaged_workbasket.py
dalecannon: I found that `select_for_update()` doesn't appear to work in conjunction with `aggregate()` unless the queryset has been evaluated first, otherwise `SELECT FOR UPDATE` doesn't get added to the resulting SQL statement.

publishing/models/packaged_workbasket.py
```diff
@@ -142,15 +142,19 @@ def create(self, workbasket: WorkBasket, **kwargs):
...
-        position = (
-            PackagedWorkBasket.objects.aggregate(
-                out=Coalesce(
-                    Max("position"),
-                    Value(0),
-                ),
-            )["out"]
-            + 1
-        )
+        packaged_workbaskets = PackagedWorkBasket.objects.select_for_update().all()
```
paulpepper-trade, line 145: Does `all()` lock the table, preventing, say, conflicting concurrent attempts at creating new `PackagedWorkBasket` instances? There's also [common.util.TableLock](https://github.com/uktrade/tamato/blob/562452836ae947cb4e181a9a5e767d5d67a52e90/common/util.py#L401).

publishing/models/packaged_workbasket.py
```diff
@@ -701,10 +712,12 @@ def demote_position(self) -> "PackagedWorkBasket":
         """Demote the instance by one position down the package processing
         queue."""
 
-        if self.position == PackagedWorkBasket.objects.max_position():
-            return
+        if self.position in {0, PackagedWorkBasket.objects.max_position()}:
+            return self
 
-        obj_to_swap = PackagedWorkBasket.objects.get(position=self.position + 1)
+        obj_to_swap = PackagedWorkBasket.objects.select_for_update(nowait=True).get(
```
paulpepper-trade, line 718: Does the object referenced by `self` need locking too?

publishing/views.py
```diff
@@ -94,45 +95,60 @@ def _unpause_queue(self, request):
         OperationalStatus.unpause_queue(user=request.user)
         return self.view_url
 
+    @atomic
     def _promote_position(self, pk):
         try:
-            packaged_work_basket = PackagedWorkBasket.objects.get(pk=pk)
+            packaged_work_basket = PackagedWorkBasket.objects.select_for_update(
```
paulpepper-trade, line 101: I think this answers [an earlier question about locking this object](https://github.com/uktrade/tamato/pull/1267/files#r1688204336). Although, could the two objects being swapped be locked within `promote_position()` to help with readability?

publishing/views.py
```diff
@@ -94,45 +95,60 @@ def _unpause_queue(self, request):
...
         ):
             # Nothing to do in the case of these exceptions.
             pass
         return self.view_url
 
+    @atomic
     def _demote_position(self, pk):
         try:
-            packaged_work_basket = PackagedWorkBasket.objects.get(pk=pk)
+            packaged_work_basket = PackagedWorkBasket.objects.select_for_update(
```
paulpepper-trade, line 117: Ditto [this comment](https://github.com/uktrade/tamato/pull/1267/files#r1688212182).

publishing/views.py
```diff
@@ -94,45 +95,60 @@ def _unpause_queue(self, request):
...
         ):
             # Nothing to do in the case of these exceptions.
             pass
         return self.view_url
 
+    @atomic
     def _promote_to_top_position(self, pk):
         try:
-            packaged_work_basket = PackagedWorkBasket.objects.get(pk=pk)
+            packaged_work_basket = PackagedWorkBasket.objects.select_for_update(
```
paulpepper-trade, line 133: Could locking be done in `promote_to_top_position()`, as per [an earlier comment](https://github.com/uktrade/tamato/pull/1267/files#r1688212182)?

publishing/views.py
```diff
@@ -94,45 +95,60 @@ def _unpause_queue(self, request):
...
         ):
             # Nothing to do in the case of these exceptions.
             pass
         return self.view_url
 
+    @atomic
     def _remove_from_queue(self, pk):
         try:
-            packaged_work_basket = PackagedWorkBasket.objects.get(pk=pk)
+            packaged_work_basket = PackagedWorkBasket.objects.select_for_update(
```
paulpepper-trade, line 149: Could locking be done in `abandon()`, as per [an earlier comment](https://github.com/uktrade/tamato/pull/1267/files#r1688212182)?

publishing/views.py
```diff
@@ -184,12 +201,15 @@ def post(self, request, *args, **kwargs):
 
         return redirect(reverse("publishing:envelope-queue-ui-list"))
 
+    @atomic
     def _process_envelope(self, pk):
         if not OperationalStatus.is_queue_paused():
-            packaged_work_basket = PackagedWorkBasket.objects.get(pk=pk)
+            packaged_work_basket = PackagedWorkBasket.objects.select_for_update(
```
paulpepper-trade, line 207: Could locking be done in `begin_processing()`, as per [an earlier comment](https://github.com/uktrade/tamato/pull/1267/files#r1688212182)?

publishing/models/packaged_workbasket.py
dalecannon: `all()` is not enough to force the queryset evaluation. However, `TableLock` seems to do the trick, removing the need for the if/else block.

publishing/views.py
dalecannon: Done! (And for the other respective instance methods).

paulpepper-trade (approved):
LGTM

PR #1273, opened 2024-07-30
TP2000-1458: Limit erga omnes exclusions to erga omnes group members

eadpearce:
# TP2000-1458: Limit erga omnes exclusions to erga omnes group members

## Why

* Jersey and Guernsey are not part of the erga omnes geo area group. This prevents users from selecting them as exclusions and causing an error

## What

* Passes the list of erga omnes members to the react component as a new prop to populate the select options

### DIFF
--- a/measures/views/wizard.py
+++ b/measures/views/wizard.py
@@ -624,7 +624,14 @@ class MeasureCreateWizard(
             .as_at_today_and_beyond()
             .order_by("description")
         )
-        exclusions_options = all_geo_areas
+        erga_omnes = GeographicalArea.objects.erga_omnes().first()
+        erga_omnes_exclusions_pks = [
+            membership.member.pk
+            for membership in GeographicalMembership.objects.filter(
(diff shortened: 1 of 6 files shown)


measures/views/wizard.py
```diff
@@ -624,7 +624,14 @@ def get_react_script(self, form):
             .as_at_today_and_beyond()
             .order_by("description")
         )
-        exclusions_options = all_geo_areas
+        erga_omnes = GeographicalArea.objects.erga_omnes().first()
+        erga_omnes_exclusions_pks = [
+            membership.member.pk
+            for membership in GeographicalMembership.objects.filter(
+                geo_group__pk=erga_omnes.pk,
+            ).prefetch_related("geo_group", "member")
```
dalecannon, line 632: Would `select_related("member")` suffice if `member` is the only (foreign-key) related object that would hit the DB in each loop iteration?

dalecannon (approved):

PR #1274, opened 2024-08-01
TP2000-1446 Improve DBT SSO

mattjamc:
# TP2000-1446 Improve DBT SSO

## Why

Update the user email to pull through SSO UUID, not just email UUID. This will allow us to utilise people finder data. This can be useful for future TAP development such as workflows and notifications

## What

This PR:
- adds an sso_uuid column to the User table to store their SSO uuid.
- overrides [DBT SSO Authbroker Backend](https://github.com/uktrade/django-staff-sso-client) to include the SSO uuid for new users and add it in for existing users too.

I can't think of a worthwhile test that works but open to suggestions.

## Checklist
- Requires migrations? Yes


common/auth_backend.py
```diff
@@ -0,0 +1,24 @@
...
+class CustomAuthbrokerBackend(AuthbrokerBackend):
+    def user_create_mapping(self, profile):
+        return {
+            "is_active": True,
+            "first_name": profile["first_name"],
+            "last_name": profile["last_name"],
+            "sso_uuid": profile["user_id"],
+        }
+
+    def confirm_sso_uuid(self, user, profile):
```
paulpepper-trade, line 16: Might `set_sso_uuid()` better describe what this function does.

common/models/user.py
```diff
@@ -12,6 +12,8 @@ class User(AbstractUser):
         blank=True,
     )
 
+    sso_uuid = models.UUIDField(unique=True, null=True, blank=True)
```
paulpepper-trade, line 15: Because it's non-obvious, it maybe worth adding a docstring or `help_text` param to explain that this attribute value is populated via our Staff SSO middleware override.

common/auth_backend.py
```diff
@@ -0,0 +1,24 @@
...
+        if not user.sso_uuid:
+            user.sso_uuid = profile["user_id"]
```
paulpepper-trade, line 18: It would be safer to either wrap this in a try / except (perhaps with a `logger.warning()` in the except block) or do:
```suggestion
            user.sso_uuid = profile.get("user_id")
```
I know the base class applies your approach, but we'd want to reasonably avoid application failure due to base class changes, and we can do that here at almost no cost.

common/auth_backend.py
```diff
@@ -0,0 +1,24 @@
...
+UserModel = get_user_model()
+
+
+class CustomAuthbrokerBackend(AuthbrokerBackend):
+    def user_create_mapping(self, profile):
+        return {
+            "is_active": True,
+            "first_name": profile["first_name"],
+            "last_name": profile["last_name"],
+            "sso_uuid": profile["user_id"],
```
paulpepper-trade, line 13: Ditto other comment about accessing the `profile` dictionary.

paulpepper-trade (commented):
> I can't think of a worthwhile test that works but open to suggestions.

Perhaps mock `user` and `profile` to test `user_create_mapping()` and `set_sso_uuid()` in isolation?

Could you mock `AuthbrokerBackend.get_or_create_user()` in order to test your override, `CustomAuthbrokerBackend.get_or_create_user()`?

paulpepper-trade (approved):

PR #1280, opened 2024-08-14
TP2000-1377: Prevent origin referencing exclusion not in geo area group

eadpearce:
# TP2000-1377: Prevent origin referencing exclusion not in geo area group

## Why

* We were getting errors in the quota edit form due to selecting an exclusion not in a group

## What

* Uses the same pattern with the react multiselect with dynamic exclusions list as used in the measure create wizard
* Deletes a bunch of code that is now obsolete!
* Also overrides styling on the react multiselect to make it look a bit more govuk-y

Screenshot 2024-08-14 at 15 10 01


quotas/views.py
```diff
@@ -307,13 +309,29 @@ class QuotaUpdateMixin(
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
         kwargs["request"] = self.request
-        kwargs["geo_area_options"] = (
+        geo_area_options = (
             GeographicalArea.objects.current()
             .prefetch_related("descriptions")
             .with_latest_description()
             .as_at_today_and_beyond()
```
mattjamc, line 316: Not specific to this PR but can _as_at_today_and_beyond_ cause issues not including legacy codes. E.g. trying to submit origin form for quota 092851 fails validation as previous origins using legacy codes now have geo area field empty as it's not part of the queryset

mattjamc: Perhaps this isn't a real use case though - maybe no-one would ever try and edit an origin for that quota?

mattjamc (approved):
Looks good to me and all seems to work smoothly!

PR #1281, opened 2024-08-20
TP2000-1463 Support SQLite dump on DBT Platform

paulpepper-trade:
# TP2000-1463 Support SQLite dump on DBT Platform

## Why

The SQLite dump process fails on DBT Platform deployments because of file write access restrictions under an application’s deployment directory.

Both django-dotenv and python-dotenv rely upon the same `import dotenv`, causing a conflict when dependended-upon packages rely upon python-dotenv.

## What

This PR:

- Adds support to generate and execute SQLite migration source files under a system's temporary file storage directory. Copilot-based service deployments run under `/tmp`, while all other deployments do not do so by default - the env var `SQLITE_MIGRATIONS_IN_TMP_DIR` may be set to `True` to force `/tmp` directory behaviour.

- Switches from using the django-dotenv package to the more ubiquitous python-dotenv package to load `.env` file variables into the applications shell environment. This increases the chances of other packages that rely upon python-dotenv to work with Tamato.

## Checklist
- Requires migrations? No
- Requires dependency updates? yes


exporter/sqlite/runner.py
```diff
@@ -16,80 +18,210 @@
...
+        logger.info(f"Entering context manager {self.__class__.__name__}")
+        return settings.BASE_DIR
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        logger.info(f"Exiting context manager {self.__class__.__name__}")
+        for file in Path(settings.BASE_DIR).rglob(SQLITE_MIGRATIONS_GLOB):
+            file.unlink()
+
+
+def try_as_octal(decimal: int):
```
dalecannon, line 64: Is this function still intended to be used?

dalecannon (approved):

exporter/sqlite/runner.py
paulpepper-trade: It was being used and now it's not. Now removed.

PR #1282, opened 2024-08-22
TP2000-1465 Have commodity hierarchy only show active comm codes

mattjamc:
# TP2000-1465 Have commodity hierarchy only show active comm codes

## Why

The commodity code hierarchy tree in the Commodity Codes view, currently shows commodity codes that are no longer active (have been end dated)

TAP should only show “live” or Commodity Codes that either:

- Don’t have an end date or,
- Have an end date greater than todays date.

## What

Commodity code hierarchy has been corrected to:
- not include end dated commodities
- not display at all if you're viewing it for an end dated commodity

image

### DIFF
--- a/commodities/jinja2/includes/commodities/tabs/hierarchy.jinja
+++ b/commodities/jinja2/includes/commodities/tabs/hierarchy.jinja
@@ -10,6 +10,9 @@
 
 <h2 class="govuk-heading-l">Commodity hierarchy</h2>
 
+{% if not current %}
+<p>This commodity has been end dated so no longer forms part of the hierarchy.</p>
+{% else %}
 {% if snapshot.commodities|length == 1 %}
   {{ leaf_commodity(snapshot.commodities[0]) }}
 {% else %}
@@ -45,4 +48,5 @@
 
 {% endif %}
 
+{% endif %}
 {% endblock %}
--- a/commodities/jinja2/includes/commodities/tabs/hierarchy.jinja
+++ b/commodities/jinja2/includes/commodities/tabs/hierarchy.jinja
@@ -10,43 +10,42 @@
 
 <h2 class="govuk-heading-l">Commodity hierarchy</h2>
 
-{% if not current %}
-<p>This commodity has been end dated so no longer forms part of the hierarchy.</p>
-{% else %}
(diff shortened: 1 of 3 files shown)


commodities/jinja2/includes/commodities/tabs/hierarchy.jinja
```diff
@@ -10,6 +10,9 @@
 
 <h2 class="govuk-heading-l">Commodity hierarchy</h2>
 
+{% if not current %}
+<p>This commodity has been end dated so no longer forms part of the hierarchy.</p>
+{% else %}
```
paulpepper-trade, line 15: Should this `else` combine with the `if` on the following line to become an `elif` (and remove the closing `endif`)? Either that, or the following `if`/`else` blocks should be indented, I think.

commodities/views.py
```diff
@@ -176,19 +176,33 @@ def get_context_data(self, *args, **kwargs):
 
         context["selected_tab"] = "hierarchy"
 
-        prefix = self.object.item_id[0:4]
-        commodities_collection = CommodityCollectionLoader(prefix=prefix).load()
+        end_date = context["commodity"].valid_between.upper
+        current = end_date is None or end_date > date.today()
```
paulpepper-trade, line 180: We often use `current` to reference an object, so perhaps `is_current` would convey that this is a bool more effectively in this context?

commodities/tests/test_views.py
```diff
@@ -662,3 +662,29 @@ def test_commodity_measures_vat_excise_get_related(
...
+    assert commodity2.item_id in str(response.content)
+    assert commodity3.item_id not in str(response.content)
+
+    # Assert the tree does not show for the end dated commodity code
+    url = reverse("commodity-ui-detail-hierarchy", kwargs={"sid": commodity3.sid})
+    response = valid_user_client.get(url)
+    assert (
+        "This commodity has been end dated so no longer forms part of the hierarchy."
+        in str(response.content)
+    )
```
paulpepper-trade, line 690: This might be better split into two tests; one for each of the success and failure cases.

paulpepper-trade (approved):

PR #1283, opened 2024-08-30
TP2000-1474 Amend view definition periods

mattjamc:
# TP2000-1474 Amend view definition periods

## Why

Currently the View Definition Periods view allows a user to view definition periods by default. If a user wants to view associations, blocking and suspension periods, they need to select the option via a radio button in a hidden menu. This is a legacy practice from a previous iteration of TAP. Instead we should show the various options as tabs.

## What

This PR:
- replaces the old radio button form on the view definitions page with tabs
- adds a test 

image

image


CPrich905 (approved):
LGTM :)

PR #1286, opened 2024-09-03
TP2000-1387 Async Bulk Edit - Python 3.12 

LaurenMullally:
# TP2000-1387 Async Bulk Edit

## Why
The moment you've all been waiting for!! Sad it only amounts to 15 file changes oof. 

As part of efforts to make things quicker in TAP, we're utilising celery tasks to take some of the processing work away from the web worker. Previously we changed over the MeasureBulkCreate journey, this PR Is for the MeasureBulkEdit journey. It's pretty much a replica process just applied to the edits.

## What
- Set up and split out the done function to either process bulk edits synchronously, or asynchronously, depending on whether the `MEASURE_ASYNC_EDIT` switch is on. 
- If editing Async, form data, kwargs, and the selected measures are stored in the DB as a `MeasuresBulkEditor` object, then a task is scheduled. The task is run asynchronously, and handles setting the state of the task, although we don't utilise it like we do in measures as the design hasn't changed this time, and then handles calling the `MeasureEditor` class' `edit_measures()` function. 
- If editing Sync, then the done function basically hands the data to the same `edit_measures()` function above, just all within the web worker. 

## Checklist
- Requires migrations? - Yes
- Requires dependency updates? - No?


measures/editors.py
```diff
@@ -0,0 +1,112 @@
+from typing import Dict
```
LaurenMullally, line 1: This is the class that now handles editing the measures

measures/forms/wizard.py
```diff
@@ -781,7 +781,7 @@ def __init__(self, *args, **kwargs):
         )
 
 
-class MeasureStartDateForm(forms.Form):
+class MeasureStartDateForm(forms.Form, SerializableFormMixin):
```
LaurenMullally, line 784: Throughout this file, I add some functions to each form, as well as the mixin, all of which aid the serializing and deserializing required for passing the form data to the celery task

measures/models/bulk_processing.py
```diff
@@ -414,3 +415,164 @@ def _log_form_errors(self, form_class, form_or_formset) -> None:
         for form_errors in errors:
             for error_key, error_values in form_errors.items():
                 logger.error(f"{error_key}: {error_values}")
+
```
LaurenMullally, line 418: This is the model that we use to store the serialized form data, kwargs, selected measures, and other info. We store it in the db so that we can drag it out within the celery task as the celery task only likes being passed simple data types.

measures/tests/conftest.py
```diff
@@ -324,12 +324,33 @@ def mock_request(rf, valid_user, valid_user_client):
     request.requests_session = requests.Session()
```
LaurenMullally, line 324: Making some fixtures in here. The tests I use them in are parameterised, but you have to do some special trick in order to get fixtures used in a parameterised test.. this meant writing a fixture for some really simple stuff, so not sure whether this the best approach or not.

measures/tests/test_views.py
```diff
@@ -58,11 +58,11 @@
 
 @pytest.fixture()
 def mocked_diff_components():
-    """Mocks `diff_components()` inside `update_measure_components()` in
+    """Mocks `diff_components()` inside `update_measure_components()` that is called in
     `MeasureEditWizard` to prevent parsing errors where test measures lack a
```
LaurenMullally, line 62: Need to change this, she's not called in the Measure Edit Wizard any more!!

measures/tests/test_views.py
```diff
@@ -2038,7 +2038,7 @@ def test_multiple_measure_edit_single_form_functionality(
         assert measure.update_type == UpdateType.UPDATE
         assert reduce(getattr, updated_attribute.split("."), measure) == expected_data
 
-
+@override_settings(MEASURES_ASYNC_EDIT=False)
```
LaurenMullally, line 2041: These tests were preexisting, and checked that the form data provided did actually end up editing things as expected. Nice as it checks that the edit function works as it should, but I need to add this settings override so that it tests it through the synchronous route, or it complains that Celery isn't running! Considering that both routes use the same edit function, I felt this was okay.. but let me know if I'm wrong!

measures/util.py
```diff
@@ -49,6 +56,10 @@ def diff_components(
     """
     from measures.parsers import DutySentenceParser
 
+    # We add in the component output type here as otherwise we run into circular import issues.
```
LaurenMullally, line 59: Oh this was annoying!!! I'm not sure what I did, I think it's the making of the BulkEditor class pehaps?? but whatever I did gave me circular import errors which were such a pain. The workaround (Big thanks Paul) was to assign the type for the component output as none, then assign it correctly when it came to actually using the function!

measures/util.py
```diff
@@ -91,3 +102,103 @@ def diff_components(
...
+    diff_components(
+        instance=measure,
+        duty_sentence=duties if duties else measure.duty_sentence,
+        start_date=measure.valid_between.lower,
+        workbasket=workbasket,
+        transaction=workbasket.current_transaction,
+    )
+
+
+def update_measure_condition_components(
```
LaurenMullally, line 122: These all lived in the MeasureEditWizard, but considering they're now shared across two routes, I think it makes sense to keep them here. I guess these could live in the measure editor class?? but idk

measures/views/mixins.py
```diff
@@ -50,3 +51,72 @@ def get_queryset(self):
         """Get the queryset for measures that are candidates for
         editing/deletion."""
         return models.Measure.objects.filter(pk__in=self.measure_selections)
+
+
+class MeasureSerializableWizardMixin():
```
LaurenMullally, line 56: Moved these functions into a mixin that can now be shared across wizards..

eadpearce (approved):

PR #1287, opened 2024-09-09
Tp2000 1475 edit sub quotas

mattjamc:
# TP-??? Your PR title here


quotas/jinja2/quota-definitions/sub-quota-definitions-updates.jinja
```diff
@@ -1,6 +1,10 @@
 {% extends "layouts/form.jinja" %}
+{% from "components/breadcrumbs.jinja" import breadcrumbs %}
+
 {% from "components/table/macro.njk" import govukTable %}
 
+{% set page_title = "Update sub quota definition and association" %}
```
eadpearce, line 6: Does "sub quota" here want to be hyphenated or not?

quotas/views.py
```diff
@@ -307,7 +307,13 @@ def suspension_periods(self):
 
     @property
     def sub_quotas(self):
-        return QuotaAssociation.objects.filter(main_quota__order_number=self.quota)
+        return (
+            QuotaAssociation.objects.approved_up_to_transaction(
```
eadpearce, line 311: Does `.current()` also work here?

quotas/views.py
```diff
@@ -1128,3 +1134,174 @@ def get_context_data(self, *args, **kwargs):
...
+            self.last_transaction,
+        ).get(sid=self.kwargs["sid"])
+
+    @property
+    def association(self):
+        return models.QuotaAssociation.objects.approved_up_to_transaction(
+            self.last_transaction,
+        ).get(sub_quota__sid=self.sub_quota.sid)
+
+    def get_main_definition(self):
```
eadpearce, line 1190: This could also be a `@property`

quotas/views.py
SteveCN7: I think we'll need .current here to show newly created but not published data if IIRC

mattjamc: So is that what current does? I wasn't sure from the docstring as it talks about the global transaction whatever that is. As long as it also includes any updated version in the user's workbasket then that's fine. It seems to from my quick testing

mattjamc: I'm not clear what .current() does hence why I have this long approved_up_to_transaction() function.

The docstring says "Returns a queryset of approved versions of the model" so that suggests to me that .current() wouldn't show newly created but not published data? Unless I'm misunderstanding what approved means.

mattjamc: On the contrary when calling .current_version on a specific object this does _not_ include updated versions in the user's workbasket. I find that very confusing

eadpearce (approved):

PR #1288, opened 2024-09-10
TP2000-1472  Revise task-related models

dalecannon:
# TP2000-1472  Revise task-related models

## Why
The existing task-related models require some changes to support the first iteration task and workflow management features.

## What
- Renames `UserAssignment` model to `TaskAssignee`
- Adds new models, `TaskCategory`, `TaskProgressState` and `TaskLog`
- Adds new fields to `Task` model, `parent_task` and `creator`
- Adds factories and unit tests for the newly added models


tasks/models.py
```diff
@@ -1,26 +1,77 @@
...
+    category = models.ForeignKey(
+        "TaskCategory",
+        blank=True,
+        null=True,
+        on_delete=models.PROTECT,
+    )
+    progress_state = models.ForeignKey(
+        "TaskProgressState",
+        blank=True,
+        null=True,
```
paulpepper-trade, line 25: Shouldn't a Task instance always have a valid `TaskProgressState`, defaulting to "TODO" on creation and perhaps setting to "DONE" for those existing tasks that have an associated published workbasket?

tasks/models.py
```diff
@@ -1,26 +1,77 @@
...
+    )
+
+    class Meta:
+        verbose_name_plural = "task categories"
+
+    def __str__(self):
+        return self.name
+
+
+class TaskProgressState(models.Model):
```
paulpepper-trade, line 64: Should we have an invariant base set of system-defined states, say, `"TODO"`, `"IN_PROGRESS"` and `"DONE"`. One of the `django.db.models.Choices` may be useful in achieving that.

tasks/models.py
```diff
@@ -1,26 +1,77 @@
...
+        on_delete=models.PROTECT,
+    )
+    parent_task = models.ForeignKey(
+        "self",
+        blank=True,
+        null=True,
+        on_delete=models.CASCADE,
+        related_name="subtasks",
+    )
     workbasket = models.ForeignKey(
```
paulpepper-trade, line 35: Not all `Task`s will have a workbasket associated with them, so `Task.workbasket` would need to be nullable.

tasks/models.py
dalecannon: Yeah, I've now made `progress_state` a required field alongside a data migration operation to give existing tasks a progress state value.

tasks/models.py
dalecannon: Yeah, I've added those states as choices in an enum on the model and used them to create initial `TaskProgressState` instances in the migration.

tasks/models.py
dalecannon: Ah I had forgotten to change it. Done now!

tasks/models.py
```diff
@@ -1,26 +1,84 @@
...
+        on_delete=models.PROTECT,
+        related_name="created_tasks",
+    )
 
     def __str__(self):
         return self.title
 
 
-class UserAssignmentQueryset(models.QuerySet):
+class TaskCategory(models.Model):
```
paulpepper-trade, line 51: In the context of this module, do we need to prefix `Category` with `Task`?

tasks/models.py
```diff
@@ -1,26 +1,84 @@
...
+
+    class Meta:
+        ordering = ["name"]
+        verbose_name_plural = "task categories"
+
+    def __str__(self):
+        return self.name
+
+
+class TaskProgressState(models.Model):
```
paulpepper-trade, line 65: Ditto previous comment about `Task` prefix.

paulpepper-trade (approved):

PR #1293, opened 2024-09-16
bugfix

CPrich905:
# Bug fix on the QuotaDefinition duplicator

## Why
During use, the Tariff Managers discovered that there was an issue with the validation:
if a QuotaDefinition did not have any sub_quotas, it would fail `check_QA5_equivalent_volumes` as the volume value count would be 0 rather than 1. 
During testing of this, identified another issue whereby if the main quota definition was open ended, the select definition pages and edit definition pages would error, and `check_QA2_dict` would fail.

## What
Fixes the above problems by:
- allowing for null values in parent definition end dates
- checking that the volume values count in `check_QA5_equivalent_volumes` is `<=1` rather than `==1`

### DIFF
--- a/quotas/jinja2/quota-definitions/sub-quota-definitions-selected.jinja
+++ b/quotas/jinja2/quota-definitions/sub-quota-definitions-selected.jinja
@@ -15,11 +15,12 @@
         </span>
       {% endset %}
       {% set main_definition = view.get_main_definition(definition.main_definition)%}
-
+      {{main_definition.valid_between.lower}}
+      {{main_definition.measurement_unit.abbreviation}}
       {{ table_rows.append([
         {"text": main_definition.sid},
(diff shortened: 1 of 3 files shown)


quotas/jinja2/quota-definitions/sub-quota-definitions-selected.jinja
```diff
@@ -15,11 +15,12 @@
         </span>
       {% endset %}
       {% set main_definition = view.get_main_definition(definition.main_definition)%}
-
+      {{main_definition.valid_between.lower}}
```
mattjamc, line 18: Were lines 18/19 added for testing? or are they supposed to be displaying here on the page?

CPrich905: Thanks! Removed

mattjamc (approved):
LGTM :)

PR #1297, opened 2024-09-19
TP2000-1352: Add react origins form to quota create page

eadpearce:
# TP2000-1352: Add react origins form to quota create page

## Why

* Fixes a workbasket transaction error caused by trying to update a quota and origins after creating a quota and origins in different transactions
* Brings the create form in line with the edit form

## What

* Reuses the react quota origins and exclusions form used on the quota edit page and adds it to the quota create page as well
* More tests to improve coverage

### DIFF
--- a/quotas/forms.py
+++ b/quotas/forms.py
@@ -51,7 +51,7 @@ START_DATE_HELP_TEXT = "If possible, avoid putting a start date in the past as t
 ORDER_NUMBER_HELP_TEXT = "The order number must begin with 05 and be 6 digits long. Licensed quotas must begin 054 and safeguards must begin 058"
 
 
-class QuotaOriginsReactMixin:
+class QuotaOriginsReactMixin(ExtraErrorFormMixin):
     """Custom cleaning and validation for QuotaUpdateForm and
     QuotaOrderNumberCreateForm."""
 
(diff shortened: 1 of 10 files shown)


quotas/forms.py
```diff
@@ -338,6 +317,7 @@ def init_layout(self, request):
 
 
 class QuotaOrderNumberCreateForm(
+    QuotaOriginsReactMixin,
```
dalecannon, line 320: Does `ExtraErrorFormMixin` also need to be included here for `self.clean_quota_origins()` to use `self.add_extra_error()`?

quotas/forms.py
```diff
@@ -375,23 +355,39 @@ class Meta:
...
 
+        origins_html = render_to_string(
+            "includes/quotas/quota-create-origins.jinja",
+            {
+                "object": self.instance,
+                "request": request,
+                "geo_area_options": self.geo_area_options,
+                "groups_with_members": self.groups_with_members,
+                "exclusions_options": self.exclusions_options,
+                "errors": self.errors,
```
dalecannon, line 382: I noticed `QuotaUpdateForm` has an `origins_initial` key that preserves submitted origins data in case the form fails validation. Is that something that can be done and is worth doing in the create form too?

quotas/forms.py
eadpearce: Yep. Not sure why I didn't notice that in testing. I'll add another to test that

dalecannon (approved):

PR #1301, opened 2024-10-04
TP2000-1515 Deleting definition deletes association too

mattjamc:
# TP2000-1515 Deleting definition deletes association too

## Why

If a definition period for either the main or sub quota in an association is deleted, then the association object must also be deleted in that same transaction.

## What

This PR:
- updates the definition delete view to also delete any linked associations in the same transaction
- adds a test
- updates QA6 so it doesn't fail when an association gets deleted as there aren't any left

image

### DIFF
--- a/quotas/business_rules.py
+++ b/quotas/business_rules.py
@@ -540,12 +540,12 @@ def check_QA6_dict(main_quota, new_relation_type, transaction=None):
         .order_by("sub_quota_relation_type")
         .distinct()
     )
-    if relation_type.count() > 1:
-        return False
+    if relation_type.count() == 0:
+        return True
     elif relation_type.count() == 1:
         return relation_type[0]["sub_quota_relation_type"] == new_relation_type
     else:
-        return True
+        return False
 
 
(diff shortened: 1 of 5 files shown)


quotas/business_rules.py
```diff
@@ -544,6 +544,8 @@ def check_QA6_dict(main_quota, new_relation_type, transaction=None):
         return False
     elif relation_type.count() == 1:
         return relation_type[0]["sub_quota_relation_type"] == new_relation_type
+    else:
```
CPrich905, line 547: In this instance, I don't think it makes much of a difference, however I think in general it might be better to have a specific check for the truthy criteria here rather than a catch-all else, ie. if the relationship count is 0 or 1, return true, else false.

mattjamc: Yeah makes sense  - I can change it to check if it's equal to 0 then return True instead. I do think something to that effect is necessary or it will just return false and fail

CPrich905 (approved):

PR #1302, opened 2024-10-04
TP2000-1493 - TAP driven quota open data export

dougmills-DIT:
# TP2000-1493 TAP driven quota open data export

## Why
 * The current open data implementation is created from an SQL lite export imported into data workspace and queried via a static query stored in data-flow. 
 * The move from data workspace to TAP generated will speed the process up and allows more stringent testing to be automated around the output. 
 * Update have been made to the current implementation several times to resolve data inconsistency issues. Moving to within TAP will help reduce future issues as TAP is updated. 
 * Moving the creation of quotas open data is the first of several open data exports that would be better placed within TAP. 

## What
 * Create process to export TAP quota data to a CSV file
 * Add methods of storage for both S3 and local 
 * Unit test everything 

## Checklist
- Requires migrations? No

Links to relevant material
See: [Description](https://uktrade.atlassian.net/browse/TP2000-1493)


SteveCN7:
Silly question - the initial volume column, are we pulling that from TAP or OTT? I think we should be pulling it from TAP really if we're not already

exporter/quotas/runner.py
```diff
@@ -0,0 +1,351 @@
+import csv
+import logging
+from datetime import date
+from datetime import timedelta
+from tempfile import NamedTemporaryFile
+from typing import List
+
+from django.db.models import Q
+
+from commodities.models import GoodsNomenclatureDescription
```
sumau, line 10:
```suggestion
from commodities.models import GoodsNomenclatureDescription
```

exporter/quotas/runner.py
```diff
@@ -0,0 +1,351 @@
...
+            if quota.measurement_unit.abbreviation != "":
+                measurement_unit_description = (
+                    measurement_unit_description
+                    + f" ({quota.measurement_unit.abbreviation})"
+                )
+            return measurement_unit_description
+        return None
+
+    @staticmethod
+    def get_api_query_date(quota):
```
DevelopwithTom, line 200: test

exporter/quotas/runner.py
```diff
@@ -0,0 +1,351 @@
...
+        Args:
+            quota: The quota to be queried
+
+        Returns:
+            date or none: a date if available or none if quota is in the future
+        """
+        api_query_dates = []
+
+        # collect possible query dates, but only for current and historical, not future
+        if quota.valid_between.lower <= date.today():
```
pacesetterplus, line 220:
```suggestion
        if quota.valid_between.lower <= specified_date:
```

PR #1304, opened 2024-10-09
Tp2000 1516 delete quota association

CPrich905:
# TP-2000 1516 Delete Quota Association

## Why
There is a requirement for TMs to be able to delete a QuotaAssociation between two QuotaDefinitions while maintaining the QuotaDefinitions

## What
This PR follows the standard Delete journey, enabling a TM to delete a QuotaAssociation object without deleting either the main or sub-quota definition:
- Adding a Delete link in the Sub-quotas view (image 1 below)
- Following the standard delete journey (images 2 & 3)
- Adding tests
There are some minor copy changes as requested by Ash/Rachel during an initial review:
- On the Quota Data/sub-quotas page: changing table headings from `Quota definition sid` to `Quota definition SID` and `Relation type` to `Relationship type`

Note: on the confirm delete page, we currently link back to the sub-quota definition. Once this ticket is in production, should our users request it, we can either change this to direct the user back to the main quota definition, or add a button so the user can return to either main or sub-quota definition.
An association can be created from either main or sub-quota definition.

image
image
image

### DIFF
--- a/quotas/jinja2/quotas/tables/sub_quotas.jinja
+++ b/quotas/jinja2/quotas/tables/sub_quotas.jinja
@@ -2,7 +2,6 @@
 <h2 class="govuk-heading-l">Sub-quotas</h2>
 
 {% set table_rows = [] %}
-{# <a class="govuk-link" href="{{ url('quota_association-ui-delete', kwargs={'main_quota__sid': object.main_quota.sid, 'sub_quota__sid': object.sub_quota.sid}) }}">Delete</a> #}
 {% if sub_quotas %}
   {% for object in sub_quotas %}
     {% set definition_link -%}
@@ -29,11 +28,11 @@
 
   {{ govukTable({
     "head": [
-      {"text": "Quota definition sid"},
+      {"text": "Quota definition SID"},
       {"text": "Sub-quota order number"},
       {"text": "Start date"},
       {"text": "End date"},
-      {"text": "Relation type"},
+      {"text": "Relationship type"},
       {"text": "Coefficient"},
       {"text": "Actions"},
     ],
--- a/quotas/jinja2/quotas/tables/sub_quotas.jinja
+++ b/quotas/jinja2/quotas/tables/sub_quotas.jinja
@@ -2,7 +2,7 @@
 <h2 class="govuk-heading-l">Sub-quotas</h2>
 
 {% set table_rows = [] %}
-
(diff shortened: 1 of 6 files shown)


quotas/views.py
```diff
@@ -1288,3 +1296,42 @@ def dispatch(self, request, *args, **kwargs):
...
+
+class QuotaAssociationConfirmDelete(
+    TrackedModelDetailView,
+):
+    template_name = "quota-associations/confirm-delete.jinja"
+    model = models.QuotaDefinition
+
+    def get_queryset(self):
+
+        return models.QuotaDefinition.objects.filter(
```
mattjamc, line 1335: I'm not sure if get_queryset is doing anything here that Django wouldn't already do itself. It has the model name and knows the sid from the URL so I think it's fine without.

quotas/jinja2/quotas/tables/sub_quotas.jinja
```diff
@@ -11,27 +10,29 @@
     {% set sub_quota_link -%}
       <a class="govuk-link" href="{{ url('quota-ui-detail', args=[object.sub_quota.order_number.sid]) }}">{{ object.sub_quota.order_number.order_number }}</a>
     {% endset %}
-    {% set edit_link -%}
+    {% set actions_html %}
       <a class="govuk-link" href="{{ object.sub_quota.version_at(request.user.current_workbasket.transactions.last()).get_association_edit_url() }}">Edit</a>
+      <a class="govuk-link" href="{{ url('quota_association-ui-delete', args=[object.pk]) }}">Delete</a>
```
mattjamc, line 15: Could this say 'Delete association' as I don't think it's clear from the UI whether this button will delete the association or the sub-quota definition. Though perhaps will be more clear if we are changing the tab name to 'Quota associations' soon

CPrich905: Sure - I've changed the `Edit` to `Edit association` as well to be consistent

quotas/views.py
```diff
@@ -121,6 +121,14 @@ def get_queryset(self):
         return models.QuotaOrderNumber.objects.approved_up_to_transaction(tx)
 
 
+# class QuotaAssociationMixin:
```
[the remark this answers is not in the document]
CPrich905: Yes

mattjamc (approved):

PR #1307, opened 2024-10-16
Tp2000 1517 implement improved quota views for users

CPrich905:
# TP2000-1517

## Why
Implement updates to the QuotaDetail and QuotaDefinitionList 

## What
- Adding a title caption based on the quotas' status as a main or sub-quota
- Displaying main and sub-quotas on relevant tabs
- Add an action column with Edit/Delete columns on the Quota data view
- Change tab title from `Sub-quotas` to `Quota associations`
- Further changes as requested by UR team as detailed on ticket 1517, including changes to Actions column as requested by @SteveCN7 

image
image
image
image
image

### DIFF
--- a/quotas/jinja2/includes/quotas/tabs/sub_quotas.jinja
+++ b/quotas/jinja2/includes/quotas/tabs/sub_quotas.jinja
@@ -43,24 +43,26 @@
 <div class="quota__sub-quotas govuk-grid-row">
     <div class="quota__sub-quotas__content govuk-grid-column-three-quarters">
       <h2 class="govuk-heading-l">Quota associations</h2>
-        {% if sub_quota_associations %}
-        <h3>Sub-quotas</h3>
-        {% elif main_quota_associations %}
-        <h3>Main quota</h3>
-        {% else %}
-          <p class="govuk-body">No active main or sub-quotas.</p>
-        {% endif %}
-        {{
-          govukTable({
-            "head": [
-              {"text": "Order number"},
-              {"text": "Relationship type"},
-              {"text": "Co-efficient"},
-              {"text": "Actions"}
-            ],
-            "rows": table_rows
-          })
-        }}
+      {% if sub_quota_associations or main_quota_associations %}
+          {% if sub_quota_associations %}
+          <h3>Sub-quotas</h3>
+          {% elif main_quota_associations %}
+          <h3>Main quota</h3>
+          {{
+            govukTable({
+              "head": [
+                {"text": "Order number"},
+                {"text": "Relationship type"},
+                {"text": "Co-efficient"},
+                {"text": "Actions"}
+              ],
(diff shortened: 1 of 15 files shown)


quotas/views.py
```diff
@@ -240,12 +240,14 @@ def get_context_data(self, *args, **kwargs):
             f"{URLs.BASE_URL.value}quota_search?order_number={self.object.order_number}"
         )
 
-        context[
-            "quota_associations"
-        ] = QuotaAssociation.objects.latest_approved().filter(
+        context["sub_quota_associations"] = QuotaAssociation.objects.all().filter(
```
dalecannon, line 243: Might `.current()` (or similar) be needed here to avoid displaying multiple versions?

quotas/jinja2/includes/quotas/tabs/sub_quotas.jinja
```diff
@@ -1,33 +1,69 @@
...
-        {% if quota_associations %}
-        {{ govukTable({
-            "head": [
-                {"text": "Sub-quota order number"},
-                {"text": "Relation type"},
+      <h2 class="govuk-heading-l">Quota associations</h2>
+      {% if sub_quota_associations or main_quota_associations %}
+          {% if sub_quota_associations %}
+          <h3>Sub-quotas</h3>
+          {% elif main_quota_associations %}
```
dalecannon, line 49: Does the table for sub-quotas need adding here?

quotas/views.py
CPrich905: I spoke with Steve C about this - the intention is to have all available versions of associations available. That said, it's a new feature, so this might change after a couple of weeks/months of use!

quotas/jinja2/includes/quotas/tabs/sub_quotas.jinja
CPrich905: Sorry, this was missing an `{% endif %}` - the column headings are the same, only the table title changes based on the presence of `sub_quota_associations` or `main_quota_associations`. Corrected in ca79f56.

quotas/jinja2/quotas/definitions.jinja
```diff
@@ -44,8 +50,14 @@
 
 {% block content %}
 
+<span class="govuk-caption-xl">
+  {% block page_subtitle %}
+    {{ page_subtitle }}
+  {% endblock %}
+</span>
```
dalecannon, line 57: Does this need to be wrapped in a if statement in case the quota is neither a main quota or a sub-quota and {{ page_subtitle }} is displayed as is without a set value?

dalecannon (approved):

quotas/jinja2/quotas/definitions.jinja
CPrich905: Done!

PR #1308, opened 2024-10-17
Implement prototype UI for tasks

dalecannon:
# Implement prototype UI for tasks

## Why
A prototype UI for tasks is useful for testing user interactions and enabling quicker design and functionality iterations.

## What
- Creates initial views and URL routes to list, view, create, update and delete tasks

- Creates initial forms to create, update and delete tasks

- Creates `TaskFilter` that enables filtering by `progress_state` & `category` and searching by `title` & `description`

- Uses placeholder templates until UI designs are finalised

- Sets up a pre_save signal for automated `TaskLog` creation following updates to `Task` and `TaskAssignee` instances

- Registers `TaskLog` model in admin to make entries viewable

- Adds two mixins, `WithSignalManagerMixin` & `WithSignalQuerysetMixin`, overriding some default Django methods that would otherwise not send pre_save signals


tasks/models.py
```diff
@@ -218,11 +218,12 @@ class AuditActionType(models.TextChoices):
     action = models.CharField(
         max_length=100,
         choices=AuditActionType.choices,
+        editable=False,
     )
     description = models.TextField(editable=False)
     task = models.ForeignKey(
         Task,
-        on_delete=models.PROTECT,
+        on_delete=models.CASCADE,
```
paulpepper-trade, line 226: Does it make sense to persist logs beyond the lifetime of the things they're referencing for auditing purposes? If so, then `on_delete=models.SET_NULL`. I'm unsure.

paulpepper-trade (approved):

paulpepper-trade (approved):

PR #1313, opened 2024-10-25
TP2000-1476 Edit capability for blocking periods

mattjamc:
# TP2000-1476 Edit capability for blocking periods

## Why

User’s can create a blocking periods for quotas, but they can’t edit existing periods. We should implement that journey using the create form as a template.

### DIFF
--- a/quotas/views.py
+++ b/quotas/views.py
@@ -302,7 +302,9 @@ class QuotaDefinitionList(SortingMixin, ListView):
 
     @property
     def blocking_periods(self):
-        return QuotaBlocking.objects.filter(quota_definition__order_number=self.quota)
+        return QuotaBlocking.objects.current().filter(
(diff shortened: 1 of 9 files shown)


quotas/views.py
```diff
@@ -1242,6 +1248,77 @@ def get_queryset(self):
         return QuotaSuspension.objects.filter(pk=self.deleted_suspension)
 
 
+class QuotaBlockingUpdateMixin(TrackedModelDetailMixin):
```
CPrich905, line 1251: Not essential, but if my refactoring of the quotas space PR is released first, you could move these into their own file with the other QuotaBlocking model views.

CPrich905 (approved):
One comment which is entirely dependent on whether this goes out before my PR or not, otherwise LGTM :)

PR #1315, opened 2024-10-28
formatting

CPrich905:
# Refactor of quotas/views.py and forms.py

## Why
Both the views.py and forms.py files are getting large (>1,000 lines long)
These are only going to grow further as we add functionality. Reducing the size of the files makes them easier to comprehend.

## What
This PR separates out the forms and views which are used for the DuplicateDefinitionsWizard and BulkQuotaDefinitionCreate processes. Note that the BulkQuotaDefinitionCreate is a work in progress.

### DIFF
--- a/quotas/forms/base.py
+++ b/quotas/forms/base.py
@@ -16,7 +16,6 @@ from django.db.models import TextChoices
 from django.template.loader import render_to_string
 from django.urls import reverse_lazy
 
-from common.fields import AutoCompleteField
 from common.forms import BindNestedFormMixin
 from common.forms import ExtraErrorFormMixin
 from common.forms import FormSet
@@ -40,8 +39,6 @@ from quotas import validators
 from quotas.constants import QUOTA_EXCLUSIONS_FORMSET_PREFIX
 from quotas.constants import QUOTA_ORIGIN_EXCLUSIONS_FORMSET_PREFIX
 from quotas.constants import QUOTA_ORIGINS_FORMSET_PREFIX
-from quotas.serializers import serialize_duplicate_data
-from workbaskets.forms import SelectableObjectsForm
 
 RELATIONSHIP_TYPE_HELP_TEXT = "Select the relationship type for the quota association"
 COEFFICIENT_HELP_TEXT = "Select the coefficient for the quota association"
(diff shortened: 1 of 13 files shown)


quotas/forms/base.py
```diff
@@ -0,0 +1,404 @@
...
+                    id = int(g[field_name])
+                    g[field_name] = GeographicalArea.objects.get(id=id)
+            initial_exclusions = new_data
+
+        initial[QUOTA_ORIGIN_EXCLUSIONS_FORMSET_PREFIX] = initial_exclusions
+
+        return initial
+
+
+class QuotaSuspensionType(TextChoices):
```
eadpearce, line 199: Could also split this out into a forms/constants.py

quotas/forms/base.py
```diff
@@ -0,0 +1,404 @@
...
+from common.forms import unprefix_formset_data
+from common.util import validity_range_contains_range
+from common.validators import SymbolValidator
+from common.validators import UpdateType
+from geo_areas.models import GeographicalArea
+from quotas import models
+from quotas import validators
+from quotas.constants import QUOTA_ORIGIN_EXCLUSIONS_FORMSET_PREFIX
+
+CATEGORY_HELP_TEXT = "Categories are required for the TAP database but will not appear as a TARIC3 object in your workbasket"
```
eadpearce, line 32: These could also go in a constants file but for now it's probably fine for them to live in here

quotas/forms/react_forms.py
```diff
@@ -0,0 +1,413 @@
...
+from django.core.exceptions import ValidationError
+from django.template.loader import render_to_string
+
+from common.forms import ExtraErrorFormMixin
+from common.forms import ValidityPeriodForm
+from common.forms import unprefix_formset_data
+from geo_areas.models import GeographicalArea
+from quotas import models
+from quotas import validators
+from quotas.constants import QUOTA_EXCLUSIONS_FORMSET_PREFIX
```
eadpearce, line 20: Looks like the only values we have in this constants file are used in forms only so I think it makes sense to move this to `quotas/forms/constants.py` and add some of the other ones that appear in here. Some might need renaming to avoid clashes

quotas/forms/wizards.py
```diff
@@ -0,0 +1,309 @@
+from crispy_forms_gds.helper import FormHelper
+
+# from crispy_forms_gds.layout import Button
```
eadpearce, line 3: Commented code

eadpearce (approved):

PR #1317, opened 2024-10-31
TP2000-1562 Quota association view update

mattjamc:
# TP2000-1562 Quota association view update

## Why

There are some issues with the current quota association view:
- it errors if someone loads the page without a workbasket
- if a sole definition has been edited in the workbasket, it currently links to an update-edit form which errors - as while an update-edit would be appropriate for the definition this does not work for the unedited association
- there is an edit link for main quota associations which doesn't reflect where the link goes (currently to the edit page of whatever the most recently edited sub quota was)

## What

This PR:
- adds require_workbasket_decorator to quota definition list view to catch and redirect anyone without a workbasket
- removes the actions column (thus edit link) for main quota associations since the form in question only handles sub quota association edits
- ensures the update (not update-edit) form is loaded if an association has not yet been edited to prevent errors

## Notes
- Whilst I could have kept the edit link for main quota associations and just linked to the edit form for whatever the sub-quota is of that relationship, this seemed overly complex and difficult to make clear to the user that that is what it would be doing.
- I've also added in some help text as the whole thing feels quite confusing to me. However, if you feel this help text actually confuses things more I'm happy to remove.

**Main quota with sub quota definitions**

image

**Sub quota with main quota definitions**

image

### DIFF
--- a/common/tests/test_models.py
+++ b/common/tests/test_models.py
@@ -409,7 +409,7 @@ def test_get_description_dates(description_factory, date_ranges):
 
 def test_trackedmodel_get_url(
     trackedmodel_factory,
-    valid_user_client,
+    client_with_current_workbasket,
     mock_quota_api_no_data,
 ):
     """Verify get_url() returns something sensible and doesn't crash."""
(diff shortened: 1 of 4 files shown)


TomMacca:
I think I get it @mattjamc, I like the help text, maybe for this kinda PR a couple screenshots might help

mattjamc:
> I think I get it @mattjamc, I like the help text, maybe for this kinda PR a couple screenshots might help

Images now added in @TomMacca

PR #1318, opened 2024-11-01
TP2000-1501 Fix comm code hierarchy bug

mattjamc:
# TP2000-1501 Fix comm code hierarchy bug

## Why

The commodity code hierarchy currently shows all versions of commodities. Therefore if one has multiple versions it will appear multiple times in the tree which is incorrect.

## What

This PR:
- adds current_only so only current commodities are shown
- adds a test to account for this


TomMacca (approved):
Looks good Matt, have seen this problem in the past

PR #1319, opened 2024-11-06
TP2000-1487: Create Subtask Form

marya-shariq:
# TP2000-1487: Create Subtask Form
## Why
* As part of the task workflow work, we need to provide users with a way of creating subtasks that are linked to parent tasks.
* This is so users can break down their tasks into smaller tasks should they choose to do so. 

## What
* Created a URL for creating a subtask from a parent task.
* Added a view to serve this URL 
* Added a Create Subtask form which extends off of TaskBaseForm, that on save passes in a parent task and assigns to the created subtask

## Checklist
- Requires migrations? - NO
- Requires dependency updates? - NO

### DIFF
--- a/tasks/views.py
+++ b/tasks/views.py
@@ -11,6 +11,7 @@ from django.views.generic.edit import UpdateView
 from common.views import SortingMixin
 from common.views import WithPaginationListView
 from tasks.filters import TaskFilter
+from tasks.forms import SubTaskCreateForm
 from tasks.forms import TaskCreateForm
 from tasks.forms import TaskDeleteForm
 from tasks.forms import TaskUpdateForm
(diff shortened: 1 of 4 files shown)


tasks/views.py
```diff
@@ -112,3 +113,23 @@ def get_context_data(self, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["page_title"] = "Create a subtask"
+        return context
+
+    def form_valid(self, form):
+        parent_task = Task.objects.filter(pk=self.kwargs["pk"]).first()
+        self.object = form.save(parent_task, user=self.request.user)
+        return HttpResponseRedirect(self.get_success_url())
```
LaurenMullally, line 132: We will fix this in another ticket!

tasks/views.py
```diff
@@ -112,3 +113,23 @@ def get_context_data(self, **kwargs):
...
+
+class SubTaskCreateView(PermissionRequiredMixin, CreateView):
+    model = Task
+    template_name = "layouts/create.jinja"
+    permission_required = "tasks.add_task"
+    form_class = SubTaskCreateForm
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["page_title"] = "Create a subtask"
```
dalecannon, line 126: It might be useful to see which parent task the subtask is being created for, either in the `page_title` or `page_label` for the time being. But that could be included in some fashion in a later iteration of this view's template.

dalecannon (approved):

PR #1320, opened 2024-11-11
TP2000-1537 Bulk create QuotaDefinitionPeriods

CPrich905:
# TP2000-1537 Bulk create QuotaDefinitionPeriods

## Why
* Currently QuotaDefinitionPeriods can only be created one by one
* A Quota may cover a significant period of time and creating definition periods to cover the duration of the quota may require a significant amount of repetitions through this journey.

## What
Following the figma design linked on the ticket and [here]([Single and multiple journeys](https://www.figma.com/proto/A3q0JK3d0bdkC22OBb0O6A/%5BTP2000-1537%5D-Bulk-create-quota-definitions?page-id=6344%3A25965&node-id=6344-26554&node-type=canvas&viewport=807%2C320%2C0.06&t=wu3ha8Tkub1BXf8R-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=6344%3A26554&show-proto-sidebar=1)), this PR 
- utilises a Django form wizard to bulk create QuotaDefinitionPeriods, and 
- implements some design changes to the journey for creating a single QuotaDefinitionPeriod

There are four stages to the journey:
Beginning the journey from the Edit Workbasket View
image

1. Start, outlining the information the users will need to have available and collecting the QuotaOrderNumber the QuotaDefinitionPeriods are to be built with.
image

2. Enter the definition information. This collects the information required to build a QuotaDefinitionPeriod object, as well as the number of definition periods to be created, and the frequency with which they will be duplicated (Once, Annually, Twice-yearly, Quarterly, Custom)
image

3. Review page. Here, the user can review the details that for an individual QuotaDefinitionPeriod via a collapsible info box, or edit any that require changing via the Edit link. On `submit` and the QuotaDefinitionPeriods are created and the user is taken to a standard success page.
image

On success, the user is redirected to a standard success page with Next steps and Further actions
image

Design changes:
1. removing the accordion sections
2. updates to default and help text for various widgets
3. changes to the submit buttons/back links - these are now passed as args from the view, enabling us to reuse the QuotaDefinitionCreateForm for updating the bulk created definitions
4. reorganisation of the Actions panel on the Quotas page
7. adding help/info text to the journey for creating a single QuotaDefinitionPeriod
8. adding a custom widget that allows the use of a suffix on a Decimal input field
image


quotas/jinja2/includes/quotas/actions.jinja
```diff
@@ -12,12 +12,13 @@
             <li><a class="govuk-link" href="{{ delete_url }}">Delete this {{object._meta.verbose_name}}</a></li>
           {% endif %}
           <br>
-          <strong class=govuk-heading-xs">Create quota data</strong>
-            <li><a class="govuk-link" href="{{ url('quota_definition-ui-create', args=[object.sid]) }}">Create definition period</a></li>
+          <strong class="govuk-heading-xs">Create quota data</strong>
+            <li><a class="govuk-link" href="{{ url('quota_definition-ui-create', args=[object.sid]) }}?order_number={{object.order_number}}">Create definition period</a></li>
+            <li><a class="govuk-link" href="{{ url('quota_definition-ui-bulk-create', args=[object.sid]) }}">The bulk creator</a></li>
```
mattjamc, line 17: I like the name but I have a feeling it wouldn't pass the GDS naming conventions - is it clear what is being bulk created here? (Seems obvious to us)

quotas/forms/definitions.py
```diff
@@ -156,9 +156,14 @@ class Meta:
             "maximum_precision",
         ]
 
+    order_number = forms.ModelChoiceField(
```
mattjamc, line 159: What is the reason for adding this field in? What was happening before to get the order number?

mattjamc: I seem to be getting an error in this form that order number doesn't have a value when I submit but can't figure out why are you are assigning it an initial value

quotas/jinja2/quota-definitions/create.jinja
```diff
@@ -1,12 +1,28 @@
...
+            {{ page_title }}
+        {% endblock %}
+        <span class="govuk-caption-xl">
+        {% block page_subtitle %}
+            {{ page_subtitle }}
+        {% endblock %}
+        </span>
+    </h1>
+  <p class="govuk-body">You can create one new quota definition period using this form. Alternatively, you can <a class="govuk-link" href="{{ url('quota_definition-ui-bulk-create') }}">create multiple quota definition periods</a> for a chose quota order number.</p>
+  {% call django_form() %}
```
mattjamc, line 25: chosen*

quotas/jinja2/quota-definitions/create.jinja
```diff
@@ -1,12 +1,28 @@
...
+{% extends 'layouts/form.jinja' %}
 
 {% set page_title = "Create a new quota definition period" %}
-
+{% set page_subtitle = "Quota order number: "~ quota_order_number %}
 {% block breadcrumb %}
   {{ breadcrumbs(request, [
       {"text": "Find and edit quotas", "href": url("quota-ui-list")},
-      {"text": "Quota order number", "href": url("quota-ui-detail", args=[view.kwargs.sid]) ~ "#definitions"},
+      {"text": "Quota order number", "href": url("quota-ui-detail", args=[quota_sid]) ~ "#definitions"},
```
mattjamc, line 8: Appreciate it was already like that but I'm not sure #definitions is a real anchor tag for that page. Maybe #definition-details or none at all

mattjamc (commented):
Looks good! Just some small comments

quotas/forms/definitions.py
CPrich905: Fixed the error, but do double check it's working for you as well!
As discussed on Teams, there was an issue testing this due to inherited forms using `transaction.atomic`. This was fixed by passing in the `quota_order_number` in the query string and then extracting it here. The error you encountered was that, while the initial value was being set, the `order_number` widget was still showing as missing a value (this error doesn't occur on the bulk creator).
The fix for this was to check if an empty value for the `order_number` widget is in the `form.errors` when `form.clean()` is called, assign the value of `cleaned_data['order_number']` as the `quota_order_number` and then remove the error.

quotas/forms/wizards.py
```diff
@@ -196,114 +281,318 @@ class Meta:
...
-        )
+        self.request.session["staged_definition_data"] = staged_definitions
 
-    def init_layout(self):
+    def clean(self):
+        cleaned_data = super().clean()
+        for field in self.required_fields:
+            if field not in cleaned_data:
+                raise ValidationError(f"A value for {field} must be provided")
+        self.save_definition_data_to_session(cleaned_data)
```
mattjamc, line 410: On reflection I'm not too sure about this being called here. It gets called with cleaned_data regardless of whether there were any form errors. So if a user submits no end date or an invalid date it will still try and do the definition calculations on a None/invalid date and then break.

Maybe I'd propose wrapping line 410 in `if self.is_valid():`

quotas/jinja2/quota-definitions/create.jinja
```diff
@@ -1,12 +1,28 @@
...
+{% extends 'layouts/form.jinja' %}
 
 {% set page_title = "Create a new quota definition period" %}
-
+{% set page_subtitle = "Quota order number: "~ quota_order_number %}
 {% block breadcrumb %}
   {{ breadcrumbs(request, [
       {"text": "Find and edit quotas", "href": url("quota-ui-list")},
-      {"text": "Quota order number", "href": url("quota-ui-detail", args=[view.kwargs.sid]) ~ "#definitions"},
+      {"text": "Quota order number", "href": url("quota-ui-detail", args=[quota_sid]) ~ "#definitions-detail"},
```
mattjamc, line 8: Other way round I think `#definition-details`

mattjamc (approved):

PR #1324, opened 2024-11-13
TP2000 1570 & 1571 Update quota create layout and quota content changes

mattjamc:
# TP2000-1570 and TP2000-1571 Update quota create layout and quota content changes

## Why

To maintain consistency across TAP we should remove the accordions from the quota create journey.

We should also make other content changes to the Create quota associations journey for a better user experience.

## What

This PR:
- Removes accordions from the create quota journey and makes some minor content updates
- Makes content updates to the Create associations journey - renaming steps, updating help text and adding styling to the table
- 
image

image


CPrich905 (approved):
LGTM! :)

PR #1328, opened 2024-11-15
TP2000-1544 Prototype workflow template create & update views

dalecannon:
# TP2000-1544 Prototype workflow template create & update views

## Why
Workflow template create and update views are required for workflow template management.

## What
- Fixes an issue with queue item reordering (ensures items are filtered by their own queue)
- Hides UI reordering buttons in the case of a single item in a queue on workflow template detail view
- Adds support for creating and updating workflow templates in the UI through new views, URL routes and forms
- Adds unit tests for the two new views

## Checklist
- Requires migrations? No

## Screenshots
Screenshot 2024-11-15 at 17 01 41
Screenshot 2024-11-15 at 17 01 34


tasks/forms.py
```diff
@@ -81,6 +82,46 @@ def save(self, parent_task, user, commit=True):
...
+
+        self.helper = FormHelper(self)
+        self.helper.label_size = Size.SMALL
+        self.helper.legend_size = Size.SMALL
+        self.helper.layout = Layout(
+            "title",
+            "description",
+            Submit(
+                "submit",
+                "Save",
```
paulpepper-trade, line 110: Does this need to be overridden in the `TaskWorkflowTemplateCreateForm` subclass to read "Create"?

paulpepper-trade (approved):

PR #1332, opened 2024-11-18
Tp 2000 1489  delete subtask

marya-shariq:
# TP-??? Your PR title here


tasks/tests/test_views.py
```diff
@@ -144,6 +146,43 @@ def test_create_subtask_form_errors_when_parent_is_subtask(valid_user_client):
...
+    client_type,
+    expected_status_code_get,
+    expected_status_code_post,
+    request,
+):
+    """Tests that attempting to delete a subtask fails for users without the
+    necessary permissions."""
+    client_type = request.getfixturevalue(client_type)
+
+    # Creating an instance of a subtask that will be deleted
```
paulpepper-trade, line 166: The comments against each variable shouldn't be required. Unless there's some unavoidable, unclear rationale that requires an explanatory note, it's usually preferable that code is written to be understood without comments.

PR #1334, opened 2024-11-19
TP2000-1488 Edit Subtasks 

LaurenMullally:
#TP2000-1488 Edit Subtasks 

## Why
As part of the Task work, we need a custom set of urls, views and templates to edit a subtask. A lot of the functionality piggybacks off creating a task. 

## What
- Cleans up urls file to be consistent in it's patterns
- Adds urls for updating and confirm updating
- Adds views for both updating and confirm update. 
- Amends the templates to be dynamic so that both tasks and subtasks can share it. 
- Adds logic to the details template to either go to the subtask update url, or the task update url depending on what type of task it is. 
- Adds tests to check this logic. 

## Checklist
- Requires migrations? - no

### DIFF
--- a/tasks/tests/test_views.py
+++ b/tasks/tests/test_views.py
@@ -108,9 +108,9 @@ def test_create_subtask_button_shows_only_for_non_parent_tasks(
     page = BeautifulSoup(response.content.decode(response.charset), "html.parser")
 
     if expected_result:
-        assert page.find("a", href=f"/tasks/{task.pk}/subtasks/create")
+        assert page.find("a", href=f"/tasks/{task.pk}/subtasks/create/")
     else:
-        assert not page.find("a", href=f"/tasks/{task.pk}/subtasks/create")
+        assert not page.find("a", href=f"/tasks/{task.pk}/subtasks/create/")
 
 
 def test_create_subtask_form_errors_when_parent_is_subtask(valid_user_client):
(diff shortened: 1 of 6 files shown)


tasks/tests/test_views.py
```diff
@@ -81,6 +81,40 @@ def test_confirm_create_template_shows_task_or_subtask(
...
+    assert response.status_code == 200
+
+    page = BeautifulSoup(response.content.decode(response.charset), "html.parser")
+
+    if update_subtask_url:
+        assert page.find("a", href=f"/tasks/subtasks/{task.pk}/update/")
+        assert not page.find("a", href=f"/tasks/{task.pk}/update/")
+    else:
+        assert page.find("a", href=f"/tasks/{task.pk}/update/")
+        assert not page.find("a", href=f"/tasks/subtasks/{task.pk}/update/")
```
paulpepper-trade, line 115: The `href` value could be created via their URL pattern names, allowing URL path changes without impacting the tests. Perhaps they could also be passed into the test as part of the `parametrize`, avoiding the conditional altogether.

tasks/tests/test_views.py
```diff
@@ -108,9 +142,9 @@ def test_create_subtask_button_shows_only_for_non_parent_tasks(
     page = BeautifulSoup(response.content.decode(response.charset), "html.parser")
 
     if expected_result:
-        assert page.find("a", href=f"/tasks/{task.pk}/subtasks/create")
+        assert page.find("a", href=f"/tasks/{task.pk}/subtasks/create/")
```
paulpepper-trade, line 145: Ditto previous comment.

paulpepper-trade (approved):

PR #1335, opened 2024-11-21
TP2000-1584  Fix duty sentence parser error with MAX expression

dalecannon:
# TP2000-1584  Fix duty sentence parser error with MAX expression

## Why
Entering the duty sentence `9.10% MAX 1.00% + 0.90 GBP / 100 kg`  in the Commodities & Duties step of the Create Measures journey raises the following form error: `Duty expressions must be used in the duty sentence in ascending order of SID`. 

## What
- Removes the matched duty expression SID in addition to SIDs of a lower order when tracking applicable duty expressions, ensuring expressions are applied in ascending SID order:
    - For example, the duty expression SID order for the above parsed duty sentence was showing as 1, 17, 4; 
    - The new SID order would be applied as 1, 17, 19

- Refactors duty sentence parser related fixtures and tests catering for duty expression SID ordering
    - Modifies both `plus_percent_or_amount` and `maximum` fixtures to return a list of objects for each possible SID
    - Factors out a helper function, `replace_with_component_instances`, that turns test param data defining expected duty components into their corresponding object instances
    - Fixes tests that were giving false positives as a result of `with pytest.raises()` returning early once an exception was raised
    - Separates duty expression applicability code validation from duty expression validation test

Links to relevant material
See: [Duty expressions - Tariff data manual](https://uktrade.github.io/tariff-data-manual/documentation/data-structures/measure-components.html#duty-expressions)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1777,15 +1777,15 @@ def percent_or_amount() -> DutyExpression:
 
 
 @pytest.fixture
-def plus_percent_or_amount() -> DutyExpression:
-    return factories.DutyExpressionFactory(
-        sid=4,
-        prefix="+",
-        description=f"+ % or amount",
-        duty_amount_applicability_code=ApplicabilityCode.MANDATORY,
-        measurement_unit_applicability_code=ApplicabilityCode.PERMITTED,
-        monetary_unit_applicability_code=ApplicabilityCode.PERMITTED,
-    )
+def plus_percent_or_amount() -> list[DutyExpression]:
+    kwargs = {
+        "prefix": "+",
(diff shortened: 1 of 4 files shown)


conftest.py
```diff
@@ -1800,6 +1800,18 @@ def plus_agri_component() -> DutyExpression:
     )
 
 
+@pytest.fixture
+def maximum() -> list[DutyExpression]:
```
paulpepper-trade, line 1804: Perhaps name this fixture to indicate that it relates to a maxium in the context of duty expressions.

paulpepper-trade (approved):

PR #1336, opened 2024-11-21
TP2000-1581 Fixing timezone errors

mattjamc:
# TP2000-1581 Fixing timezone errors

## Why

When running tests there is a lot of noise and warnings which distract from the useful output. Lots of these warnings are timezone related. For example:

geo_areas/tests/test_views.py::test_geo_area_detail_version_control_view
  tamato/venv3.12/lib/python3.12/site-packages/django/db/models/fields/init.py:1595: RuntimeWarning: DateTimeField TrackedModelCheck.created_at received a naive datetime (1970-01-01 00:00:00) while time zone support is active.
    warnings.warn(

Django has time zone support enabled by default meaning it stores datetime information in UTC in the database and uses time-zone-aware datetime objects. These warnings arise when we assign naive datetimes or basic dates to datetime fields.

## What

This PR looks at where we are using naive datetime objects or simple dates and replaces them with timezone aware dates to reduce the warnings.

`make_aware()` is a Django utility function used to make datetimes timezone aware. It sets the timezone to that defined in the project settings - in this case 'Europe/London'.
- https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.timezone.make_aware

This allows more flexibility rather than hardcoding a timezone.

Pytest log output is down from 2000+ lines to ~1580.

DateTimeField errors are down from 161 to 3. The last 3 I can't pinpoint exactly but they seem to relate to migrations in the pytest setup and do not appear when pytest --no-migrations is run. Given that we shouldn't edit old migrations anyway I've left these for now.


paulpepper-trade (approved):
If you have any references or rationale for using `django.utils.timezone.make_aware()` to fix the warnings, then perhaps worth popping into the PR description.

A note on the need and correctness of swapping `date` objects for `datetime` may be useful too.

mattjamc:
> If you have any references or rationale for using `django.utils.timezone.make_aware()` to fix the warnings, then perhaps worth popping into the PR description.
> 
> A note on the need and correctness of swapping `date` objects for `datetime` may be useful too.

Sure, PR updated

PR #1337, opened 2024-11-21
TP2000-1590 TaskWorkflow summary info support

paulpepper-trade:
# TP2000-1590 TaskWorkflow summary info support

## Why

TaskWorkflow summary information is required to present a workflow's information in a Task-like way.

TaskWorkflow summary info must be presentable alongside Task info, in order to support at-a-glance list views that combine the two types of information.

## What

This PR:

* Adds a summary Task instance to the Workflow model, referenced by its `Workflow.summary_task` field.
* Includes migrations to create the new field in the database and populate with dummy data. The dummy data is only a consideration while this work is in development, since workflows will be newly introduced and migrations squashed before moving to the master, main line.
* Adds three new filter functions to `TaskQueryset`: `non_workflow()`, `workflow_summary()` and `top_level()`. These should cater for the need to extract the Task instances that provide summary information for various views in a follow-up PR.
* Adds and updates model factories and fixtures for use in related tests.
* Provides unit tests for new queryset filtering.

## Checklist
- Requires migrations? No


tasks/tests/test_models.py
```diff
@@ -78,6 +79,40 @@ def test_task_assignee_workbasket_reviewers_queryset(
...
+    assert Task.objects.count() == 3
+    assert task == non_workflow_tasks.get()
+
+
+def test_workflow_summary_task_queryset(task, task_workflow_single_task_item):
+    """Test correct behaviour of TaskQueryset.workflow_summary()."""
+
+    """Return a queryset of TaskWorkflow summary Task instances, i.e. those
+    with a non-null related_name=taskworkflow."""
+    # models.Q(taskworkflow__isnull=False),
```
dalecannon, line 96: Leftover comment and docstring?

tasks/tests/test_models.py
```diff
@@ -78,6 +79,40 @@ def test_task_assignee_workbasket_reviewers_queryset(
...
+
+    non_workflow_tasks = Task.objects.workflow_summary()
```
dalecannon, line 98: It looks like this variable is meant to represent workflow summary tasks & standalone tasks.

tasks/models/workflow.py
```diff
@@ -72,9 +59,29 @@ class TaskItem(QueueItem):
...
+    """
+    description = models.TextField(
+        blank=True,
+        help_text="Description of what this workflow template is used for. ",
+    )
+    """
+    Description of what the instance is used for.
+
+    This isn't the same as the description that may be applied to a TaskWorkflow
+    instance generated from a template.
```
dalecannon, line 82: Would a `TaskWorkflow` instance's title and description be derived from its summary task now that those fields are being removed from the model?

dalecannon (approved):

tasks/tests/test_models.py
paulpepper-trade: Removed.

tasks/tests/test_models.py
paulpepper-trade: Corrected var name.

PR #1338, opened 2024-11-25
Tp2000 1589  - temporary list views tile

marya-shariq:
# TP-1589 - TEMPORARY workflow list views tile
## Why

 * At the moment, there is no straight forward navigation to task workflow related pages for non-dev users
 * As task workflow functionality increases, there needs to be an easier way for non-dev users to find and view task workflow related pages

## What
 * A new TEMPORARY navigation tile has been added to the homepage which links to tasks list views
 * WorkflowTemplates list view is WIP so '#TODO' has been set as a placeholder

## Checklist
- Requires migrations? - No

### DIFF
--- a/common/jinja2/common/homepage.jinja
+++ b/common/jinja2/common/homepage.jinja
@@ -150,8 +150,8 @@
           <a href="{{ url('workflow:task-ui-list') }}"
             class="govuk-link"
           >Tasks</a>
-          </p>
-          <p class="govuk-body"></p>
+        </p>
+        <p class="govuk-body"></p>
           <a href=#TODO
             class="govuk-link"
           >Workflows</a>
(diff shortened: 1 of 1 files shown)


common/jinja2/common/homepage.jinja
```diff
@@ -144,6 +144,25 @@
...
     {% endif %}
 
+      <article class="masonry-card">
+        <h3 class="govuk-heading-m">Tasks and Workflows - TEST ONLY</h3>
+        <p class="govuk-body">
+          <a href="{{ url('workflow:task-ui-list') }}"
+            class="govuk-link"
+          >Tasks</a>
+          </p>
+          <p class="govuk-body"></p>
```
paulpepper-trade, line 154: A little indent oddness.

PR #1339, opened 2024-11-26
Update TaskWorkflowTemplate.create_task_workflow() method and unit test

dalecannon:
# Update TaskWorkflowTemplate.create_task_workflow() method

## Why
The [task-workflow megabranch](https://github.com/uktrade/tamato/pull/1298) has a failing test following changes to `TaskWorkflow` model: `tasks/tests/test_workflow_models.py::test_create_task_workflow_from_task_workflow_template`

## What
- Adds `summary_data` param to `TaskWorkflowTemplate.create_task_workflow()` to populate the details of the `TaskWorkflow.summary_task` instance (since it has become a required field), replacing the old way of using the workflow template's title and description
- Updates unit test based on the above change
- Fixes application of `tasks/migrations/0012_taskworkflow_assign_summary_task.py` by saving the modified object instances in the forward migration operation

## Checklist
- Requires migrations? Yes

### DIFF
--- a/tasks/models/workflow.py
+++ b/tasks/models/workflow.py
@@ -91,13 +91,17 @@ class TaskWorkflowTemplate(Queue):
         )
 
     @atomic
-    def create_task_workflow(self) -> "TaskWorkflow":
-        """Create a workflow and it subtasks, using values from this template
-        workflow and its task templates."""
+    def create_task_workflow(self, summary_data: dict) -> "TaskWorkflow":
+        """
+        Create a workflow and it subtasks, using values from this template
+        workflow and its task templates.
 
+        The `summary_data` param is used to populate the details of the `TaskWorkflow.summary_task` instance.
+        """
+
+        summary_task = Task.objects.create(**summary_data)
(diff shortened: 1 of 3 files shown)


tasks/models/workflow.py
```diff
@@ -91,13 +91,17 @@ def get_task_templates(self) -> models.QuerySet:
         )
 
     @atomic
-    def create_task_workflow(self) -> "TaskWorkflow":
-        """Create a workflow and it subtasks, using values from this template
-        workflow and its task templates."""
+    def create_task_workflow(self, summary_data: dict) -> "TaskWorkflow":
```
paulpepper-trade, line 94: Might it be preferable, better documenting this API, to be explicit stating which fields are required or expected to create a `TaskWorkflow`, with `TaskWorkflow.summary_task` being kept as an internal detail?
```suggestion
    def create_task_workflow(self, title: str, description: str) -> "TaskWorkflow":
```

paulpepper-trade (approved):

PR #1343, opened 2024-11-27
TP2000-1582  Enable customisable queue field name on QueueItem subclasses

dalecannon:
# TP2000-1582  Enable customisable queue field name on QueueItem subclasses

## Why
By subclassing `QueueItem`, `TaskItem` and `TaskItemTemplate` must define a queue ForeignKey field. It would be preferable to allow the field to be named some other value that better reflects the purpose of the queue.

## What
- Adds customisable `queue_field` attribute to concrete subclasses
- Updates `QueueItemMetaClass` validation ensuring a queue field has been defined properly
- Adds `QueueItemMetaClass` validation unit tests


tasks/models/queue.py
```diff
@@ -107,6 +139,10 @@ class Meta:
         abstract = True
         ordering = ["queue", "position"]
 
+    queue_field = "queue"
```
paulpepper-trade, line 142:
```suggestion
    queue_field: str = "queue"
```

tasks/models/queue.py
```diff
@@ -107,6 +139,10 @@ class Meta:
...
+    """The name of the ForeignKey field relating this instance to a Queue
+    instance."""
```
paulpepper-trade, line 144: Perhaps offer a little more info about how to use this class-level attribute? For instance, that it can be kept named as is, `queue`, or overriden in subclasses to better reflect the purpose and use of the queue in those subclasses.

tasks/models/workflow.py
```diff
@@ -42,6 +42,8 @@ class TaskItem(QueueItem):
     """Task item queue management for Task instances (these should always be
     subtasks)."""
 
+    queue_field = "queue"
```
paulpepper-trade, line 45: Would it now make sense to name this something like `task_workflow` or just `workflow`?
```suggestion
    queue_field = "workflow"
```
and then
```
    queue = models.ForeignKey(
        TaskWorkflow,
        related_name="queue_items",
        on_delete=models.CASCADE,
    )
```
would become
```
    workflow = models.ForeignKey(
        TaskWorkflow,
        related_name="workflow_items",
        on_delete=models.CASCADE,
    )
```
It'd break a whole load of code, so could be pulled out into a separate PR.

tasks/models/workflow.py
```diff
@@ -142,6 +144,8 @@ def get_url(self, action: str = "detail"):
 class TaskItemTemplate(QueueItem):
     """Queue item management for TaskTemplate instances."""
 
+    queue_field = "queue"
```
paulpepper-trade, line 147: Ditto [previous comment](https://github.com/uktrade/tamato/pull/1343/files#r1862503490).

tasks/models/workflow.py
```diff
@@ -191,9 +195,11 @@ def get_url(self, action: str = "detail"):
 class TaskItemTemplate(QueueItem):
     """Queue item management for TaskTemplate instances."""
 
-    queue = models.ForeignKey(
+    queue_field = "workflow"
+
+    workflow = models.ForeignKey(
```
paulpepper-trade, line 200: I'm not entirely sure, would `workflow_template` be preferable as the queue name?

paulpepper-trade: If you felt it was preferable, then `related_name` should probably become `worflow_template_items`.

tasks/tests/test_queue_models.py
```diff
@@ -59,6 +63,62 @@ def three_item_queue(queue) -> TestQueue:
     return queue
 
 
+def test_queueitem_metaclass_validation_missing_queue_field():
+    """Tests that a concrete sublass of `QueueItem` must provide a queue field
+    on the model."""
+    with pytest.raises(RequiredFieldError) as error:
+
+        class TestQueueItemSubclass(QueueItem):
```
paulpepper-trade, line 71: Wouldn't the expected behaviour be to have an inherited and working `QueueItem.queue` ForeignKey field if `QueueItem.queue_field` remains unchanged as I think it does here? Am I missing something?

paulpepper-trade (approved):

PR #1344, opened 2024-11-28
TP2000-1600 Add Subtask filter to Task admin view

LaurenMullally:
#TP2000-1600 Add Subtask filter to Task admin view

## Why
This PR adds some helpful details to identify subtasks amongst a list of tasks. As Tasks and Subtasks are the same model, creating an admin view for Subtasks will take a bit more thought around how it's done. In the interim, I've added a couple of changes to the Task Admin view to help identify and filter Subtasks from Tasks. 

## What
- Add an `is_subtask` column to the Task details list. 
- Add a filter to the Task list to filter Tasks and Subtasks 

## Checklist
- Requires migrations? - no 


tasks/admin.py
```diff
@@ -29,10 +29,29 @@ def link_to_workbasket(self, workbasket):
...
+    def lookups(self, request, model_admin):
+        return (
+            ("TRUE", "Subtasks"),
+            ("FALSE", "Tasks"),
+        )
+
+    def queryset(self, request, queryset):
+        value = self.value()
+        if value == "TRUE":
+            return queryset.exclude(parent_task=None)
```
paulpepper-trade, line 45: Could we extend our `TaskQueryset` to include these filters. Maybe something like `TaskQueryset.parents()` and `TaskQueryset.subtasks()`. We'll need to decide how we're categorising those types of `Task` and whether or how we want to distinguish them from summary tasks on workflows.

paulpepper-trade (approved):

paulpepper-trade (approved):

PR #1348, opened 2024-12-05
Tp2000 1592 workflow template list view

marya-shariq:
# TP-2000-1592 - TaskWorkflowTemplate List View

## Why
 * As we continue to build our TAP task management functionality, users can create and view their own workflow templates
 * This PR provides a list view for all of the list view templates which provides summary information such as the title, description, and creator of the template
 * Users can search for and filter templates via their title, description, or display name of the creator

## What
 * Adds new view and URL route for task workflow templates list view
 * Added 'creator' field to WorkflowTemplate model to store information about users and created migration file
 * Adds new 'TaskWorkflowTemplate' filter for text searches through 'title', 'description', and 'creator' fields
 * Added 'list' URL for WorkflowTemplate class' 'get_url' method

Optionally let reviewers know they need to run migrations or update dependencies before
testing by adding the following section:
## Checklist
- Requires migrations? - YES 


tasks/forms.py
```diff
@@ -196,8 +196,16 @@ def __init__(self, *args, submit_title, **kwargs):
 
 class TaskWorkflowTemplateCreateForm(TaskWorkflowTemplateBaseForm):
     def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop("request", None)
```
dalecannon, line 199: This line can be removed if `request` isn't being used in the form.

tasks/jinja2/tasks/workflows/create.jinja
```diff
@@ -8,7 +8,7 @@
   {{ breadcrumbs(
     request,
     [
-      {"text": "Find and view workflow templates", "href": "#TODO"},
+      {"text": "Find and view workflow templates", "href": url("workflow:task-workflow-template-ui-list")},
```
dalecannon, line 11: I think since both `TaskWorkflowCreateView` and `TaskWorkflowTemplateCreateView` use this template, the breadcrumb's title and URL could be made to handle both list views by setting context variables and doing something like:

`{"text": "Find and view "~ verbose_name ~ "s", "href": list_url},`

tasks/migrations/0015_taskworkflowtemplate_creator.py
```diff
@@ -0,0 +1,27 @@
...
+from django.db import migrations
+from django.db import models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ("tasks", "0014_remove_taskworkflow_description_and_more"),
+    ]
```
dalecannon, line 14: It looks like this migration file (and its dependency) clashes with `0015_alter_taskworkflow_options_and_more.py` and might need to be renamed or recreated.

tasks/models/workflow.py
```diff
@@ -64,6 +65,10 @@ def get_url(self, action: str = "detail"):
             return reverse(
                 "workflow:task-workflow-ui-create",
             )
+        elif action == "list":
+            return reverse(
+                "workflow:task-workflow-template-ui-list",
+            )
```
dalecannon, line 71: Is this meant to be added to the `TaskWorkflowTemplate` model?

tasks/views.py
```diff
@@ -452,7 +465,7 @@ def get_context_data(self, **kwargs):
                 "verbose_name": "workflow",
                 "deleted_pk": self.kwargs["pk"],
                 "create_url": reverse("workflow:task-workflow-ui-create"),
-                "list_url": "#NOT-IMPLEMENTED",
+                "list_url": reverse("workflow:task-workflow-template-ui-list"),
```
dalecannon, line 468: Should this remain as it was until the workflow list view is implemented?

tasks/views.py
```diff
@@ -501,11 +514,6 @@ class TaskWorkflowTemplateCreateView(PermissionRequiredMixin, CreateView):
     template_name = "tasks/workflows/create.jinja"
     form_class = TaskWorkflowTemplateCreateForm
 
-    def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
-        context["verbose_name"] = "workflow template"
-        return context
-
```
dalecannon, line 508: Is `verbose_name` still needed in `workflows/create.jinja` to differentiate between workflow templates and workflows?

tasks/models/workflow.py
marya-shariq: Yes, Paul and I discussed this earlier :)

tasks/views.py
marya-shariq: Yes, I think you're right! I accidentally added it in there whilst searching for other instances of 'list_url'. Thanks for spotting that, Dale :)

dalecannon (approved):

PR #1352, opened 2024-12-09
TP2000-1154 Transaction reordering bug

mattjamc:
# TP2000-1154 Transaction reordering bug

### DIFF
--- a/common/models/transactions.py
+++ b/common/models/transactions.py
@@ -279,6 +279,7 @@ class Transaction(TimestampedMixin):
     # The default order ("partition", "order") gives global ordering.
     order = models.IntegerField()
 
+    # Originally containing order, partition and workbasket pk / envelope id. Now a random unique string to avoid duplicate values when reordering / deleting transactions.
     composite_key = models.CharField(max_length=16, unique=True)
 
     objects = TransactionManager.from_queryset(TransactionQueryset)()
--- a/taric_parsers/importer.py
+++ b/taric_parsers/importer.py
@@ -1,3 +1,4 @@
+from os import urandom
 from typing import Generator
 from typing import List
 from typing import Optional
@@ -344,14 +345,14 @@ class TaricImporter:
             None
         """
 
-        envelope = Envelope.new_envelope()
+        Envelope.new_envelope()
 
         transaction_order = 1
 
         for parsed_transaction in self.parsed_transactions:
             # create transaction
             transaction_inst = Transaction.objects.create(
-                composite_key=f"{envelope.envelope_id}{transaction_order}",
+                composite_key=urandom(8).hex(),
                 workbasket=self.workbasket,
                 order=transaction_order,
             )
--- a/workbaskets/models.py
+++ b/workbaskets/models.py
@@ -4,6 +4,7 @@ import importlib
 import logging
 from abc import ABCMeta
 from abc import abstractmethod
+from os import urandom
 from typing import Optional
 from typing import Tuple
 
@@ -557,9 +558,8 @@ class WorkBasket(TimestampedMixin):
             )
 
         if "composite_key" not in kwargs:
-            kwargs["composite_key"] = (
-                f"{self.pk}-{kwargs['order']}-{kwargs['partition']}"
-            )
+            # Generating a random 16 digit hexadecimal composite key to avoid duplicates
+            kwargs["composite_key"] = urandom(8).hex()
 
         # Get Transaction model via transactions.model to avoid circular import.
         return self.transactions.model.objects.create(workbasket=self, **kwargs)


taric_parsers/importer.py
```diff
@@ -344,14 +345,14 @@ def commit_data(self):
             None
         """
 
-        envelope = Envelope.new_envelope()
+        Envelope.new_envelope()
```
mattjamc, line 348: Not sure if this is even needed now

common/models/transactions.py
```diff
@@ -279,6 +279,7 @@ class Meta:
     # The default order ("partition", "order") gives global ordering.
     order = models.IntegerField()
 
+    # Originally containing order, partition and workbasket pk / envelope id. Now a random unique string to avoid duplicate values when reordering / deleting transactions.
     composite_key = models.CharField(max_length=16, unique=True)
```
paulpepper-trade, line 283: Maybe document this as a docstring (using triple quotes directly beneath the attribute) so that it forms part of the class's docs.

paulpepper-trade: Where and how `composite_key` is used has confounded us, so perhaps mention what's been uncovered about that as part of the docstring against this attribute.

workbaskets/models.py
```diff
@@ -557,9 +558,8 @@ def new_transaction(self, **kwargs):
             )
 
         if "composite_key" not in kwargs:
-            kwargs["composite_key"] = (
-                f"{self.pk}-{kwargs['order']}-{kwargs['partition']}"
-            )
+            # Generating a random 16 digit hexadecimal composite key to avoid duplicates
```
paulpepper-trade, line 561: Perhaps worth mentioning the probably of a clash.

workbaskets/models.py
```diff
@@ -557,9 +558,8 @@ def new_transaction(self, **kwargs):
...
+            kwargs["composite_key"] = urandom(8).hex()
```
paulpepper-trade, line 562: Are we also including the transaction's workbasket PK as a prefix in the composite_key? (Of course it would need some care to choose a reasonable future-proof width when converting the `int` to a zero padded `str`.)

common/models/transactions.py
mattjamc: I can stick it in triple quotes underneath the attribute but this won't be accessible in any way will it? As opposed to the class docstring where you can call Transaction.__doc__

paulpepper-trade: Documentation generation tools will generally pick up docstrings (triple quote convention) but not comments (hash convention). I have a vague memory that the TAP documentation generation capability does pick them up, but I can't find that right just now. Still, if we feel attribute would benefit from documenting in this way, then I don't think there's a downside to presenting that documentation within a docstring.

workbaskets/models.py
mattjamc: Yes that's now done `'{workbasket.pk}{random-12-digit-string}'`. I have not separated them with any delimiter so as not to lose another character and decrease uniqueness further.

workbaskets/models.py
```diff
@@ -557,9 +558,12 @@ def new_transaction(self, **kwargs):
             )
 
         if "composite_key" not in kwargs:
-            kwargs["composite_key"] = (
-                f"{self.pk}-{kwargs['order']}-{kwargs['partition']}"
-            )
+            # Forms a composite key of workbasket id and a randomly generated hexadecimal 12 character string.
+            # The probability of a collision is miniscule unless workbasket transactions are in the millions. It becomes likely (>50%) around 20 million.
+            composite_key = f"{self.pk}{urandom(6).hex()}"
```
paulpepper-trade, line 563: `WorkBasket.pk` is an [AutoField](https://docs.djangoproject.com/en/dev/ref/models/fields/#autofield) that returns an `int` type in Python, so it will vary in width when stringified.

To create a `composite_key` of consistent width for tests, workbaskets created in the future and TAP used in unexpected (to us right now) ways in the future, the stringified `WorkBasket.pk` portion of `composite_key` should be set to a zero-padded fixed width.

That means deciding upon a reasonable range of values that `WorkBasket.pk` can take. Values from `1` (for Postgresql) up to the 10s of thousands (`99,999`) might be a reasonable range, giving a width of five characters for the pk portion of `composite_key`. If  `99,999` were choosen as the max value of the pk portion (i.e. `WorkBasket.pk` exceeds 99,999), then, at a minimum, this function should emit a warning log.

workbaskets/models.py
```diff
@@ -557,9 +558,15 @@ def new_transaction(self, **kwargs):
...
-            kwargs["composite_key"] = (
-                f"{self.pk}-{kwargs['order']}-{kwargs['partition']}"
-            )
+            # Forms a composite key of a zero-padded 5 digit workbasket id + a randomly generated hexadecimal 11 character string.
+            # The probability of a collision is miniscule. It becomes 1% around 600,000 transactions and likely (>50%) at around 5 million.
+            workbasket_pk = str(self.pk).zfill(5)
+            if len(workbasket_pk) > 5:
+                raise ValueError(
+                    "Workbasket PK cannot be bigger than 5 digits for composite_key generation.",
+                )
```
paulpepper-trade, line 567: It might read better to directly apply the length test against `str(self.pk)` as it's its length that we're really interesting in checking.
```suggestion
            if len(str(self.pk)) > 5:
                # Take warning or error action.
            workbasket_pk = str(self.pk).zfill(5)
```

PR #1356, opened 2024-12-10
Tp2000 1594 ordering of workflow templates list view

marya-shariq:
# TP-2000-1594 - Ordering of WorkflowTemplates List View

## Why
 * This PR provides more summary information (date created and date last updated) to the workflow templates list view 
 * Task workflow templates can now be sorted by date created and/or date updated
 * Users can now order workflow templates by the date created or the date that a template was last updated
 
## What
 * Added datetime support for Task Workflow Template model using TimeStampMixin 
 * Added 'Date created' and 'Last updated' columns to list view table
 * Added sortable anchor to 'Date created' and 'Last updated' so templates can be ordered in both descending and ascending order

Optionally let reviewers know they need to run migrations or update dependencies before
testing by adding the following section:
## Checklist
- Requires migrations? - YES

Workflow Templates sorted by 'date created' and ordered in ascending order
Screenshot 2024-12-10 at 14 24 39

Workflow Templates sorted by 'date created' and ordered in descending order
Screenshot 2024-12-10 at 14 24 44


tasks/views.py
```diff
@@ -311,15 +311,24 @@ def get_queryset(self):
         return queryset
 
 
-class TaskWorkflowTemplateListView(PermissionRequiredMixin, WithPaginationListView):
+class TaskWorkflowTemplateListView(
+    PermissionRequiredMixin,
+    SortingMixin,
```
paulpepper-trade, line 316: Sorting appears to override and lose any filtering that has been set in the view. I expect might therefore need to intergrate or in some way coordinate the side-effects of `SortingMixin` with with those of `FilterSet` / `WorkflowTemplateFilter`.

paulpepper-trade (approved):

PR #1357, opened 2024-12-10
TP2000-1615 Various errors

paulpepper-trade:
# TP2000-1615 Various errors

## Why

A few errors found during manual testing:

- Tasks without workbaskets raise exceptions in Task list view.
- Tasks without a creator raise exception in Task detail view.
- Task and workflow ordering is inconsistent in list views with a default listing order that is neither by Id or date created.

## What

Fix errors listed in the what section, above.

## Checklist
- Requires migrations? No

### DIFF
--- a/tasks/jinja2/tasks/detail.jinja
+++ b/tasks/jinja2/tasks/detail.jinja
@@ -99,7 +99,7 @@
       Created by
     </dt>
     <dd class="govuk-summary-list__value">
-      {{ object.creator.get_displayname() }}
+      {{ object.creator.get_displayname() if object.creator else "-" }}
     </dd>
   </div>
 </dl>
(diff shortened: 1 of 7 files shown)


tasks/jinja2/tasks/detail.jinja
```diff
@@ -99,7 +99,7 @@
       Created by
     </dt>
     <dd class="govuk-summary-list__value">
-      {{ object.creator.get_displayname() }}
+      {{ object.creator.get_displayname() if object.creator else "-" }}
```
marya-shariq, line 102: Is "-" the default placeholder value for a lack of user display name across the TAP repo?  I have used 'Unknown' in the workflow templates list view, but I'm happy to change that for consistency.

paulpepper-trade: I'm not sure it's a default across the whole of Tamato, and I had used unknown, but changed to "-" for consistency within the tasks application.

marya-shariq: Sounds good!

marya-shariq (approved):

PR #1359, opened 2024-12-13
TP2000-1478: Missing measures check

eadpearce:
# TP2000-1478: Missing measures check

## Why

* To reduce errors in workbaskets by adding a check to ensure every commodity code has at minimum a type 103 measure associated (Most Favoured Nation or MFN)

## What

* Adds new tab layout design for workbasket checks
* New page where users can run the missing measures check and see results
* New database models `MissingMeasuresCheck` and `MissingMeasureCommCode` added
* Extra field added to `WorkBasket` model - `missing_measures_check_task_id`. Functions the same as `rule_check_task_id`
* The check blocks the publishing of the workbasket if it doesn't pass, like the business rules check
* The check runs asynchronously in a celery task
* The check can be stopped and re-run

## Still to do
- [x] Add green and red boxes to match latest design and business rules check page

## Checklist
- Requires dependency updates? **No**

Screenshot 2024-12-19 at 12 13 24
Screenshot 2024-12-19 at 12 13 37
Screenshot 2024-12-19 at 12 14 31


common/static/common/scss/_utils.scss
```diff
@@ -6,4 +6,13 @@
 
 .background-light-grey {
     background: govuk-colour("light-grey");
+}
+
+.align-right {
```
paulpepper-trade, line 11: Will this conflict with [an existing CSS declaration](https://github.com/uktrade/tamato/blob/26e153d8b163302c474f4a6329a5ed78683a4786/common/static/common/scss/_layout.scss#L65) with the same class spec?

workbaskets/forms.py
```diff
@@ -425,3 +425,7 @@ def __init__(self, *args, **kwargs):
                 css_class="govuk-button-group",
             ),
         )
+
+
+class WorkbasketMeasuresCheckForm(forms.Form):
+    pass
```
paulpepper-trade, line 431: Is this empty Form subclass only defined to provide a mandatory `form` attribute on the `WorkBasketCommCodeChecks` form view or is it needed for somthing else?

paulpepper-trade (commented):
Some screen shots of views and forms in the PR cover note could be useful.

common/static/common/scss/_utils.scss
eadpearce: Oh that's annoying... I'll rename this class

workbaskets/forms.py
eadpearce: Yeah pretty much. There might be a better way to do this

paulpepper-trade: Ah, makes sense. I suppose it might be possible to use `django.forms.Form` directly to set `WorkBasketCommCodeChecks.form_class`. A small reduction in redundancy and code if so.

paulpepper-trade: Sounds like a good way of documenting purpose and reducing duplication.

checks/models.py
```diff
@@ -156,3 +156,32 @@ def rule_code(self):
...
+
+    commodity = models.ForeignKey(
+        "commodities.GoodsNomenclature",
+        on_delete=models.CASCADE,
+    )
+    missing_measures_check = models.ForeignKey(
+        MissingMeasuresCheck,
+        on_delete=models.CASCADE,
+        related_name="model_checks",
+    )
```
paulpepper-trade, line 187: This seems to take a different approach to business rule checks. 

Business rule checks have a matching `TrackedModelCheck.successful=True` and new timestamp for each WorkBasket `TrackedModel` when all business rules have passed. But if there's an incomplete set of `TrackedModelCheck`s, or one or more with a `successful` attribute set to `False`, or timestamp older than `TrackedModel`updated_at` , then business rules are considered to have failed.

Goods are unlikely to change in a workbasket. But measures may do, which would make a clean `MissingMeasuresCheck` instance invalid. Is staleness and check completeness validation required or supported in this implementation?

Either way, a docstring would be useful, perhaps in `MissingMeasuresCheck`, to say a little about the checking strategy because of the differing approaches.

paulpepper-trade (approved):

PR #1363, opened 2024-12-18
Tp2000 1616  workflow task create view

marya-shariq:
# TP2000-1616 : Workflow Task Create View
## Why
 * As part of the task workflow management journey, users should be able to create tasks for existing workflows
 * Users can create both standalone tasks and tasks as part of workflows, thus giving them greater flexibility for task management and organisation

## What
 * Two URLs for task workflow task create and confirm create 
 * Two views  to serve the above URLs:`TaskWorkflowTaskCreate` and `TaskWorkflowTaskConfirmCreate`
 * Both views use existing Task create form and confirm create success page templates
 * A `TaskItem` instance is created for each Task upon completion to link it to its corresponding workflow and manage queue position
 * Two tests - `test_create_workflow_task_view` and `test_workflow_delete_view_deletes_related_tasks`
 
## Checklist
- Requires migrations? - No


tasks/forms.py
```diff
@@ -25,6 +25,13 @@
 class TaskBaseForm(ModelForm):
     class Meta:
         model = Task
+        include = [
```
paulpepper-trade, line 28: There's no `include` attribute on `ModelForm` instances. I suspect this should be `fields` ([Selecting the fields to use](https://docs.djangoproject.com/en/4.2/topics/forms/modelforms/#selecting-the-fields-to-use))?

If that's so, and that is Django's suggested preferred approach to selecting model fields in this type of form, then the `exclude` attribute should be removed.

tasks/views.py
```diff
@@ -553,6 +553,53 @@ def get_context_data(self, **kwargs):
...
+        """Get the associated TaskWorkflow via its pk in the URL."""
+        return TaskWorkflow.objects.get(
+            pk=self.kwargs["task_workflow_pk"],
+        )
+
+    def get_context_data(self, **kwargs) -> dict:
+        context = super().get_context_data(**kwargs)
+
+        context["page_title"] = "Create a task"
+        # context["task_workflow"] = self.get_task_workflow()
```
paulpepper-trade, line 572: Redundant comment?

tasks/views.py
```diff
@@ -553,6 +553,53 @@ def get_context_data(self, **kwargs):
...
+
+class TaskWorkflowTaskConfirmCreateView(PermissionRequiredMixin, DetailView):
+    model = Task
+    template_name = "tasks/confirm_create.jinja"
+    permission_required = "tasks.add_task"
+
+    def get_context_data(self, **kwargs) -> dict:
+        context = super().get_context_data(**kwargs)
+        context["verbose_name"] = "task"
+        # context["task_workflow"] = self.get_object().taskitem.workflow
```
paulpepper-trade, line 599: Redundant comment?

tasks/views.py
marya-shariq: Thanks for catching that!

tasks/urls.py
```diff
@@ -111,6 +111,16 @@
         views.TaskWorkflowConfirmDeleteView.as_view(),
         name="task-workflow-ui-confirm-delete",
     ),
+    path(
+        "<int:task_workflow_pk>/task/create/",
+        views.TaskWorkflowTaskCreateView.as_view(),
+        name="task-workflow-task-ui-create",
+    ),
+    path(
+        "task/confirm-create/<int:pk>",
```
paulpepper-trade, line 120: For URL pattern consistency with other confirm-create views (for instance, subtask creation uses `"subtasks/<int:pk>/confirm-create/",`):
```suggestion
        "task/<int:pk>/confirm-create/",
```

tasks/views.py
```diff
@@ -553,6 +553,51 @@ def get_context_data(self, **kwargs):
...
+        return context
+
+    def form_valid(self, form) -> HttpResponseRedirect:
+        with transaction.atomic():
+            self.object = form.save(user=self.request.user)
+            TaskItem.objects.create(
+                workflow=self.get_task_workflow(),
+                task=self.object,
+            )
+        return HttpResponseRedirect(self.get_success_url(), self.object.pk)
```
paulpepper-trade, line 581: Is the second parameter to the `HttpResponseRedirect` constructor correct or needed?

tasks/tests/test_views.py
```diff
@@ -815,3 +815,78 @@ def test_task_and_workflow_list_view(valid_user_client, task, task_workflow):
...
+        kwargs={"task_workflow_pk": task_workflow.pk},
+    )
+
+    form_data = {
+        "title": factory.Faker("sentence"),
+        "description": factory.Faker("sentence"),
+        "progress_state": progress_state.pk,
+    }
+    create_response = valid_user_client.post(create_url, form_data)
+    created_workflow_task = task_workflow.get_tasks().get()
```
paulpepper-trade, line 839: There's an assert that checks for a single task count further down, so if this blows up because there isn't exactly one task instance, then that assert will never be applied. The assertion could be performed against task count before getting the (single) task.

Checking the response's status code straight after the call to `post()` would also usefully catch failures in a more descriptive way, rather than, say, the `reverse()` failing in a less straightforward way.

tasks/tests/test_views.py
```diff
@@ -815,3 +815,78 @@ def test_task_and_workflow_list_view(valid_user_client, task, task_workflow):
...
+
+    assert create_response.status_code == 302
+    assert task_workflow.get_tasks().count() == 1
+    assert create_response.url == confirmation_url
+
+    confirmation_response = valid_user_client.get(confirmation_url)
+
+    soup = BeautifulSoup(str(confirmation_response.content), "html.parser")
+
+    assert confirmation_response.status_code == 200
```
paulpepper-trade, line 853: Similar to previous comment, before attempting to access the response's content, the `assert` should first perform its validation checking.

Placement of `asserts` in the other test, `test_workflow_delete_view_deletes_related_tasks`, is better because it places them close to the side-effect or invariance they're validating.

marya-shariq: Thank you for taking the time to explain, Paul!

paulpepper-trade (approved):

PR #1368, opened 2024-12-23
TP2000-1503-speed-up-auto-end-measures

mattjamc:
# TP2000-1503-speed-up-auto-end-measures

## Why

User testing of the auto end-dating measures in UAT found that the page times out with a 504 error. This does not happen locally but it can take about 15 seconds to load.

## What

This PR moves the .effective_valid_between() part of the function to the end so that it is applied on the final smallest queryset. As this is the most time consuming bit of the query it reduces the time to load the page.


CPrich905 (approved):
DWISOTT

PR #1370, opened 2024-12-24
Tp2000 1640 new ref doc data importer

dougmills-DIT:
# TP2000-1640 Create new reference documents data importer
 * Updates to reference document alignment checks 
 * Create web based data importer for reference documents using new  format 
 * improve UI to display passes and failures 
 * divide up results based on year for display

## Why
 * Previous import process was done via command line and was not accessible by most users
 * Some alignment checks did not operate as expected
 * Review of alignment check was was not easy to navigate

## What
 * Updates to reference document alignment checks 
 * Create web based data importer for reference documents using new  format 
 * improve UI to display passes and failures 
 * divide up results based on year for display

## Checklist
- Requires migrations? yes

Links to relevant material
See: [ticket](https://uktrade.atlassian.net/browse/TP2000-1604)


reference_documents/tests/test_csv_upload_forms.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.mark.reference_documents
+class TestReferenceDocumentCreateCsvUploadForm:
+    pass
```
Luisella21, line 8: Do we need this file?

reference_documents/check/base.py
```diff
@@ -710,6 +815,7 @@ def tap_recursive_comm_code_check(
         Args:
             snapshot: CommodityTreeSnapshot, A snapshot from self.ref_rate.comm_code and children
             parent_item_id: str, The parent comm code item_id to check
+            parent_item_suffix: int, The parent comm code suffix to check
```
Luisella21, line 818: The parent_item_suffix is a str, not an int.

reference_documents/csv_importer/importer.py
```diff
@@ -0,0 +1,468 @@
...
+    def find_reference_document_version(
+        self,
+        reference_document_version: float,
+        reference_document: ReferenceDocument,
+        status=None,
+    ):
+        """
+
+        Args:
+            reference_document_version: the version of the reference document as a string
```
Luisella21, line 100: reference_document_version is a float in the function definition.

reference_documents/check/ref_rates.py
```diff
@@ -18,7 +18,7 @@ def run_check(self):
         """
         # comm code live on EIF date
         if not self.tap_comm_code():
-            message = f"{self.ref_rate.commodity_code} {self.tap_geo_area_description()} comm code not live"
+            message = f"Rate {self.ref_rate.commodity_code} {self.ref_rate.valid_between}: commodity code not found for period."
```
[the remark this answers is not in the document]
dougmills-DIT: yes this is correct, the error message occurs when a preferential rate associated with a commodity code check fails because there is no commodity codes spanning a validity period for the rate so knowing the date range that it was not covered for and the commodity code, we can investigate TAP data. Before there was not enough information to check from the error

reference_documents/models.py
```diff
@@ -887,48 +929,107 @@ class AlignmentReportCheck(TimestampedMixin):
 
     ref_quota_definition = models.ForeignKey(
         "reference_documents.RefQuotaDefinition",
-        on_delete=models.PROTECT,
+        on_delete=models.CASCADE,
```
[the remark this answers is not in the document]
dougmills-DIT: yes, this was an intentional change, if we delete a ref doc quota definition we want the deletion to cascade to related records including the alignment checks, otherwise when deleting the record it give an error. This is not TAP data with all that tracked model around it - it's just plain old ORM

reference_documents/tests/test_csv_upload_forms.py
dougmills-DIT: no - removed

reference_documents/check/base.py
dougmills-DIT: good spot, updated

reference_documents/csv_importer/importer.py
dougmills-DIT: good spot - updated

Luisella21 (approved):
LGTM

PR #1371, opened 2024-12-30
Updates to QA5 to prevent False positives from deleted associations

dougmills-DIT:
# TOPS-1874 
 * Add pytest mark to all business rules for easier testing business rules in isolation 
 * Add test to check deleted associations are not being incorrectly checked
 * Update QA5 

## Why
 * The rule incorrectly raised rule violations 
 * No test checked that deleted associations were not being checked

## What
 * Update QA5 
 * Test to verify the rule is operating correctly

## Checklist
- Requires migrations? No

Links to relevant material
See: [Description](https://uktrade.atlassian.net/browse/TOPS-1874)

### DIFF
--- a/quotas/business_rules.py
+++ b/quotas/business_rules.py
@@ -36,7 +36,7 @@ class PreventDeletingLinkedQuotaDefinitions(BusinessRule):
         if quota_definition.update_type == UpdateType.DELETE:
             kwargs = {f"{self.sid_prefix}sid": quota_definition.sid}
             if related_model.objects.approved_up_to_transaction(
-                    transaction=quota_definition.transaction,
+                transaction=quota_definition.transaction,
             ).filter(**kwargs):
                 raise self.violation(quota_definition)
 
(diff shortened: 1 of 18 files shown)


dougmills-DIT:
the only file that has changed that will have an effect in the running application is quotas/business_rules.py the rest are test updates

quotas/business_rules.py
```diff
@@ -471,13 +473,16 @@ def validate(self, association):
...
-                .count()
-                > 1
-            ):
+
+            quota_associations = QuotaAssociation.objects.approved_up_to_transaction(
+                association.transaction
+            ).filter(main_quota=association.main_quota)
+            volumes = []
+            for association in quota_associations:
+                if association.sub_quota.volume not in volumes:
```
mattjamc, line 482: Looks good! Just a question about this business rule - is it the volume or the initial_volume that need to be the same? Depending on the answer we may need to change how we validate the form as we updated that to compare initial_volumes

CPrich905 (approved):
👍

PR #1375, opened 2025-01-07
TP2000-1639 new year envelope publishing failure

paulpepper-trade:
# TP2000-1639 new year envelope publishing failure

## Why

Tamato is failing to publish new envelopes, beginning with this new year’s 25xxxx envelope ID naming pattern, to the tariffs-api service. This is because the last envelope of 2024 was rejected but was still being considered as the last published envelope.

## What

This PR:

- Makes a bare call to Tamato's `publish_to_api` Django management benign. Instead of performing a publishing action it now lists (with additional detail) the envelopes that are awaiting publishing to tariffs-api.
- Adds the `--publish-now` and `--publish-async` flags to `publish_to_api`. The former now runs the publishing process in the context of the console session, allowing an engineer to immediately see logs. The latter performs the publishing action asynchronously.
- Changes a function name from `last_unpublished_envelope_id` to `last_published_envelope_id` to correctly give the intent of the function's action.
- Uses `EnvelopeQuerySet.last_envelope_for_year()` when filtering for the last published Envelope instance of the previous year. This prevents rejected envelopes being considered when getting that last published Envelope instance.
- Updates the Celery Beat crontab setting to run the Crown Dependencies publishing process every two hours, night and day and every day.

## Checklist
- Requires migrations? No

### DIFF
--- a/publishing/management/commands/publish_to_api.py
+++ b/publishing/management/commands/publish_to_api.py
@@ -53,7 +53,7 @@ class Command(BaseCommand):
             for i, crowndependencies in enumerate(incomplete, start=1):
                 self.print_envelope_details(
                     position=i,
-                    envelope=crowndependencies.packagedworkbaskets.last(),
+                    envelope=crowndependencies.packagedworkbaskets.last().envelope,
                 )
         if unpublished:
             self.stdout.write(
--- a/publishing/management/commands/publish_to_api.py
+++ b/publishing/management/commands/publish_to_api.py
@@ -8,28 +8,40 @@ from publishing.tasks import publish_to_api
 
 
 class Command(BaseCommand):
-    help = "Upload unpublished envelopes to the Tariff API."
+    help = (
+        "Manage envelope uploads to Tariff API. Without arguments, this "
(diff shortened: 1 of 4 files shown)


publishing/management/commands/publish_to_api.py
```diff
@@ -39,22 +51,33 @@ def list_unpublished_envelopes(self):
                 f"{incomplete.count()} envelope(s) not completed publishing:",
             )
             for i, crowndependencies in enumerate(incomplete, start=1):
-                self.stdout.write(
-                    f"{i}: {crowndependencies.packagedworkbaskets.last().envelope}",
+                self.print_envelope_details(
+                    position=i,
+                    envelope=crowndependencies.packagedworkbaskets.last(),
```
dalecannon, line 56: Should the packaged workbasket's envelope be passed in here?

dalecannon (approved):

publishing/management/commands/publish_to_api.py
paulpepper-trade: Fixed.

CPrich905 (approved):

PR #1378, opened 2025-01-08
TP2000-1642 Missing application security header

mattjamc:
# TP2000-1642 Missing application security header

## Why

Penetration testing highlighted that we are not using the `X-Permitted-Cross-Domain-Policies` HTTP response header. It is recommended to use this header and set it to none.

Usage of this header is less common since Adobe Flash Player and Microsoft Silverlight have been deprecated. However, some security checks like penetration testing still check for this header because it can mitigate the risk of an overly-permissive policy file added to your site by accident or through malicious actions [[Source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Permitted-Cross-Domain-Policies)].

## What

This PR extends the SecurityMiddleware to include and set the X-Permitted-Cross-Domain-Policies HTTP response header as none.


CPrich905 (approved):
lgtm

PR #1379, opened 2025-01-10
TP2000 1650 commodities snapshot behaviour

Luisella21:
TP2000-1650 commodities snapshot behaviour with effective_only=True

Using CommodityCollectionLoader.load(effective_only=True) gives an error, because a filter is applied to a non existing field.
Added a small test showing the issue, and changed the code to fix it.
- Requires migrations: No
- Requires dependency updates: No


.envrc
dalecannon, line 1: Might this file be better served as an addition to `.gitignore` to allow for local, personalised `.envrc` configurations?

commodities/tests/test_commodity_tree_snapshot.py
```diff
@@ -92,3 +92,11 @@ def test_get_dependent_measures_works_with_wonky_archived_measure(
     assert wonky_archived_measure.generating_regulation == old_regulation
     assert target_commodity in commodities_collection.commodities
     assert target_commodity in target.commodities
+
+
+def test_CommodityCollectionLoader(seed_database_with_indented_goods):
+    # Test that effective_only does not crash the code
```
dalecannon, line 98: Perhaps the comment could form part of the test name to reflect what in particular is being tested.

dalecannon (approved):

commodities/tests/test_commodity_tree_snapshot.py
```diff
@@ -92,3 +92,11 @@ def test_get_dependent_measures_works_with_wonky_archived_measure(
     assert wonky_archived_measure.generating_regulation == old_regulation
     assert target_commodity in commodities_collection.commodities
     assert target_commodity in target.commodities
+
+
+def test_CommodityCollectionLoader(seed_database_with_indented_goods):
```
eadpearce, line 97: can this be snake case please 🙏  eg `test_commodity_collection_loader`

.envrc
Luisella21: Absolutely correct. I removed it from the repository and amended .envrc

commodities/tests/test_commodity_tree_snapshot.py
Luisella21: Absolutely!

PR #1384, opened 2025-01-13
Created new index.

Luisella21:
TP2000-1654 Missing Index on model VersionGroup

The field current_version in the model VersionGroup is used extensively. An index on it will improve access.
Defined the new index on the model.

Requires migrations? YES
Requires dependency updates? NO


paulpepper-trade (approved):
Before merging to the main line, it's probably worth building and deploying this PR's branch to a non-production environment. That would allow testing and timing the migration phase of a deployment (lots of transactions to migrate!), so that we can manage / tweak things if necessary.

PR #1387, opened 2025-01-16
TP2000-1632 Auto-end-date footnote associations

mattjamc:
# TP2000-1632 Auto-end-date footnote associations

## Why

When a commodity is end-dated by the EU, tariff managers need to also end date  or delete any footnote associations on that commodity code.

## What

This PR:
-  extends the auto end-dating of measures functionality to also end-date footnote associations by changing it to a generic function which works for any tracked model
- updates the layout of the page to a tabbed view so the user can see both measures and footnote associations to be ended
- adds a new test for footnote associations and refactors and parameterizes 2 existing tests so they account for associations as well as measures

https://github.com/user-attachments/assets/3ea91a47-66f1-449b-bd5a-77f87e563350


workbaskets/models.py
```diff
@@ -779,6 +779,44 @@ def get_measures_to_end_date(self) -> QuerySet:
...
+        from commodities.models.orm import GoodsNomenclature
+
+        end_dated_commodities = GoodsNomenclature.objects.current().filter(
+            transaction__workbasket=self,
+            valid_between__upper_inf=False,
+        )
+        commodity_dict = {
+            commodity.sid: commodity.valid_between
+            for commodity in end_dated_commodities
+        }
```
paulpepper-trade, line 795: Is this code a duplicate of that used in `get_measures_to_end_date()`? If so, because measures and footnotes are ended within the scope of a single HTTP request, the commodity dictionary could, say, be shared in some way (perhaps with `cached_property`?).

workbaskets/jinja2/workbaskets/auto_end_objects.jinja
```diff
@@ -0,0 +1,66 @@
+{% extends "layouts/detail.jinja" %}
+
+{% from "components/button/macro.njk" import govukButton %}
+{% from "components/tabs/macro.njk" import govukTabs %}
+{% from "components/table/macro.njk" import govukTable %}
+{% from "components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
+
+
+{% set page_title %}
+  Workbasket {{ workbasket.pk }} - Auto end measures
```
paulpepper-trade, line 10: Because the submit perfoms end-dating of both measures and footnote associations in a single click (single HTTP POST), would it be preferable to keep the title the same across both tabs? The page title changes across the two tabs made me initially think the submit button was bound one to each tab and required one submit (HTTP POST) per tab. A common, shared page title indicating the whole outcome of the submit (the tabs have their own descriptive titles) might avoid unexpected behaviour.

workbaskets/models.py
```diff
@@ -779,6 +779,44 @@ def get_measures_to_end_date(self) -> QuerySet:
             db_effective_valid_between__not_gt=F("commodity_valid_between"),
         )
 
+    def get_footnote_associations_to_end_date(self):
+        """Queries all end-dated commodites in a workbasket and finds related
+        footnote associations so they can be end-dated too."""
+        from commodities.models.orm import FootnoteAssociationGoodsNomenclature
+        from commodities.models.orm import GoodsNomenclature
```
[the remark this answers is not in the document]
mattjamc: Yes it was necessary to put both there from what I recall

workbaskets/models.py
mattjamc: Yeah the functions are very similar and only different in that measures use the effective_valid_between rather than their actual valid_between. I could factor out the commodity dictionary if worth it?

workbaskets/jinja2/workbaskets/auto_end_objects.jinja
mattjamc: There is a shared H1 across both pages which is 'Auto end measures' but the page H2 does change for each tab.

This page_title bit is only used in the breadcrumbs which stays the same for each tab.

I'm open to changing the title. Steve's view was for the overarching name to still be Auto end measures since that is the primary function and footnote associations are an edge case.

workbaskets/models.py
paulpepper-trade: My comment about HTTP scoped querying is probably only relevant if the functions were in the View. You could move them there if the refactor gains (deduplicated code and DB query reduction) outweigh the usefulness of them being available on the WorkBasket model.

mattjamc: I've now moved these functions to the view and factored out the commodity_dictionary into a separate property. it has made the testing a bit more complex but they still work.

workbaskets/views/ui.py
```diff
@@ -1887,7 +1894,25 @@ def workbasket(self) -> WorkBasket:
...
     @property
     def measures(self):
-        return self.workbasket.get_measures_to_end_date()
+        return self.get_measures_to_end_date()
+
+    @property
+    def footnote_associations(self):
+        return self.get_footnote_associations_to_end_date()
+
+    @property
```
paulpepper-trade, line 1903: Could [cached_property](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.functional.cached_property) be used to avoid duplicate DB queries?

paulpepper-trade (approved):

PR #1388, opened 2025-01-17
Add missing measures check to rule-check task queue

eadpearce:
# TP2000-1478 Hotfix

## Why

* Task was not running in UAT env as it was being sent to the default 'celery' queue and we don't have a worker running this queue https://docs.celeryq.dev/en/stable/userguide/routing.html 

## What

* Added task to the rule-check queue


CPrich905 (approved):
DWISOTT

PR #1389, opened 2025-01-20
TP2000-1672 Measure current version bug

mattjamc:
# TP2000-1672 Measure current version bug

## Why

If a workbasket has 2 updates of the same object and it is queued and dequeued, revert_current_version() incorrectly assigns the current version to one of the updates which is about to be turned to draft.

Current version should be either the latest approved version or None.

## What

This PR updates revert_current_version() so when it's querying existing approved versions to revert to, it does not include those in the same workbasket. Also adds a test to confirm the fix works.

### DIFF
--- a/common/models/transactions.py
+++ b/common/models/transactions.py
@@ -7,6 +7,7 @@ from logging import getLogger
 
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
+from django.db.models import Q
 from django.db.transaction import atomic
 from django_fsm import FSMIntegerField
 from django_fsm import transition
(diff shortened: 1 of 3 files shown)


common/models/transactions.py
```diff
@@ -202,7 +203,10 @@ def revert_current_version(self):
             versions = (
                 version_group.versions.has_approved_state()
                 .order_by("-pk")
-                .exclude(pk=obj.pk)
+                .exclude(
+                    Q(pk=obj.pk)
+                    | Q(transaction__workbasket=obj.transaction.workbasket),
```
paulpepper-trade, line 208: Would the returned tracked models need restricting further to only those that are in either `SEED_FILE` or `REVISION` partition transactions and therefore excluding `DRAFT` transactions (i.e. they've been published)?

mattjamc: I believe this is what the .has_approved_state() is doing already

paulpepper-trade (approved):

PR #1395, opened 2025-01-24
TP2000-1678 Measure condition duty bug update

mattjamc:
# TP2000-1678 Measure condition duty bug update

## Why

The validation applied to the `applicable_duty` field in the measure condition form could be applied to all duty sentence fields.

Also the test I previously wrote could be consolidated into an existing test as a new parameter.

## What

This PR:
- moves the check for no more than 3 decimal places from the MeasureCondition form to the Duty Sentence Parser so it can be applied to all duty sentence fields
- moves a test to be a parameter of an existing test

### DIFF
--- a/measures/duty_sentence_parser.py
+++ b/measures/duty_sentence_parser.py
@@ -379,6 +379,12 @@ class DutyTransformer(Transformer):
                 f"Measurement unit qualifier {qualifier.abbreviation} cannot be used with measurement unit {unit.abbreviation}.",
             )
 
+    def validate_duty_amount(self, duty_amount):
+        if len(str(duty_amount).split(".")[-1]) > 3:
+            raise ValidationError(
+                f"The reference price cannot have more than 3 decimal places.",
+            )
+
     def validate_phrase(self, phrase):
         # Each measure component can have an amount, monetary unit and measurement.
         # Which expression elements are allowed in a component is controlled by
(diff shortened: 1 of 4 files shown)


measures/tests/test_forms.py
```diff
@@ -1228,8 +1229,9 @@ def test_measure_forms_conditions_wizard_form_invalid_duty(date_ranges):
     factories.MeasurementUnitFactory.create()
     factories.MeasurementUnitQualifierFactory.create()
     factories.MeasureConditionComponentFactory.create()
-    factories.DutyExpressionFactory.create(sid=99)
     action = factories.MeasureActionFactory.create()
+    if not DutyExpression.objects.filter(sid=99):
+        factories.DutyExpressionFactory.create(sid=99)
```
dalecannon, line 1234: This test looks similar to `test_measure_forms_conditions_wizard_invalid_duty`. Could the reference price and expected error message be added there?

measures/duty_sentence_parser.py
```diff
@@ -379,6 +379,12 @@ def validate_measurement(self, unit, qualifier):
                 f"Measurement unit qualifier {qualifier.abbreviation} cannot be used with measurement unit {unit.abbreviation}.",
             )
 
+    def validate_duty_amount(self, duty_amount):
+        if len(str(duty_amount).split(".")[-1]) > 3:
+            raise ValidationError(
+                f"The reference price cannot have more than 3 decimal places.",
+            )
```
dalecannon, line 386: It looks like the condition will also be satisfied if a duty amount has no decimal places but has more than 3 digits. An alternative way of getting the decimal places might be to use `Decimal.as_tuple().exponent`.

measures/tests/test_forms.py
mattjamc: Yes! I hadn't noticed that one - much better

dalecannon (approved):

PR #1398, opened 2025-01-28
TP2000-1686 Relocate imported commodities view

mattjamc:
# TP2000-1686 Relocate imported commodities view

## Why

The _Review commodities_ tab pulls commodity data from the imported Taric file. This provides an accurate audit of what came through in the import file although it is not exactly a summary of what commodities are in the workbasket. This should be moved to another location in TAP, such as from the EU Taric import list. 

After this PR, the Review commodities can be updated to a tabbed view which queries current commodities and related items in the same workbasket, in line with the other review tabs

## What

This PR moves code around but very little is added.

Specifically:
- moves the _Review imported goods_ view to the importer app - users can navigate to the page via the EU Taric import list for a successfully imported workbasket with commodities
- updates the current _Review commodities_ page to a holding message directing users to the review imported good page - this page will later become a tabbed ListView of commodities and related items in the workbasket

See updated link to view imported goods:
image

Leading to a new page with the same content as review commodities previously showed:
image

Temporary holding page:
image


workbaskets/jinja2/workbaskets/review-goods.jinja
```diff
@@ -3,80 +3,5 @@
...
-        {"text": "Start date"},
-        {"text": "End date"},
-        {"text": "Comments"},
-      ],
-      "rows": rows,
-    }) }}
-  {% endif %}
-
-  {% set items_name = "goods changes" %}
+  <p>Visit the <a href="{{url('review-imported-goods', kwargs={'pk':workbasket.pk})}}">Review imported goods</a> page to see commodities imported into this workbasket.</p>
```
eadpearce, line 6: Instead of doing this could we just set up a temporary redirect on the URL?

importer/tests/test_views.py
```diff
@@ -322,3 +322,9 @@ def test_notify_channel_islands_redirects(
         "goods-report-notify-success",
         kwargs={"pk": completed_goods_import_batch.pk},
     )
+
+
+def test_view_imported_goods_renders(valid_user_client, workbasket):
```
eadpearce, line 327: Not for this PR but leaving a note here to remind that we should definitely add more tests to this view in future as I don't think any were added when it was originally implemented

workbaskets/jinja2/workbaskets/review-goods.jinja
mattjamc: Have updated to a temporary redirect

PR #1400, opened 2025-01-29
TP2000-1660: Add staleness checking to missing measures check

eadpearce:
# TP2000-1660: Add staleness checking to missing measures check

## Why

* The missing measures check should be invalidated when the contents of the workbasket change

## What

* If a missing measures check has been run, checks the state of commodity and measure objects in the workbasket and compares it to the hash stored on the check. If they are different, the check is invalidated and must be run again

## Checklist
- Requires migrations? ✅ 

Screenshot 2025-01-29 at 11 32 51


workbaskets/jinja2/workbaskets/checks/missing_measures.jinja
```diff
@@ -88,7 +88,21 @@
         {# check was not successful #}
         {% else %}
 
-
+        {% if missing_measures_check.hash != workbasket.commodity_measure_changes_hash %}
```
dalecannon, line 91: This conditional may get skipped even if the hash has changed if the last check was successful.

dalecannon, line 91: Might this staleness check be useful as a boolean property?

workbaskets/models.py
```diff
@@ -456,6 +457,36 @@ def rule_check_progress(self) -> Tuple[int, int]:
...
+        check_workbasket_for_missing_measures.
+        """
+        # avoid circular import
+        from commodities.models.orm import GoodsNomenclature
+
+        changes = [
+            item
+            for item in self.tracked_models.all().order_by("pk")
+            if isinstance(item, GoodsNomenclature) or isinstance(item, Measure)
+        ]
```
dalecannon, line 477: The queryset here and in `def has_commodities()` could be filtered for subclasses:

```python
qs = self.tracked_models.filter(
    Q(instance_of=GoodsNomenclature) | Q(instance_of=Measure)
).order_by("pk")
```

workbaskets/jinja2/workbaskets/checks/missing_measures.jinja
eadpearce: Oh yeah you're right. I'll see if I can rewrite this template to be a bit less confusing

workbaskets/models.py
eadpearce: That's much neater thank you!

dalecannon (approved):

PR #1404, opened 2025-02-05
TP2000-1631 Update developer manual branding

mattjamc:
# TP2000-1631 Update developer manual branding

## Why

The TaMaTo developers manual in the public domain is still using DIT branding - We must update this to DBT.

## What

This PR updates references of DBT to DIT and updates the crown logo.

We use the [GOV.UK styled Sphinx documentation generator](https://github.com/ukgovdatascience/govuk-tech-docs-sphinx-theme) which does not account for updating the crown SVG. Its repo was archived over 2 years ago suggesting it is no longer maintained. So I've added in a new crown file to override the default logo.

New header:
image


CPrich905 (approved):
LGTM

PR #1406, opened 2025-02-05
TP2000-1645 Restrict Django Admin Login Page Access

LaurenMullally:
# TP2000-1645 Restrict Django Admin Login Page Access

## Why
The recent pen test on Tap highlighted a vulnerability with the django admin login page being available to everyone. This leaves it open for brute force attacks, even though it's not used as in hosted environments, Staff SSO handles all the logins. 

## What
It's wild but this makes essentially a custom django admin app, which is really just an extension off the default where we override the login functionality to throw a 404 if `SSO_ENABLED` is on. I've not added comments, but If you think it makes sense for me to add some comments to why we've got this custom admin config, then let me know and I'll chuck some in. There's also a nice test included for it.

## Checklist
- Requires migrations? - no 


settings/common.py
```diff
@@ -77,7 +77,7 @@
 # -- Application
 
 DJANGO_CORE_APPS = [
-    "django.contrib.admin",
+    "admin.apps.TamatoAdminConfig",
```
paulpepper-trade, line 80: Might this be better placed in the `DOMAIN_APP` list as it's a Tamato app rather than a Django core app?

paulpepper-trade (approved):

common/tests/test_views.py
```diff
@@ -380,3 +380,17 @@ def test_resources_view_displays_resources(heading, expected_url, valid_user_cli
     page = BeautifulSoup(response.content.decode(response.charset), "html.parser")
     assert page.find("h3", string=heading)
     assert page.find("a", href=expected_url)
+
+
+@override_settings(SSO_ENABLED=True)
+def test_admin_login_shows_403_when_sso_enabled(superuser_client):
```
eadpearce, line 386: Test name says 403 but docstring says 404

PR #1407, opened 2025-02-05
TP2000-1583 Add task assignee update support

paulpepper-trade:
# TP2000-1583 Add task assignee update support

## Why

There’s currently no way to set and view assignees to tasks and subtasks.

## What

This PR:

- Enables adding multiple assignees to tasks and subtasks.
- Follows the UI approach taken for legacy workbasket assignment, dynamically displaying a form to add and remove task assignees.
- Implements displaying task assignees.
- Adds a docstring to the `TaskAssignee` model with an approach to separating old style workbasket assignments (worker and reviewer). This is necessary to because of incompatibility between the two assignment approaches - workbasket assignment vs ticket / task assignment.
- Delays implementing unit tests for another ticket until further UI design sign-off is completed. This allows trialling current task assignment and fixing broken test capability.

## Checklist
- Requires migrations? No


common/models/user.py
```diff
@@ -2,9 +2,25 @@
 from django.db import models
 
 
+class UserQuerySet(models.QuerySet):
+    def active_tms(self) -> models.QuerySet | list["User"]:
+        """Return a QuerySet of active users that may take the tariff manager
+        role."""
+        return (
+            self.filter(
+                models.Q(groups__name__in=["Tariff Managers", "Tariff Lead Profile"])
```
dalecannon, line 11: A few users are in both groups and will be duplicated in the results.

tasks/jinja2/tasks/detail.jinja
```diff
@@ -101,28 +144,63 @@
...
+        {{ display_assignees(current_assignees) }}
+      </div>
+    </dd>
+    <dd class="govuk-summary-list__actions">
+      <ul class="govuk-summary-list__actions-list">
+        <li class="govuk-summary-list__actions-list-item">
+          <a
+            id="assign-users"
+            class="govuk-link"
+            href="#"
```
dalecannon, line 167: Is it desirable to include a link to the form to keep it accessible in the event JavaScript is unavailable on the page?

common/models/user.py
```diff
@@ -2,9 +2,25 @@
...
+            )
+            .filter(is_active=True)
+            .order_by("first_name", "last_name")
+        )
+
+
 class User(AbstractUser):
     """Custom user model."""
 
+    objects = UserQuerySet.as_manager()
```
dalecannon, line 22: A custom user manager might be needed to maintain support for creating users or editing inherited user fields.

common/models/user.py
paulpepper-trade: Added `distinct()`.

tasks/jinja2/tasks/detail.jinja
paulpepper-trade: Definitely. Now done.

common/models/user.py
paulpepper-trade: Custom manager also added along with migration.

dalecannon (approved):

PR #1408, opened 2025-02-06
TP2000-1673: Add missing measures check to processing queue view

eadpearce:
# TP2000-1673: Add missing measures check to processing queue view

## Why

* We want to be able to see the progress of the new missing measures check on the processing queue page

## What

* Modifies TAPTasks class to support business rule checks and missing measure checks
* Refactors TAPTasks a bit to improve test coverage 

## Checklist
- Requires migrations? ✅ 

Screenshot 2025-02-06 at 15 35 57


workbaskets/tests/test_views.py
```diff
@@ -2529,7 +2529,7 @@ def test_clean_tasks():
                 "hostname": "celery@1",
                 "time_start": None,
                 "acknowledged": False,
-                "delivery_info": {},
+                "delivery_info": {"routing_key": "rule-check"},
```
mattjamc, line 2532: Now that there is a _test_tap_tasks_ file would this test go better in there?

mattjamc (approved):
Looks good 🥳

workbaskets/tests/test_views.py
eadpearce: oh you're right. i had a look over these initially and some of them make a request to the url so i figured i'd keep those ones in here. i'll look through again and move the ones that don't into that new file

PR #1409, opened 2025-02-07
TP2000-1633 Auto delete measures and associations

mattjamc:
# TP2000-1633 Auto delete measures and associations

## Why

Now that TAP can automatically end date/delete measures where the commodity code has been end-dated, we need to cover the scenario where a commodity code is being deleted.

In this scenario, all measures and footnote associations linked to these commodity codes must be deleted.

## What

This PR:
- Extends the auto end-dating functionality to also check for deleted commodity codes and delete any related measures or footnote associations
- Adds a new column to the template so users can see whether each row is referencing a deleted or end-dated commodity code
- introduces a new property `current_version_inc_draft` on goods_nomenclature to solve the complexity in getting the current version of an object which also includes drafts in the workbasket, like the .current() filter does but the existing `current_version` does not
- Extends existing tests and adds some new ones

image


mattjamc:
> LGTM, only point to note (as discussed face to face) is that, if a measure is deleted and then updated with an end date, your query set will increase dramatically as it meets the requirements for 'Deleted' and 'To be end dated' image

@SteveCN7 This is an important point - this all works _unless_ somehow a commodity is deleted but then gets created/updated after the delete. Is there any chance the EU will do this?

SteveCN7:
> > LGTM, only point to note (as discussed face to face) is that, if a commodity is deleted and then updated with an end date, your query set will increase dramatically as it meets the requirements for 'Deleted' and 'To be end dated' image
> 
> @SteveCN7 This is an important point - this all works _unless_ somehow a commodity is deleted but then gets created/updated after the delete. Is there any chance the EU will do this?

That's fine @mattjamc  - if the EU delete a commodity code and they need to recreate it, they'll make a new one with a new unique SID

CPrich905 (approved):
Code LGTM, had issues running on my local but worked on testing and for other devs

PR #1410, opened 2025-02-10
TP2000-1643 Input validation not enforced by application

CPrich905:
# TP-2000-1643 Input validation not enforced by application 

## Why
The recent Pen test demonstrated that some of our current validation checks could be bypassed by a malicious user.

## What
Adds an additional check that the uploaded file name matches the expected format of either `DITXXXXXX.xml` or `TGBXXXXX.xml` and provides an error message.
Updates tests to work with the latest fix.
Removes the original `validate_file_name` as it's no longer relevant


importer/forms.py
```diff
@@ -323,6 +340,12 @@ def clean(self):
             else None
         )
         if taric_file:
+            mime_type = get_mime_type(taric_file)
+            if mime_type not in ["text/xml", "application/xml"]:
```
mattjamc, line 344: Is this check not already being done by the ImporterV2FormMixin?

mattjamc (approved):
LGTM :) Maybe at some point the page content should be updated too.

_Valid TARIC files... usually have a .xml file name extension_ - surely they always need this now not just usually

Also maybe a hint to let them know what file names are accepted since at the moment it doesn't tell you what's wrong with the name.

importer/forms.py
```diff
@@ -67,6 +68,23 @@ def process_file(
             )
         return chunk_count
 
+    def validate_taric_file_name(self, file_name):
+        """
+        This replaces validate_file_name which was vulnerable to exploitation:
+        Input Validation Not Enforced by Application
```
dalecannon, line 74: I think `validate_filename` was added afterwards, around 18 July.

importer/forms.py
CPrich905: Yes, I meant to remove this duplication - thanks for catching!

importer/forms.py
dalecannon: Yeah, it was related to CodeQL alert at the time. Enforcing a file naming convention is a good idea.

CPrich905:
> LGTM :) Maybe at some point the page content should be updated too.
> 
> _Valid TARIC files... usually have a .xml file name extension_ - surely they always need this now not just usually
> 
> Also maybe a hint to let them know what file names are accepted since at the moment it doesn't tell you what's wrong with the name.

Good shout on the page content, I've changed `usually` to `must`. RE: the hinting, I'm less inclined to change this for a few reasons (happy to change if you/anyone else has a strong argument otherwise though):
1. the help text specifies it must be an `.xml` file type
2. any legitimate user should in theory know the naming conventions
3. we don't want to give a malicious user too much help if they encounter this error
Again, happy to be swayed otherwise if you have a strong argument/preference!

PR #1416, opened 2025-02-20
TP2000-1719 Ticket list view

mattjamc:
# TP2000-1719 Ticket list view

## Why
The ticket (previously workflow) list view needs updating to match the [new design](https://www.figma.com/design/kb4Q9sC1SED1KX5auCGqXy/%5BTP2000-1609%5D-Viewing-and-filtering-all-tasks?node-id=0-1&p=f&t=rexNxnqnZ5zKlVGC-0) and term changes.

## What
 * Adding new columns to list view ('entry into force date', 'assignee') in line with new design changes
 * Sorting capability by 'ID' and 'entry into force date' by ascending or descending order
 * Status tags added
 
<details open>
<summary> New ticket list view</summary>
<br>
Screenshot 2025-03-06 at 10 16 33
</details>

## To revisit
- Update the assignment column and create the assignment filtering once the new assignment functionality is created
- Any new status tags that need adding and a coloured label creating

## Checklist
- Requires migrations? - No


tasks/filters.py
```diff
@@ -51,9 +56,27 @@ class TaskWorkflowFilter(TamatoFilter):
         widget=CheckboxSelectMultiple,
     )
 
+    category = ModelChoiceFilter(
+        label="Work type",
+        queryset=Category.objects.all(),
+        widget=widgets.Select(),
+    )
```
dalecannon, line 63: In the UI design, is work type intended to filter workflows by creator template (i.e the template the ticket was created from)?

tasks/filters.py
```diff
@@ -51,9 +56,27 @@ class TaskWorkflowFilter(TamatoFilter):
...
+
+    assignees = ModelChoiceFilter(
+        label="Assignees",
+        field_name="assignees",
+        queryset=User.objects.all(),
```
dalecannon, line 68: Could this queryset be filtered for TMs only (e.g` User.objects.active_tms()`)?

tasks/jinja2/tasks/workflows/list.jinja
```diff
@@ -3,7 +3,13 @@
 {% from "components/table/macro.njk" import govukTable %}
 {% from "components/create_sortable_anchor.jinja" import create_sortable_anchor %}
 
-{% set page_title = "Find and view workflows" %}
+{% macro status_tag(task_status) %}
+    <span class="{{status_tag_generator(task_status).tag_class}}">
+            {{ status_tag_generator(task_status).text }}
+        </span>
+{% endmacro %}
```
dalecannon, line 10: Could the `display_status` macro in `display_details.jinja` be used in place of this?

tasks/jinja2/tasks/workflows/list.jinja
```diff
@@ -45,35 +51,28 @@
...
-                class="govuk-link"
-                href="{{ url("workbaskets:workbasket-ui-detail", kwargs={"pk": object.workbasket.pk}) }}"
-              >{{ object.workbasket.pk }}</a>
-            {%- else -%}
-              -
-            {%- endif -%}
+          {%- set assignees -%}
+            {% for assignee in object.assignees.all()%}
+            {{assignee.user.get_displayname()}}<br>
+            {% endfor%}
```
dalecannon, line 57: The queryset may need to use `.assigned()` to filter out unassigned users. Perhaps the `display_assignees` macro could also be used here too.

tasks/filters.py
marya-shariq: Yesss, I saw that Paul had added in the `active_tms` method to the `UserQuerySet` class and had included it in a previous commit but was unsure why it wasn't working. I think that's definitely a more sensible option and I will implement it.

mattjamc: Sorry yes - the assignment stuff is all quite unfinished at the moment since we are waiting for the new assignment process to be determined

tasks/filters.py
marya-shariq: Yes, thank you for spotting!

tasks/jinja2/tasks/workflows/list.jinja
marya-shariq: Would it be better to have the macro display "None" as opposed to "-" for tickets without an assignee?

tasks/jinja2/tasks/workflows/list.jinja
```diff
@@ -45,35 +47,28 @@
...
-            {% if object.workbasket %}
-              <a
-                class="govuk-link"
-                href="{{ url("workbaskets:workbasket-ui-detail", kwargs={"pk": object.workbasket.pk}) }}"
-              >{{ object.workbasket.pk }}</a>
-            {%- else -%}
-              -
-            {%- endif -%}
+          {%- set assignees -%}
+            {% for assignee in object.assignees.all()%}
```
dalecannon, line 51: This can be removed now that the macro is being used instead.

tasks/filters.py
```diff
@@ -51,9 +55,27 @@ class TaskWorkflowFilter(TamatoFilter):
...
+        queryset=TaskWorkflowTemplate.objects.all(),
+        field_name="taskworkflow__creator_template",
+        widget=widgets.Select(),
+    )
+
+    assignees = ModelChoiceFilter(
+        label="Assignees",
+        field_name="assignees",
+        queryset=User.objects.active_tms(),
+    )
```
dalecannon, line 69: I'd imagine there'll need to be a custom filter method for assignee lookup. E.g
```python
def filter_by_assignee(self, queryset, name, value):
    assigned_tasks = TaskAssignee.objects.filter(user=value).assigned().values_list("task__id", flat=True)
    return queryset.filter(id__in=assigned_tasks)
```

dalecannon (approved):

PR #1418, opened 2025-02-25
TP2000-1646  tamato HMRC_PACKAGING_SEED_ENVELOPE_ID year unaware

CPrich905:
# TP-2000-1646  TaMaTo HMRC_PACKAGING_SEED_ENVELOPE_ID year unaware

## Why
On 1/1/25, the first envelope of the year failed to publish to the tariffs-api service because it was seeded with an incorrect `Envelope.envelope_id` (ie. it was starting `250045` rather than `250001`)
The Tamato setting, `HMRC_PACKAGING_SEED_ENVELOPE_ID`, is used to provide a “seed” envelope ID value. It has two intended uses:
1. Provide a seed envelope ID for use when transitioning from manual envelope creation and publishing (as previously occurred - see @lauren.qurashi about “the gulag”) to an automated publishing system, as is now the case.
2. Provide a way of skipping envelope IDs during automated publishing. This allows automated publishing to be paused, manually publish to HMRC’s CDS system, set HMRC_PACKAGING_SEED_ENVELOPE_ID to the next envelope ID, then switch back to automated publishing, allowing correct automated envelope ID generation for the next auto-published envelope.

Currently, HMRC_PACKAGING_SEED_ENVELOPE_ID is implemented in a year unaware way. This leads to the envelope seed value that it holds being applied once each year, outside the two uses given above. This is incorrect behaviour. 

## What
This PR is formed of three parts.
1. config change to update the `HMRC_PACKAGING_SEED_ENVELOPE_ID` to be year aware, ie. from `0044` to `230044` (the id, with year suffix, for when automatic publishing began)
2. updating the `Envelope.next_envelope_id()` class method to account for a year aware seed id. This ensures we are getting the correct next `envelope_id` for the first envelope of the year and for envelopes after automatic publishing has been paused.
3. updating the `next_expected_to_api()` check on the PackagedWorkBasket object. This returns a boolean value based on whether the previous envelope id matches the expected value whether on the first of the year, during automated publishing, or when automated publishing resumes after manually publishing an envelope.

As well as the main Jira ticket, additional context can be found [here](https://uktrade.atlassian.net/browse/TP2000-1639) and [here](https://uktrade.atlassian.net/browse/TP2000-1635)

## Dependencies
The `HMRC_PACKAGING_SEED_ENVELOPE_ID` environment variable in our parameter store will need updating from four digits to the new six digit length to align with this PR


publishing/models/envelope.py
```diff
@@ -256,28 +256,33 @@ class Meta:
...
+
+        if (previous_envelope is None) and (seed_id[:2] != current_year):
+            # First envelope of the year in the format of YYCCCC.
+            counter = current_year + "0001"
         else:
-            year = int(envelope.envelope_id[:2])
             counter = max(
-                int(envelope.envelope_id[2:]) + 1,
-                int(settings.HMRC_PACKAGING_SEED_ENVELOPE_ID),
+                int(previous_envelope.envelope_id or 0) + 1,
```
paulpepper-trade, line 275: What's being guarded against by the `previous_envelope.envelope_id or 0`? I'm not sure what circumstances would result in `previous_envelope` being truthy / valid, but it's `envelope_id` being falsy.

paulpepper-trade (commented):
To avoid future confusion, it's probably worth updating the `HMRC_PACKAGING_SEED_ENVELOPE_ID` environment variable in our parameter store from four digits to the new six digit length to align with this PR. Perhaps mention that as a dependency in the PR cover note?

publishing/models/envelope.py
```diff
@@ -256,28 +256,33 @@ class Meta:
 
     @classmethod
     def next_envelope_id(cls) -> str:
-        """Get packaged workbaskets where proc state SUCCESS."""
-        envelope = Envelope.objects.last_envelope_for_year()
+        """
+        Get packaged workbaskets where proc state SUCCESS.
```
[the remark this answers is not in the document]
CPrich905: 👍  Removed

publishing/models/envelope.py
CPrich905: Interesting.... I had an edge case while building, but I can no longer replicate it so this is removed

paulpepper-trade: ... but could `previous_envelope` itself be `None` / falsy?

paulpepper-trade: Somethig like this perhaps?
```suggestion
                (int(previous_envelope.envelope_id) + 1) if previous_envelope else 1
```

publishing/models/packaged_workbasket.py
```diff
@@ -401,39 +400,33 @@ def next_expected_to_api(self) -> bool:
...
-            )
+        previous_id = PackagedWorkBasket.objects.last_published_envelope_id() or 0
+        current_id = self.envelope.envelope_id
+        current_year = str(datetime.now().year)[-2:]
+        if previous_id == 0 or (int(previous_id[:2]) == int(current_year) - 1):
+            return current_id == current_year + "0001"
         else:
-            expected_previous_id = str(int(self.envelope.envelope_id) - 1)
+            expected_previous_id = max(
+                int(previous_id or 0),
```
paulpepper-trade, line 421: Why the `previous_id or 0`?
Doesn't
```previous_id = PackagedWorkBasket.objects.last_published_envelope_id() or 0```
deal with the possible falsey return from `last_published_envelope_id()`, setting `previous_id` to `0`.
And then the statement
```if previous_id == 0 or (int(previous_id[:2]) == int(current_year) - 1):```
deals with the case where `previous_id` is `0`.

Could you just safely do this?
```int(previous_id)```

File: publishing/models/packaged_workbasket.py [hunk 4]
```diff
@@ -401,39 +400,33 @@ def next_expected_to_api(self) -> bool:
...
+                int(settings.HMRC_PACKAGING_SEED_ENVELOPE_ID),
+            )
 
-        if previous_id and previous_id != expected_previous_id:
+        if previous_id and int(previous_id) != expected_previous_id:
```
paulpepper-trade, line 425: By the same [reasoning in the previous comment](https://github.com/uktrade/tamato/pull/1418/files#r1977726994), won't `previous_id` always be truthy by this point? If so, then this could be simplified to:
```suggestion
        if int(previous_id) != expected_previous_id:
```

PR #1419, opened 2025-02-26
TP200-1722 Create New Ticket

LaurenMullally:
# TP200-1722 Create New Ticket

## Why
The new designs rename a workflow to a ticket. This PR brings in the changes required to make this happen! 

## What
- UI references to workflow have been changed to Tickets excluding the urls. 
- Form fields have been renamed to reflect the UI titles on the form fields. 
- New form fields have been added to the forms, and ultimately to the Workflow model. 
- Form redirects to the Ticket detail page on submission, rather than a success page. 
- Tests amended to reflect these changes. 

This may be a bit confusing however, as I've only amended what I have to in order to not break too much. This leaves things like the Create a workflow form taking in a ticket name on the form, but a ticket name being saved as a title within the database / on the models. 

EIF date and Policy Contact have been added to the forms. These items are saved on the Workflow model, as Paul and I felt they related to a workflow specifically. The rest of the information continues to be saved on the summary task. 

The designs ask for an assignee to be added to a workflow. We don't have this functionality yet, and the work to add it has been carved out into a separate ticket, so that explains why the assignee fields are missing from the form. 

Link to designs: https://www.figma.com/proto/WXm47Am68r22NtC5xAxf6C/%5BTP2000-1606%5D-Create-a-task?page-id=6578%3A13392&node-id=6578-13393&p=f&viewport=-793%2C-223%2C0.25&t=yGF2xA3HH2UNmif6-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=6578%3A13393&show-proto-sidebar=1

## Checklist
- Requires migrations? - yes


marya-shariq (commented):
Looks good to me :)

dalecannon (approved):

PR #1423, opened 2025-03-04
Models for open data simplified database

Luisella21:
# TP2000-1740 Open data database

There is a desire to use TAP data in open data analysis. The data currently published contain all the historical information, and analysis are complex.

This PR create a set of 'current' data that can be exported every day and used to create reports within TAP.

It creates a shadow model of each tracked model in TAP. The shadow model contains only the approved and current version. 
Where feasible, models have been merged: for instance, descriptions are in the described model, not in a different one. 

The tables supporting the models are created in a different partition of the database, so they can easily be excluded from the daily back up of the database.

There is a mechanism to populate the data : this will eventually happen daily, but it is not part of the current PR.

- Requires migrations: Yes
- Requires dependency updates - No


TomMacca:
Really nice work @Luisella21 

One possible addition is saving the date of the simplified snapshot somewhere. Could a snapshot_date column on all the tables? Or it could be in the table names themselves? Or a metadata table? 🤔

TomMacca (approved):
Approved from me, looks great @Luisella21

PR #1434, opened 2025-03-13
TP2000-1767 Add required param to create_bucket()

paulpepper-trade:
# TP2000-1767 Add required param to create_bucket()

## Why

The unit `test workbaskets/tests/test_views.py::test_download` has started to fail or failing intermittently because of a missing mandatory parameter to a Boto3 API call, `create_bucket()`.

## What

Provides the mandatory parameter to fix the failing unit test.

## Checklist
- Requires migrations? No


mattjamc (approved):
LGTM :)

PR #1436, opened 2025-03-13
TP2000-1737 Move assignments and comments into workbaskets

paulpepper-trade:
# TP2000-1737 Move assignments and comments into workbaskets

## Why

As part of preparing the way for tasks and workflows, the models used for user assignment and comments require moving to the workbaskets module.

## What

This PR:
- Pulls the `AssignmentType` model out from inner class to a module-level class. It simplifies its use and overcomes an intractable data migration problem accessing the inner class via a migrator's `ProjectState` instance.

## Checklist
- Requires migrations? Yes


CPrich905 (approved):
LGTM, thanks for the run throughs on Friday :)

PR #1438, opened 2025-03-19
TP2000-1744 Support configurable ticket ID prefix

mattjamc:
# TP2000-1744 Support configurable ticket ID prefix

## Why

A ticket naming prefix may be desirable as a way of including contextual information when referring to tickets by their ID. For instance, if a prefix of “TC-” (Tariff Change, or other preferred prefix, including the empty string) is adopted, then by referring to TC-123 it’s clear that ticket 123 is being referenced.

## What

This PR adds a configurable TICKET_PREFIX env variable used as the prefix to a ticket PK to form the prefixed_id which is rendered to users whenever a ticket ID is referenced.

It also updates filtering so correct results are returned regardless of case or whether the prefix is included in search.

image

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -948,4 +948,4 @@ MEASURES_ASYNC_EDIT = is_truthy(os.environ.get("MEASURES_ASYNC_EDIT", "true"))
 
 
 # Ticket naming prefix
-TICKET_PREFIX = "TC-"
+os.environ.get("TICKET_PREFIX", "TC-")
--- a/settings/common.py
+++ b/settings/common.py
@@ -948,4 +948,4 @@ MEASURES_ASYNC_EDIT = is_truthy(os.environ.get("MEASURES_ASYNC_EDIT", "true"))
 
 
 # Ticket naming prefix
-os.environ.get("TICKET_PREFIX", "TC-")
+TICKET_PREFIX = os.environ.get("TICKET_PREFIX", "TC-")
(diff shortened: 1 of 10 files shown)


tasks/tests/test_filters.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pytest.mark.parametrize(
+    "search_term, expected_result",
+    [
+        (f"{settings.TICKET_PREFIX}1234", True),
+        (f"{settings.TICKET_PREFIX.lower()}1234", True),
+        (f"{settings.TICKET_PREFIX[:-1]}1234", True),
+        ("1234", True),
+        ("4321", False),
```
paulpepper-trade, line 19: Perhaps it's worth varying the value of `settings.TICKET_PREFIX` to cater for cases outside those that are currently in Tamato's settings. For instance, prefixes that contain numberals, like dates (e.g. "TC2025").

tasks/tests/test_models.py
```diff
@@ -259,3 +260,8 @@ def test_task_is_subtask_property(task_factory):
 def test_task_is_summary_task_property(create_task_fn):
     task = create_task_fn()
     assert bool(hasattr(task, "taskworkflow")) == task.is_summary_task
+
+
+def test_prefixed_id(task_workflow):
+    expected_prefixed_id = f"{settings.TICKET_PREFIX}{str(task_workflow.id)}"
```
paulpepper-trade, line 266: Is the `str` cast needed in the format string?

tasks/filters.py
```diff
@@ -68,6 +74,21 @@ class TaskWorkflowFilter(TamatoFilter):
         queryset=User.objects.active_tms(),
     )
 
+    def get_search_term(self, value):
+        """Looks for a pattern matching a ticket ID and removes any non
+        numerical prefix so only the number is searched."""
+        value = value.strip()
+        if self.search_regex:
```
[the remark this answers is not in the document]
mattjamc: Responding to your other comments I've now changed how this works

settings/common.py
```diff
@@ -948,4 +948,4 @@
 
 
 # Ticket naming prefix
-TICKET_PREFIX = "TC-"
+os.environ.get("TICKET_PREFIX", "TC-")
```
paulpepper-trade, line 951: Sorry, I missed the assignment in my previous suggestion!
```suggestion
TICKET_PREFIX = os.environ.get("TICKET_PREFIX", "TC-")
```

paulpepper-trade (approved):

PR #1440, opened 2025-03-20
TP2000-1758 Step Detail View

LaurenMullally:
# TP2000-1758 Step Detail View

## What
This PR adds design changes to the Task Detail view: 
- Rename any mentions of Tasks to Steps
- Amend and add headers as per design
- Change and format table details
- Remove additional action buttons, and move delete button to top right. 

This PR also includes some clean up over the breadcrumbs, but there will be a separate ticket to clean the breadcrumbs up completely. 

Also includes tests to check that the page looks as it should, and amendments to the existing tests!

## Checklist
- Requires migrations? no 

Old look: 
image

New look: 
image


paulpepper-trade (approved):
A few comments, but otherwise LGTM!

PR #1444, opened 2025-03-25
TP2000-1775  Amend breadcrumbs on ticket and step pages

dalecannon:
# TP2000-1775  Amend breadcrumbs on ticket and step pages

## Why
A coherent page hierarchy and user navigation experience is required on ticket and step pages.

## What
- Amends the breadcumbs on all ticket, ticket template, step and step template pages
- Updates button text and links on confirmation pages
- Renames old references to workflows and tasks on pages

## Checklist
- Requires migrations? No

### DIFF
--- a/tasks/views.py
+++ b/tasks/views.py
@@ -628,6 +628,11 @@ class TaskWorkflowConfirmCreateView(PermissionRequiredMixin, DetailView):
     template_name = "tasks/workflows/confirm_create.jinja"
     permission_required = "tasks.add_taskworkflow"
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["verbose_name"] = "ticket"
+        return context
+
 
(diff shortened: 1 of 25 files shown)


tasks/views.py
```diff
@@ -203,7 +197,10 @@ class TaskConfirmDeleteView(PermissionRequiredMixin, TemplateView):
     def get_context_data(self, **kwargs):
         context_data = super().get_context_data(**kwargs)
         context_data["deleted_pk"] = self.kwargs["pk"]
-        context_data["verbose_name"] = "task"
+        context_data["ticket"] = TaskWorkflow.objects.get(
+            pk=self.request.session["ticket_pk"],
+        )
+        del self.request.session["ticket_pk"]
```
paulpepper-trade, line 203: "Flash values" will be lost if the user refreshes the page. Better to pass the ticket's PK via a URL param I think.

PR #1446, opened 2025-03-26
TP2000-1749: Geo area exclusion real edits

eadpearce:
# TP2000-1749: Geo area exclusion real edits

## Why

* Makes the workflow smoother for users

## What

* Plugs in the make_real_edit function in place of create or new_version methods to take into account changes in the current workbasket


CPrich905 (approved):
LGTM 🌍

PR #1447, opened 2025-03-26
Tp2000 1764  implement comm code design changes

CPrich905:
# TP-2000-1746 - Implementing Commodity Code design changes

## Why
As part of the overall redesign of the Commodity Code hierarchy redesign, we are implementing styling changes to the  Commodity Code/Hierarchy view page to make it easier to understand the relationship between Commodities.

## What
The design has been iterated over more than once in discussion with Tariff Managers, but the first image below shows the final agreed design:
image

The following changes to this were agreed via Teams chat:
1) not having an underline for the expandable sections
2) any text overflow should be included in the row as per the screen shot below
3) the grey banner to indicate the selected commodity code would be monochromatic

image


common/static/common/scss/_commodities.scss
```diff
@@ -1,8 +1,115 @@
...
+}
+
+.govuk-details__summary {
+    display: block;
+    margin-bottom: 0px;
+}
+
+.govuk-details__summary-text {
+    text-decoration: none;
+}
```
paulpepper-trade, line 21: Won't these classes be changed globally across all site pages, not just commodities-related pages? Is that the intent? If so, then they should probably be placed in an SCSS file that isn't specific to commodities, but is intended for site-wide styling.

CPrich905: 🤦 you are correct - nested into under a new class

common/static/common/scss/_commodities.scss
```diff
@@ -1,8 +1,115 @@
-.commodity-hierarchy {
+// Overrides of govuk styles to get everything else displaying correctly
+.govuk-body,
+.comment p,
+body,
+.govuk-body-m {
+    margin-bottom: 0;
+}
```
paulpepper-trade, line 7: [Those](https://github.com/uktrade/tamato/pull/1447#discussion_r2024524633) and these too?

CPrich905: Ah sorry, just saw your second comment. I'd removed the parent class at the start of this, it's now back in.

paulpepper-trade (approved):

PR #1448, opened 2025-03-28
TP-2000-1784 annual volume changes for recurring QuotaDefinitionPeriods

CPrich905:
# TP-2000-1784 Incremental volume increase for annually recurring QuotaDefinitionPeriods

## Why
Annually recurring QuotaDefinitionPeriods can increase or decrease by a numerical or percentage value from the previous year.

## What
This feature calculates the value of a QuotaDefinitionPeriod volume based on the value of the previous year's volume and a provided figure, either as a percentage or an amount.
The widget used is a custom RadioSelect widget that uses elements of two GDS radio select widgets, the radio select with text divider and the radio select with conditionally revealed inputs. Documentation on these can be found [here](https://design-system.service.gov.uk/components/radios/)

The user is presented with five options, no change or for the figure to increase or decrease by amount or percentage. Choosing an option that changes the volume opens an input field for a value, either amount or percentage.
image

Choosing a change option but not entering a value triggers a validation error.

Calculated amounts are then displayed on the summary page
image

Note 1: the widget has required false set twice because not all of the options have a secondary input and submitting this option resulted in a false failure. The validation for this widget runs on form submission.

Note 2: I would like to have made this a generic widget that could be re-used and have built with this in mind. At the moment, it's custom to the bulk create process and so the template is bespoke. The criteria for generalisation I believe are as follows:
1. can pass through a divider class (ie. or)
2. can pass through **optional** secondary inputs
3. can pass through a suffix for the secondary inputs, e.g. `%` 
This is quite low priority as it works for the current scenario and I don't know of anywhere else it would be used at present.

### DIFF
--- a/common/widgets.py
+++ b/common/widgets.py
@@ -27,6 +27,17 @@ class RadioNestedWidget(RadioSelect):
         self.nested_forms = forms
 
 
+class CommodityVolumeRadioSelect(RadioSelect):
+    option_template_name = "common/widgets/radio_conditional_commodities.html"
+
+    def get_context(self, name, value, attrs):
+        ctx = super().get_context(name, value, attrs)
+
+        ctx.update({"optgroups": self.choices})
+
+        return ctx
+
+
 class FormSetFieldWidget(widgets.Widget):
     template_name = "common/widgets/formset_field.html"
     input_type = ""
--- a/common/widgets.py
+++ b/common/widgets.py
@@ -31,11 +31,11 @@ class CommodityVolumeRadioSelect(RadioSelect):
     option_template_name = "common/widgets/radio_conditional_commodities.html"
 
     def get_context(self, name, value, attrs):
-        ctx = super().get_context(name, value, attrs)
+        context = super().get_context(name, value, attrs)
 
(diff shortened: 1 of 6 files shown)


common/widgets.py
```diff
@@ -27,6 +27,17 @@ def bind_nested_forms(self, forms):
         self.nested_forms = forms
 
 
+class CommodityVolumeRadioSelect(RadioSelect):
+    option_template_name = "common/widgets/radio_conditional_commodities.html"
+
+    def get_context(self, name, value, attrs):
```
LaurenMullally, line 33: I feel like It would be clearer if you had ctx as context instead

quotas/tests/test_views.py
```diff
@@ -2874,6 +2878,77 @@ def test_bulk_create_creates_definition(
         assert len(models.QuotaDefinition.objects.all()) == 1
 
 
+# @pytest.mark.parametrize(
```
LaurenMullally, line 2881: Cheeky comments left in

LaurenMullally (approved):
Delightful

PR #1450, opened 2025-04-04
Step automation foundations and workbasket creation

paulpepper-trade:
# TP2000-1776 & TP2000-1777 Step automation foundations and workbasket creation

## Why

Many types of ticket steps could be automated. Rather than a user entering data into web forms or performing multiple actions across several web pages or services, TAP automation could be used on behalf of the user.

A simple example: a “create workbasket” step would normally require the user to navigate to the “create workbasket” web form, enter valid information (by copying from elsewhere) about the workbasket and finally copy new workbasket details to the ticket. This could be automated by TAP because the information is available within the context of the ticket.

## What

This PR:
- Creates the "foundation" implementation support for dynamically defining and executing step automations.
- Defines an `Automation` base class from which step automation subclasses inherit.
- Defines `StateChoices`, only one of which an automation must be in and report via its `get_state()` method.
- Defines `Automation.rendered_state()` that must be implemented by subclasses to provide their rendered state with a step's details.
- Implements a step automation, `CreateWorkBasketAutomation`, inheriting from the `Automation` base class and implementing methods required by the base class.
- Provides model tests for `CreateWorkBasketAutomation`.
- Provides tests for the view used to run `CreateWorkBasketAutomation`.
- Adds test support with fixtures and a factory class.

## Checklist
- Requires migrations? Yes


LaurenMullally (approved):
Amazing

PR #1454, opened 2025-04-14
Tp2000 1743 ticket transition rules 3

LaurenMullally:
# TP2000-1743 Ticket Transition Rules

## Why
A Task's `progress_state` has previously been a model, but in order to keep things simpler and introduce rules around how a task can move through states, it needed to adapt into a FSM Field (Finite State Machine). This allows us to create transition methods that dictate how a state can be changed. 

## What
This PR: 
- Changes `progress_state` on a Task to be a FSM field stored on the task itself, rather than a Foreign Key to a ProgressState model. 
- Introduces transitions for changing the value of the `progress_state` field. 
- Amends all the areas of the codebase that call on or use this field now that the state is a field not an FK
- Amends all the tests that were broken as a result of this change 
- Amends the UpdateTaskView and corresponding form to enable manual change of a progress state rather than directly changing it in the database on form save. 

- Requires migrations? - Yes 
- Requires dependency updates?  - Nah


tasks/tests/test_views.py
```diff
@@ -84,11 +80,11 @@ def test_task_update_view_update_progress_state(valid_user_client):
     response = valid_user_client.post(url, form_data)
     assert response.status_code == 302
 
-    instance.refresh_from_db()
+    refreshed_instance = Task.objects.get(pk=instance.pk)
```
paulpepper-trade, line 83: What's the reason for replacing `Model.refresh_from_db()`? Transition problems?

tasks/forms.py
```diff
@@ -139,6 +129,24 @@ def init_layout(self):
             ),
         )
 
+    def update_status(self, task) -> bool:
```
paulpepper-trade, line 132: Perhaps name this function to reflect which field it's updating?
```suggestion
    def update_progress_state(self, task: Task) -> bool:
```

tasks/forms.py
```diff
@@ -139,6 +129,24 @@ def init_layout(self):
...
+        if task.progress_state != new_status:
+            match new_status:
+                case ProgressState.TO_DO:
+                    task.to_do()
+                case ProgressState.IN_PROGRESS:
+                    task.in_progress()
+                case ProgressState.DONE:
+                    task.done()
+        else:
+            raise ValidationError(
```
paulpepper-trade, line 143: Is it legitimate to allow a user to save without making changes to the progress state? If that were so, then I guess there'd be no need raise a `ValidationError`.

paulpepper-trade: If submitting with no change of progress state, then perhaps the validation could be performed in `progress_state`'s field-specific clean method, `clean_progress_state()`?

PR #1471, opened 2025-05-19
TP2000 1469 Measures as defined and Commodities reports available on TAP.

Luisella21:
# TP-2000-1469 Add futere measures to MAD report

## Why

The users need to see the full list of measures (Measures as Defined, MAD), including the measures not yet into force. The report available on open data does not show the future measures. Also,  because the report is in a different 'space', it is not so easy to access.

This PR adds the MAD report to the reports available from the report page on TAP. This MAD report includes the future measures too.
Because of the complexity of the data used, the report is not generated when the user requests it, but it is created when the simplified  open data are generated, and stored in a table.  The validity date is displayed on the report page.
The report is only available as a download, because the number of rows involved is too big.

NOTE: 
The system is still missing the night job required to populate the open data tables.
For the moment, the data can be created running a management command.

## Checklist
- Requires migrations? Yes


open_data/reports/commodities.py
```diff
@@ -0,0 +1,40 @@
+from open_data.models import ReportGoodsNomenclature
+from open_data.models.report_models import ReportCommodityReport
+
+
+def create_commodities_report(verbose=True):
```
TomMacca, line 5: Hmm, I have an idea, what if we wrap all these reports in try, except and atomic transactions e.g. for `create_commodities_report`

```
def create_commodities_report(verbose=True):
    try:
        with transaction.atomic():
            ReportCommodityReport.objects.all().delete()
            [rest of the function]
     except Exception as e:
         print(f"An error occured while creating the Commodities Report: {e}")
```

Im thinking if something like `commodity.parent_trackedmodel_ptr` doesn't exist, the current report will be half written but then`AttributeError` out. If we added the try, except and atomic transaction it would save None to the field and continue to generate the rest of the report (I think). What do you think?

TomMacca (approved):

open_data/models/report_models.py
```diff
@@ -0,0 +1,79 @@
...
+
+class ReportCommodityReport(models.Model):
+    # It represents the commodity report, as stored on Data Workspace
+    # The model is populated when the open data is refresh
+    # Using a mdel so it can easily be included in the sqlite export
+    # The column names match the current report in Data Workspace
+    # Confusing, because the column names of other tables are translated
+    # inside Dataworkspace, but it is convenient to use the final names here
+
+    id = models.IntegerField(
```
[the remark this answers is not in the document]
Luisella21: I can't use AutoField because the reports expect an ID starting from 1. An autofield will always increase, even after deleting everything in the table. But I add a comment about it. 'unique' should be true, I am surprised I did not get an error, as the field is the primary key.

open_data/reports/commodities.py
Luisella21: Very good point. I'll change the code.

